Issue #13, opened 2020-11-04
docker-compose up: "./manage.py": stat ./manage.py: no such file or directory: unknown


yarikoptic:
Trying to try out/explore recent developments.  What should be the workflow for that?

Following current (old?) README.md, and having removed all volumes and all `dandi` images etc (I think everything relevant), 
```
docker-compose run --rm django ./manage.py migrate
```
ends with
```
Step 7/7 : WORKDIR /opt/django
 ---> Running in 62ba4ff93148
Removing intermediate container 62ba4ff93148
 ---> d3eeb0d939ef
Successfully built d3eeb0d939ef
Successfully tagged dandi-api_django:latest
WARNING: Image for service django was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "./manage.py": stat ./manage.py: no such file or directory: unknown
docker-compose run --rm django ./manage.py migrate  15.46s user 5.82s system 8% cpu 4:18.14 total
```

some COPY was forgotten?  or may be it is /opt/django vs /opt/django-project paths 
```
$> git grep /opt/django           
dev/django-public.Dockerfile:WORKDIR /opt/django
dev/django-public.Dockerfile:COPY . /opt/django/
dev/django.Dockerfile:COPY ./setup.py /opt/django-project/setup.py
dev/django.Dockerfile:RUN pip install --editable /opt/django-project[dev]
dev/django.Dockerfile:WORKDIR /opt/django
docker-compose.override.yml:      - .:/opt/django-project
docker-compose.override.yml:      - .:/opt/django-project
```
(not sure what/why is the difference)

PR #16, opened 2020-11-16
Upload/Verification flow

dchiquit:
* Add views for upload/validation
  * upload-initialize: presigns all the part upload requests for an object to be uploaded
  * upload-complete: presigns a complete upload request
  * validate: begins a validation (celery task) on an existing object
  * get-validation: gets the current status of a validation
* Add asset create method to viewset to create the asset from a validation
* Add Validation object that describes the state of a validation run
* Add tests


yarikoptic:
woohoo - I would be excited to test it out, but ideally would need #13 "addressed" somehow so I could try on my "own" instance without torturing any central instance

dchiquit:
@yarikoptic I will look at that tomorrow so you can start breaking things ASAP

yarikoptic:
;-) I will meanwhile lubricate my favorite demolition tools for smooth "operation" . Thank you in advance @dchiquit

PR #20, opened 2020-11-19
Swagger returns

dchiquit:
We have many viewsets with custom actions. Swagger cannot dynamically determine the response shape of these actions, so it must be specified manually using `@swagger_auto_schema`.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -2,6 +2,7 @@ from django.core.validators import RegexValidator
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django_filters import rest_framework as filters
+from drf_yasg import openapi
 from drf_yasg.utils import swagger_auto_schema
 from guardian.utils import get_40x_or_None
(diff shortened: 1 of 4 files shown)


dchiquit:
For context:
> Each REST endpoint is defined as a python function. Since Python isn't strongly typed, there's no way for a static code analyzer (which the swagger generator is) to determine what the response for a given endpoint is. The swagger generator can make good guesses for the really boilerplate stuff, but for more customized actions like publish or upload, it just assumes (incorrectly) that the response will be structured the same as the request. The @swagger_auto_schema decorator is used to explicitly inform the swagger generator what the request and responses look like for a certain REST endpoint.

yarikoptic:
is "python type hints" would be of help in the long run, e.g. see https://sobolevn.me/2019/08/typechecking-django-and-drf ?

dchiquit:
Very likely not, as [drf_yasg](https://drf-yasg.readthedocs.io/en/stable/) would need to be set up to read from those mypy types, which I have seen no indication that it does. Typechecking is definitely a good thing for development, though.

PR #37, opened 2020-12-15
Make it possible to list all assets starting from the root of a version

jwodder:
Assuming that asset paths are not supposed to start with a slash, then the empty string is the only valid prefix for all asset paths in a version, yet the code currently makes it impossible to query by this prefix via the `/dandisets/{dandiset_id}/versions/{version}/assets/paths/` endpoint, as the endpoint forces `path_prefix` to end with a slash.  This patch permits the use of the empty string as `path_prefix`, which should make it possible to list all assets via the endpoint.


dchiquit:
I'm OK with merging this, but my preference would be requiring `/` to be prepended to all roots. That is a larger decision that requires a larger discussion though, so this is fine in the meantime.

dchiquit (approved):

yarikoptic:
> I'm OK with merging this, but my preference would be requiring `/` to be pre-pended to all roots.

I failed to find prior discussion:  I had the same question/desire awhile back - to not have leading `/` in the returned paths (by girder IIRC).  My argument pretty much was "because they are relative paths".  The counter-argument IIRC boiled down to "but it is easier to append such path in JavaScript to the top path which doesn't not end with /".  Is that pretty much the reason for this preference in your case @dchiquit or there is more to it?

I have disliked this `/` prefix in other JS programs (e.g. https://github.com/bids-standard/bids-validator/issues/294) where if such paths are exposed to the user (without stripping away `/`) -- they are confusing to the user.  For a developer they also require additional "special handling" since now you need to retain in the code logic that what looks like an absolute path is actually not an absolute path but path relative to the top of the dataset.

So I would have preferred not having leading `/` unless it is some form of a convention for APIs.

dchiquit:
I have no practical objection to not requiring the root `/`, so I'm OK with merging this.

PR #41, opened 2020-12-18
Data migration to create default application

dchiquit:
Adds a migration that creates a default OAuth application if one does not already exist (with the name `Frontend Provider`, or with the given client ID).


dandiapi/api/migrations/0005_default_oauth_application.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 3.0.9 on 2020-12-17 23:59
+# flake8: noqa N806
+
+from django.db import migrations
+from django.db.models import Q
+
+
+def create_application(apps, schema_editor):
+    Application = apps.get_model('oauth2_provider', 'Application')
+    name = 'Frontend Provider'
```
brianhelba, line 10:
```suggestion
    name = 'DANDI GUI'
```

This seems like a little more descriptive name. Feel free to pick something else, if we have a better term for the Vue-based frontend app, but using the word "Provider" here is misleading.

brianhelba (changes requested):

dandiapi/api/migrations/0005_default_oauth_application.py
```diff
@@ -0,0 +1,46 @@
...
+    # This is specified so it matches the default value used in the frontend.
+    client_id = 'Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl'
+    if not Application.objects.filter(Q(name=name) | Q(client_id=client_id)).exists():
+        # Use the first admin user
+        User = apps.get_model('auth', 'User')
+        user = User.objects.filter(is_superuser=True).order_by('id')[0]
```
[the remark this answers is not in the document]
dchiquit: I did not realize it was optional, I think it's better to remove it.

dandiapi/api/migrations/0005_default_oauth_application.py
dchiquit: I just spitballed that, I don't have strong opinions on the name.

dandiapi/api/migrations/0005_default_oauth_application.py
```diff
@@ -0,0 +1,42 @@
...
+            client_secret='',
+            name=name,
+            # This can be turned off in production if appropriate.
+            skip_authorization=False,
+        )
+        application.save()
+
+
+def reverse_create_application(apps, schema_editor):
+    # Deleting the application would destroy any customization applied to it.
```
brianhelba, line 31: Just FYI, destroying customization is an expected result of reversing a migration, and shouldn't be a reason not to write a reverse migration.

However, reverse migrations aren't used very often, and it may not be worth the burden to write one.

brianhelba (approved):

PR #45, opened 2020-12-30
Add the ability to sort dandisets when listing

dchiquit:
Add a custom filter that orders dandisets by created date and name (actually the name of the most recent published version).

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -1,10 +1,11 @@
 from django.contrib.auth.models import User
+from django.db.models import OuterRef, Subquery
 from django.http import Http404
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -20,10 +21,36 @@
...
+        'Which field to use when ordering the results. '
+        'Options are created, -created, name, and -name.'
+    )
+
+    def filter_queryset(self, request, queryset, view):
+        orderings = self.get_ordering(request, queryset, view)
+        if orderings:
+            ordering = orderings[0]
+            # ordering can be either 'created' or '-created', so test for both
+            if ordering[-7:] == 'created':
```
waxlamp, line 36: Just a Python tip: you can use a string's `endswith()` method to avoid having to slice the string: `if ordering.endswith('created'): ...`.

PR #59, opened 2021-01-19
User migration scripts

dchiquito:
Things should be done in roughly this order:
* Copy dandisets from Girder to Django (presumably using CLI).
* Run `create_placeholder_users` script. This will recognize and dandisets that exist in both Girder and Django, create placeholder user accounts for each owner as needed, and assign ownership to the users/placeholders.
* Inform users that they will need to log in to GitHub again and authorize on api.dandiarchive.org. This is honestly not hard at all, but maybe they should be alerted.
* When each user logs in, a post login hook will run and assign the user all of the ownerships that their placeholder had, then delete the placeholder.

The `depose_placeholders` script will also attempt to remove all possible placeholders, but hopefully that won't be necessary.

The `list_placeholders` script generates a space separated list of all the emails for users who still have placeholders, in case we want to send reminder emails.


scripts/create_placeholder_users.py
```diff
@@ -0,0 +1,92 @@
...
+            assign_perm('owner', django_user, dandiset)
+
+
+def migrate(gc):
+    girder_dandisets = get_girder_dandisets(gc)
+    for dandiset_identifier in girder_dandisets:
+        # Do not use the dandiset identifier, it is an autoincremented primary key
+        # and cannot be trusted.
+        # We are also assuming that no two dandisets will have metadata with the same identifier.
+        # TODO make sure this query works with all identifier formats
```
waxlamp, line 71: You may want to file an issue linking back to this line of code, just so nothing gets lost.

scripts/create_placeholder_users.py
```diff
@@ -0,0 +1,92 @@
...
+        version = Version.objects.filter(metadata__metadata__identifier=dandiset_identifier).first()
```
waxlamp, line 72: I assume that, by design, `version` would be either an empty list or a singleton list. Is there any condition that could lead to multiple versions being found here?

dandiapi/api/__init__.py
```diff
@@ -0,0 +1,2 @@
+# TODO remove this after migration is complete
```
waxlamp, line 2: Could you file an issue for this and mark it with `migration` or whatever label may already exist? That will make it easier to get a complete list of tasks together in a formal migration plan.

dandiapi/api/user_migration.py
```diff
@@ -0,0 +1,39 @@
...
+
+    placeholder_email = 'placeholder_' + user.email
+    try:
+        placeholder_user = User.objects.get(email=placeholder_email)
+    except User.DoesNotExist:
+        # No placeholder user, nothing to do
+        return
+
+    print(f'Replacing {placeholder_user} with {user}')
+    copy_ownership(placeholder_user, user)
```
waxlamp, line 31: If this step fails partway through for some reason (like, network failure, etc.), will that cause this script to properly abort so we don't end up deleting the placeholder user prematurely?

scripts/create_placeholder_users.py
```diff
@@ -0,0 +1,94 @@
...
+        if version is None:
+            print(f'Dandiset {dandiset_identifier} does not exist in Django')
+            continue
+        print(f'Found dandiset {dandiset_identifier} in both Girder and Django')
+        user_ids = girder_dandisets[dandiset_identifier]
+        migrate_dandiset(gc, version.dandiset, user_ids)
+
+
+def run(*args):
```
waxlamp, line 81: Style question: why not just make the arg list `api_url`, `username`, and `password`? Is that a requirement of how these Django scripts work?

waxlamp (approved):
Looks great. All of my comments are either non-urgent questions, or minor, non-code tasks.

dandiapi/api/user_migration.py
dchiquito: Any python exception should interrupt execution, yes. I'm reasonably confident that all operations are idempotent, but I don't want to spend too much time proving it for these functions that will hopefully only be run once.

scripts/create_placeholder_users.py
dchiquito: This finds an arbitrary version which has metadata that contains a dandiset identifier. There could hypothetically be multiple published versions to match against.

It's ultimately OK if there are multiple versions, since we are actually only using `version.dandiset`, which should hopefully be the same for all versions which claim to belong to a dandiset.

scripts/create_placeholder_users.py
dchiquito: Yep https://django-extensions.readthedocs.io/en/latest/runscript.html#passing-arguments

PR #67, opened 2021-01-25
Map ids endpoint

dchiquito:
* Copy the `map_ids.py` script into the `api` package so it can be referenced by the new views
* Add views to run `map_ids` against a specific dandiset or a single dandiset
* Do some really ugly stuff to log `map_ids` progress into a list of strings rather than stdout, then return it from the views.

### DIFF
--- a/dandiapi/api/dandiset_migration.py
+++ b/dandiapi/api/dandiset_migration.py
@@ -1,0 +1,69 @@
+"""
+Maps the identifiers of all Dandisets to the identifier listed in their metadata.
+
+Identifier cycles are not addressed.
+If you have Dandiset 000001 with {"identifier":000002} and Dandiset 000002
+with {"identifier":000001}, manual intervention is required.
+"""
+from dandiapi.api.models.dandiset import Dandiset
+
+
+def get_new_identifier(dandiset):
+    metadata = dandiset.most_recent_version.metadata.metadata
+    if 'identifier' not in metadata:
+        print(f'Dandiset {dandiset.identifier} does not specify an identifier')
+        return None
+    try:
+        identifier = int(metadata['identifier'])
+    except ValueError:
+        print(
+            f'Dandiset {dandiset.identifier} has a bad metadata identifier {metadata["identifier"]}'
+        )
+        return None
+
+    if 0 > identifier or identifier > 999999:
+        print(
+            f'Dandiset {dandiset.identifier} has a bad metadata identifier {metadata["identifier"]}'
+        )
+        return None
+
+    if dandiset.id == identifier:
+        # The identifier already matches
+        return None
+
+    if Dandiset.objects.filter(id=identifier):
+        print(
+            f'Dandiset {dandiset.identifier} cannot be copied because {identifier} already exists'
+        )
+        return None
+
+    return identifier
+
+
+def move(dandiset, new_id):
+    # Grab the old dandiset's relevant information
+    owners = dandiset.owners
+    versions = list(dandiset.versions.all())
+
+    # Create a new dandiset with the new identifier
+    new_dandiset = Dandiset(id=new_id)
+    new_dandiset.save()
+
+    # Copy the owners
+    new_dandiset.set_owners(owners)
+    new_dandiset.save()
+
+    # Copy the versions
+    for version in versions:
+        version.dandiset = new_dandiset
+        version.save()
+
+    # Delete the old dandiset
+    dandiset.delete()
+
+
+def move_if_necessary(dandiset):
+    new_id = get_new_identifier(dandiset)
+    if new_id:
+        print(f'Copying dandiset {dandiset.identifier} to {new_id}')
+        move(dandiset, new_id)
--- a/dandiapi/api/dandiset_migration.py
+++ b/dandiapi/api/dandiset_migration.py
@@ -8,21 +8,25 @@ with {"identifier":000001}, manual intervention is required.
 from dandiapi.api.models.dandiset import Dandiset
 
 
+# Track log messages instead of logs.appending so they can be output in rest response
+logs = []
+
+
 def get_new_identifier(dandiset):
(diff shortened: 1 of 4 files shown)


dandiapi/api/dandiset_migration.py
```diff
@@ -0,0 +1,75 @@
...
+        version.dandiset = new_dandiset
+        version.save()
+
+    # Delete the old dandiset
+    dandiset.delete()
+
+
+def move_if_necessary(dandiset):
+    global logs
+    logs = []
```
AlmightyYakob, line 70: Usually I'd say we should do this in a different way, but since this is temporary code anyway, I don't think it's an issue.

dandiapi/urls.py
```diff
@@ -80,6 +81,9 @@ def to_url(self, value):
     ),
     path('api/users/me/', users_me_view),
     path('api/users/search/', users_search_view),
+    # TODO remove these endpoints after migration
+    path('api/scripts/map_ids/', map_ids_view, name='map-ids'),
+    re_path(r'^api/scripts/map_ids/(?P<dandiset_id>[0-9]{6})/$', map_id_view, name='map-id'),
     path('accounts/', include('allauth.urls')),
```
AlmightyYakob, line 87: This might make it a little more clear:

```suggestion
    path('api/users/search/', users_search_view),
    
    # TODO remove these endpoints after migration
    path('api/scripts/map_ids/', map_ids_view, name='map-ids'),
    re_path(r'^api/scripts/map_ids/(?P<dandiset_id>[0-9]{6})/$', map_id_view, name='map-id'),
    
    path('accounts/', include('allauth.urls')),
```

dandiapi/api/dandiset_migration.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def move(dandiset, new_id):
+    # Grab the old dandiset's relevant information
+    owners = dandiset.owners
+    versions = list(dandiset.versions.all())
+
+    # Create a new dandiset with the new identifier
+    new_dandiset = Dandiset(id=new_id)
+    new_dandiset.save()
```
AlmightyYakob, line 53: Why save here if you just save again below?

dandiapi/api/dandiset_migration.py
```diff
@@ -0,0 +1,75 @@
...
+    return identifier
+
+
+def move(dandiset, new_id):
+    # Grab the old dandiset's relevant information
+    owners = dandiset.owners
+    versions = list(dandiset.versions.all())
+
+    # Create a new dandiset with the new identifier
+    new_dandiset = Dandiset(id=new_id)
```
AlmightyYakob, line 52: This is likely just due to my inexperience, but does specifying the `id` here have any adverse effects on the ordering/numbering of new dandisets? Put another way, what are the effects of creating a new dandiset normally, and how does specifying the `id` here change that?

AlmightyYakob (commented):
Looks good, just a few comments.

dandiapi/api/dandiset_migration.py
dchiquito: Django provides an auto-incrementing `id` field [for free](https://docs.djangoproject.com/en/3.1/topics/db/models/#automatic-primary-key-fields). If it's not specified, it just picks the next int in the sequence.

Regardless of whether you used the auto-incremented value or explicitly specified an id as we did here, creating an instance of the model does not do any collision checking. Even if dandiset `000007` already exists, there's no problem with calling `Dandiset(id=7)` (we use `id` to generate `identifier`, which is the 0-padded string). If, however, you try to call `save()` on that duplicate dandiset, a `Dandiset.IntegrityError` is thrown with some info on the underlying Postgres error (in this case, duplicate unique columns).

Technically, we should do some error handling here in case there is a collision. However, we already checked for an existing dandiset as the last step in `get_new_identifier`, which is good enough for this script. There are some problems with race conditions during simultaneous runs, but that's not a consideration for a script like this.

dandiapi/api/dandiset_migration.py
dchiquito: I didn't notice I did it twice ¯\_(ツ)_/¯ it's technically redundant, but whatever

dandiapi/api/dandiset_migration.py
dchiquito: agreed

dandiapi/urls.py
dchiquito: This violates linting, sadly

AlmightyYakob (approved):

yarikoptic:
I see some "move" etc, scared to try.  @dchiquito would you be so kind to walk me through on what I should invoke to instantiate all dandisets from https://gui.dandiarchive.org/ on https://gui-beta-dandiarchive-org.netlify.app/ ?  Then I will 
- 1. upload dandiset level metadata to all of those
- 2. upload assets to some of those

dchiquito:
@yarikoptic It's safe to use in the wild, I just verified it. The intended flow is:
* Download the dandiset from https://gui.dandiarchive.org/ that you want to upload to  https://gui-beta-dandiarchive-org.netlify.app/.
* Create a new dandiset in  https://gui-beta-dandiarchive-org.netlify.app/. It will almost definitely have the wrong identifier.
* Use the CLI to upload the dandiset into the wrongly-identified dandiset on https://gui-beta-dandiarchive-org.netlify.app/. Make sure the metadata contains the correct identifier. The CLI will need a way to upload the dandiset to the "wrong" identifier.
* Use the [swagger page](https://api.dandiarchive.org/swagger/) to run `/scripts/map_ids/{wrong_id}` to change that dandiset's identifier to the one listed in its metadata. Alternatively, just run `/scripts/map_ids/` to do that to all dandisets.

If you run into any cycles (i.e. 000001 needs to moved to 000002, and vice versa), you'll need admin intervention.

PR #73, opened 2021-01-27
Use request_body=no_body on VersionViewSet.publish

### LINKED ISSUE #72: Remove data field from Swagger description of publish endpoint
On the Swagger page, the entry for the `publish` action includes a field `data` which specifies that the request should include an object which looks like a Version. This is misleading, as the endpoint completely ignores the request body.

Remove the `data` field from swagger.

dchiquito:
`drf-yasg` generates the endpoints on the Swagger page automatically, but sometimes makes incorrect guesses. In this case, it assumes that all POST endpoints require request bodies, which is not the case for the `publish` endpoint.

The solution is to use the `@swagger_auto_schema` decorator to indicate that the `request_body` is `no_body`. I originally tried to this by using `None`, but `no_body` is required to explicitly disable the request body.

Resolves #72

### DIFF
--- a/dandiapi/api/views/version.py
+++ b/dandiapi/api/views/version.py
@@ -1,4 +1,4 @@
-from drf_yasg.utils import swagger_auto_schema, no_body
+from drf_yasg.utils import no_body, swagger_auto_schema
 from guardian.utils import get_40x_or_None
 from rest_framework import status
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):
Tested locally :+1:

PR #74, opened 2021-01-27
Throw meaningful duplicate asset error

### LINKED ISSUE #68: Duplicate Asset error reporting
All assets are unique (uuid is minted, we never modify the asset, only create new ones, and eventually gc unused).... Server should error out informatively nevertheless if there is path "collision"

_Originally posted by @yarikoptic in https://github.com/dandi/dandi-api/issues/65#issuecomment-767226092_

Currently, trying to create a new Asset with the same version+path as an existing Asset results in an error 500, with no additional context in the response. This should be a more informative error and should be documented on the Swagger page.

dchiquito:
Assets have a `unique_together` constraint on parent version + path, for good reason. It's possible to request a new Asset with the same parent version + path as an existing Asset, which resulted in a 500 error because of the DB throwing an `IntegrityError`.

Catch that `IntegrityError` and throw an HTTP 400.

Fixes #68

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -1,4 +1,5 @@
 from django.core.validators import RegexValidator
+from django.db.utils import IntegrityError
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django_filters import rest_framework as filters
@@ -85,7 +86,10 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
             metadata=asset_metadata,
             version=version,
         )
-        asset.save()
+        try:
+            asset.save()
+        except IntegrityError:
+            return Response('asset already exists', status=status.HTTP_400_BAD_REQUEST)
 
         serializer = AssetDetailSerializer(instance=asset)
         return Response(serializer.data, status=status.HTTP_200_OK)
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -88,8 +88,13 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
         )
         try:
             asset.save()
-        except IntegrityError:
-            return Response('asset already exists', status=status.HTTP_400_BAD_REQUEST)
+        except IntegrityError as e:
+            # https://stackoverflow.com/questions/25368020/django-deduce-duplicate-key-exception-from-integrityerror
+            # https://www.postgresql.org/docs/13/errcodes-appendix.html
+            # Postgres error code 23505 == unique_violation
+            if e.__cause__.pgcode == '23505':
+                return Response('asset already exists', status=status.HTTP_400_BAD_REQUEST)
+            raise e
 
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/asset.py
```diff
@@ -85,7 +86,10 @@ def create(self, request, version__dandiset__pk, version__version):
             metadata=asset_metadata,
             version=version,
         )
-        asset.save()
+        try:
+            asset.save()
+        except IntegrityError:
+            return Response('asset already exists', status=status.HTTP_400_BAD_REQUEST)
```
yarikoptic, line 92: is that the only `IntegrityError` case whenever it could be raised?  may be so ATM, but would it be always the only one? ;)
I think it would be quite bad if error message is not factually correct...
from https://github.com/dandi/dandi-api/issues/65 it looks like 
```
django.db.utils.IntegrityError: duplicate key               
Jan 25 15:15:40 dandi-api app/web.1                     value violates unique constraint                            
Jan 25 15:15:40 dandi-api app/web.1                     "api_asset_path_version_id_0267c3d9_uniq"                   
Jan 25 15:15:40 dandi-api app/web.1                     DETAIL:  Key (path,                                         
Jan 25 15:15:40 dandi-api app/web.1                     version_id)=(sub-RAT123/sub-RAT123.nwb, 10)                 
Jan 25 15:15:40 dandi-api app/web.1                     already exists
```

I hate doing matching in outputs or messages, but may be first match for `duplicate key  value violates unique constraint api_asset_path_version_id` and if that matches -- return the above response, but ideally include that path into it (could help debugging while trying to upload many files and some file failing).  If no match -- just a `str` of that exception?

yarikoptic (changes requested):

dandiapi/api/views/asset.py
yarikoptic: re
> return the above response

I would also improve that as `f'there is already an asset at {path}'` or alike.  "asset already exists" is a bit of, since in my mind assets are uniq, and I am just creating a new asset, so how could it already "exist"? ;)

dchiquito: Turns out there is a semi-formal way to accomplish this: https://stackoverflow.com/questions/25368020/django-deduce-duplicate-key-exception-from-integrityerror

dandiapi/api/views/asset.py
```diff
@@ -85,7 +86,15 @@ def create(self, request, version__dandiset__pk, version__version):
...
         )
-        asset.save()
+        try:
+            asset.save()
+        except IntegrityError as e:
+            # https://stackoverflow.com/questions/25368020/django-deduce-duplicate-key-exception-from-integrityerror
+            # https://www.postgresql.org/docs/13/errcodes-appendix.html
+            # Postgres error code 23505 == unique_violation
+            if e.__cause__.pgcode == '23505':
+                return Response('asset already exists', status=status.HTTP_400_BAD_REQUEST)
```
AlmightyYakob, line 96: Minor but since it is an API response: 
```suggestion
                return Response('Asset Already Exists', status=status.HTTP_400_BAD_REQUEST)
```

AlmightyYakob (approved):

PR #75, opened 2021-01-27
Delete asset endpoint

### LINKED ISSUE #70: need DELETE /dandisets/{version__dandiset__pk}/versions/{version__version}/assets/{uuid}/
Without this we cannot delete a file in a draft version, or upload (Follow up to #65) a new version of the file (ref: https://github.com/dandi/dandi-cli/issues/345)

For a regular user only "draft" `{version__version}`  should be allowed

edit: just to make clear -- it should not really delete an asset, just delete association of that asset with given dandiset version

dchiquit:
Fixes #70

[ViewSets](https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions) handle most of the boilerplate around generate API endpoints. All we have to do to add the DELETE endpoint for assets is to define the appropriate `destroy` method on the existing `AssetViewSet`.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -124,6 +124,19 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
         serializer = AssetDetailSerializer(instance=asset)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
+    # @permission_required_or_403('owner', (Dandiset, 'pk', 'version__dandiset__pk'))
+    def destroy(self, request, **kwargs):
+        asset = self.get_object()
+
+        # TODO @permission_required doesn't work on methods
+        # https://github.com/django-guardian/django-guardian/issues/723
+        response = get_40x_or_None(request, ['owner'], asset.version.dandiset, return_403=True)
+        if response:
+            return response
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/asset.py
```diff
@@ -124,6 +124,19 @@ def update(self, request, **kwargs):
         serializer = AssetDetailSerializer(instance=asset)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
+    # @permission_required_or_403('owner', (Dandiset, 'pk', 'version__dandiset__pk'))
+    def destroy(self, request, **kwargs):
```
yarikoptic, line 128: why `destroy` and not `delete` for consistency?

yarikoptic:
what is your thinking @dchiquit on what else todo to have it merged and "tried"?

dandiapi/api/views/asset.py
dchiquit: `django-rest-framework` defines these ViewSet classes that correspond to REST "entities". The `list` method is mapped to `GET assets/`, `get` to `GET assets/{uuid}`, and `destroy` to `DELETE assets/{uuid}`.

dchiquit:
@AlmightyYakob care to take a peek?

PR #77, opened 2021-01-29
Automatically populate name and identifier in VersionMetadata on Version save

### LINKED ISSUE #63: create: should populate identifier
ATM creation of new dandiset via 'Create' button (at least that is what I did) does not populate `identifier` (only name and description), which then leads to `dandi-cli` downloaded `dandiset.yaml` (which is just a metadata record)` lack that identifier information.  Should look like
```
identifier:
  propertyID: DANDI
  value: '000009'
```
AFAIK according to our current schema and samples of converted dandiset level metadata in  https://github.com/dandi/dandi-api-datasets.git

AlmightyYakob:
Fixes #63

The main change here is an overriding of the `Version.save` method, to include automatically populating the `VersionMetadata.metadata` field with `name` and `identifier`. The format of `identifier` is dependent on [this](https://github.com/dandi/dandi-cli/pull/348) schema change.

### DIFF
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -3,7 +3,7 @@ import pytest
 
 from dandiapi.api.models import Dandiset
 
-from .fuzzy import DANDISET_ID_RE, TIMESTAMP_RE
+from .fuzzy import DANDISET_ID_RE, DANDISET_SCHEMA_ID_RE, TIMESTAMP_RE
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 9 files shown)


AlmightyYakob:
Changed back to a draft PR because I need to fix the tests

dandiapi/api/tests/test_dandiset.py
```diff
@@ -129,7 +129,11 @@ def test_dandiset_rest_create(api_client, user):
     assert dandiset.versions.count() == 1
     assert dandiset.most_recent_version.version == 'draft'
     assert dandiset.most_recent_version.metadata.name == name
-    assert dandiset.most_recent_version.metadata.metadata == metadata
+    assert dandiset.most_recent_version.metadata.metadata == {
+        **metadata,
+        'name': name,
+        'identifier': DANDISET_SCHEMA_ID_RE,
```
dchiquito, line 135:
```suggestion
        'identifier': response.data['identifier'],
```

dchiquito (approved):

dandiapi/api/tests/test_dandiset.py
AlmightyYakob: This won't match because the value in the metadata isn't the raw identifier, but adheres to [this](https://github.com/satra/dandi-cli/blob/2f4755f19f8c6f5272313c6d440306ec1955a5ef/dandi/models.py#L677-L683) schema change (of form `DANDI:XXXXXX`)

PR #82, opened 2021-02-02
Remove failing test and add comment

AlmightyYakob:
As a part of #77, I added a new test, `test_search_metadata`, which tests searching arbitrary version metadata. This seems to fail sporadically, and the search endpoint is intentionally limited anyway, so this PR removes that test.


dchiquito (approved):
oh well

PR #83, opened 2021-02-03
Populate `schema_version` metadata, add `info` endpoint

### LINKED ISSUE #64: create: should populate schemaVersion so it is returned along with metadata record for a dandiset
all metadata records must have `schemaVersion` field defined (ATM the `1.0.0-rc1`, @satra please correct me if I am wrong).

So `create` should define
```
schemaVersion: 1.0.0-rc1
```
ATM

examples are among https://github.com/dandi/dandi-api-datasets.git

AlmightyYakob:
Fixes #64.

This addresses the main issue of populating the `schemaVersion` field, but also introduces some extra steps that need to be taken in order to coordinate the client and server. Namely, a new `info` endpoint that returns the schema version, along with the schema URL ~~(yet to be added)~~. My thought is that this endpoint is meant to convey general information about the server, as it relates to operation. I didn't think it quite fit in the `stats` endpoint, since that's a self-contained thing. Let me know what you think.

The schema URL field is a bit extra, but in my opinion necessary in the long term, since we'll eventually need to do metadata validation on the server, and will need to know where to retrieve the schema from.


dandiapi/api/views/info.py
```diff
@@ -0,0 +1,15 @@
+from django.conf import settings
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+schema_url = f'https://raw.githubusercontent.com/dandi/schema/master/releases/{settings.DANDI_SCHEMA_VERSION}/dandiset.json'
```
AlmightyYakob, line 5: This is a bit smelly, but IMO adding another environment variable for this wouldn't really be worth it, as I don't expect this URL to change, and it would add more complexity.

dchiquito: I think it's fine, since this is derived from another environment variable anyway.

AlmightyYakob:
@dchiquito This should be good for review now

yarikoptic:
now it needs conflicts resolution.  Would be lovely to see this merged, so we could add support in dandi-cli

dandiapi/api/views/info.py
```diff
@@ -0,0 +1,18 @@
...
+
+schema_url = (
+    'https://raw.githubusercontent.com/dandi/schema/master/'
+    f'releases/{settings.DANDI_SCHEMA_VERSION}/dandiset.json'
+)
+
+
+@api_view()
+def info_view(self):
+    return Response(
```
dchiquito, line 13: Tragically this will not self-document on the swagger page, `/api/info/` will just show some generic 200 response placeholder.

The solution would involve defining a `Serializer` that has `schemaVersion` and `schemaURL` defined as fields, then invoking it like this:
```
from drf_yasg.utils import no_body, swagger_auto_schema
...
    @swagger_auto_schema(
        request_body=no_body,
        responses={
            200: ApiInfoSerializer(),
        },
    )
    ```
    
You should probably also use that serializer when generating the response, but this endpoint is so simple it hardly matters.

dchiquito (approved):

PR #84, opened 2021-02-03
Fix object keys

### LINKED ISSUE #81: Replacing an asset in a draft affects published versions
As of dandi-api commit 3f0e49181880aa6bfd1ee10a943736ffd7072dcd, replacing an asset in a Dandiset's "draft" version by deleting and uploading a different asset also affects any assets with the same path in published versions.  This can be seen by running the following MVCE against a dandi-api instance:

<details>
<summary>MCVE</summary>

```python
"""
Dependency::

    git+https://github.com/dandi/dandi-cli@gh-345

Run as::

    python3 <this-script> <DANDI API base URL> <auth token>
"""

from   pathlib        import Path
import sys
from   tempfile       import TemporaryDirectory
from   dandi.dandiapi import DandiAPIClient

api_url, token = sys.argv[1:]

client = DandiAPIClient(api_url=api_url, token=token)

with TemporaryDirectory() as tmpdir, client.session():
    dandiset_id = client.create_dandiset("Test Dandiset", {})["identifier"]

    tmp_path = Path(tmpdir)
    (tmp_path / "upload").mkdir()
    (tmp_path / "upload" / "file.txt").write_text("This is test text.\n")

    client.upload(dandiset_id, "draft", "file.txt", {}, tmp_path / "upload" / "file.txt")
    print('Uploaded file.txt to draft with contents "This is test text."')

    version_id = client.publish_version(dandiset_id, "draft")["version"]
    print("Draft published as version", version_id)

    (tmp_path / "download1").mkdir()
    client.download_asset_bypath(
        dandiset_id,
        version_id,
        "file.txt",
        tmp_path / "download1" / "file.txt",
    )
    print("file.txt downloaded from", version_id)
    print("Contents:")
    print(f"    Got: {(tmp_path / 'download1' / 'file.txt').read_text()!r}")
    print("    Expected: 'This is test text.\\n'")

    client.delete_asset_bypath(dandiset_id, "draft", "file.txt")

    (tmp_path / "upload" / "file.txt").write_text("This is different text.\n")
    client.upload(dandiset_id, "draft", "file.txt", {}, tmp_path / "upload" / "file.txt")
    print('Uploaded file.txt to draft with contents "This is different text."')

    (tmp_path / "download2").mkdir()
    client.download_asset_bypath(
        dandiset_id,
        version_id,
        "file.txt",
        tmp_path / "download2" / "file.txt",
    )
    print("file.txt downloaded from", version_id)
    print("Contents:")
    print(f"    Got: {(tmp_path / 'download2' / 'file.txt').read_text()!r}")
    print("    Expected: 'This is test text.\\n'")
```

</details>

Output when I ran it:

```
Uploaded file.txt to draft with contents "This is test text."
Draft published as version 0.210202.2121
file.txt downloaded from 0.210202.2121
Contents:
    Got: 'This is test text.\n'
    Expected: 'This is test text.\n'
Uploaded file.txt to draft with contents "This is different text."
file.txt downloaded from 0.210202.2121
Contents:
    Got: 'This is different text.\n'
    Expected: 'This is test text.\n'
```

dchiquito:
Fixes #81 

The upload process will now append a UUID to the `object_key`, so all uploads that go through the validation process are guaranteed to not overwrite eachother.

We do not use the `sha256` value because at upload time, it is given by the client. If we allow the client to give all the information that generates the object key, malicious clients could potentially give the `sha256` of a different file, overwriting the blob.

### DIFF
--- a/dandiapi/api/models/validation.py
+++ b/dandiapi/api/models/validation.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from uuid import uuid4
+
 from django.conf import settings
 from django.core.files.storage import Storage
 from django.core.validators import RegexValidator
@@ -14,8 +16,7 @@ def _get_validation_blob_storage() -> Storage:
 
 
 def _get_validation_blob_prefix(instance: Validation, filename: str) -> str:
-    # return f'{instance.version.dandiset.identifier}/{instance.version.version}/{filename}'
-    return filename
+    return f'{filename}/{uuid4()}'
 
 
 class Validation(TimeStampedModel):
--- a/dandiapi/api/tests/test_upload.py
+++ b/dandiapi/api/tests/test_upload.py
@@ -22,7 +22,7 @@ def test_upload_initialize(api_client, user):
         {'file_name': file_name, 'file_size': file_size},
         format='json',
     ).data == {
-        'object_key': file_name,
+        'object_key': Re(f'{file_name}/[a-z0-9\\-]+'),
         'upload_id': UUID_RE,
         'parts': [
             {
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -96,7 +96,9 @@ def upload_initialize_view(request: Request) -> HttpResponseBase:
     # TODO The first argument to generate_filename() is an instance of the model.
     # We do not and will never have an instance of the model during field upload.
     # Maybe we need a different generate method/upload_to with a different signature?
-    object_key = Validation.blob.field.storage.generate_filename(upload_request['file_name'])
+    object_key = Validation.blob.field.storage.generate_filename(
+        Validation.blob.field.upload_to(None, upload_request['file_name'])
+    )
 
     initialization = MultipartManager.from_storage(Validation.blob.field.storage).initialize_upload(
         object_key, upload_request['file_size']


yarikoptic:
Let me ask: If the same 1Tb file is uploaded twice - do we get two copies on s3?

yarikoptic:
> sha256 of a different file, overwriting the blob.

If blob is already there - there should be nothing to upload, and nothing to overwrite

satra:
i thought in our design discussion we had the following:
1. upload to staging area with uuid (so indeed two identical blobs could be uploaded if the client is not checking)
2. on upload finalize call, run checksum computation to finish validation
3. move the blob to blob area with checksum as key

dchiquito:
Our immediate priority is to do the migration. Currently, this bug is breaking the API, and this will fix the bug. The external API will be functional again, even if things aren't perfect under the hood.

Right now the staging area and the draft bucket are the same, and files are not copied after they finish validation. This is a high priority feature that definitely needs to be implemented, but it is not a blocker for the migration. We can change all of this after we finish jettisoning the old Girder.

mgrauer:
I've been trying to impress on @dchiquito and @AlmightyYakob the importance of landing the migration as soon as we can, and so to help @yarikoptic and @jwodder cut a clear path, and then doing clean up and improvements after.

@dchiquito could you create an issue to track this improvement post-migration?

dchiquito:
https://github.com/dandi/dandi-api/issues/85

satra:
@mgrauer - but doesn't migration imply the CLI and server are working as well, which means validating (which hasn't been done thus far, and we have issues in datasets right now with missing files). also i thought based on our last meeting that migration would happen through the CLI + some server side things (e.g., users). and that the focus right was on making sure the API server + CLI played nice to create, update and publish dandisets. (and then do migration, which would require out of band activities).

AlmightyYakob (approved):
Tested locally :+1:

yarikoptic:
> I've been trying to impress on @dchiquito and @AlmightyYakob the importance of landing the migration as soon as we can, and so to help @yarikoptic and @jwodder cut a clear path, and then doing clean up and improvements after.
> 
> @dchiquito could you create an issue to track this improvement post-migration?

echoing @satra 's reply : we would need this feature before we "flip the switch" and do the final migration. But  before that we would need to test it quite extensively to ensure that it is working as expected.  So, we could proceed with current solution just to get things moving and establish the testable server/client workflow and migration routines, but proper implementation should not be postponed until after migration.

waxlamp:
> > I've been trying to impress on @dchiquito and @AlmightyYakob the importance of landing the migration as soon as we can, and so to help @yarikoptic and @jwodder cut a clear path, and then doing clean up and improvements after.
> > @dchiquito could you create an issue to track this improvement post-migration?
> 
> echoing @satra 's reply : we would need this feature before we "flip the switch" and do the final migration. But before that we would need to test it quite extensively to ensure that it is working as expected. So, we could proceed with current solution just to get things moving and establish the testable server/client workflow and migration routines, but proper implementation should not be postponed until after migration.

Let's discuss how to handle this during today's scheduled discussion about migration operations.

yarikoptic:
And it was discussed ;) To keep the ball rolling, I think it would be of benefit to have this merged, thus have #81 addressed, and dandi-cli unblocked so we could proceed with solidifying dandi-cli/dandi-api  interactions. RF with placement into the "blob store" according to checksum could follow and #85 (although current title is not exactly describing that) seems to be about that.
By the power vested in me, and encouraged by an approval, I am clicking the green button.

PR #87, opened 2021-02-05
Create dandiset with identifier

dchiquito:
If `identifier` is in the metadata when creating a new dandiset, attempt to use that identifier instead of using the auto-incrementing sequence.

### DIFF
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -140,6 +140,71 @@ def test_dandiset_rest_create(api_client, user):
 
 
 @pytest.mark.django_db
+def test_dandiset_rest_create_with_identifier(api_client, user):
+    api_client.force_authenticate(user=user)
+    name = 'Test Dandiset'
+    identifier = '123456'
+    metadata = {'foo': 'bar', 'identifier': f'DANDI:{identifier}'}
+
+    response = api_client.post(
+        '/api/dandisets/',
+        {'name': name, 'metadata': metadata},
+        format='json',
+    )
+    assert response.data == {
+        'identifier': identifier,
+        'created': TIMESTAMP_RE,
+        'modified': TIMESTAMP_RE,
+        'most_recent_version': {
+            'version': 'draft',
+            'name': name,
+            'asset_count': 0,
+            'size': 0,
+            'dandiset': {
+                'identifier': identifier,
+                'created': TIMESTAMP_RE,
+                'modified': TIMESTAMP_RE,
+            },
+            'created': TIMESTAMP_RE,
+            'modified': TIMESTAMP_RE,
+        },
+    }
+
+    # Creating a Dandiset has side affects.
+    # Verify that the user is the only owner.
+    dandiset = Dandiset.objects.get(id=identifier)
+    assert list(dandiset.owners.all()) == [user]
+
+    # Verify that a draft Version and VersionMetadata were also created.
+    assert dandiset.versions.count() == 1
+    assert dandiset.most_recent_version.version == 'draft'
+    assert dandiset.most_recent_version.metadata.name == name
+
+    # Verify that name and identifier were injected
+    assert dandiset.most_recent_version.metadata.metadata == {
+        **metadata,
+        'name': name,
+        'identifier': f'DANDI:{identifier}',
+    }
+
+
+@pytest.mark.django_db
+def test_dandiset_rest_create_with_duplicate_identifier(api_client, user, dandiset):
+    api_client.force_authenticate(user=user)
+    name = 'Test Dandiset'
+    identifier = dandiset.identifier
+    metadata = {'foo': 'bar', 'identifier': f'DANDI:{identifier}'}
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -94,8 +95,21 @@ def create(self, request):
         if created:
             version_metadata.save()
 
-        dandiset = Dandiset()
-        dandiset.save()
+        if 'identifier' in serializer.validated_data['metadata']:
```
AlmightyYakob, line 98: I actually think that if we're going to support this feature first hand, we should specify it at the top level (next to `name`). What do you think?

dandiapi/api/views/dandiset.py
```diff
@@ -94,8 +95,21 @@ def create(self, request):
...
+            identifier = serializer.validated_data['metadata']['identifier']
+            if identifier.startswith('DANDI:'):
+                identifier = identifier[6:]
+            dandiset = Dandiset(id=int(identifier))
```
AlmightyYakob, line 102: We should probably add a try-except here (since something like `DANDI:qwerty` would throw an exception here), and if an exception is thrown, return a Response that informs them that the passed identifier is invalid.

dandiapi/api/views/dandiset.py
```diff
@@ -94,8 +95,21 @@ def create(self, request):
...
+        else:
+            dandiset = Dandiset()
+        try:
+            dandiset.save(force_insert=True)
+        except IntegrityError as e:
+            print(e.__cause__.pgcode)
```
AlmightyYakob, line 108: I'm guessing this is leftover?
```suggestion
```

dandiapi/api/views/dandiset.py
```diff
@@ -94,8 +95,21 @@ def create(self, request):
...
+            if e.__cause__.pgcode == '23505':
```
AlmightyYakob, line 109: A comment here explaining what this code means would be nice.

dandiapi/api/views/dandiset.py
```diff
@@ -94,8 +95,21 @@ def create(self, request):
...
-        dandiset.save()
+        if 'identifier' in serializer.validated_data['metadata']:
+            identifier = serializer.validated_data['metadata']['identifier']
+            if identifier.startswith('DANDI:'):
+                identifier = identifier[6:]
+            dandiset = Dandiset(id=int(identifier))
+        else:
+            dandiset = Dandiset()
+        try:
+            dandiset.save(force_insert=True)
```
AlmightyYakob, line 106: It'd be good to leave a comment explaining why force_insert is required.

dandiapi/api/tests/test_dandiset.py
```diff
@@ -139,6 +139,71 @@ def test_dandiset_rest_create(api_client, user):
...
+
+    response = api_client.post(
+        '/api/dandisets/',
+        {'name': name, 'metadata': metadata},
+        format='json',
+    )
+    assert response.status_code == 400
+    assert response.data == f'Dandiset {identifier} Already Exists'
+
+
```
AlmightyYakob, line 206: To go along with my other comment about specifying invalid identifiers, a test here for that is probably a good idea.

dandiapi/api/tests/test_dandiset.py
```diff
@@ -139,6 +139,71 @@ def test_dandiset_rest_create(api_client, user):
...
+    api_client.force_authenticate(user=user)
+    name = 'Test Dandiset'
+    identifier = '123456'
+    metadata = {'foo': 'bar', 'identifier': f'DANDI:{identifier}'}
+
+    response = api_client.post(
+        '/api/dandisets/',
+        {'name': name, 'metadata': metadata},
+        format='json',
+    )
```
AlmightyYakob, line 153: If we decide to change how `identifier` is specified, this (along with the other test) would probably become:

```
    identifier = '123456'
    metadata = {'foo': 'bar'}

    response = api_client.post(
        '/api/dandisets/',
        {'name': name, 'identifier': identifier, 'metadata': metadata},
        format='json',
    )
```

AlmightyYakob (changes requested):
Looks good, left some suggestions.

waxlamp (commented):
This looks good to me. I agree with the discussion points Jake raised in his review. When he's happy with this, I will be as well.

dandiapi/api/views/dandiset.py
yarikoptic: FWIW I think this feature is worthwhile to have only during migration although may be we could find some use later but it then must be restricted to admin only... As such I think may be it is not worth "promoting" it etc by placing to the top level.

dchiquito: Requiring `identifier` as a top level field would involve tweaking the `VersionMetadataSerializer` and changing the CLI upload code, which doesn't seem like a worthwhile investment of time for a temporary tool like this.

AlmightyYakob (approved):

PR #104, opened 2021-02-11
Staging area for uploaded blobs

### LINKED ISSUE #114: inclusion of path within object id is likely the one rendering upload impossible
As pin pointed in https://github.com/dandi/dandi-api/issues/98#issuecomment-779444764 as the likely cause for that particular 500. 
#104 (if merged) is likely to address this so I will adjust that PR description to close this issue.

dchiquito:
Uploads now are now stored in the bucket at `uploads/{uuid}`, then are copied to `blobs/{sha256}` after validation succeeds.

This requires copying blobs, which is an operation that is not provided by anything we currently have installed, hence `copy.py`. After the copy is completed, the original blob is deleted to avoid keeping duplicate copies of all files.

Multipart copy is used to avoid the 5GB limit on normal copy operations.

@yarikoptic The only API change is that it will no longer be necessary to specify `filename` when calling `POST /uploads/initialize/`. Nothing will break if you continue to provide `filename`, it is simply no longer necessary.

@yarikoptic edit: likely Closes #114

### DIFF
--- a/dandiapi/api/tasks.py
+++ b/dandiapi/api/tasks.py
@@ -48,7 +48,6 @@ def validate(validation_id: int) -> None:
         validation.error = None
         validation.save()
 
-        # TODO separate storages for Validations and Assets require a copy at this point
         asset_blob, created = AssetBlob.from_validation(validation)
         if created:
             asset_blob.save()
--- a/dandiapi/api/copy.py
+++ b/dandiapi/api/copy.py
@@ -12,6 +12,9 @@ except ImportError:
     MinioStorage = type('FakeMinioStorage', (), {})
 
 
+PART_SIZE = 500 * 1024 * 1024  # 500 MB
+
+
 def copy_object(validation: Validation, dest_key: str):
     storage = Validation.blob.field.storage
     source_bucket = storage.bucket_name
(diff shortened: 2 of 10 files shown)


dandiapi/api/tasks.py
```diff
@@ -48,7 +48,6 @@ def validate(validation_id: int) -> None:
         validation.error = None
         validation.save()
 
-        # TODO separate storages for Validations and Assets require a copy at this point
         asset_blob, created = AssetBlob.from_validation(validation)
         if created:
             asset_blob.save()
```
yarikoptic, line 53: below, if validation fails (but not internal error) I think we should just kill that upload right here?  or do you see a use case where user/client can do something about validation which would make it pass later on?

yarikoptic: just to make sure -- is this "validation of a checksum" or does it include "validation of metadata" as provided by the client (or extracted?) as well?

dchiquito: I think that's a question that can be answered later when we get around to solving garbage collection more generally.

yarikoptic: note: that was the comment to the initial questions but I was too full of them so some are yet to be answered ;-)

dchiquito: Right now, validation only consists of verifying that the given checksum is correct. We will eventually add the CLI as a dependency and use it to validate the asset metadata as well. If either of those fails, the validation saves the error message so the client can report on what went wrong. I think we're all on the same page there.

Right now, if validation fails, the uploaded blob is basically orphaned. Even if the user wants to upload the exact same data and try again, the system will force them to upload all the data again. This will leave the data in S3 in case it might be useful for debugging, rather than cleaning it up aggressively as I think you are suggesting. 

Leaving the upload in place also forces us to make a decision later about how best to garbage collect it. It's an important thing to solve, but not urgent, so I deliberately left it for later.

Does that answer all the questions?

yarikoptic: ;-) yes

>  We will eventually add the CLI as a dependency and use it to validate the asset metadata as well

we should do that at the stage of initiating the upload (based on metadata provided), validation of the blob should indeed just verify checksum only -- it would be a boomer if lengthy upload "fails" if some metadata apparently is not kosher according to server side decision making.  I think we are doomed to trust client to provide "untainted" metadata upon initiation of the upload.

dchiquito: I made #117 to address this.

That would be doable, but I don't think it's a blocker for this feature. For now, I think we can trust our users.

PR #96, opened 2021-02-09
Error when path in asset metadata conflicts with asset path

jwodder:
Closes #95.

PR #105, opened 2021-02-12
Populate asset path from metadata

### LINKED ISSUE #95: Server should error when asset assigned metadata with wrong path
When a `POST /dandisets/{dandiset_id}/versions/{version_id}/assets/` request is made and the `metadata` field in the request body contains a `path` field whose value differs from the `path` field at the top level of the request body, the server should respond with an error (I believe 400 is the most appropriate status).

See https://github.com/dandi/dandi-cli/issues/368#issue-804701654.

dchiquito:
Fixes #95 

@jwodder This removes `path` from the asset create/update request body and pulls it from the metadata instead. The CLI should simply stop providing `path` in the request body. Can you confirm that this will resolve #96 to your satisfaction?

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -84,6 +84,24 @@ class Asset(TimeStampedModel):
     def sha256(self):
         return self.blob.sha256
 
+    def _populate_metadata(self):
+        new: AssetMetadata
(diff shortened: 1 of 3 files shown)


jwodder (approved):

yarikoptic:
I will wait for @AlmightyYakob to review and/or click Merge ;)

PR #106, opened 2021-02-12
Upgrade to cookiecutter v0.17

brianhelba:


setup.py
```diff
@@ -40,23 +40,30 @@
...
         'httpx',
         # Production-only
+        'django-composed-configuration[prod]',
+        'django-s3-file-field[boto3]',
         'django-storages[boto3]',
         'gunicorn',
-        # Development-only
-        'django-debug-toolbar',
+        # Development-only, but required
         'django-minio-storage',
```
brianhelba, line 58: `django-minio-storage` is directly referenced in local project code, but I really don't think it needs to be. The current content of `storage.py` can probably be cut, especially once `django` is upgraded.

@dchiquito We should talk more about this.

PR #111, opened 2021-02-12
Replace DeconstructableFileField with upstream FileFile

brianhelba:
Now that Django 3.1.2 is released, this class is no longer necessary, as its bugfixes are available upstream.

No migration is necessary, as the updated `FileField` is identical to `DeconstructableFileField`.


dchiquito (approved):
I look forward to seeing FieldFields

PR #120, opened 2021-02-16
Asset size field

dchiquito:
Fixes https://github.com/dandi/dandiarchive/issues/570

Store the size of AssetBlobs rather than using `self.blob.size`, which queries S3.


dandiapi/api/models/asset.py
```diff
@@ -34,14 +34,11 @@ class AssetBlob(TimeStampedModel):
         blank=True, storage=_get_asset_blob_storage, upload_to=_get_asset_blob_prefix
     )
     sha256 = models.CharField(max_length=64, validators=[RegexValidator(f'^{SHA256_REGEX}$')])
+    size = models.BigIntegerField()
```
brianhelba, line 37:
```suggestion
    size = models.PositiveBigIntegerField()
```

This does require Django 3.1: https://docs.djangoproject.com/en/3.1/ref/models/fields/#positivebigintegerfield

brianhelba (approved):
One note about the field type. Otherwise LGTM.

dandiapi/api/models/asset.py
dchiquito: Sadly this was added to `django-stubs` in November, but the last release was in October. There will be some editor confusion until that makes it in.

PR #131, opened 2021-02-24
Simplify search

dchiquito:
Turns out I was really vastly overthinking it, postgres supports `to_tsvector` on `JSONB` columns exactly like text columns, so we can actually just use the built in DRF SearchFilter to provide a `search` query parameter that works on a whole ViewSet.


dandiapi/api/tests/test_dandiset.py
```diff
@@ -376,3 +376,24 @@ def test_dandiset_rest_add_malformed(api_client, dandiset, user):
...
+
+@pytest.mark.django_db
+def test_dandiset_rest_search_identifier(api_client, version):
+    results = api_client.get('/api/dandisets/', {'search': version.dandiset.identifier}).data[
+        'results'
+    ]
+    assert len(results) == 1
+    assert results[0]['identifier'] == version.dandiset.identifier
+    assert results[0]['most_recent_version']['version'] == version.version
+    assert results[0]['most_recent_version']['name'] == version.name
```
AlmightyYakob, line 399: It might be good to test search on other metadata too, (like something that's only represented in metadata and not at a higher level) but it's not really a huge deal, I'm okay with this as is.

AlmightyYakob (approved):
Looks good, tested locally :+1:

PR #143, opened 2021-03-02
Asset rework

dchiquito:
This PR will fundamentally change how assets are related to versions, which will result in data loss for any existing assets. Rather than try to reconcile the differences, I recommend just wiping the data from production (as we planned to do anyway) and start from scratch.


yarikoptic:
FWIW: +1 on "wipe out and start from scratch"

May be while at it we should also relocate to use a prefix on our public bucket, instead of the private one.  

What is the current "structure" within the private bucket? (can't `ls` it with some dandi s3 credentials I have)  I guess smth like
- `blobs/`
- `dandisets/` (ref: https://github.com/dandi/dandi-api/issues/139#issuecomment-787520447)

I wonder if to future-proof/simplify migrations we should add an additional prefix (e.g. `dandiarchive-v1/`) or a suffix  (e.g. `blobs-v1/`)?  @satra - may be it is time to try out that S3 AMI policy spec to provide permissions for this new location (only) while keeping girder safe?

mgrauer:
@dchiquito Could you summarize what the changes for how assets relate to versions?

dchiquito:
@yarikoptic Right now the bucket has a `uploads` folder for uploads and a `blobs` folder for validated blobs.
@mgrauer 
* Versions <-> Assets is now a many to many relationship. Each version can have many assets, and each asset can belong to many versions.
* `POST .../assets/` creates a new `Asset`, exactly the same as it used to.
* `PUT .../assets/{uuid}/` creates a new `Asset` with the given data, mints a new UUID for it, then replaces the existing asset identified by the URL parameter `uuid` in the context of the given version. The old `Asset` still exists and other versions can still point to it.
* `DELETE .../assets/{uuid}/` will drop an `Asset` from the DB, which I am now realizing is not correct. It should instead remove the asset from the given version, leaving the `Asset` in place in case other versions refer to it.
* `POST .../versions/draft/publish` will create a new `Version`, then iterate over all `Asset`s in the old version and add them to the new version as well.

yarikoptic:
> @yarikoptic Right now the bucket has a `uploads` folder for uploads and a `blobs` folder for validated blobs.

thanks. adjusted my [description](https://github.com/dandi/dandi-api/pull/143#issuecomment-789210538)

yarikoptic:
FWIW, with the fix to `DELETE` description sounds correct to me!  Obviously we will need some `gc` later on assets which aren't in any version (as well as blobs which aren't in any asset).

satra:
@dchiquito - thank you for this PR.

@yarikoptic - since `dandiarchive` is the name of the bucket. we could indeed have a prefix such as `/dandiarchive/v1/[blobs,dandisets]` . perhaps it just acts as a namespace. 

possible `/dandiarchive/uploads` could be it's own prefix since that is ephemeral and can be removed, and we may in the future consider uploading directly into `blobs`

waxlamp (approved):
This all looks good to me.

The one design comment I have is a fairly pedantic point of RESTful design: the `PUT` endpoint that copies an asset to a new UUID seems like it should actually be a `POST` with a `/copy` suffix, so something like `POST .../assets/{uuid}/copy`. I mention this because `PUT` generally has semantics of replacing the named resource in place.

dchiquito:
@waxlamp I see your point, but I think the line is sort of fuzzy there. Technically we are updating an asset in a dandiset in place, but in the process of updating, the asset gets a new UUID. From an internal perspective, the `Asset` object was copied, but from the API perspective, it's still the same asset, just with some tweaked data and a new name tag.

waxlamp:
> @waxlamp I see your point, but I think the line is sort of fuzzy there. Technically we are updating an asset in a dandiset in place, but in the process of updating, the asset gets a new UUID. From an internal perspective, the `Asset` object was copied, but from the API perspective, it's still the same asset, just with some tweaked data and a new name tag.

I guess for me, an asset that is mostly the same, but with tweaked data, really is a new asset, as evidenced by the fact that it does in fact get a new name tag (i.e., a new UUID). But I'm not interested in defending my position here too violently. It's something to think about, and perhaps to change in the future. Carry on.

PR #150, opened 2021-03-09
Doc design blob UUID & upload process

yarikoptic:
Extract from https://github.com/dandi/dandi-api/issues/146#issuecomment-793114637 with only minimal editing and placing "Check" first without changing any substance for now.  Re [@satra 's comment](https://github.com/dandi/dandi-api/issues/146#issuecomment-794256248): kept numbering as is to not fight markdown renderers which might decide differently while rendering separated numbered lists.

TODOs as I see it
- [x] establish parallels to current API endpoints/procedure
- [x] refine following the feedback

### DIFF
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -14,7 +14,7 @@ A blob is referenced by a key generation function. the value of this key is used
     2. If it exists, return the `blobs:blob_id`.
     3. It it doesn't exist, return 404.
 2. If 404, start upload process.
-3. if key, call `POST .../assets/` with the key to register the new asset.
+3. if `blob_id`, call `POST .../assets/` with the `blob_id` to register the new asset.
 
 ### Blob upload process:
 1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -19,7 +19,7 @@ A blob is referenced by a key generation function. the value of this key is used
 4. API saves initial data (size, ETAG, key) to Upload table
 3. API returns presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
-5. CLI uploads to presigned URL and checks part ETags on return. Any failures can be retried without involving API.
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. CLI sends part info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
 8. API validates the size and that the initially reported ETag matches the actual ETag
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -32,7 +32,7 @@ A blob is referenced by a key generation function. the value of this key is used
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
         2. there is collision - deletes `blobs:key` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
     4. kicks off background process to calculate checksums
-9. CLI calls `POST .../assets/` with the key to register the new asset.
+9. CLI calls `POST .../assets/` with the `blob_id` to register the new asset.
 
 ## Garbage Collection process
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -18,7 +18,7 @@ A blob is referenced by a key generation function. the value of this key is used
 
 ### Blob upload process:
 1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
-   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:key` in the header (`Location`?).
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:blob_id` in the header (`Location`?).
    2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
    3. saves initial data (size, ETag, key) to Upload table
    4. API returns `uploads:upload_id` and presigned URLs
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -26,7 +26,7 @@ A blob is referenced by a key generation function. the value of this key is used
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
-8. CLI `POST /uploads/{upload_id}/validate/` <-- `{}` --> `blobs:key`. API
+8. CLI `POST /uploads/{upload_id}/validate/` <-- `{}` --> `blobs:blob_id`. API
     1. validates the size and that the initially reported ETag matches the actual ETag
     2. checks for collision, since some other task could have finished by this time with the same object
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -28,6 +28,7 @@ A blob is referenced by a key generation function. the value of this key is used
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
         2. there is collision - delete key record and file in blob store, will return key of an existing record
     4. kicks off background process to calculate checksums
+9. CLI calls `POST .../assets/` with the key to register the new asset.
 
 ## Garbage Collection process
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -11,7 +11,7 @@ A blob is referenced by a key generation function. the value of this key is used
 ### Check process (for de-duplication):
 1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-etag", "value": "123etag123"}`. API:
     1. Looks for an existing AssetBlob with the given digest(s).
-    2. If it exists, return the key (UUID).
+    2. If it exists, return the `blobs:blob_id`.
     3. It it doesn't exist, return 404.
 2. If 404, start upload process.
 3. if key, call `POST .../assets/` with the key to register the new asset.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -20,7 +20,7 @@ A blob is referenced by a key generation function. the value of this key is used
 3. API returns presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
-6. CLI sends part info to API, API responds with presigned completion URL
+6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
 8. API validates the size and that the initially reported ETag matches the actual ETag
 9. Check for collision, since some other task could have finished by this time with the same object
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -26,7 +26,7 @@ A blob is referenced by a key generation function. the value of this key is used
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
-8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:key`. API
+8. CLI `POST /uploads/{upload_id}/validate/` <-- `{}` --> `blobs:key`. API
     1. validates the size and that the initially reported ETag matches the actual ETag
     2. checks for collision, since some other task could have finished by this time with the same object
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -9,7 +9,7 @@ A blob is referenced by a key generation function. the value of this key is used
 ## Upload (from a CLI)
 
 ### Check process (for de-duplication):
-1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-s3-etag", "value": "123etag123"}`. API:
+1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-etag", "value": "123etag123"}`. API:
     1. Looks for an existing AssetBlob with the given digest(s).
     2. If it exists, return the key (UUID).
     3. It it doesn't exist, return 404.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -24,9 +24,9 @@ A blob is referenced by a key generation function. the value of this key is used
    4. API returns `uploads:upload_id` and presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
-6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
-8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
+8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:key`. API
     1. validates the size and that the initially reported ETag matches the actual ETag
     2. checks for collision, since some other task could have finished by this time with the same object
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -9,7 +9,7 @@ A blob is referenced by a key generation function. the value of this key is used
 ## Upload (from a CLI)
 
 ### Check process (for de-duplication):
-1. CLI `GET /blobs/` <-- expected digest with digest type (e.g., ETag). API:
+1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-s3-etag", "value": "123etag123"}`. API:
     1. Looks for an existing AssetBlob with the given digest(s).
     2. If it exists, return the key (UUID).
     3. It it doesn't exist, return 404.
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -14,10 +14,10 @@ A blob is referenced by a key generation function. the value of this key is used
 3. if exists, returns key, if not returns an error.
 
 ### Blob upload process:
-1. CLI initiates upload sending contentSize + ETag
-2. API generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
-4. API saves initial data (size, ETAG, key) to Upload table
-3. API returns presigned URLs
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
+   1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   2. saves initial data (size, ETag, key) to Upload table
+   3. API returns presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -36,9 +36,8 @@ A blob is referenced by a key generation function. the value of this key is used
 
 1. Receives key from API and queues up a job to process the blob
 2. performs a set of tasks on blob including multi checksum computation.
-3. adds checksums to object metadata
-4. pings api that checksums are ready
-5. API adds checksums to row in blob db
+3. adds checksums to the AssetBlob row in the DB (initially they are null or something)
+4. API injects checksums into metadata whenever it is requested: initially it won't be there while being calculated, but will magically appear after calculations finish.
 
 References:
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -1,4 +1,4 @@
-A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>`
+A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
 
 ** database table of identities **: indexed by the key and contains columns for as many types of checksums as we want. these checksums constitute identities of the blob and should use functions that will always return the same value.
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -18,9 +18,10 @@ A blob is referenced by a key generation function. the value of this key is used
 
 ### Blob upload process:
 1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
-   1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
-   2. saves initial data (size, ETag, key) to Upload table
-   3. API returns `uploads:upload_id` and presigned URLs
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT.
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -20,7 +20,7 @@ A blob is referenced by a key generation function. the value of this key is used
 1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
    1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
    2. saves initial data (size, ETag, key) to Upload table
-   3. API returns presigned URLs
+   3. API returns `uploads:upload_id` and presigned URLs
 4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -22,10 +22,12 @@ A blob is referenced by a key generation function. the value of this key is used
 5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
 6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
-8. API validates the size and that the initially reported ETag matches the actual ETag
-9. Check for collision, since some other task could have finished by this time with the same object
-10. API adds key + ETag to AssetBlob table, delete from Upload table
-11. API kicks off background process to calculate checksums
+8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
+    1. validates the size and that the initially reported ETag matches the actual ETag
+    2. checks for collision, since some other task could have finished by this time with the same object
+        1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
+        2. there is collision - delete key record and file in blob store, will return key of an existing record
+    4. kicks off background process to calculate checksums
 
 ## Garbage Collection process
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -30,7 +30,7 @@ A blob is referenced by a key generation function. the value of this key is used
     1. validates the size and that the initially reported ETag matches the actual ETag
     2. checks for collision, since some other task could have finished by this time with the same object
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
-        2. there is collision - deletes `blobs:key` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
+        2. there is collision - deletes `blobs:blob_id` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
     4. kicks off background process to calculate checksums
 9. CLI calls `POST .../assets/` with the `blob_id` to register the new asset.
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -26,7 +26,7 @@ A blob is referenced by a key generation function. the value of this key is used
     1. validates the size and that the initially reported ETag matches the actual ETag
     2. checks for collision, since some other task could have finished by this time with the same object
         1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
-        2. there is collision - delete key record and file in blob store, will return key of an existing record
+        2. there is collision - delete key record and file in uploads, will return key of an existing record
     4. kicks off background process to calculate checksums
 9. CLI calls `POST .../assets/` with the key to register the new asset.
 
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -2,35 +2,39 @@ A blob is referenced by a key generation function. the value of this key is used
 
 ** database table of identities **: indexed by the key and contains columns for as many types of checksums as we want. these checksums constitute identities of the blob and should use functions that will always return the same value.
 
-**key generation function**: the current suggestion (from satra) for this function is UUID4. 
+**key generation function**: the current suggestion (from satra) for this function is UUID4.
 
 **DANDI ETag generation function**: this is an agreement between API server and DANDI CLI for now to ensure there is an algorithm that consistently returns the correct number of parts given a `contentSize` in bytes. this should in theory also be the AWS ETag generation function.
 
-**upload process**:
+## Upload (from a CLI)
+
+### Check process (for de-duplication):
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
 1. CLI initiates upload sending contentSize
 2. generate key
 3. API returns presigned URLs
 4. CLI verifies that the number of URLs match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
 5. CLI uploads to presigned URL and checks ETag on return
 6. Completes upload and contacts API
-7. API validates size and ensures object prefix matches computed ETag. 
+7. API validates size and ensures object prefix matches computed ETag.
 8. Check for collision, since some other task could have finished by this time with the same object.
 9. API moves the object into `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>` where key is returned by the key generation function. API returns this key to CLI.
 10. API adds key + ETag into db.
 11. API sends new key to trusted background process
 
-**check process**
-1. CLI sends expected digest with digest type (e.g., ETag)
-2. API checks for existence of ETag in identity table. 
-3. if exists, returns key, if not returns an error.
+## Garbage Collection process
 
-** garbage collection process **
 1. if dangling key's are found, API does garbage collection on keys that are older than 24 hours.
 2. sends background process info about deleted keys
 
-** out of band background process **
+## Out of band background process
+
 1. Receives key from API and queues up a job to process the blob
-2. performs a set of tasks on blob including multi checksum computation. 
+2. performs a set of tasks on blob including multi checksum computation.
 3. adds checksums to object metadata
 4. pings api that checksums are ready
 5. API adds checksums to row in blob db
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -24,7 +24,7 @@ A blob is referenced by a key generation function. the value of this key is used
 7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
 8. API validates the size and that the initially reported ETag matches the actual ETag
 9. Check for collision, since some other task could have finished by this time with the same object
-10. API adds key + ETag to Asset table, delete from Upload table
+10. API adds key + ETag to AssetBlob table, delete from Upload table
 11. API kicks off background process to calculate checksums
 
 ## Garbage Collection process
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -1,0 +1,40 @@
+A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>`
+
+** database table of identities **: indexed by the key and contains columns for as many types of checksums as we want. these checksums constitute identities of the blob and should use functions that will always return the same value.
+
+**key generation function**: the current suggestion (from satra) for this function is UUID4. 
+
+**DANDI ETag generation function**: this is an agreement between API server and DANDI CLI for now to ensure there is an algorithm that consistently returns the correct number of parts given a `contentSize` in bytes. this should in theory also be the AWS ETag generation function.
+
+**upload process**:
+1. CLI initiates upload sending contentSize
+2. generate key
+3. API returns presigned URLs
+4. CLI verifies that the number of URLs match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URL and checks ETag on return
+6. Completes upload and contacts API
+7. API validates size and ensures object prefix matches computed ETag. 
+8. Check for collision, since some other task could have finished by this time with the same object.
+9. API moves the object into `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>` where key is returned by the key generation function. API returns this key to CLI.
+10. API adds key + ETag into db.
+11. API sends new key to trusted background process
+
+**check process**
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table. 
+3. if exists, returns key, if not returns an error.
+
+** garbage collection process **
+1. if dangling key's are found, API does garbage collection on keys that are older than 24 hours.
+2. sends background process info about deleted keys
+
+** out of band background process **
+1. Receives key from API and queues up a job to process the blob
+2. performs a set of tasks on blob including multi checksum computation. 
+3. adds checksums to object metadata
+4. pings api that checksums are ready
+5. API adds checksums to row in blob db
+
+References:
+
+- original issue/discussion: https://github.com/dandi/dandi-api/issues/146#issuecomment-793114637
--- a/doc/design/blob-uuid-1.md
+++ b/doc/design/blob-uuid-1.md
@@ -9,9 +9,12 @@ A blob is referenced by a key generation function. the value of this key is used
 ## Upload (from a CLI)
 
 ### Check process (for de-duplication):
-1. CLI sends expected digest with digest type (e.g., ETag)
-2. API checks for existence of ETag in identity table.
-3. if exists, returns key, if not returns an error.
+1. CLI `GET /blobs/` <-- expected digest with digest type (e.g., ETag). API:
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
 
 ### Blob upload process:
 1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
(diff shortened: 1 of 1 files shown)


dchiquito (commented):
After some experimentation it seems that ETag of the final object is not calculable in practice. There is some 3rd party documentation that claims that the ETag of a multipart upload is simply the MD5 of the concatenation of the MD5s of the parts, with a suffix indicating how many parts were concatenated. However, the ETags of uploads that we did from the CLI do not adhere to this format. S3 does not make any promises as to the final ETag of a multipart upload, and apparently Amazon is exercising that right here.

For what it's worth, the ETags of each part in the multipart upload do seem to be the MD5 of the content, so we can still verify that each part was uploaded correctly. Presumably we can trust S3 to stitch parts together correctly, so I think that will be sufficient for integrity checking.

Since the ETag cannot be determined by clients, I recommend that the CLI calculate the SHA256 or MD5 of the entire blob as it does currently and provide that to the server, which for now will simply trust that it is correct. We can add a validation step that verifies the provided checksums at a later date.

satra:
@dchiquito - did you check the uploaded completed ETag in the staging area? or was the check after the move to blob storage?  it's possible that the ETag changes after move, where amazon doesn't guarantee the same chunks for moving.

yarikoptic:
damn you @satra -- where have you been all this time?!  I bet you nailed it and it is due to the move from staging into the blob store -- how could I forget that there is this part of the dance... damn, I feel old

yarikoptic:
Confirming!  With additional debugging output to dandi-cli (yet to PR) I see :
```
drogon:~/.cache/dandi-cli/log
*$> grep -e Uploading.*parts -e 'completed.*ETag' 20210310134557Z-31993.log
grep: warning: GREP_OPTIONS is deprecated; please use an alias or script
2021-03-10T08:47:17-0500 [DEBUG   ] dandi 31993:139725637817728 Uploading sub-YutaMouse51_ses-YutaMouse51-160515_behavior+ecephys.nwb in 260 part
s
2021-03-10T08:57:58-0500 [DEBUG   ] dandi 31993:139725637817728 sub-YutaMouse51/sub-YutaMouse51_ses-YutaMouse51-160515_behavior+ecephys.nwb: Uplo
ad completed. Response content: b'<?xml version="1.0" encoding="UTF-8"?>\n\n<CompleteMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/200
6-03-01/"><Location>https://dandi-api-dandisets-testing.s3.amazonaws.com/uploads%2F73b6b478-5fa4-4afa-8997-c48326c38053</Location><Bucket>dandi-a
pi-dandisets-testing</Bucket><Key>uploads/73b6b478-5fa4-4afa-8997-c48326c38053</Key><ETag>&quot;ecd3b91111edec976726aeb46b79ad7e-260&quot;</ETag>
</CompleteMultipartUploadResult>'
```
so final ETag `ecd3b91111edec976726aeb46b79ad7e-260` corresponds to the originally planned 260 parts upload. And with the script from https://github.com/dandi/dandi-api/issues/146  I get
```
$> python3 /tmp/aws-etag.py sub-YutaMouse51/sub-YutaMouse51_ses-YutaMouse51-160515_behavior+ecephys.nwb
ETag: ecd3b91111edec976726aeb46b79ad7e-260
```
- a perfect match.  So it is indeed just a matter of api server using larger parts while moving from staging area to blobstore.  Mystery resolved -- thanks @satra for pointing obvious ;)

dchiquito:
https://github.com/dandi/dandi-api/pull/150/commits/21fe5e50e051f39f1e96c9f683b73290430e52e4 I tweaked the API behavior slightly. I think this is implementable.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,45 @@
...
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
+1. CLI initiates upload sending contentSize + ETag
+2. API generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+4. API saves initial data (size, ETAG, key) to Upload table
+3. API returns presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URL and checks part ETags on return. Any failures can be retried without involving API.
```
yarikoptic, line 22:
```suggestion
5. CLI uploads to presigned URLs, and for each part provides Content-MD5 header field  and checks ETag on return. Any part upload failure can be retried without involving API.
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,45 @@
...
+6. CLI sends part info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. API validates the size and that the initially reported ETag matches the actual ETag
+9. Check for collision, since some other task could have finished by this time with the same object
+10. API adds key + ETag to Asset table, delete from Upload table
```
yarikoptic, line 27: in 8.-10. results in a return a record with blob UUID (either a new one or "reused"), right?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Upload (from a CLI)
+
+### Check process (for de-duplication):
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
+1. CLI initiates upload sending contentSize + ETag
```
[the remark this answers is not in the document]
yarikoptic: currently we do
```
            resp = self.post(
                "/uploads/initialize/",
                json={
                    "file_name": f"{dandiset_id}/{version_id}/{asset_path}",
                    "file_size": total_size,
                },
            )
```
so already conflating "pure blob" upload with an asset upload (has a file_name which could be different across blobs).  So I wonder if uploads should just be initiated from  `dandiset/.../versions/draft` end point?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,45 @@
...
+4. API saves initial data (size, ETAG, key) to Upload table
+3. API returns presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URL and checks part ETags on return. Any failures can be retried without involving API.
+6. CLI sends part info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. API validates the size and that the initially reported ETag matches the actual ETag
+9. Check for collision, since some other task could have finished by this time with the same object
+10. API adds key + ETag to Asset table, delete from Upload table
+11. API kicks off background process to calculate checksums
```
yarikoptic, line 28: a side-note that if we stay true to current design -- addition of any new checksum should change asset UUID unless we keep those outside of the metadata records and treat them as some "dynamic" overlay ...

doc/design/blob-uuid-1.md
dchiquito: The API server needs to specify all headers when presigning the request, so to provide the `Content-MD5` header instead of checking the resulting ETag would require the CLI to pass all of the part MD5s to the API when initializing the upload.

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
satra: that is weird. i can see filename without all those other things being useful, since that can be put on the s3 object metadata for content-disposition, but what is the full file_name being used for in the upload process (dandiset/version seems completely unnecessary)

doc/design/blob-uuid-1.md
dchiquito: My thought was that ETag would be a good "identity" key for blobs, since all blobs necessarily have an ETag, and ETag has to be used to check if a blob already exists. We could use UUID instead, but it might get confusing managing blob UUIDs and asset UUIDs.

doc/design/blob-uuid-1.md
satra: that is correct. i would suggest that this step only involve the CLI doing the check without sending content-MD5 to aws.

doc/design/blob-uuid-1.md
satra: i would think so.

doc/design/blob-uuid-1.md
dchiquito: I think dynamically injecting the checksums like we already do with asset path info would be better. I can imagine someone uploading some new data to a dandiset, checking that it's there in the UI, then publishing it immediately before the background checksum calculations complete.

doc/design/blob-uuid-1.md
satra: @dchiquito - what's the Asset Table here? there is no asset in this upload process.

doc/design/blob-uuid-1.md
satra: any new checksum is an overlay on the uuid - should not change uuid of a blob.

doc/design/blob-uuid-1.md
yarikoptic: Right, forgot about that, adjusted my suggestion

doc/design/blob-uuid-1.md
dchiquito: My mistake, I mean AssetBlob. At this point the data is in S3 and is known to be valid, so it needs to be canonicalized in the DB.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,44 @@
...
+2. API generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+4. API saves initial data (size, ETAG, key) to Upload table
+3. API returns presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. CLI sends part info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. API validates the size and that the initially reported ETag matches the actual ETag
+9. Check for collision, since some other task could have finished by this time with the same object
+10. API adds key + ETag to Asset table, delete from Upload table
```
dchiquito, line 27:
```suggestion
10. API adds key + ETag to AssetBlob table, delete from Upload table
```

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
dchiquito: `file_name` is actually unused by the API, it was required at some point but has been removed. Apparently it was never removed from the CLI, and having extra JSON fields doesn't cause any problems, the data is simply ignored. That line can be safely removed, it is meaningless.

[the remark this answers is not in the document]
dchiquito: I think planning for different upload pipelines at this point is not productive. For now we have a well defined problem, uploading a file from a client machine using the CLI. If we want to support different ways to ingest files in the future, we can build a different upload flow at that time.

My understanding is that to get to release faster, we were going to just trust the CLI that content is valid for now. Handling metadata and validating during initialization would make things substantially more complicated. Is this a migration blocker, or can we leave it out for now @yarikoptic ?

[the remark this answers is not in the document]
satra: i think @yarikoptic - simply meant sending the info in the json header in that post (and not really full metadata validation). but i can also see sending contentsize/digest info in the call to finalize upload, since that's where validation would happen. 

and i do agree that making that json match with the subset of assetmetadata keys would probably be a good thing, but not essential at this point.

so the payload on `/uploads/validate` the upload would be something like this.

```json
{ 
  "identifier": uuid,
  "contentSize":  12345,
  "digest":  {"dandi:dandi-s3-etag": "aadd-123"}
}
```

this formulation still needs a change in the cli.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,44 @@
+A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>`
+
+** database table of identities **: indexed by the key and contains columns for as many types of checksums as we want. these checksums constitute identities of the blob and should use functions that will always return the same value.
```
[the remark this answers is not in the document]
satra: @dchiquito - shouldn't this be indexed by the UUID with the ETag being one of the identities? only because two ETag's could collide (in theory, the probably for that is higher than the probability for UUID4s to collide on the same system)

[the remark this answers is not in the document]
yarikoptic: if we are to have dedicated columns for checksums, I thought they all could just go into "blobs" table with an index, UUID (unless UUID to be used as index, but I thought it is inferior), ETag (which we might want to name 'dandietag' or alike to make more explicit), sha256 columns to start with, i.e. that there would be no need in another table with 1-to-1 relationship to blobs table.  or what am I missing?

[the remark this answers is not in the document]
satra: i'm fine with `index, uuid, dandietag, sha256, blake3, <future checksums>` whether the index is built in or uuid, i don't care a whole lot. but there may be performance penalties for search, which would likely be given a checksum, find uuid, or given a uuid, find checksums.

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
dchiquito: I think `identifier` not be included, or at least be optional:
```
{ 
  "contentSize":  12345,
  "digest":  {"dandi:dandi-s3-etag": "aadd-123"}
}
```
When the CLI needs to upload file, the first step is to calculate some digest (in this case ETag) and ask the API using this endpoint if it already exists; there's no way for the CLI to know the UUID.
Since we already need to support that use case, I don't see what specifying `identifier` gains us.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,44 @@
...
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
+1. CLI initiates upload sending contentSize + ETag
+2. API generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+4. API saves initial data (size, ETAG, key) to Upload table
+3. API returns presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. CLI sends part info to API, API responds with presigned completion URL
```
yarikoptic, line 23:
```suggestion
6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,44 @@
...
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. API validates the size and that the initially reported ETag matches the actual ETag
+9. Check for collision, since some other task could have finished by this time with the same object
+10. API adds key + ETag to AssetBlob table, delete from Upload table
+11. API kicks off background process to calculate checksums
```
yarikoptic, line 28:
```suggestion
8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blob:uuid`. API
    1. validates the size and that the initially reported ETag matches the actual ETag
    2. checks for collision, since some other task could have finished by this time with the same object
    3. adds key + ETag to AssetBlob table, delete from Upload table
    4. kicks off background process to calculate checksums
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,44 @@
...
+### Check process (for de-duplication):
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
+1. CLI initiates upload sending contentSize + ETag
+2. API generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+4. API saves initial data (size, ETAG, key) to Upload table
+3. API returns presigned URLs
```
yarikoptic, line 20:
```suggestion
1. CLI `/uploads/initiate` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
   1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
   2. saves initial data (size, ETAG, key) to Upload table
   3. API returns presigned URLs
```

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
yarikoptic: FWIW re `file_path` and content-disposition: I would stand that blobs have no filename (IMHO UUID is better than a wrong/misleading/disclosing one from the first ever uploaded copy of the blob):
- Pre-signed urls produced from `*/assets/{uuid}/download` endpoint can have appropriate for the asset content-disposition
- git-annex doesn't care
- "published" manifests would have proper paths and thus filenames for each of the blobs

So indeed, no need to provide file path or name anywhere here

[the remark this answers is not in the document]
satra: > Pre-signed urls produced from */assets/{uuid}/download endpoint can have appropriate for the asset content-disposition

has this been tested. let me add it as an issue.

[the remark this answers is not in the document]
dchiquito: @yarikoptic I agree with returning the blob ID only from `/uploads/validation/`, but I don't see the need for an intermediate `upload-id`. To support deduplication, `/uploads/validate/` needs to accept a request body like:
```
{ 
  "contentSize":  12345,
  "digest":  {"dandi:dandi-s3-etag": "aadd-123"}
}
```
Since the ETag is already there (and is already being relied on to uniquely identify blobs), we can just look up the upload using that. I don't see what including an `upload-id` gains us.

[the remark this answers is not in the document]
satra: what if two people are uploading the same blob?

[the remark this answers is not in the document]
satra: would you randomly choose the first uuid?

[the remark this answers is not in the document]
dchiquito: To have a collision, the second person would need to call `/uploads/initialize/` with the same ETag as the first person. This request should fail because there is already an upload.

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
dchiquito: How about this:
```suggestion
8. CLI `/uploads/validate` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API
    1. Checks for an existing AssetBlob. If it exists, no need to validate: return the key (UUID).
    2. Looks up the Upload from the DB with the given ETag. If not present, return 404.
    3. Verify that the object described by the Upload exists, is the right size, and has the right ETag. If not, return 400.
    4. Adds key + ETag + size to AssetBlob table and delete the Upload.
    4. kicks off background process to calculate checksums
    5. Return key (UUID) of AssetBlob so CLI can do `POST .../assets/` to mint new asset
```
The big difference is that the ETag is used to lookup the Upload, and there is some more explicit checking for existing info in the DB.

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
satra: i thought we allowed for parallel uploads of the same content (since we are using UUID, and the ETag check does not happen on the upload table) - which if we check with ETag during upload stage while nice, would make the person A wait for person B to finish uploading. this would also ensure that the fastest uploaded blob gains precedence.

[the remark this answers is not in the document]
yarikoptic: > I don't see what including an upload-id gains us

- a clear identification of the upload in question, and avoidance of matching by metadata provided in `/upload/initialize/` and thus avoiding problems with duplicate uploads as @satra has mentioned - we can just allow them to happen and the first one to complete would be the one the `key` of which would be used for that blob
- anything about upload (cancel etc) should be done via its `upload_id` instead of matching uploads and possibly canceling some other parallel one (ATM we do not have `DELETE` on uploads but I guess we should get one eventually)
That is why below I would prefer [my](https://github.com/dandi/dandi-api/pull/150#discussion_r591868008) over [@dchiquito ](https://github.com/dandi/dandi-api/pull/150#discussion_r591902088) proposal below.

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
yarikoptic: > Checks for an existing AssetBlob. If it exists, no need to validate 

if so -- still would need to remove an upload (so 2.) from uploads.  But matching could lead to upload of others:

> here is some more explicit checking for existing info in the DB.

how is it "more explicit" if it is not by an `upload_id`?  matching by metadata could match uploads of others

doc/design/blob-uuid-1.md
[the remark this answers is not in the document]
dchiquito: OK, I think at this point we're really overloading `/upload/initialize/`. How about breaking it into two endpoints:
 * `GET /blobs/` checks for an existing asset with an arbitrary digest (but ETag for now):
 ```
 { 
  "contentSize":  12345,
  "digest":  {"dandi:dandi-s3-etag": "aadd-123"}
}
```
* `POST /uploads/validate/{upload_id}/` starts validation for an upload, identified by upload ID.

File: doc/design/blob-uuid-1.md [hunk 10]
```diff
@@ -0,0 +1,46 @@
...
+**key generation function**: the current suggestion (from satra) for this function is UUID4.
+
+**DANDI ETag generation function**: this is an agreement between API server and DANDI CLI for now to ensure there is an algorithm that consistently returns the correct number of parts given a `contentSize` in bytes. this should in theory also be the AWS ETag generation function.
+
+## Upload (from a CLI)
+
+### Check process (for de-duplication):
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
```
dchiquito, line 14:
```suggestion
1. CLI `GET /blobs/` <-- expected digest with digest type (e.g., ETag). API:
    1. Looks for an existing AssetBlob with the given digest(s).
    2. If it exists, return the key (UUID).
    3. It it doesn't exist, return 404.
2. If 404, start upload process.
3. if key, call `POST .../assets/` with the key to register the new asset.
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,46 @@
...
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
+    1. validates the size and that the initially reported ETag matches the actual ETag
+    2. checks for collision, since some other task could have finished by this time with the same object
+        1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
+        2. there is collision - delete key record and file in blob store, will return key of an existing record
+    4. kicks off background process to calculate checksums
+
```
dchiquito, line 31:
```suggestion
    4. kicks off background process to calculate checksums
9. CLI calls `POST .../assets/` with the key to register the new asset.

```

doc/design/blob-uuid-1.md
yarikoptic: That is pretty much how it is now -- a responsibility of a good client to explicitly check first. 

I think we should leave above "Check" possibility as is for its own life, and make `/uploads/initialize` with similar request as to check, and do check internally (thus returning e.g. HTTP 409 Conflict with a Location or some other header field to point to existing blob key).  This would preclude "bad" clients to not check first, cause traffic, and us only later returning them a key for an already existing blob at `validate`.  And that is why I liked it when you (I believe) added to have `/uploads/initialize` to get the size and digest.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,46 @@
...
+## Garbage Collection process
+
+1. if dangling key's are found, API does garbage collection on keys that are older than 24 hours.
```
yarikoptic, line 35:
```suggestion
1. if dangling keys (not used by any asset) are found, API does garbage collection on keys that are older than 24 hours.
```

doc/design/blob-uuid-1.md
dchiquito: @yarikoptic That would work, but it doesn't resolve the overloading of HTTP actions. I liked the logical separation of having either
`GET /blobs/` -> `POST .../assets/` (no upload involved, so no calls to `/upload/...`)
or
`GET /blobs/` -> `POST /upload/initialize+complete+validate/` -> `POST .../assets/`
That keeps the upload process as a more distinct pathway.

This is just splitting hairs over nitty gritty HTTP design, though. I can implement it your way if you feel strongly.

yarikoptic: "strongly" would be the wrong word since I am not an experienced API designer ;)  
but it feels that it could help us avoid unnecessary waste of bandwidth so feels "more right" to me.

Note that a client could still do `GET /blobs/` first to proceed current/first path, but going directly to `POST /upload/initialize` (and handling a possible error indicating presence of the blob) would avoid that additional initial `GET /blobs/` call completely

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,47 @@
...
+   3. API returns presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
+    1. validates the size and that the initially reported ETag matches the actual ETag
+    2. checks for collision, since some other task could have finished by this time with the same object
+        1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
+        2. there is collision - delete key record and file in blob store, will return key of an existing record
```
yarikoptic, line 29:
```suggestion
        2. there is collision - delete key record and file in uploads, will return key of an existing record
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,47 @@
+A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<restofkey>`
```
yarikoptic, line 1:
```suggestion
A blob is referenced by a key generation function. the value of this key is used to index the blob in the database table of identities and on the s3 storage. `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
```
this would be consistent with the form used below in upload 1.1, and the form I would generally prefer even though at the little cost of duplicating first 6 digits in the directory names and the key name.  I find such a form easier to reuse later on and possibly avoid the file name to start with `-` if it happens to be the 7th digit in the key.

yarikoptic: but I do not feel too strong about it (I do not immediately see where using `restofkey` would really fail) so feel free to reject but then adjust path below in upload 1.1 to match.

doc/design/blob-uuid-1.md
dchiquito: How about adding the `GET /blobs/` endpoint so clients have a dedicated, reliable way to get look up existing assets, and making `/upload/initialize/` simply fail when attempting to upload a digest that's already registered?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,47 @@
...
+### Check process (for de-duplication):
+1. CLI sends expected digest with digest type (e.g., ETag)
+2. API checks for existence of ETag in identity table.
+3. if exists, returns key, if not returns an error.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
+   1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   2. saves initial data (size, ETag, key) to Upload table
+   3. API returns presigned URLs
```
yarikoptic, line 20:
```suggestion
   3. API returns `uploads:upload_id` and presigned URLs
```

doc/design/blob-uuid-1.md
yarikoptic: Yes, that is what [I had in mind above as well](https://github.com/dandi/dandi-api/pull/150#discussion_r592744335): `/blobs` is generally good to have, but not needed for "upload logic". Just that `/upload/initialize` should "simply fail" informatively - provide `blobs:key` in the return headers so that client could proceed with it to ```9. CLI calls `POST .../assets/` with the key to register the new asset.```
Note that it does leave possibility for a race condition (GC removes blob:key before CLI calls POST it for a new asset) but I guess it would be very unlikely... may be also there could be some timestamp increment on `blobs` record for that `key` so GC does not even consider blobs which were recently "shown interest in"? Is there something like "access time" for DB records or overhead would be too high?

dchiquito: I think we're on the same page, I'm committing this suggestion.

I think race conditions related to the garbage collector are garbage collector problems, which are not in the scope of this design doc.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,50 @@
...
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
+   1. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   2. saves initial data (size, ETag, key) to Upload table
+   3. API returns `uploads:upload_id` and presigned URLs
```
dchiquito, line 23:
```suggestion
   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT.
   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
   3. saves initial data (size, ETag, key) to Upload table
   4. API returns `uploads:upload_id` and presigned URLs
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+1. CLI `GET /blobs/` <-- expected digest with digest type (e.g., ETag). API:
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT.
```
yarikoptic, line 21:
```suggestion
   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with a key in the header (`Location`?).
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:key` in the header (`Location`?).
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
```
dchiquito, line 29:
```suggestion
8. CLI `POST /uploads/validations/{uploads:upload_id}/` --> `blobs:key`. API
```

dchiquito (commented):
(Hopefully) final round of changes, I have a working implementation that has this API surface.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+### Check process (for de-duplication):
+1. CLI `GET /blobs/` <-- expected digest with digest type (e.g., ETag). API:
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{contentSize , digests: { "dandi-s3-etag": "..." }`. API:
```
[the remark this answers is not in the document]
satra: my only comment here is that we will want `contentSize` here and in the remaining places, since that's in the asset metadata. also the current digest doesn't look like above, but hopefully we will make that adjustment soon.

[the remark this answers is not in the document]
dchiquito: `file_size` is here, which is a synonym for `contentSize`.
I typo'd the `algorithm`, it should be fixed now.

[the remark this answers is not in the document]
yarikoptic: we have contentSize in the schema, blobs are data blobs so might even have no file (at least with a name) so I would stick with `contentSize` without adding synonyms.

doc/design/blob-uuid-1.md
yarikoptic: Why? Do you expect a table of validations to be also queried for some kind of status?

We have `/initialize` and `/complete`  actions, and from the description below action should be atomic enough to not need any `validations` table.  For background processes I think we did talk about `/jobs` where checksums calculation could be a type of job associated with a `key`.

dchiquito: The endpoint basically takes an upload ID, looks up the Upload object, and verifies that the upload was completed correctly by the client. 
I see now that `POST /uploads/validations/{uploads:upload_id}/` is misleading, I will change my suggestion to `POST /uploads/validate/{uploads:upload_id}/`.
As an alternative, it could also be `POST /uploads/validate/` <-- `{"uuid": "1234...5678"}`, but I like that less than putting the ID in the URL.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+** database table of identities **: indexed by the key and contains columns for as many types of checksums as we want. these checksums constitute identities of the blob and should use functions that will always return the same value.
+
+**key generation function**: the current suggestion (from satra) for this function is UUID4.
+
+**DANDI ETag generation function**: this is an agreement between API server and DANDI CLI for now to ensure there is an algorithm that consistently returns the correct number of parts given a `contentSize` in bytes. this should in theory also be the AWS ETag generation function.
+
+## Upload (from a CLI)
+
+### Check process (for de-duplication):
+1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-s3-etag", "value": "123etag123"}`. API:
```
dchiquito, line 12:
```suggestion
1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-etag", "value": "123etag123"}`. API:
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-s3-etag", "value": "..."}`. API:
```
dchiquito, line 20:
```suggestion
1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
```

dchiquito (commented):
@yarikoptic changing the algorithms to `dandi:dandi-etag` as per https://github.com/dandi/dandi-api/issues/157#issuecomment-801356336

doc/design/blob-uuid-1.md
yarikoptic: > As an alternative, it could also be POST /uploads/validate/ <-- {"uuid": "1234...5678"}, but I like that less than putting the ID in the URL.

I think `POST /uploads/validate/ <-- {"uploads:upload_id": "1234...5678"}` (note: not  `uuid` AKA key which isn't yet returned) would keep it more consistent with other actions like `/initialize` and `/complete`, and consistency would be good since otherwise need to remember where an id goes into URL and where into the load for some imperative `action/` endpoint.  For direct operations like `GET` or `DELETE /uploads/{upload_uuid}` it indeed makes a good sense to have it in a URL, while again `PUT` or `POST` suggest a data record to be sent, so why then in URL?...  could then even be `GET /uploads/validate/{uploads:upload_id}/` and that would be really odd ;-)

Having said that, if was `/uploads/initialize -> upload_id`, `/uploads/{upload_id}/complete/` and `/uploads/{upload_id}/validate/` - I might have not complained since it would have paralleled better with `.../assets` endpoints

yarikoptic: I like this `/uploads/{upload_uuid}{,/<operation>}` idea!

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:key` in the header (`Location`?).
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `/uploads/validate` <-- `uploads:upload_id` --> `blobs:key`. API
```
dchiquito, line 29:
```suggestion
6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:key`. API
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-etag", "value": "123etag123"}`. API:
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:key` in the header (`Location`?).
```
yarikoptic, line 21: @dchiquito so is it to be `Location` header field in response and thus we should remove `?` or something else?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
```
yarikoptic, line 28: what does "aborts the upload" would entail -- removing `uploads:upload_id` record and target `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>` or something else?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:key`. API
```
yarikoptic, line 29:
```suggestion
8. CLI `GET /uploads/{upload_id}/validate/` --> `blobs:key`. API
```
or should be some POST?

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+    1. validates the size and that the initially reported ETag matches the actual ETag
+    2. checks for collision, since some other task could have finished by this time with the same object
+        1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
+        2. there is collision - delete key record and file in uploads, will return key of an existing record
```
yarikoptic, line 33:
```suggestion
        2. there is collision - deletes `blobs:key` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
```

yarikoptic (commented):
@dchiquito I left some nit-picking suggested changes which might be just what you are implementing and might need doc tune up. Please approve/merge or comment on

doc/design/blob-uuid-1.md
dchiquito: I implemented it to use the `Location` header, but I'm happy to leave the `?` until the CLI half is implemented in case it turns out something else would be preferable.

doc/design/blob-uuid-1.md
dchiquito: I think if the CLI wants to abort at any point, it should just stop. The user could kill the CLI at any point, or the CLI could run into some random CLI bug. It's the server's job to have lifecycle rules in place to clean up stale multipart uploads that were not completed, and garbage collection to clean up uploads that were never minted into Assets.

doc/design/blob-uuid-1.md
dchiquito: It should be POST, as it is doing things, not just returning information.

yarikoptic: adjusted suggestion with an explicitly empty record to POST

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+3. if key, call `POST .../assets/` with the key to register the new asset.
+
+### Blob upload process:
+1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:key` in the header (`Location`?).
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
```
jwodder, line 26: What exactly is the client supposed to check the ETag against?  The md5 digest of the part?

doc/design/blob-uuid-1.md
jwodder, line 28: "checks final ETag" — Where is this ETag that is being checked?

doc/design/blob-uuid-1.md
yarikoptic: yes AFAIK

doc/design/blob-uuid-1.md
yarikoptic: after you POST (IIRC) to that "completion URL" you should receive final ETag, and it should correspond to the `dandi-etag` computed before initiating upload

doc/design/blob-uuid-1.md
dchiquito:
```suggestion
8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:blob_id`. API
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+**key generation function**: the current suggestion (from satra) for this function is UUID4.
+
+**DANDI ETag generation function**: this is an agreement between API server and DANDI CLI for now to ensure there is an algorithm that consistently returns the correct number of parts given a `contentSize` in bytes. this should in theory also be the AWS ETag generation function.
+
+## Upload (from a CLI)
+
+### Check process (for de-duplication):
+1. CLI `POST /blobs/digests/` <-- `{"algorithm": "dandi:dandi-etag", "value": "123etag123"}`. API:
+    1. Looks for an existing AssetBlob with the given digest(s).
+    2. If it exists, return the key (UUID).
```
dchiquito, line 14:
```suggestion
    2. If it exists, return the `blobs:blob_id`.
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+    3. It it doesn't exist, return 404.
+2. If 404, start upload process.
+3. if key, call `POST .../assets/` with the key to register the new asset.
```
dchiquito, line 17:
```suggestion
3. if `blob_id`, call `POST .../assets/` with the `blob_id` to register the new asset.
```

doc/design/blob-uuid-1.md
dchiquito, line 21:
```suggestion
   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:blob_id` in the header (`Location`?).
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `/uploads/{upload_id}/validate/` <-- `uploads:upload_id` --> `blobs:key`. API
+    1. validates the size and that the initially reported ETag matches the actual ETag
+    2. checks for collision, since some other task could have finished by this time with the same object
+        1. no collision - adds key + ETag to AssetBlob table, delete from Upload table.
+        2. there is collision - deletes `blobs:key` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
```
dchiquito, line 33:
```suggestion
        2. there is collision - deletes `blobs:blob_id` record, `uploads:upload_id` record, and `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`, will return key of an existing record
```

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+    4. kicks off background process to calculate checksums
+9. CLI calls `POST .../assets/` with the key to register the new asset.
```
dchiquito, line 35:
```suggestion
9. CLI calls `POST .../assets/` with the `blob_id` to register the new asset.
```

dchiquito (commented):
Rename `blobs:key` to `blobs:blob_id` to reflect how upload was implemented.

doc/design/blob-uuid-1.md
```diff
@@ -0,0 +1,51 @@
...
+1. CLI `/uploads/initialize` <-- `{"content_size": 12345 , "digest": {"algorithm": "dandi:dandi-etag", "value": "..."}`. API:
+   1. Checks if there is already an asset blob with the given digest. If so, return 409 CONFLICT with the `blobs:blob_id` in the header (`Location`?).
+   2. generates key, determines upload location `s3://dandiarchive/blob/<3digitskey>/<next3digitskey>/<key>`
+   3. saves initial data (size, ETag, key) to Upload table
+   4. API returns `uploads:upload_id` and presigned URLs
+4. CLI verifies that the number+size of parts match the calculated number of parts (i.e. server and CLI are using the same ETag generation function)
+5. CLI uploads to presigned URLs, and for each part checks ETag on return. Any part upload failure can be retried without involving API.
+6. `/uploads/{upload_id}/complete/`: CLI sends parts info to API, API responds with presigned completion URL
+7. CLI completes upload and checks final ETag. Mismatch aborts the upload.
+8. CLI `POST /uploads/{upload_id}/validate/` <-- `{}` --> `blobs:key`. API
```
dchiquito, line 29:
```suggestion
8. CLI `POST /uploads/{upload_id}/validate/` <-- `{}` --> `blobs:blob_id`. API
```

dchiquito:
I think we have reached a consensus, so I'm going to merge this. We can always revisit it in a new PR if more changes are required.

Issue #160, opened 2021-03-18
Use dandi.core.digests.DandiETag for upload parts


yarikoptic:
which is being developed in https://github.com/dandi/dandi-cli/pull/474 and yet to be finished/merged/released, so it is a blocker for dandi-api ATM to adopt it in the main code and tests (I remember seeing some plain md5 calls in #156)

Blocked by
- [x] https://github.com/dandi/dandi-cli/pull/474
- [x] release of dandi-api with that PR merged

yarikoptic:
0.12.0 release of dandi-cli is now available from pypi and conda-forge.  It provides https://github.com/dandi/dandi-cli/blob/master/dandi/core/digests/dandietag.py with desired functionality/interfaces

PR #171, opened 2021-03-23
Change minimum contentSize to 0

### LINKED ISSUE #168: DJANGO_API: allow for upload of 0-length file
initially detected/reported in https://github.com/dandi/dandi-cli/pull/479#issuecomment-804871890

```
Error 400 while sending POST request to http://localhost:8000/api/uploads/initialize/: {"contentSize":["Ensure this value is greater than or equal to 1."]}
```

although it is an odd use case and would exist solely in a singular blob instance due to deduplication, api backend should allow for an upload of such a file which I expect us to encounter whenever we allow uploads of non-nwb files.  As such it is not a priority but should be done/fixed.

dchiquito:
Fixes #168

### DIFF
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -25,7 +25,7 @@ class DigestSerializer(serializers.Serializer):
 
 
 class UploadInitializationRequestSerializer(serializers.Serializer):
-    contentSize = serializers.IntegerField(min_value=1)  # noqa: N815
+    contentSize = serializers.IntegerField(min_value=0)  # noqa: N815
     digest = DigestSerializer()
 
 


yarikoptic:
wasn't enough?

dchiquito:
Nope, wasn't. `_iter_part_sizes` needs to handle the case where file_size=0, which requires adjusting the upstream for now. As soon as we move that function to the CLI though, this should be easier.

yarikoptic:
well, we have it in dandi-cli already  https://github.com/dandi/dandi-cli/blob/master/dandi/core/digests/dandietag.py#L111 and should release (https://github.com/dandi/dandi-cli/issues/487) to make it available to dandi-api .
The question would be whether we should fix it up to return 1 part aiming for 0 length upload, or jumping to upload completion url would be enough?

yarikoptic:
actually we could just find an answer after dandi-api migrates to use dandi-cli and thus fix up in a single spot (dandi-api)... so let's release dandi-cli and then address #160 here and fixup in dandi-cli if needed

satra:
just a thought: for the 0 content length, could we prepopulate that use case (in the db and in the blob storage), such that no one every has to try initializing an upload for that?

dchiquito:
I vote for returning 1 part with 0 length, so that both the client and the server don't have to handle any special cases.

I would rather have a system that always works instead of relying on manually maintaining the health of the system state.

yarikoptic:
+1 on having "native" support for 0-length.

> I vote for returning 1 part with 0 length, so that both the client and the server don't have to handle any special cases.

I do not think there would be any special cases -- both server and client would just get an empty list of parts  and client loop would not loop and just proceed  to completion stage.  Or what am I missing @dchiquito ?

PR #174, opened 2021-03-23
Strip double quotes from S3 ETags

dchiquito:
https://github.com/dandi/dandi-cli/pull/479#issuecomment-805056962

S3 wraps ETags in double quotes, so strip them before testing that expected ETag matches actual ETag.


waxlamp (commented):
This seems like it will work just fine, but also something like `json.loads()` might be better, just because it's more "standard" in some way.

Is there a reason you need to check to see if the etag value has quotes around it?

Issue #90, opened 2021-02-08
Mint DOIs upon "release"


yarikoptic:
TODOs
- [x] @yarikoptic to provide @dchiquito with details on metadata record and interfacing with datacite (sandbox to be used while developing)
- [x] trigger minting DOI during publish/release of a dandiset and add it as a `doi` field within released dandiset
     - DOI should point to our versioned "permalink" provided by redirector (not web UI URL)

yarikoptic:
- test instance https://doi.test.datacite.org/repositories/dartlib.dandi , so repository is `dartlib.dandi`
- api reference: https://support.datacite.org/reference/introduction and in particular https://support.datacite.org/docs/api-create-dois for minting DOIs
  - there seems to be no token auth, so password needs to be provided, sent via email to @dchiquito and @satra

dchiquito:
For future reference, the prefix for `dartlib.dandi` is `10.80507`.

dchiquito:
Quoting @yarikoptic from https://github.com/dandi/dandi-api/pull/175#discussion_r600006179:
> Also relating to above comment on f'{prefix}/{dandiset_id}/{version_id}' - I did not know that it is legit for suffix (everything after {prefix}/) to contain / . I do see though that there is now DOI 10.80507/000001/0.210323.2052 on that test sandbox instance and according to https://www.doi.org/doi_handbook/2_Numbering.html
> 
> 2.6.3 URN presentation
> 
> To enable the use of DOIs in workflows that have already standardized on URNs, the DOI proxy servers understand the substitution of a colon in place of the initial slash in a DOI name. DOI names may therefore be expressed as URNs in the doi.org domain by writing, for example, the DOI name 10.123/456 in the form https://doi.org/urn:doi:10.123:456. Note, however, that a DOI suffix is allowed to contain other slashes, and where these occur they must be percent-encoded rather than replaced with a colon: for example, the DOI name 10.123/456ABC/zyz would become https://doi.org/urn:doi:10.123:456ABC%2Fzyz, with the final slash character encoded as %2F.
> 
> So may be this "complication" is why typically people avoid using / in the DOI suffixes? E.g. openneuro has 10.18112/openneuro.ds003569.v1.0.0 for https://openneuro.org/datasets/ds003569/versions/1.0.0
> 
> @satra -- are you aware of gotchas with / in the DOI suffix and why it seems generally is avoided?

dchiquito:
[...]

PR #175, opened 2021-03-23
Mint DOIs on publish

dchiquito:


dandiapi/api/doi.py
```diff
@@ -0,0 +1,63 @@
...
+import requests
+
+from dandiapi.api.models import Version
+
+
+def _generate_doi_data(version: Version) -> Dict:
+    prefix = settings.DANDI_DOI_API_PREFIX
+    dandiset_id = version.dandiset.identifier
+    version_id = version.version
+    doi = f'{prefix}/{dandiset_id}/{version_id}'
```
yarikoptic, line 14: Does it show for all 3?

dandiapi/api/doi.py
```diff
@@ -0,0 +1,63 @@
...
+        return
+    request_body = _generate_doi_data(version)
+    requests.post(
+        settings.DANDI_DOI_API_URL,
+        json=request_body,
+        auth=requests.auth.HTTPBasicAuth(
+            settings.DANDI_DOI_API_USER,
+            settings.DANDI_DOI_API_PASSWORD,
+        ),
+    ).raise_for_status()
```
yarikoptic, line 63: I guess (didn't check) that as a result of that post you should receive an actual DOI and that is the one then which should be assigned to the `doi` field within metadata of the published dandiset.

Also relating to above comment on `f'{prefix}/{dandiset_id}/{version_id}'` - I did not know that it is legit for suffix (everything after `{prefix}/`) to contain `/` .  I do see though that there is now DOI 10.80507/000001/0.210323.2052 on that test sandbox instance and according to https://www.doi.org/doi_handbook/2_Numbering.html

2.6.3 URN presentation

To enable the use of DOIs in workflows that have already standardized on URNs, the DOI proxy servers understand the substitution of a colon in place of the initial slash in a DOI name. DOI names may therefore be expressed as URNs in the doi.org domain by writing, for example, the DOI name 10.123/456 in the form https://doi.org/urn:doi:10.123:456. Note, however, that a DOI suffix is allowed to contain other slashes, and where these occur they must be percent-encoded rather than replaced with a colon: for example, the DOI name 10.123/456ABC/zyz would become https://doi.org/urn:doi:10.123:456ABC%2Fzyz, with the final slash character encoded as %2F.

So may be this "complication" is why typically people avoid using `/` in the DOI suffixes?  E.g. openneuro has `10.18112/openneuro.ds003569.v1.0.0` for https://openneuro.org/datasets/ds003569/versions/1.0.0

@satra -- are you aware of gotchas with `/` in the DOI suffix and why it seems generally is avoided?

dandiapi/api/doi.py
dchiquito: what?

An example DOI generated from this is `10.80507/000001/0.210323.2052`.

dandiapi/api/doi.py
dchiquito: I'm happy to change this to whatever. I thought `10.80507/000001/0.210323.2052` would match how the `dandiset.yaml` files are stored.

dchiquito:
@yarikoptic I am going to merge this so I can verify it works in production, and so you guys can try the feature. I'm moving the discussion back to #90 .

dandiapi/api/doi.py
yarikoptic: I also like how it looks, but as I have mentioned, I did not even know that it is possible since I never saw any DOI with `/` in the suffix.  So we need to research on that more.  @effigies may be you know or can refer us to someone who would be able to answer the question  
"why (is there any pragmatic reason) openneuro has DOI like 10.18112/openneuro.ds003569.v1.0.0 and not e.g. 10.18112/openneuro/ds003569/v1.0.0"?

yarikoptic:
sure, go ahead with the merge -- we are still torturing the test instance anyways.

effigies:
Hmm. @rwblair (on vacation) might know. Would guess aesthetics.

dandiapi/api/doi.py
rwblair: Regarding openneuro DOIs, when the code to mint DOIs was first written I didn't know that '/' could be used in the suffix and for whatever reason I didn't look it up. '.' as a delimiter was common in the Datacite documentation so I defaulted to that.

yarikoptic: Thank you @rwblair!

PR #183, opened 2021-03-25
Set up integration tests with dandi-cli

### LINKED ISSUE #42: Establish integration testing with dandi-cli
May be not right away if some breaking changes are still to be expected, but if not or just to be "friendly" to dandi-cli developers right away, it would be great to have a CI run (github actions are used in dandi-cli) which would run dandi-cli tests against the state of dandi-api changes. This way we would at least be made aware of changes in dandi-api which would break dandi-cli to proactively mitigate them.

jwodder:
Closes #42.

Blocked by https://github.com/dandi/dandi-cli/pull/500.

### DIFF
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -35,9 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - 3.7
           - 3.8
-          - 3.9
         dandi-version:
           - release
           - master
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           docker build \
             -t dandiarchive/dandiarchive-api \
-            -f dev/django.Dockerfile \
+            -f dev/django-public.Dockerfile \
             .
 
       - name: Set up Python ${{ matrix.python }}
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -5,8 +5,32 @@ on:
   pull_request:
 
 jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out this repository
+        uses: actions/checkout@v2
+
+      - name: Build Docker image
+        run: |
+          docker build \
+            -t dandiarchive/dandiarchive-api \
+            -f dev/django-public.Dockerfile \
+            .
+
+      - name: Export Docker image
+        run: |
+          docker image save -o dandiarchive-api.tgz dandiarchive/dandiarchive-api
+
+      - name: Upload Docker image tarball
+        uses: actions/upload-artifact@v2
+        with:
+          name: dandiarchive-api.tgz
+          path: dandiarchive-api.tgz
+
   test:
     runs-on: ubuntu-latest
+    needs: build-image
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


jwodder:
The failing tests all seem to have failed due to the Girder provisioning container failing.  I do not know why that happened or why only some of them failed.

yarikoptic:
@jwodder let's test here with just one version of python - the one used by dandi-api in production. Should reduce stress on docker compose and hopefully eliminate sporadic fails. Which python is used  by dandi-api @dchiquito ?

dchiquito:
The API server uses `python-3.8.7`.

yarikoptic:
@dchiquito could you bless our curse those PR please? ;-)

dchiquito:
@yarikoptic I think this would be more useful in `dandi-cli`. Adding this as a CI step means that we are blocked from merging if a change would break the CLI, which I think we would still want to do, at least until we have a staging server. Seeing that the CLI CI step failed is pretty meaningless to us, so we would just have to ping a CLI developer out of band to resolve the problem. Running these `dandi-cli` tests in `dandi-cli` means that you will be better informed.

I would suggest either nightly runs or using something like https://github.com/peter-evans/repository-dispatch to trigger it.

yarikoptic:
@dchiquito 

>  I think this would be more useful in dandi-cli

well -- we do have it there already

> Adding this as a CI step means that we are blocked from merging if a change would break the CLI, which I think we would still want to do, at least until we have a staging server.

There is no police writing out tickets or forbidding merges, so PR could "in principle" be merged even if `dandi-cli` tests are failing if it is decided that "it is ok". And then we all would know which change in dandi-api broke dandi-cli.

Otherwise, PR in dandi-api can (and did in the past) break dandi-cli and nobody would notice until we run tests in dandi-cli or a user (after "migration") reports that a problem. Then CLI developers need to inspect and/or bisect changes in dandi-api to figure out what broke it and how to handle the change in dandi-api behavior.  IMHO that is a time wasted by CLI developers since could be avoided at minimal (if any cost) to dandi-api developers.

> Seeing that the CLI CI step failed is pretty meaningless to us, so we would just have to ping a CLI developer out of band to resolve the problem.

I would strongly disagree.  We are developing DANDI archive with dandi-api to be the product to deliver, and not a collection of independent projects. With the "migration" in the immediate horizon, we will really need to avoid breaking dandi-cli (and web ui, but no automated tests, so insofar will rely on user reports) operation (for users) by non-coordinated with CLIs changes in dandi-api.  After "migration" - if CLI CI step fails for legit reason due to changes in dandi-api, in most of the cases PR must not be merged/deployed before dandi-cli is adjusted.

Note that dandi-api doesn't test the entire API surface of dandi-api, so many things could still be changed in dandi-api without triggering dandi-cli tests failures (we do use `publish` endpoint though I believe).

> at least until we have a staging server.

well, we do not have staging server, we do have dandi-api server which is intended to go in "production" directly from `master` branch here very soon, and we do have dandi-cli  which should work with dandi-api.  We can adjust CI setup accordingly whenever staging server is deployed.

yarikoptic:
Why not to merge it (costs dandi-api folks nothing insofar) and see how it goes and either it would be more pain or more gain?  It will be very easy to delete it later on if decided to be "not useful".

waxlamp:
> Why not to merge it (costs dandi-api folks nothing insofar) and see how it goes and either it would be more pain or more gain? It will be very easy to delete it later on if decided to be "not useful".

It seems like we could run exactly these tests in the dandi-cli repository itself using the [repository dispatch](https://github.com/marketplace/actions/repository-dispatch) action. This seems like a more natural thing to do so that the test runs remain in the dandi-cli repository scope along with its other CI.

The exception to this is if you *do* want a breaking CLI to block merges in the API codebase. In that case, I'd want to discuss a bit to figure out the right workflow here (but in the meantime, we could merge this PR as an experiment, as you've suggested).

PR #187, opened 2021-03-26
Supply content-disposition header for asset downloads

### LINKED ISSUE #152: Add content-disposition to download URLs
Pre-signed urls produced from */assets/{uuid}/download endpoint should have appropriate for the asset content-disposition

waxlamp:
Done with a ton of help from @dchiquito.

Closes #152.

I am not sure who else should review this.

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -1,3 +1,4 @@
+import os.path
 from uuid import uuid4
 
 from guardian.shortcuts import assign_perm
@@ -303,7 +304,7 @@ def test_asset_download(api_client, storage, version, asset):
     download = requests.get(download_url)
     cd_header = download.headers.get('Content-Disposition')
 
-    assert cd_header == f'attachment; filename="{asset.path}"'
+    assert cd_header == f'attachment; filename="{os.path.basename(asset.path)}"'
 
     with asset.blob.blob.file.open('rb') as reader:
         assert download.content == reader.read()
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -2,8 +2,9 @@ from uuid import uuid4
 
 from guardian.shortcuts import assign_perm
 import pytest
+import requests
 
-from dandiapi.api.models import Asset
+from dandiapi.api.models import Asset, AssetBlob
 
 from .fuzzy import HTTP_URL_RE, TIMESTAMP_RE, UUID_RE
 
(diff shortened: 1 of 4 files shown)


dandiapi/api/tests/test_asset.py
```diff
@@ -282,6 +284,32 @@ def test_asset_rest_delete_not_an_owner(api_client, user, version, asset):
...
+
+    download_url = response.get('Location')
+    assert download_url == HTTP_URL_RE
+
+    download = requests.get(download_url)
+    cd_header = download.headers.get('Content-Disposition')
+
+    assert cd_header == f'attachment; filename="{os.path.basename(asset.path)}"'
+
+    with asset.blob.blob.file.open('rb') as reader:
```
yarikoptic, line 309: Just an observation unrelated to pr changes since I lack sufficient knowledge: Eh, we used to have .repo.repo in Datalad - it was always fun to debug and also a good memory excercise .

dandiapi/api/views/asset.py
```diff
@@ -1,3 +1,16 @@
+try:
+    from storages.backends.s3boto3 import S3Boto3Storage
+except ImportError:
+    # This should only be used for type interrogation, never instantiation
+    S3Boto3Storage = type('FakeS3Boto3Storage', (), {})
+try:
+    from minio_storage.storage import MinioStorage
+except ImportError:
+    # This should only be used for type interrogation, never instantiation
+    MinioStorage = type('FakeMinioStorage', (), {})
```
AlmightyYakob, line 10: Is the point of this dual try-catch to always import either one or the other?

AlmightyYakob (commented):
A couple clarifying comments, but otherwise LGTM

dandiapi/api/views/asset.py
waxlamp: I believe the idea is that in testing or local environments, `MinioStorage` will be available, and boto otherwise, but we need this code path to work in both situations.

If we could just have both that would make the code simpler, but maybe there's good reason to avoid that. @dchiquito will know details here.

dchiquito: This code traces its lineage back to S3FF. I believe the original idea was that if an app was only going to use Minio, it should only need to pip install the minio package, not boto3, and vice versa. In DANDI I see that we are just requiring both all the time, but it's a nice thought.

dandiapi/settings.py
```diff
@@ -65,6 +65,7 @@ class DevelopmentConfiguration(DandiMixin, DevelopmentBaseConfiguration):
 
 class TestingConfiguration(DandiMixin, TestingBaseConfiguration):
     MINIO_STORAGE_MEDIA_BUCKET_NAME = 'test-django-storage'
+    MINIO_STORAGE_MEDIA_URL = 'http://minio:9000/test-django-storage'
```
dchiquito, line 68:
```suggestion
    MINIO_STORAGE_MEDIA_URL = 'http://localhost:9000/test-django-storage'
```
This fixes the tests for me locally. @waxlamp do you recall why we needed the container name in the `TestingConfiguration`? I would think it would only be necessary when running in development.

dchiquito (approved):

PR #204, opened 2021-04-01
No schema version

### LINKED ISSUE #172: should not add schema_version; should support/check for range of schemaVersions in metadata received
I have freshly recreated 000004 on dandi-api instance, and queried its metadata but it provides old schema_version and new schemaVersion:
```
$> curl --silent -X GET "https://api.dandiarchive.org/api/dandisets/000004/versions/draft/" -H "accept: application/json" | jq . | grep -i version
  "version": "draft",
    "schemaVersion": "0.1.1",
    "schema_version": "1.0.0-rc1",
```

dchiquito:
Fixes #172

* Rename `schema_version` to `schemaVersion` in version metadata.
* No longer inject `schemaVersion` to the version metadata whenever the version is saved.
* Inject `schemaVersion` into the version metadata when creating the initial metadata during dandiset creation. The API can also specify a `schemaVersion` in the metadata when calling `POST /dandisets/`, which will take precedence.

@satra @yarikoptic does that change list sound right?

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import datetime
 
-from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.validators import RegexValidator
 from django.db import models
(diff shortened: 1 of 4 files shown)


waxlamp (approved):
Code changes all look fine. Should wait for @yarikoptic or @satra's feedback on if this is acceptable for solving the problem at hand.

satra:
@dchiquito -  i don't think the API-server should ever set the version, it should simply check that the version is a valid version. the version should be set by the clients whether UI or CLI, when POSTing dandiset metadata into the server. what the server should be doing is just "validating" according to the schema.

it seems in this PR that the API server is setting the schemaVersion value.

PR #212, opened 2021-04-05
First draft of DOi design doc

dchiquito:


doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,26 @@
...
+The API will pass the following to the DOI generator:
+* A DOI identifier. This will be of the form `f'{prefix}/{dandiset_id}/{version_id}'`, for example `10.80507/000001/0.210402.1835`.
+* A URL for the dandiset. This will be a redirector URL of the form `f'https://dandiarchive.org/dandiset/{dandiset_id}/{version_id}'`.
+* The dandiset metadata, as a python dictionary.
+
+## dandi.core.doi ?
+
+This function will generate the body of the DOI and return it.
+
+TODO: fill me in
```
djarecka, line 18: * checking `meta = DandisetMeta(meta)` or `meta = PublishedDandisetMeta (meta)` (if `publishedBy` and `datePublished` is available):
      - if fields are missing (e.g. `contributors`) it would return error even today
      - I would add checks if `doi`  (and possibly other fields) is available
*  running `datacite_json = to_datacite(meta)`
      - this should not return an error

djarecka:
I've added a comment, and will wait for discussion. If that looks good, I can create a PR to your branch

doc/design/doi-generation-1.md
dchiquito: On the previous design doc we were using the GitHub suggestion feature to inject changes rather than writing PRs for the PR:
```suggestion
* checking meta = DandisetMeta(meta) or meta = PublishedDandisetMeta (meta) (if publishedBy and datePublished is available):
  * if fields are missing (e.g. contributors) it would return error even today
  * I would add additional checks if doi (and possibly other fields) is available
* running datacite_json = to_datacite(meta)
  * this should not return an error
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,26 @@
+# DOI generation process
+
+## dandi-api
+
+The DOI generation process happens inside a call to `POST /dandisets/{dandiset_id}/draft/publish/`.
+DOI generation is the first step of the publish process, because it is the only step to rely on an external service.
+If the call to the Datacite API fails, there are no files in S3 that need to be cleaned up.
+
+The API will pass the following to the DOI generator:
+* A DOI identifier. This will be of the form `f'{prefix}/{dandiset_id}/{version_id}'`, for example `10.80507/000001/0.210402.1835`.
```
yarikoptic, line 10:
```suggestion
* A DOI identifier. This will be of the form `f'{prefix}/{dandiset_id}/{version_id}'`, for example `10.80507/dandi.000001.v0.210402.1835`.
```
Given the discussion elsewhere we might want to mention dandi and avoid /

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,26 @@
+# DOI generation process
```
satra, line 1:
```suggestion
# Publish process

This process is predicated on a few tasks that are expected to receive valid input and assuming that only certain errors could happen that are out of our control.

- publish enabled only if dandiset is valid (so all validation checks have passed)
- on publish trigger:
   - create the versioned release in the db (this prevents any asset/blob belonging to the release from being deleted)
   - insert all the relevant metadata (version, doi, url, manifest location/checksum) into the metadata record of the new version 
  - send this new metadata record to mint the doi (there should be no reason why this should fail unless the external resource fails. if it fails notify admins since this is likely a new bug)
   - write out to s3 bucket (even on any datacite failure)
   - switch over ownership of the release to dandi (unless releases are not modifiable by the dandiset creator)
   
## DOI generation process
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,26 @@
...
+* A DOI identifier. This will be of the form `f'{prefix}/{dandiset_id}/{version_id}'`, for example `10.80507/000001/0.210402.1835`.
+* A URL for the dandiset. This will be a redirector URL of the form `f'https://dandiarchive.org/dandiset/{dandiset_id}/{version_id}'`.
+* The dandiset metadata, as a python dictionary.
+
+## dandi.core.doi ?
+
+This function will generate the body of the DOI and return it.
+
+TODO: fill me in
+TODO: what exceptions might be thrown? Missing creator is one.
```
satra, line 19: missing creator should never happen at this stage. any metadata should be valid before the time this function is called.

doc/design/doi-generation-1.md
dchiquito: I thought this PR would be scoped to DOIs so Dorota and I could get a cohesive design. If you want to cover the entire publish process, I can add:
* Metadata validation (so the publish button isn't available if metadata is invalid)
* Web GUI UX
* Creating the published version
* Writing the manifests

satra: yes let's scope the whole thing. because the datacite conversion and doi generation sits within the context of that.

doc/design/doi-generation-1.md
djarecka: ok

djarecka:
I want to check with you what exactly mean valid - as I mentioned on the slack, running validator gives me additional errors. Some of them I could deal within the function, e.g. if a person has a contribution type that is not in the schema I could ignore it, or if someone has more roles I could take the first role only. Another example is if `relatedResource.identifier` has `doi` instead of `DOI` I could fix it.
But I'm not sure, if these kind of issues should be "fixed" within the function or meta should be considered as invalid

satra:
i think these should be adjusted in the function (doi vs DOI should not be an issue in general). you can also add additional patterns or constraints in the model if you want to restrict the fields in some way. so changes to model code are welcome, but not changes like restricting the set of roles. 

does datacite validate the roles?

djarecka:
yes, officially datacide expects a string from this list `[ContactPerson, DataCollector, DataCurator, DataManager, Distributor, Editor, HostingInstitution, Producer, ProjectLeader, ProjectManager, ProjectMember, RegistrationAgency, RegistrationAuthority, RelatedPerson, Researcher, ResearchGroup, RightsHolder, Sponsor, Supervisor, WorkPackageLeader, Other]`

As I wrote on slack, I was able to push to  `https://doi.test.datacite.org/ my previous version that took everything from meta, but I just realized that datacite simply ignored this field.

satra:
you may want to look into this blog post: https://blog.datacite.org/contributor-information-in-datacite-metadata/ (there is a section on creator and contributorType)

djarecka:
I've learned how to create the `contributorType`: https://doi.test.datacite.org/dois/10.80507%2F2%2F000004

Perhaps this is not the best place to have this discussion, but it's not clear to me when should I say that meta is not valid. From your comment I understand that I should not restricting the set of roles, so the meta is considered "valid" even if I'm not able to use it in `datacite`

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,38 @@
...
+
+This process is predicated on a few tasks that are expected to receive valid input and assuming that only certain errors could happen that are out of our control.
+
+- publish enabled only if dandiset is valid (so all validation checks have passed)
+- on publish trigger:
+   - create the versioned release in the db (this prevents any asset/blob belonging to the release from being deleted)
+   - insert all the relevant metadata (version, doi, url, manifest location/checksum) into the metadata record of the new version 
+  - send this new metadata record to mint the doi (there should be no reason why this should fail unless the external resource fails. if it fails notify admins since this is likely a new bug)
+   - write out to s3 bucket (even on any datacite failure)
+   - switch over ownership of the release to dandi (unless releases are not modifiable by the dandiset creator)
```
dchiquito, line 11: @satra Can you clarify this? "Owners" in `dandi-api` means a set of users who allowed to make changes through the API to a dandiset. Ownership only applies to dandisets as a whole, not versions. There's also "ownership" of the objects in the S3 bucket, but all objects belong to the sponsored bucket so I don't think that's what is meant here.

Do you mean that owners should not be allowed to modify published versions through the API?

doc/design/doi-generation-1.md
dchiquito:
```suggestion
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,42 @@
+# Publish process
+
+This process is predicated on a few tasks that are expected to receive valid input and assuming that only certain errors could happen that are out of our control.
+
+- publish enabled only if dandiset is valid (so all validation checks have passed)
```
dchiquito, line 5:
```suggestion
Dandisets have a new attribute added to track their validation status: `VALID`, `INVALID`, and `VALIDATING`.
Whenever a dandiset is modified (metadata changed or files uploaded/removed), it is set to `VALIDATING` and an asynchronous task is triggered to run metadata validation on it. The task will update the state of the dandiset accordingly when it finishes.

- publish enabled only if a dandiset is `VALID`. The button will be greyed out in the UI and the API endpoint will return 405.
```

doc/design/doi-generation-1.md
satra: correct. the only people who should be able to call PATCH/DELETE/PUT on non-draft dandisets are admins.

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,43 @@
...
+Dandisets have a new attribute added to track their validation status: `VALID`, `INVALID`, and `VALIDATING`.
+Whenever a dandiset is modified (metadata changed or files uploaded/removed), it is set to `VALIDATING` and an asynchronous task is triggered to run metadata validation on it. The task will update the state of the dandiset accordingly when it finishes.
+
+- publish enabled only if a dandiset is `VALID`. The button will be greyed out in the UI and the API endpoint will return 405.
+- on publish trigger:
+   - create the versioned release in the db (this prevents any asset/blob belonging to the release from being deleted)
+   - insert all the relevant metadata (version, doi, url, manifest location/checksum) into the metadata record of the new version 
+  - send this new metadata record to mint the doi (there should be no reason why this should fail unless the external resource fails. if it fails notify admins since this is likely a new bug)
+   - write out to s3 bucket (even on any datacite failure)
+   - switch over ownership of the release to dandi (unless releases are not modifiable by the dandiset creator)
```
dchiquito, line 13:
```suggestion
```

Explicit switching of ownership isn't a thing, published versions are not modifiable.

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,42 @@
...
+  * I would add additional checks if doi (and possibly other fields) is available
+* running datacite_json = to_datacite(meta)
+  * this should not return an error
+
+## dandi-api
+
+* If DOI generation raised any errors, return a 400 error describing the problem.
+* Otherwise, attempt to register the DOI identifier/body with Datacite.
+* If that raises any errors, log them and return a 400 error.
+* Finally, continue with the publish process.
```
dchiquito, line 42:
```suggestion
* If that raises any errors, log them so they are reported in Sentry. Regardless of the outcome, continue with the publish process.
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+This process is predicated on a few tasks that are expected to receive valid input and assuming that only certain errors could happen that are out of our control.
+Dandisets have a new attribute added to track their validation status: `VALID`, `INVALID`, and `VALIDATING`.
+Whenever a dandiset is modified (metadata changed or files uploaded/removed), it is set to `VALIDATING` and an asynchronous task is triggered to run metadata validation on it. The task will update the state of the dandiset accordingly when it finishes.
+
+- publish enabled only if a dandiset is `VALID`. The button will be greyed out in the UI and the API endpoint will return 405.
+- on publish trigger:
+   - create the versioned release in the db (this prevents any asset/blob belonging to the release from being deleted)
+   - insert all the relevant metadata (version, doi, url, manifest location/checksum) into the metadata record of the new version 
+  - send this new metadata record to mint the doi (there should be no reason why this should fail unless the external resource fails. if it fails notify admins since this is likely a new bug)
+   - write out to s3 bucket (even on any datacite failure)
```
satra, line 12:
```suggestion
   - write out to s3 bucket (even on any datacite failure)
      - add checksums for dandiset.yaml and assets.yaml which should be stored in a checksums.json in the published directory.
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
+# Publish process
+
+This process is predicated on a few tasks that are expected to receive valid input and assuming that only certain errors could happen that are out of our control.
+Dandisets have a new attribute added to track their validation status: `VALID`, `INVALID`, and `VALIDATING`.
+Whenever a dandiset is modified (metadata changed or files uploaded/removed), it is set to `VALIDATING` and an asynchronous task is triggered to run metadata validation on it. The task will update the state of the dandiset accordingly when it finishes.
+
+- publish enabled only if a dandiset is `VALID`. The button will be greyed out in the UI and the API endpoint will return 405.
+- on publish trigger:
+   - create the versioned release in the db (this prevents any asset/blob belonging to the release from being deleted)
+   - insert all the relevant metadata (version, doi, url, manifest location/checksum) into the metadata record of the new version
```
satra, line 10:
```suggestion
   - insert all the relevant metadata (version, doi, url, manifest location) into the metadata record of the new version 
    -  manifest url == dandiarchive api url to GET on assets for the versioned dataset.
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+## DOI generation process
+
+## dandi-api
+
+The DOI generation process happens inside a call to `POST /dandisets/{dandiset_id}/draft/publish/`.
+DOI generation is the first step of the publish process, because it is the only step to rely on an external service.
+If the call to the Datacite API fails, there are no files in S3 that need to be cleaned up.
+
+The API will pass the following to the DOI generator:
+* A DOI identifier. This will be of the form `f'{prefix}/{dandiset_id}/{version_id}'`, for example `10.80507/dandi.000001.v0.210402.1835`.
```
djarecka, line 23: assuming that we are using "." instead of "/" on purpose?  Should we change the form to something like this?
`f'{prefix}/dandi.{dandiset_id}.v.{version_id}'

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+* A URL for the dandiset. This will be a redirector URL of the form `f'https://dandiarchive.org/dandiset/{dandiset_id}/{version_id}'`.
+* The dandiset metadata, as a python dictionary.
+
+## dandi.core.doi ?
```
djarecka, line 27:
```suggestion
## dandi.core.to_datacite
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+
+This function will generate the body of the DOI and return it.
+
+* checking meta = DandisetMeta(meta) or meta = PublishedDandisetMeta (meta) (if publishedBy and datePublished is available):
+  * if fields are missing (e.g. contributors) it would return error even today
+  * I would add additional checks if doi (and possibly other fields) is available
+* running datacite_json = to_datacite(meta)
```
djarecka, line 34:
```suggestion
* creating a mapping between datacite's metadata and datacite's dictionary
```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+  * this should not return an error
```
djarecka, line 35:
```suggestion
  * this should not return an error due to the problems with metadata (should be validated before)
```

doc/design/doi-generation-1.md
djarecka: please see https://github.com/dandi/dandi-cli/pull/596

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,41 @@
...
+
+## dandi-api
+
+* If DOI generation raised any errors, return a 400 error describing the problem.
+* Otherwise, attempt to register the DOI identifier/body with Datacite.
+* If that raises any errors, log them so they are reported in Sentry. Regardless of the outcome, continue with the publish process.
```
dchiquito, line 41:
```suggestion
* If that raises any errors, log them so they are reported in Sentry. Regardless of the outcome, continue with the publish process.

## Validation process

### Assets
* Whenever an asset is created (I think this is the only way to set asset metadata), it kicks off an out of band job to validate the asset metadata.
* The task will grab the appropriate schema for the schema version (currently just the normal `asset.json`, but hopefully soon it will use a more restrictive `publishAsset.json`) and verify that the asset metadata validates against that schema. The result is stored on the asset.
* If it fails to validate, any draft dandisets which contain that asset will show in the API as `INVALID`, even if they would validate otherwise.

### Dandisets
* Whenever a new dandiset metadata is created, it kicks off an out of band job to validate the dandiset metadata. This is identical to the asset process, but uses the dandisets JSON schema.
* If any of the assets in a dandiset are `INVALID`, the dandiset is also `INVALID`. 

```

doc/design/doi-generation-1.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Validation process
+
+### Assets
+* Whenever an asset is created (I think this is the only way to set asset metadata), it kicks off an out of band job to validate the asset metadata.
```
satra, line 46:
```suggestion
* Whenever an asset is created (I think this is the only way to set asset metadata), it kicks off an out of band job to validate the asset metadata. The asset metadata should be augmented with whatever is needed to satisfy the publishasset requirements. Thus this task may involve several out of band operations (e.g, calculate sha256, add other required metadata).
```

satra:
@dchiquito - there are a bunch of suggestions that has been around for a bit. please review/commit.

doc/design/doi-generation-1.md
dchiquito: I'm not clear on the value of checksumming the manifests. It doesn't provide any kind of security since it's stored in the same place as the content being checksummed. It doesn't really give any additional data integrity during the download process that HTTPS doesn't already give. Hypothetically it could safe guard against the API server uploading corrupted data, but this is not the best way to solve that problem.

Could you clarify what this is supposed to achieve?

satra: i'm not sure aws provides checksums for things. this would mostly be a way for any user to verify that the dandiset.yaml or the assets.yaml have not changed. if the aws header already provides this, great, otherwise it's just another file containing the checksums of the files. i guess the AWS e-tag would be sufficient for the moment.

PR #220, opened 2021-04-06
Add rudimentary version reporting

dchiquito:


dandiapi/api/views/info.py
```diff
@@ -22,10 +24,17 @@ class ApiInfoSerializer(serializers.Serializer):
 )
 @api_view()
 def info_view(self):
+    # Right now the production deployment is tracking master, which makes tags basically useless.
+    # TODO: Deploy from tags, and report that tag here.
+    resp = requests.get('https://api.github.com/repos/dandi/dandi-api/releases')
+    tags = [release['tag_name'] for release in resp.json()]
+    tag = tags[0]
+
```
yarikoptic, line 32: May be versioneer used by dandi-cli could be used here too.
With this setup here provided version information is really nothing to rely on (outdated container used in tests would happily report last release, no information on how far in commits after the tag, etc)

dchiquito: We would like to do that ASAP, but it would require an overhaul of how we deploy the project, so this is an absolute minimum effort stopgap measure.
We are currently using Heroku's automatic deploy feature, which uses webhooks to redeploy the app whenever master changes. This is really useful for fast, iterative development like we've been doing.
Deploying from a tag in production would involve making a release, making a code change to the terraform in `dandi-infrastructure`, and deploying that change. We'd also have to add something to pass the deployed version into the deployed app. For tests that use `dandi-api` as a container, we'd have to figure out how to build new images for tags and specify those in the test config, and define a process for keeping those versions up to date.
All of that is possible but time consuming, and it will only slow down development once it's in place. Will this solution work until migration is complete?

yarikoptic: I don't know how heroku/webhooks deploy it, but FWIW, if it is just `pip install PATH-or-GITURL` and git is available, then I think versioneer would just work without release:
```shell
$> py=3; d=venvs/dev$py; virtualenv --python=python$py --system-site-packages $d && source $d/bin/activate && pip3 install  ~/proj/dandi/dandi-cli
...
$> dandi --version
0.10.0+72.g2ba68eb.dirty
```
(in your case it should not be .dirty) so it would include information from git on how far from the release and even exact committish. And that is a semver:
```ipython
In [3]: semver.parse('0.10.0+72.g2ba68eb.dirty')
Out[3]: 
OrderedDict([('major', 0),
             ('minor', 10),
             ('patch', 0),
             ('prerelease', None),
             ('build', '72.g2ba68eb.dirty')])
```
complimentary, could provide both
```json
{
  'version': '0.10.0',
  'build': '0.10.0+72.g2ba68eb.dirty'
}
```
although I guess would have just used original one as the version since already semver.

> Will this solution work until migration is complete?

my comment is just comment/recommendation. I don't think we would use/rely on it ATM since it could be just misleading and anyways no guarantees yet on API stability yet within 0. series, right? (or could we rely that 0.MINOR would be getting boost whenever breaking dandi-cli changes introduced and thus needing newer client?)
Meanwhile added an idea on having `dandi-cli-min-version` field: https://github.com/dandi/dandi-api/issues/36

dchiquito: https://github.com/dandi/dandi-api/pull/221 should resolve this I hope

PR #221, opened 2021-04-07
Add versioneer

dchiquito:
@yarikoptic does this satisfy your needs?


yarikoptic:
> @yarikoptic does this satisfy your needs?

;-) Let's merge and hope so?

PR #223, opened 2021-04-08
Bucket prefix

### LINKED ISSUE #214: Move away from `dev/` prefix on the sponsored S3 bucket

dchiquito:
Rather than remove the `"dev/"` from all the bucket paths, I opted to replace it with a new setting, `DANDI_DANDISETS_BUCKET_PREFIX`. This moves where the files are stored from code to configuration, so we won't need any more PRs like this one to switch it on/off.

Fixes #214


waxlamp (commented):
Yes, great idea.

PR #228, opened 2021-04-09
Disable formatting for generated versioneer files

dchiquito:
Running `tox -e format` will attempt to format the files generated by versioneer, which adds unintended cruft to PRs. Rather than keep those generated files formatted properly, just flag those files so that `black` and `isort` skip them.


yarikoptic:
it is better to avoid modifying those files in themselves, so you could later inject newer version by simply copying it in.

FWIW we have in dandi-cli:
```
$> cat .pre-commit-config.yaml 
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v3.4.0
    hooks:
    -   id: trailing-whitespace
    -   id: end-of-file-fixer
    -   id: check-yaml
    -   id: check-added-large-files
-   repo: https://github.com/psf/black
    rev: 20.8b1
    hooks:
    -   id: black
        exclude: ^(dandi/_version\.py|dandi/due\.py|versioneer\.py)$
-   repo: https://github.com/PyCQA/isort
    rev: 5.7.0
    hooks:
    -   id: isort
        exclude: ^(dandi/_version\.py|dandi/due\.py|versioneer\.py)$
```

dchiquito:
@yarikoptic Good point, I will move the exemption to the config file.

PR #229, opened 2021-04-09
Update version modified date when assets are modified

dchiquito:


yarikoptic:
FWIW, checked on https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/010005/draft by uploading new assets, "Last updated" reflected the change -- good, thanks!

PR #236, opened 2021-04-15
Version metadata endpoint

dchiquito:
The version detail endpoint
(/dandisets/{dandiset_id}/versions/{version_id}) has been changed to
only return the version metadata. To avoid changing everything in the
frontend to use the metadata schema, a new version action info was
added.

Requires dandi/dandiarchive#628 to keep the UI from breaking.

Requires dandi/dandi-cli#575 to keep the CLI from breaking. I believe a new CLI release will also fix the tests.

### DIFF
--- a/dandiapi/api/tests/test_version.py
+++ b/dandiapi/api/tests/test_version.py
@@ -52,8 +52,18 @@ def test_version_rest_list(api_client, version):
 
 @pytest.mark.django_db
 def test_version_rest_retrieve(api_client, version):
+    assert (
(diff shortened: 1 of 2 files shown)


yarikoptic:
I guess we are ready to proceed on this one? would be nice to bring dandi-cli CI green again ;)

yarikoptic:
could this be merged to bring `dandi-cli` into green zone?

waxlamp (approved):

PR #237, opened 2021-04-19
Add script to clean up old blobs

dchiquito:


yarikoptic:
who/how (is it done via some django end point or just from command line) is to run this script? 
shouldn't it (or a similar one) also remove  `dev/` and `publish/` while at it?

PR #242, opened 2021-04-21
Add asset digests to asset metadata

dchiquito:
I'm holding off on merging this until the migration upload is finished to avoid any interruptions.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import List, Set
+from typing import Dict, List, Set
 import uuid
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


dandiapi/api/models/asset.py
```diff
@@ -53,6 +53,13 @@ class Meta:
     def references(self) -> int:
         return self.assets.count()
 
+    @property
+    def digest(self) -> Dict[str, str]:
+        digest = {'dandi:dandi-etag': self.etag}
+        if self.sha256:
+            digest['dandi:sha256'] = self.sha256
```
satra, line 60:
```suggestion
            digest['dandi:sha2-256'] = self.sha256
```

dandiapi/api/tests/test_asset.py
```diff
@@ -72,6 +72,35 @@ def test_asset_rest_retrieve(api_client, version, asset):
             f'/versions/{version.version}/assets/{asset.asset_id}/download/',
             HTTP_URL_RE,
         ],
+        'digest': {
+            'dandi:dandi-etag': asset.blob.etag,
+            'dandi:sha256': asset.blob.sha256,
```
satra, line 77:
```suggestion
            'dandi:sha2-256': asset.blob.sha256,
```

yarikoptic:
migration upload is finished and redone for no effect.  It was not yet explicitly validated though against girder, but I think it would be ok to unleash this offband fury

PR #246, opened 2021-04-23
Create a default OAuth application so devs can log in

### LINKED ISSUE #186: Restore the default OAuth application migration
I recently reset our migrations, and inadvertently deleted `0005_default_oauth_application.py` along with the rest. This migration set up a default OAuth application, which is necessary for new installations to be able to log in from the GUI. @waxlamp recently ran afoul of this after resetting his DB.

dchiquito:
Fixes #186 

Restores https://github.com/dandi/dandi-api/commit/e62310cc11fe7ddd5aa7663c78e8f466b23e8daa#diff-7fab13f3a7b88ebfee8c8c0742[…]1b1fe9856211c9cdbb8e0c4ba471fcc

### DIFF
--- a/dandiapi/api/migrations/0006_default_oauth_application.py
+++ b/dandiapi/api/migrations/0006_default_oauth_application.py
@@ -1,0 +1,42 @@
+# flake8: noqa N806
+
+from django.db import migrations
+from django.db.models import Q
+
+
+def create_application(apps, schema_editor):
+    Application = apps.get_model('oauth2_provider', 'Application')
+    name = 'DANDI GUI'
+    # This is specified so it matches the default value used in the frontend.
+    client_id = 'Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl'
+    if not Application.objects.filter(Q(name=name) | Q(client_id=client_id)).exists():
+        application = Application(
+            # Production instances should change this.
+            client_id='Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl',
+            # Production instances must change this.
+            redirect_uris='http://localhost:8085/',
+            # These values should not be modified.
+            client_type='public',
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):
Tested locally

dandiapi/api/migrations/0006_default_oauth_application.py
```diff
@@ -0,0 +1,42 @@
...
+    Application = apps.get_model('oauth2_provider', 'Application')
+    name = 'DANDI GUI'
+    # This is specified so it matches the default value used in the frontend.
+    client_id = 'Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl'
+    if not Application.objects.filter(Q(name=name) | Q(client_id=client_id)).exists():
+        application = Application(
+            # Production instances should change this.
+            client_id='Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl',
+            # Production instances must change this.
+            redirect_uris='http://localhost:8085/',
```
yarikoptic, line 17: set env var to provide it?

yarikoptic:
could you describe how it is to be used? or ideally provide implementation within/for/instead https://github.com/dandi/dandi-api-webshots/pull/11

dandiapi/api/migrations/0006_default_oauth_application.py
dchiquito: `http://localhost:8085/` is always correct for this context.

dchiquito:
@yarikoptic This is for setting up dev environments. Since OAuth is the required login scheme now, you need to have an OAuth application in the DB to authenticate with when you are spinning up a local environment. Otherwise, when you click `Login` in a local web GUI, the server doesn't have anything to log you in to.

This just automates a step that users would otherwise have to do manually.

PR #247, opened 2021-04-23
Update steps in dev/django.Dockerfile

AlmightyYakob:
There were some missing steps in `dev/django.Dockerfile` which caused `docker-compose build` to fail, as it couldn't fine `versioneer.py` or `setup.cfg`.


waxlamp (approved):
This solved my problem. Thanks @AlmightyYakob!

dchiquito (approved):

PR #261, opened 2021-05-05
ENH: add "extended" option to /paths endpoint

yarikoptic:
ATM there is a growing number of issues which related to deficiencies  in
listing of assets, and some of which relate to parity with older girder-based
interface, e.g

- no sizes for assets (or directories) shown:
  https://github.com/dandi/dandiarchive/issues/644
  https://github.com/dandi/dandiarchive/issues/586

- getting a listing with download buttons takes way too long:
  https://github.com/dandi/dandiarchive/issues/638

/paths endpoint already loops through the list of available assets but returns
only deduced paths.  IMHO this would be the logical place to improve it to
(optionally) make returned list virtually at no processing cost, and not only
providing a solution to aforementioned issues but also adding possibility to
enhance "View data" interface with

- displaying sizes not only for assets but directories as well
  (even beating old girder-based implementation which IIRC was only
  displaying size of files immediately under the path)
- sorting by size, created, modified

If decided to proceed this way, TODOs to finalize:

- [ ] unit-test
- [?] better sorting for "extended" - should pair "asset_id" (or None
      for folders) and "path" so we get subfolders first and then assets. edit: although -- might better remove sorting altogether since ATM it doesn't care about sorting folders first (thus we UI probably does it anyways). If desired, a "sort" option could be introduced.

PS I have not tried it locally since failed to re-bootstrap development environment following README.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -125,25 +125,51 @@ class Asset(TimeStampedModel):
         return Asset(path=asset.path, blob=asset.blob, metadata=asset.metadata)
 
     @classmethod
-    def get_path(cls, path_prefix: str, qs: List[str]) -> Set:
+    def get_path(cls, path_prefix: str, qs: List[dict], extended: bool) -> List:
         """
         Return the unique files/directories that directly reside under the specified path.
 
-        The specified path must be a folder (must end with a slash).
+        Parameters
+        ----------
+        path_prefix: str
+          The specified path must be a folder (must end with a slash).
+        extended: bool, optional
+          If True, will return a list of records: original asset records or summary
+          records for directories
         """
         if not path_prefix:
             path_prefix = '/'
         prefix_parts = [part for part in path_prefix.split('/') if part]
-        paths = set()
+        paths = dict() if extended else set()
         for asset in qs:
             path_parts = [part for part in asset['path'].split('/') if part]
 
             # Pivot index is -1 (include all path parts) if prefix is '/'
             pivot_index = path_parts.index(prefix_parts[-1]) if len(prefix_parts) else -1
             base_path, *remainder = path_parts[pivot_index + 1 :]
-            paths.add(f'{base_path}/' if len(remainder) else base_path)
-
-        return sorted(paths)
+            if not extended:
+                paths.add(f'{base_path}/' if len(remainder) else base_path)
+            elif len(remainder):
+                if base_path not in paths:
+                    # Initiate the record from the asset.  It will only miss
+                    # asset_id field and otherwise be the same with cumulative information
+                    paths[base_path] = {
+                        "path": f'{base_path}/',
+                        "size": asset['size'],
+                        "created": asset['created'],
+                        "modified": asset['modified'],
+                    }
+                else:
+                    # update record
+                    rec = paths[base_path]
+                    # Assuming we store in UTC but might better
+                    rec['size'] += asset['size']
+                    rec['created'] = min(rec['created'], asset['created'])  # earliest
+                    rec['modified'] = max(rec['modified'], asset['modified'])  # latest
+            else:
+                # asset
+                paths[base_path] = asset
+        return sorted(paths.values() if extended else paths)
 
     @classmethod
     def total_size(cls):
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Dict, List, Set
+from typing import Dict, List
 import uuid
 
 from django.conf import settings
@@ -125,7 +125,7 @@ class Asset(TimeStampedModel):
         return Asset(path=asset.path, blob=asset.blob, metadata=asset.metadata)
 
     @classmethod
-    def get_path(cls, path_prefix: str, qs: List[dict], extended: bool=False) -> List:
+    def get_path(cls, path_prefix: str, qs: List[dict], extended: bool = False) -> List:
         """
         Return the unique files/directories that directly reside under the specified path.
 
@@ -154,10 +154,10 @@ class Asset(TimeStampedModel):
                     # Initiate the record from the asset.  It will only miss
                     # asset_id field and otherwise be the same with cumulative information
                     paths[base_path] = {
-                        "path": f'{base_path}/',
-                        "size": asset['size'],
-                        "created": asset['created'],
-                        "modified": asset['modified'],
+                        'path': f'{base_path}/',
+                        'size': asset['size'],
+                        'created': asset['created'],
+                        'modified': asset['modified'],
                     }
                 else:
                     # update record
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -125,7 +125,7 @@ class Asset(TimeStampedModel):
         return Asset(path=asset.path, blob=asset.blob, metadata=asset.metadata)
 
     @classmethod
-    def get_path(cls, path_prefix: str, qs: List[dict], extended: bool) -> List:
+    def get_path(cls, path_prefix: str, qs: List[dict], extended: bool=False) -> List:
         """
         Return the unique files/directories that directly reside under the specified path.
 
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -252,7 +252,8 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
 
     @swagger_auto_schema(
         manual_parameters=[
-            openapi.Parameter('path_prefix', openapi.IN_QUERY, type=openapi.TYPE_STRING)
+            openapi.Parameter('path_prefix', openapi.IN_QUERY, type=openapi.TYPE_STRING),
+            openapi.Parameter('extended', openapi.IN_QUERY, type=openapi.TYPE_BOOLEAN)
         ],
         responses={
             200: openapi.Schema(
@@ -270,11 +271,12 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
         (to refer to the root folder) must be the empty string.
         """
         path_prefix: str = self.request.query_params.get('path_prefix') or ''
+        extended: bool = self.request.query_params.get('extended') or False
         # Enforce trailing slash
         if path_prefix and path_prefix[-1] != '/':
             path_prefix = f'{path_prefix}/'
         qs = self.get_queryset().filter(path__startswith=path_prefix).values()
 
-        return Response(Asset.get_path(path_prefix, qs))
+        return Response(Asset.get_path(path_prefix, qs, extended=extended))
 
     # TODO: add create to forge an asset from a validation


dandiapi/api/views/asset.py
```diff
@@ -270,11 +271,12 @@ def paths(self, request, **kwargs):
...
         """
         path_prefix: str = self.request.query_params.get('path_prefix') or ''
+        extended: bool = self.request.query_params.get('extended') or False
         # Enforce trailing slash
         if path_prefix and path_prefix[-1] != '/':
             path_prefix = f'{path_prefix}/'
         qs = self.get_queryset().filter(path__startswith=path_prefix).values()
 
-        return Response(Asset.get_path(path_prefix, qs))
+        return Response(Asset.get_path(path_prefix, qs, extended=extended))
```
yarikoptic, line 280: @dchiquito Does `Response` take advantage of a generator, in that e.g. would return first "page" whenever it is ready without waiting for all values to be generated?  then `get_path` could become a generator, yielding unsorted entries and it would be for UI to implement any sorting strategy it desires (folders first, size based, etc). This could then even reduce memory consumption by simply yielding individual assets results without memorizing them and only complimenting results at the end with folders' entries

yarikoptic:
<details>
<summary>ailing test looks relevant so I must have introduced a bug... </summary> 

```shell
_______________________ test_upload_validate_wrong_size ________________________

api_client = <rest_framework.test.APIClient object at 0x7f779da35490>
user = <User: williamsjessica@example.net>, upload = <Upload: Upload object (9)>

    @pytest.mark.django_db
    def test_upload_validate_wrong_size(api_client, user, upload):
        api_client.force_authenticate(user=user)
    
        wrong_content = b'not 100 bytes'
>       upload.blob.save(upload.blob.name, ContentFile(wrong_content))

api_client = <rest_framework.test.APIClient object at 0x7f779da35490>
upload     = <Upload: Upload object (9)>
user       = <User: williamsjessica@example.net>
wrong_content = b'not 100 bytes'

dandiapi/api/tests/test_upload.py:231: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/test/lib/python3.8/site-packages/django/db/models/fields/files.py:88: in save
    name = self.field.generate_filename(self.instance, name)
        content    = <ContentFile: Raw content>
        name       = 'test-prefix/example_yD4jW19.dat'
        save       = True
        self       = <FieldFile: test-prefix/example_yD4jW19.dat>
.tox/test/lib/python3.8/site-packages/django/db/models/fields/files.py:316: in generate_filename
    filename = validate_file_name(filename)
        filename   = 'test-prefix/example_yD4jW19.dat'
        instance   = <Upload: Upload object (9)>
        self       = <django.db.models.fields.files.FileField: blob>
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'test-prefix/example_yD4jW19.dat'

    def validate_file_name(name):
        if name != os.path.basename(name):
>           raise SuspiciousFileOperation("File name '%s' includes path elements" % name)
E           django.core.exceptions.SuspiciousFileOperation: File name 'test-prefix/example_yD4jW19.dat' includes path elements

name       = 'test-prefix/example_yD4jW19.dat'

.tox/test/lib/python3.8/site-packages/django/core/files/utils.py:8: SuspiciousFileOperation
```
</details>
but nope -- wasn't me -- happened on master too recently (nightly) https://app.circleci.com/pipelines/github/dandi/dandi-api/594/workflows/36072dfe-0c63-416c-a894-712fa52a04cd/jobs/601

dandiapi/api/models/asset.py
```diff
@@ -125,25 +125,51 @@ def copy(cls, asset):
...
+                paths.add(f'{base_path}/' if len(remainder) else base_path)
+            elif len(remainder):
+                if base_path not in paths:
+                    # Initiate the record from the asset.  It will only miss
+                    # asset_id field and otherwise be the same with cumulative information
+                    paths[base_path] = {
+                        'path': f'{base_path}/',
+                        'size': asset['size'],
+                        'created': asset['created'],
+                        'modified': asset['modified'],
```
satra, line 160: wouldn't it be useful to add assetid and download link here as well? mostly because someday we will show metadata of asset. adding download makes the display super fast.

yarikoptic: on one hand: web ui can produce those based on knowing the dataset id (just as fast). 
on the other: I do not think we even know the dataset_id or version here (in models) and download urls are dataset/version specific (to provide content-disposition).  But it could be done in the `view` portion: it could enhance those records with urls in `extended` mode indeed making web-ui really just a renderer of information and making it possible for other clients (e.g. `dandi-cli`) to readily provide that information or for API to become easier to use for similar use cases/views.  So - let's do it.
I guess for consistency with the model that field should be `contentUrl` and contain a list (of a single ATM) URL (or should we just keep it a singular string).  And it will be populated only for assets, not for directories (ATM at least, although could be some dandi:// url consumable by dandi-cli, ;-) )

yarikoptic: and just spotted:

> wouldn't it be useful to add assetid ...

this is for a folder -- there is no assetid for it.  For assets - records will contain it and we are just adding/returning them as is (nice and cheap ;))

satra: folder's don't have creaded/modified. the above code specifically is about info for an asset. so i'm assuming you have asset details here. remember, you can't construct the download path without knowing assetid. so that would at least be essential for the ui to create the path.

yarikoptic: > folder's don't have creaded/modified

correct. That is why I am taking min of created and max of modified across assets.  IMHO it is useful information to display. We should just document somewhere/how that it is not datetime of the folder but a "summary" across assets.

> the above code specifically is about info for an asset.

not entirely -- it traverses assets to figure out paths, including "folders/"  which are to be displayed.  It would be super useful to provide `size` for each folder, and it would come virtually "for free".

> remember, you can't construct the download path without knowing assetid. so that would at least be essential for the ui to create the path.

asset_ids are known for assets here.  code above (in view) also knows about which dandiset/version and can quickly construct contentUrl for each asset (and skip "folder/" paths records of which will lack "asset_id")

satra: > That is why I am taking min of created and max of modified across assets.

i think that's happening in a different branch of this else. where i put this comment, i don't think asset uuid is being added. this code comment is specific to a single asset not a folder.

yarikoptic: it is impossible to comprehend now the indentation levels due to this comment, here is the code snippet:
```python
            if not extended:
                paths.add(f'{base_path}/' if len(remainder) else base_path)
            elif len(remainder):
                if base_path not in paths:
                    # Initiate the record from the asset.  It will only miss
                    # asset_id field and otherwise be the same with cumulative information
                    paths[base_path] = {
                        'path': f'{base_path}/',
                        'size': asset['size'],
                        'created': asset['created'],
                        'modified': asset['modified'],
                    }
                else:
                    # update record
                    rec = paths[base_path]
                    # Assuming we store in UTC but might better
                    rec['size'] += asset['size']
                    rec['created'] = min(rec['created'], asset['created'])  # earliest
                    rec['modified'] = max(rec['modified'], asset['modified'])  # latest
            else:
                # asset
                paths[base_path] = asset
```
I think it is correct.  First block -- no extended, 2nd - extended directory, 3rd - extended asset

satra: ok  yes. i see it better now.

satra:
on looking more closely at this and thinking of where it is used, we should discuss this function more. this query runs through all assets everytime, which is going to bog down the api, when multiple view requests are being made. if we forgo folder size for now, i think the query can still return (id, size for assets) and path for folders by using some of the postgres functions available to django, which would also make it more efficient.

yarikoptic:
I am all for optimization. But given currently maximal  time of 2.3 sec for listing top dir of heaviest 00020, I think there is no immediate dire need, but we do have immediate need for getting download urls which those PR would enable

satra:
got carried away a bit, haven't compared to the existing func

```python
import requests
dandiset = 26
url = f"https://api.dandiarchive.org/api/dandisets/{dandiset:06}/versions/draft/assets/"
querystring = {"page_size":"100"}
headers = {"Accept": "application/json"}
resp = requests.request("GET", url, headers=headers, params=querystring).json()
results = resp["results"]
while resp["next"]:
    resp = requests.request("GET", resp["next"], headers=headers).json()
    results.extend(resp["results"])
assert len(results) == resp["count"]
```

this is the two part function that returns any level of data. i'm not sure but may be what i'm doing on the dataframe can be done in postgres.

```python
def split_part(x, delim, ind):
    splits = x.split(delim)
    return splits[ind - 1] if ind <= len(splits) else np.nan

def get_paths(res, prefix):
    df = pd.DataFrame(res)
    return df.groupby(df.path[df.path.str.startswith(prefix)] \
       .apply(split_part, args=("/", len(prefix) - len(prefix.replace("/", "")) + 1))) \
       .apply(lambda x: x[["asset_id", "size"]].to_dict(orient='records') \
               if len(x) == 1 and 
                  not x.path.str.replace(prefix, '').str.contains('/').any() 
               else x["size"].sum())
```

and here are some results. i'm not timing the first part, which i'm assuming should be quick on the django + postgres side

for 26
```python
In [147]: %time get_paths(results, "")
CPU times: user 8.75 ms, sys: 926 µs, total: 9.67 ms
Wall time: 8.99 ms
Out[147]: 
path
derivatives     62098107735
rawdata        836971584508
dtype: int64

In [148]: %time get_paths(results, "rawdata")
CPU times: user 8.11 ms, sys: 747 µs, total: 8.86 ms
Wall time: 8.36 ms
Out[148]: 
path
rawdata    836971584508
dtype: int64

In [149]: %time get_paths(results, "rawdata/")
CPU times: user 9.47 ms, sys: 715 µs, total: 10.2 ms
Wall time: 9.56 ms
Out[149]: 
path
sub-EXC022      1893938166
sub-I46       828556490369
sub_I48          307238533
sub_I55         6213917440
dtype: int64
```

for 20
```python
In [155]: %time get_paths(results, "")
CPU times: user 211 ms, sys: 4.06 ms, total: 215 ms
Wall time: 212 ms
Out[155]: 
path
sub-1001658946     36804772
sub-1001658958    133321673
sub-599387254      17881801
sub-601462915      22031019
sub-601462926      19310074
                    ...    
sub-992385481     186360022
sub-992385491     237421314
sub-992385538      47048954
sub-992385550     137127488
sub-992435487      49776506
Length: 1040, dtype: int64

In [156]: %time get_paths(results, "sub-992385481")
CPU times: user 13 ms, sys: 1.53 ms, total: 14.6 ms
Wall time: 13.1 ms
Out[156]: 
path
sub-992385481    186360022
dtype: int64

In [157]: %time get_paths(results, "sub-992385481/")
CPU times: user 17.2 ms, sys: 1.26 ms, total: 18.5 ms
Wall time: 17.3 ms
Out[157]: 
path
sub-992385481_ses-993243447_icephys.nwb    [{'asset_id': 'c1b4ebb4-2e54-469f-bda6-1b5f5df...
sub-992385481_ses-993245664_icephys.nwb    [{'asset_id': '33a268f1-b0f0-4a33-87e7-0367ea3...
sub-992385481_ses-993254986_icephys.nwb    [{'asset_id': 'bb441134-5849-49c1-8f47-f4819e6...
sub-992385481_ses-993257566_icephys.nwb    [{'asset_id': 'd44beed2-a967-47bb-a7b2-9aa2fdc...
dtype: object
```

satra:
don't worry about the timings here. the code that @yarikoptic wrote runs fine. the only thing i was trying to do with pandas was to get it closer to postgres semantics to see which pieces could be executed on the db itself.

dandiapi/api/models/asset.py
```diff
@@ -125,25 +125,51 @@ def copy(cls, asset):
...
-        return sorted(paths)
+            if not extended:
+                paths.add(f'{base_path}/' if len(remainder) else base_path)
+            elif len(remainder):
+                if base_path not in paths:
+                    # Initiate the record from the asset.  It will only miss
+                    # asset_id field and otherwise be the same with cumulative information
+                    paths[base_path] = {
+                        'path': f'{base_path}/',
+                        'size': asset['size'],
```
satra, line 158:
```suggestion
                        'size': asset['size'],
                        'identifier': asset['identifier'],
```

dandiapi/api/models/asset.py
```diff
@@ -125,25 +125,51 @@ def copy(cls, asset):
...
+                        'created': asset['created'],
+                        'modified': asset['modified'],
+                    }
+                else:
+                    # update record
+                    rec = paths[base_path]
+                    # Assuming we store in UTC but might better
```
satra, line 165:
```suggestion
                    # Assuming we store in UTC but might better
                    del rec['identifier']
```

satra:
@yarikoptic  and @dchiquito - any reason not to make this the extended option the default? this isn't a lot of extra bytes to send since this page is quite specific to a given level.  

also let's please add asset id at the file level. i left some suggestions. this will also allow the UI to put a link to display  metadata about a file.

yarikoptic:
> @yarikoptic and @dchiquito - any reason not to make this the extended option the default? this isn't a lot of extra bytes to send since this page is quite specific to a given level.

Breaking change.  Would need coordination with web-ui etc. not to mentioned the boost of major to the API version  (if we are to stay kosher) ;-)  Since easy to specify the option, I would keep it as non-default

dandiapi/api/models/asset.py
yarikoptic: What would be the purpose/use-case for "assign asset's identifier to a folder if there is only a single asset in the folder"?

yarikoptic: Have you watched https://www.youtube.com/watch?v=uLx7-jrji9o already?

yarikoptic:
> also let's please add asset id at the file level. i left some suggestions. this will also allow the UI to put a link to display metadata about a file.

an entire asset record is passed along in the last 
```
            else:
                # asset
                paths[base_path] = asset
```
so we should have each asset_id, previous block is only for a folder which should not have any id since we do not have an entity folder.

satra:
> Breaking change. 

i would just coordinate this with the UI and CLI now. we haven't really exposed the API properly to folks. hence i would rather see it cleaned rather than thinking backwards compatibility right this minute.

satra:
i'm not reading code well. i missed that the whole asset was indeed returned for file level things.

yarikoptic:
> i would just coordinate this with the UI and CLI now. 

FWIW CLI would not care too much, besides needing to relax regex a bit (doesn't matter if it is extended or not -- AFAIK we do not use that endpoint ATM) ... while at it - filed https://github.com/dandi/dandi-cli/issues/617 ;)

> we haven't really exposed the API properly to folks. hence i would rather see it cleaned rather than thinking backwards compatibility right this minute.

I thought to argue with: when do you think would be the moment in time when we decide that "**now** we should adhere to semver"?  But hey -- dandi-api tagged as 0.0.1, we return `0+unknown` for its version: https://github.com/dandi/dandi-api/issues/222 , so I guess indeed we still can break things left and right although I would have preferred not to.  But at least we should accompany it with 0.1.0 release of dandi-api!

yarikoptic:
> i'm not reading code well. i missed that the whole asset was indeed returned for file level things.

yeah, I even thought to proceed with aforementioned RF to make it a cleaner spaghetti ... but reluctant to do it until @dchiquito chimes in - may be he would love my current singular noodle   ;)

dandiapi/api/views/asset.py
dchiquito: `Response` is unpaginated. You would need to explicitly create a new `DandiPagination`, use it (I forget the exact magic words), and return a `Response` with the results.

yarikoptic: forget that I have mentioned pagination since it seems it is some other layer somewhere since I see that `/paths` is supporting pagination according to interface in swagger.  The questions then are
- Can `Response` be provided a generator and not a list?
- Can we stop sorting here (at dandi-api level) since web-ui must be sorting anyways?

satra: the endpoint does not support pagination yet, the swagger interface is misleading at the moment.

```
curl -X GET "https://api.dandiarchive.org/api/dandisets/000026/versions/draft/assets/paths/?page_size=5&path_prefix=rawdata%2Fsub-I46%2Fses-SPIM%2F" -H  "accept: application/json"
```

yarikoptic: oh, filed https://github.com/dandi/dandi-api/issues/266 . Questions stand.

AlmightyYakob:
@dchiquito and I have discussed this offline, and have a path forward that will address the concerns raised here. It'll be easiest for me to do this as a separate issue + PR, and reference this for posterity.

waxlamp:
> @dchiquito and I have discussed this offline, and have a path forward that will address the concerns raised here. It'll be easiest for me to do this as a separate issue + PR, and reference this for posterity.

Can we close this PR then?

yarikoptic:
But is there ETA for that alternative implementation?  IMHO might be faster/more efficient just to guide me on this one to have it finalized to your guys liking @waxlamp .

waxlamp:
> But is there ETA for that alternative implementation? IMHO might be faster/more efficient just to guide me on this one to have it finalized to your guys liking @waxlamp .

@AlmightyYakob is going to come up with an action plan soon (within a couple of days) and then we'll have a clearer idea of ETA.

Issue #267, opened 2021-05-07
add context to generated metadata for dandiset/asset


satra:
this would involve adding the following keyvalue to the returned metadata for both dandisets and assets:

`"@context": "https://raw.githubusercontent.com/dandi/schema/master/releases/0.3.0/context.json"`

where 0.3.0 is the schemaversion in the metadata.

PR #270, opened 2021-05-10
Add more computed fields to version metadata

### LINKED ISSUE #262: add citation and a few other computed fields to returned dandiset metadata
this is a quick pass. (also addresses: https://github.com/dandi/dandiarchive/issues/645)

to add to dandiset metadata:
- version (currently missing)
- id = f"{identifier}/{version}"
- url =  f"https://identifiers.org/{id}"
- citation (see funcs below)

here is a simple piece of code. if we add this to api and add the citation field to the returned metadata. that would be the easiest thing to start with. and if we add "id" addition to the API which is identifier + version

we may want to store the gif in the assets of the web ui.

```python
from datetime import datetime

def orcid(contrib, publish=False):
    if "identifier" in contrib and contrib["identifier"] and not publish:
        return f'<a href="https://orcid.org/{contrib["identifier"]}"><img src="https://info.orcid.org/wp-content/uploads/2020/12/orcid_16x16.gif" /></a>'
    return ''

def citation(metadata):
    cl = "; ".join([val["name"]+orcid(val) for val in metadata["contributor"] if val["includeInCitation"]])
    YEAR = datetime.now().year
    citestr = f"{cl} ({YEAR}) {metadata['name']}. Online: {metadata['url']}"
    return citestr
```

dchiquito:
* version
* id: i.e. 000000/draft
* url: i.e. https://identifiers.org/000000/draft
* citation: Jane Doe; John Doe (2021) Dandiset. Online: ...

Fixes #262 , #267

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -88,7 +88,7 @@ class Version(TimeStampedModel):
         year = datetime.datetime.now().year
         name = metadata['name']
         url = metadata['url']
-        if 'contributor' not in metadata:
+        if 'contributor' not in metadata or not metadata['contributor']:
             return f'{name} ({year}). Online: {url}'
         cl = '; '.join([val['name'] for val in metadata['contributor'] if val['includeInCitation']])
         citation = f'{cl} ({year}) {name}. Online: {url}'
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -91,8 +91,8 @@ class Version(TimeStampedModel):
         if 'contributor' not in metadata:
             return f'{name} ({year}). Online: {url}'
         cl = '; '.join([val['name'] for val in metadata['contributor'] if val['includeInCitation']])
-        citestr = f'{cl} ({year}) {name}. Online: {url}'
-        return citestr
+        citation = f'{cl} ({year}) {name}. Online: {url}'
+        return citation
 
     def _populate_metadata(self):
         metadata = {
@@ -101,7 +101,7 @@ class Version(TimeStampedModel):
             'identifier': f'DANDI:{self.dandiset.identifier}',
             'version': self.version,
             'id': f'{self.dandiset.identifier}/{self.version}',
-            'url': f'https://identifiers.org/{self.dandiset.identifier}/{self.version}',
+            'url': f'https://dandiarchive.org/{self.dandiset.identifier}/{self.version}',
         }
         metadata['citation'] = self.citation(metadata)
 
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -104,6 +104,12 @@ class Version(TimeStampedModel):
             'url': f'https://dandiarchive.org/{self.dandiset.identifier}/{self.version}',
         }
         metadata['citation'] = self.citation(metadata)
+        if 'schemaVersion' in metadata:
+            schema_version = metadata['schemaVersion']
+            metadata['@context'] = (
+                'https://raw.githubusercontent.com/dandi/schema/master/releases/'
+                f'{schema_version}/context.json'
+            )
 
         new: VersionMetadata
         new, created = VersionMetadata.objects.get_or_create(
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -88,10 +88,18 @@ class Version(TimeStampedModel):
         year = datetime.datetime.now().year
         name = metadata['name']
         url = metadata['url']
-        if 'contributor' not in metadata or not metadata['contributor']:
-            return f'{name} ({year}). Online: {url}'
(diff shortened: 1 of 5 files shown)


dandiapi/api/models/version.py
```diff
@@ -83,15 +83,32 @@ def make_version(cls, dandiset: Dandiset = None) -> str:
     def copy(cls, version):
         return Version(dandiset=version.dandiset, metadata=version.metadata)
 
+    @classmethod
+    def citation(cls, metadata):
+        year = datetime.datetime.now().year
+        name = metadata['name']
+        url = metadata['url']
+        if 'contributor' not in metadata:
```
AlmightyYakob, line 91: For when `metadata['contributors']` exists, but is an empty list
```suggestion
        if 'contributor' not in metadata or not metadata['contributor']:
```

dandiapi/api/tests/test_version.py
```diff
@@ -109,10 +111,16 @@ def test_version_rest_update(api_client, user, draft_version):
...
+    year = datetime.now().year
+    url = f'https://identifiers.org/{draft_version.dandiset.identifier}/draft'
     saved_metadata = {
         **new_metadata,
         'name': new_name,
         'identifier': f'DANDI:{draft_version.dandiset.identifier}',
+        'id': f'{draft_version.dandiset.identifier}/draft',
+        'version': 'draft',
+        'url': url,
+        'citation': f'{new_name} ({year}). Online: {url}',
```
AlmightyYakob, line 123: I could see an argument for just inserting the values from the existing metadata here, since (except for citation, due to year) you're really just testing that the values haven't changed. However, I don't feel strongly about this, just an observation.

dandiapi/api/models/version.py
```diff
@@ -83,15 +83,32 @@ def make_version(cls, dandiset: Dandiset = None) -> str:
...
+            return f'{name} ({year}). Online: {url}'
+        cl = '; '.join([val['name'] for val in metadata['contributor'] if val['includeInCitation']])
+        citestr = f'{cl} ({year}) {name}. Online: {url}'
+        return citestr
+
     def _populate_metadata(self):
+        metadata = {
+            **self.metadata.metadata,
```
satra, line 99:
```suggestion
            **self.metadata.metadata,
            "@context": f"https://raw.githubusercontent.com/dandi/schema/master/releases/{self.metadata.metadata['schemaVersion']}/context.json"
```

dandiapi/api/tests/test_version.py
```diff
@@ -156,10 +164,16 @@ def test_version_rest_update_large(api_client, user, draft_version):
...
     }
```
AlmightyYakob, line 178: Following my other comment, it might be good to have a specific test for the `citation` field, to test the possible conditions (I'd think mainly missing, empty, and populated).

satra (commented):
looks great. added some suggestions. one is to address #267 while this is being changed.

AlmightyYakob (changes requested):
Looks good! Just a few things

dandiapi/api/tests/test_version.py
dchiquito: They are inserted on line 188? `**new_metadata`?

dandiapi/api/tests/test_version.py
```diff
@@ -26,6 +28,77 @@ def test_version_make_version_save(mocker, dandiset, published_version_factory):
...
+    )
+
+
+@pytest.mark.django_db
+def test_version_metadata_contaxt(version):
+    version.metadata.metadata['schemaVersion'] = '6.6.6'
+    version.save()
+
+    assert version.metadata.metadata['@context'] == (
+        'https://raw.githubusercontent.com/dandi/schema/master/releases/6.6.6/context.json'
```
satra, line 98: some dangerous versions you are using here.

dandiapi/api/models/version.py
```diff
@@ -83,15 +83,46 @@ def make_version(cls, dandiset: Dandiset = None) -> str:
...
+                    val['name']
+                    for val in metadata['contributor']
+                    if 'includeInCitation' in val and val['includeInCitation']
+                ]
+            )
+            if cl:
+                citation = f'{cl} ({year}) {name}. Online: {url}'
+        return citation
+
     def _populate_metadata(self):
```
satra, line 105: this is only called for `draft` right? 

for any published dandiset, we would return the stored metadata without any changes. i.e. on publish, the metadata record returned by this + any published added info would be stored as the dandiset metadata. i.e. the metadata becomes static.

dchiquito: This method is only called when metadata is saved. As long as you don't somehow modify the metadata for the published version, the metadata won't be modified.

AlmightyYakob (approved):

satra (approved):

AlmightyYakob:
@dchiquito could you add documentation somewhere explaining that the script added in 331f1ff needs to be run manually? Maybe in `README.md`?

PR #251, opened 2021-04-27
Remove Girder code from API

jwodder:
The Girder code in dandiapi appears to be unused, and it needs to be removed so that the Docker test setup in dandi-cli can get rid of the Girder containers.

PR #272, opened 2021-05-10
Remove settings for integration with Girder 3

brianhelba:


brianhelba:
Closing in favor of #251, which was opened first and removes more code.

PR #278, opened 2021-05-12
Switch from social_account_added to user_signed_up

dchiquito:
It turns out that django-allauth only sends the `social_account_added`
when a new social account is connected to an existing user, not when a
new SocialAccount + User are registered. Instead we need to use the
`user_signed_up` signal and look up all of the social accounts for the
new user.

The change list is intimidating, but most of it is changing arguments from a `sociallogin` which contained a user+socialaccount, to directly specifying `user` and `socialaccount`.

### DIFF
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -1,6 +1,6 @@
 from typing import List
 
-from allauth.socialaccount.signals import social_account_added
+from allauth.account.signals import user_signed_up
 from django.core import mail
 from django.dispatch import receiver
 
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
The code looks fine to me, how are you testing this locally? Or are you able to at all?

dchiquito:
I set up a temporary real genuine GitHub OAuth application that was set to redirect to `http://localhost:8000/`, then signed up with my GitHub account. I was surprised it was allowed, but it worked fine. The emails sent on registration exactly as expected.

PR #279, opened 2021-05-13
Asset Metadata Validation

dchiquito:
This is just asset metadata validation, dandiset metadata validation is coming soon in a different PR for ease of consumption.


dandiapi/api/models/asset.py
```diff
@@ -117,6 +132,34 @@ def save(self, *args, **kwargs):
...
+        )
+
+        blob_url = self.blob.blob.url
+        metadata = {
+            **self.metadata.metadata,
+            'identifier': str(self.asset_id),
+            'contentUrl': [download_url, blob_url],
+            'contentSize': self.blob.size,
+            'digest': self.blob.digest,
+        }
```
satra, line 160: a few comments on this section. 

1. we will likely add a few things here similar to dandiset metadata (`id`, `context`, `schemaKey`)
1. is self.blob.blob.url now devoid of a signed url? it can be a versioned url, but we want to drop signed urls from the metadata return and metadata export on publish. the signed urls should only be used in the context of the UI, because content-disposition needs to be added, unless we can figure a different way out to do so.

dandiapi/api/tasks.py
```diff
@@ -56,7 +63,81 @@ def write_yamls(version_id: int) -> None:
...
+        if 'schemaVersion' not in metadata:
+            logger.info('schemaVersion not specified in metadata for asset %s', asset_id)
+            asset.status = AssetStatus.INVALID.name
+            asset.validation_error = 'schemaVersion not specified'
+            asset.save()
+            return
+        schema_version = metadata['schemaVersion']
+        schema_url = (
+            'https://raw.githubusercontent.com/dandi/schema/master/'
+            f'releases/{schema_version}/asset.json'
```
satra, line 120: just a heads up there will be a published-asset.json here, which will also include any fields required for publishing, which are likely to be computed fields anyway.

dandiapi/api/tasks.py
```diff
@@ -56,7 +63,81 @@ def write_yamls(version_id: int) -> None:
...
+        )
+        request = requests.get(schema_url)
+        request.raise_for_status()
+        schema = request.json()
+        jsonschema.validate(instance=metadata, schema=schema)
+    except jsonschema.exceptions.ValidationError as ve:
+        logger.info('Validation error for asset %s', asset_id)
```
satra, line 127: one of the issues with jsonschema validationa is that it returns errors one at a time (unless you know of a way to make this return all possible errors),. this would be quite inefficient for any end user. we may want to consider using the model from the CLI for validation, which returns all errors at once.

dandiapi/api/tasks.py
```diff
@@ -56,7 +63,81 @@ def write_yamls(version_id: int) -> None:
...
+# depending on which version the asset belongs to.
+def validate_asset_metadata(self, version_id: int, asset_id: int) -> None:
+    logger.info('Validating asset metadata for asset %s, version %s', asset_id, version_id)
+    asset = Asset.objects.get(id=asset_id)
+    version = Version.objects.get(id=version_id)
+
+    if asset.status == AssetStatus.VALIDATING.name:
+        # Another task is currently validating, let it finish
+        logger.info('Asset %s is already being validated, wait for it to complete', asset_id)
+        raise self.retry()
```
AlmightyYakob, line 103: What's the point in retrying if there's already a validation happening?

dandiapi/api/tasks.py
AlmightyYakob, line 120: If we're using this in multiple places in the code, maybe we should have a `schema_release_url` util function that accepts a `schema_version` parameter and returns the release folder url (where you could then append `asset.json`, or `dandiset.json`).

dandiapi/api/tasks.py
AlmightyYakob, line 127: A more specific message that the schema is invalid might be nice, unless you think that'll be covered by the error message itself. Otherwise it reads pretty similar to the message below.

dandiapi/api/tasks.py
```diff
@@ -56,7 +63,81 @@ def write_yamls(version_id: int) -> None:
...
     storage.save(assets_yaml_path, ContentFile(assets_yaml))
 
     logger.info('Wrote dandiset.yaml and assets.yaml for version %s', version_id)
+
+
+def format_as_index(indices):
+    """Render a JSON schema path as a series of indices."""
+    if not indices:
+        return ''
+    return '[%s]' % ']['.join(repr(index) for index in indices)
```
AlmightyYakob, line 78: This might make it a bit more readable, just a suggestion though feel free to ignore
```suggestion
    return '[' + ']['.join(repr(index) for index in indices) + ']'
```

dandiapi/api/models/asset.py
```diff
@@ -117,6 +132,34 @@ def save(self, *args, **kwargs):
...
+    def generate_metadata(self, version):
+        """
+        Generate the correct metadata for this asset given a version.
+
+        The route used to download an asset will be different depending on which version it belongs
+        to. Since the download url is included in the asset metadata, it must be injected rather
+        than stored. This method should always be used instead of asset.metadata.metadata.
+        """
+        # TODO use http://localhost:8000 for local deployments
+        download_url = 'https://api.dandiarchive.org' + reverse(
```
AlmightyYakob, line 144: We could configure this with an environment variable but that also seems messy. Not sure what to do here.

AlmightyYakob (changes requested):

dandiapi/api/tasks.py
dchiquito: The meditor should report validation errors, so users should rarely be surprised by errors here. Also, importing the correct version of the pydantic model would be considerably more difficult, so I think this will work sufficiently to enable/disable the publish button. We can work on more sophisticated error reporting later.

dchiquito: This line specifically is to note in the logs what the outcome of the task was. The error message saved to `validation_error` is quite specific, you can check the tests for a few examples.
The catchs below catch more generic errors, like a 404 on the schema or whatever.

satra: this may be less of a problem. the asset validation here should not result in issues that should have already been caught by the CLI. i.e. as long as people did not skip validation to upload, this issue should be moot. 

the meditor only reports validation errors for dandiset metadata not asset metadata, while this PR is largely about asset validation.

dandiapi/api/models/asset.py
satra: outside the PR is fine, but that should be done before this PR is merged. 

the blob url bit is important since those presigned urls expire and would definitely not be useful as a publish link. if we are validating the info should be valid/proper.

dandiapi/api/tests/test_tasks.py
```diff
@@ -105,8 +108,120 @@ def test_write_assets_yaml_already_exists(storage: Storage, version: Version, as
...
+    asset.metadata.save()
+
+    tasks.validate_asset_metadata(version.id, asset.id)
+
+    asset.refresh_from_db()
+
+    assert asset.status == Asset.Status.INVALID
+    assert asset.validation_error == (
+        '404 Client Error: Not Found for url: '
+        'https://raw.githubusercontent.com/dandi/schema/master/releases/xxx/asset.json'
```
AlmightyYakob, line 169: I wonder if we should catch this and report it as an invalid schema version?

AlmightyYakob (approved):
Left one small comment, but besides that you've addressed my concerns, so this looks good.

dandiapi/api/tests/test_tasks.py
dchiquito: We could, but when relying on an external API like this we still need to handle any possible API errors anyway, so I don't see much point in a special error message. `schemaVersion` will always be set automatically by the meditor or whatever generated the metadata, users shouldn't be setting it themselves anyway.

PR #281, opened 2021-05-14
Version metadata validation

dchiquito:
* Add `status` and `validation_error` columns to `Version` model
* Add task to validate version metadata
* Add property `valid` to `Version` to describe whether or not the version is valid for publish (valid metadata + valid metadata for all assets)
* Require valid version for publish to be enabled
* Add `status` and `validation_error` to version serializers

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -80,6 +80,7 @@ class Version(TimeStampedModel):
         # Import here to avoid dependency cycle
         from .asset import Asset
 
+        # Return False if any asset is not VALID
         return not self.assets.filter(~models.Q(status=Asset.Status.VALID)).exists()
 
     @staticmethod
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -36,6 +36,12 @@ def _get_default_version() -> str:
 class Version(TimeStampedModel):
     VERSION_REGEX = r'(0\.\d{6}\.\d{4})|draft'
 
+    class Status(models.TextChoices):
+        PENDING = 'Pending'
+        VALIDATING = 'Validating'
+        VALID = 'Valid'
+        INVALID = 'Invalid'
+
     dandiset = models.ForeignKey(Dandiset, related_name='versions', on_delete=models.CASCADE)
     metadata = models.ForeignKey(VersionMetadata, related_name='versions', on_delete=models.CASCADE)
     version = models.CharField(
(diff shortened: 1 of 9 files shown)


dandiapi/api/models/version.py
```diff
@@ -60,6 +72,16 @@ def name(self):
...
 
+    @property
+    def valid(self) -> bool:
+        if self.status != Version.Status.VALID:
+            return False
+
+        # Import here to avoid dependency cycle
+        from .asset import Asset
+
+        return not self.assets.filter(~models.Q(status=Asset.Status.VALID)).exists()
```
AlmightyYakob, line 83: A comment would be good, since this is slightly confusing to read at first glance
```suggestion
        # Return false if any of this version's assets have a status other than `VALID`
        return not self.assets.filter(~models.Q(status=Asset.Status.VALID)).exists()
```

dandiapi/api/views/version.py
```diff
@@ -101,6 +101,9 @@ def publish(self, request, **kwargs):
                 status=status.HTTP_405_METHOD_NOT_ALLOWED,
             )
 
+        if not old_version.valid:
+            return Response('Metadata is not valid', status=status.HTTP_400_BAD_REQUEST)
```
AlmightyYakob, line 105: It might be better to specify that this can happen if the version metadata _or_ any asset metadata is invalid.

dandiapi/api/tasks.py
```diff
@@ -146,3 +146,51 @@ def validate_asset_metadata(version_id: int, asset_id: int) -> None:
...
     asset.save()
+
+
+@shared_task
+@atomic
+def validate_version_metadata(version_id: int) -> None:
+    logger.info('Validating dandiset metadata for version %s', version_id)
+    version = Version.objects.get(id=version_id)
+
+    # Begin the validation process so no other validation tasks will run simultaneously
```
AlmightyYakob, line 157: Is this true? I don't see a check anywhere against the current state of the version validation.

dandiapi/api/tests/test_tasks.py
```diff
@@ -225,3 +225,92 @@ def test_validate_asset_metadata_malformed_keywords(version: Version, asset: Ass
...
+        # 'encodingFormat': 'application/x-nwb',
+    }
+    version.metadata.save()
+
+    tasks.validate_version_metadata(version.id)
+
+    version.refresh_from_db()
+
+    print(version.metadata.metadata)
+    print(version.validation_error)
```
AlmightyYakob, line 248:
```suggestion
```

AlmightyYakob (changes requested):
Just a few comments, looks good overall

dandiapi/api/tasks.py
dchiquito: It is not, I will remove it.

PR #282, opened 2021-05-14
ENH: disallow non-super users creating dandisets with a specified identifier

yarikoptic:
We have got dandiset 000105 leaping from 000070.  It seems that operation was initiated
from web-ui so in principle the issue is probably not related to this change.  But especially
now that migration is over, I think it would be wise to allow for creation of dandisets
with a given identifier to be allowed only to the superusers, so someone does not
create 999999 dandiset thus making us run of the "dandiset id namespace"

I am not quite sure if my use of `request.user.is_superuser` is kosher but I saw
```
dandiapi/api/views/version.py:        if not request.user.is_superuser:
```
so decided that it should be ok

### DIFF
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -222,8 +222,8 @@ def test_dandiset_rest_create(api_client, user):
 
 
 @pytest.mark.django_db
-def test_dandiset_rest_create_with_identifier(api_client, user):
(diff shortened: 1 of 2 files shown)


yarikoptic:
test failures are legit as long as CI setup does not use non-superuser for interactions.  Not sure which way we would like to go there.

dchiquito (approved):

PR #294, opened 2021-05-18
Actually validate version metadata

dchiquito:
The validate_version_metadata task was all set up, but was not actually
triggered when version metadata was modified.


AlmightyYakob (approved):
LGTM

PR #296, opened 2021-05-19
Published schemas

dchiquito:
Bump schema version to 0.3.1 so we can use `published-dandiset.json` and `published-asset.json`. This also requires updating tests to use valid metadata.

To support that, I changed the metadata factories to use a valid example Satra provided instead of randomized Faker metadata.


dandiapi/api/tests/factories.py
```diff
@@ -116,7 +151,30 @@ class AssetMetadataFactory(factory.django.DjangoModelFactory):
...
         model = AssetMetadata
 
-    metadata = factory.Faker('pydict', value_types=['str', 'float', 'int'])
+    @factory.lazy_attribute
+    def metadata(self):
+        return {
+            'schemaVersion': '0.3.1',
+            'encodingFormat': 'application/x-nwb',
+            'publishedBy': {
+                'id': 'urn:uuid:08fffc59-9f1b-44d6-8e02-6729d266d1b6',
```
satra, line 160:
```suggestion
                'id': uuid4().urn,
```

dandiapi/api/tests/factories.py
```diff
@@ -116,7 +151,30 @@ class AssetMetadataFactory(factory.django.DjangoModelFactory):
...
+                'name': 'DANDI publish',
+                'startDate': '2021-05-18T19:58:39.310338-04:00',
+                'endDate': '2021-05-18T19:58:39.310361-04:00',
```
satra, line 163: these two should reflect the start and end of the publishing activity.

dandiapi/api/tests/factories.py
```diff
@@ -116,7 +151,30 @@ class AssetMetadataFactory(factory.django.DjangoModelFactory):
...
+                'wasAssociatedWith': [
+                    {
+                        'id': 'RRID:SCR_017571',
+                        'name': 'DANDI API',
+                        'version': '0.1.0',
```
satra, line 168: this should pick up the version of the API.

dchiquito: For now the API is unversioned, but as soon as it is it will be added here.

dandiapi/api/tests/test_tasks.py
```diff
@@ -120,13 +120,6 @@ def test_write_assets_yaml_already_exists(storage: Storage, version: Version, as
 def test_validate_asset_metadata(version: Version, asset: Asset):
     version.assets.add(asset)
 
-    # This is the minimum asset metadata for schema version 0.3.0
-    asset.metadata.metadata = {
-        'schemaVersion': '0.3.0',
-        'encodingFormat': 'application/x-nwb',
-    }
-    asset.metadata.save()
-
```
AlmightyYakob, line 129: There are several places in this file for this old format for metadata is still used. It'd probably be good (for consistency) to also change them to only add/modify the things that they need to.

AlmightyYakob (commented):
LGTM, just left one comment. I like the use of `factory.lazy_attribute`.

dandiapi/api/tests/factories.py
dchiquito: So the whole `publishedBy` block needs to be populated at validation time without modifying the draft versions, then included in the newly published metadata? Are there any other fields that need this treatment?

satra: i believe these are all the fields that change on the publish trigger:
`id, publishedBy, datePublished, url, version, doi, manifestLocation, citation`

i think almost all of them can be set on version save as a function of the clicking/posting publish. the publishedBy has an endDate, which should correspond to datePublished

for all datetimes i was using `datetime.astimezone(datetime.now()).isoformat()` to preserve timezone info.

dchiquito: I will address these in a new PR, I don't want to dilute this one too much.

dandiapi/api/tests/test_tasks.py
dchiquito: I don't want to dilute this PR more than it already is, so I will file an issue

PR #298, opened 2021-05-19
BF: total size by asking for unique blobs associated with some version

### LINKED ISSUE #260: Discrepancy between size estimates/calculations
when i use the stats api, i get:
```
{
"dandiset_count":65
"published_dandiset_count":0
"user_count":66
"size":16363640643492
}
```
when i sum all the content of the size of dandisets by going through, i get:
15414647730404

```python
import requests
url = "https://api.dandiarchive.org/api/dandisets/?page_size=100"
response = requests.request("GET", url)
dandisets = response.json()["results"]
ds = []
for val in dandisets:
    ds.append({"identifier": val["identifier"], "size": val["draft_version"]["size"], 
               "updated": val["draft_version"]["modified"]})
df = pd.DataFrame(ds)
df["size"].sum()
```

yarikoptic:
Closes #260

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -148,7 +148,8 @@ class Asset(TimeStampedModel):
     @classmethod
     def total_size(cls):
         return (
-            cls.objects.values('blob__id', 'blob__size')
+            cls.objects.filter(versions__id__isnull=False)
+            .values('blob__id', 'blob__size')
             .distinct()
             .aggregate(size=models.Sum('blob__size'))['size']
             or 0
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -147,4 +147,9 @@ class Asset(TimeStampedModel):
 
     @classmethod
     def total_size(cls):
-        return cls.objects.aggregate(size=models.Sum('blob__size'))['size'] or 0
+        return (
+            cls.objects.values('blob__id', 'blob__size')
+            .distinct()
+            .aggregate(size=models.Sum('blob__size'))['size']
+            or 0
+        )
--- a/dandiapi/api/tests/test_stats.py
+++ b/dandiapi/api/tests/test_stats.py
@@ -38,7 +38,16 @@ def test_stats_user(api_client, user):
 
 
 @pytest.mark.django_db
-def test_stats_asset(api_client, asset):
-    stats = api_client.get('/api/stats/').data
+def test_stats_asset(api_client, asset, draft_version, published_version):
+    def get_size():
+        return api_client.get('/api/stats/').data['size']
 
-    assert stats['size'] == asset.size
+    # asset is not associated with any version
+    assert get_size() == 0
+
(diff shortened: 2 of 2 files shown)


yarikoptic:
well, tests pass.  Since no coverage report I have no clue if tests cover that code path. So it would need knowledgeable reviewer to see if what I have done makes any sense/has hope to work.

yarikoptic:
woohoo -- with a local instance running where I created two dandisets to which I uploaded the same singular asset from 000027, `master` version:
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H  "accept: application/json" 
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":37584}%  
```
and this PR:
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":18792}%
```
which 1. does not crash! 2. reflects reality;

then I replaced that file with another one of size `148880` in one of the dataset.  This branch
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":167672}%   
```
which is correct (148880 + 18792) and master
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":186464}%    
```
is wrong. then I replaced asset in the other dandiset with the same blob. `master`:
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":335344}%  
```
and this 
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":167672}% 
```
which is odd since I expected it to go back to 148880 since original blob was replaced in both dandisets ... khm.

yarikoptic:
pushed the fix for that, and now it reports:
```
$> curl -X GET "http://127.0.0.1:8000/api/stats/" -H "accept: application/json"
{"dandiset_count":2,"published_dandiset_count":0,"user_count":5,"size":148880}
```

so I think it is working.  will merge monday if no objections or further comments

yarikoptic:
finally the test has failed and it should have -- that asset is loose , i.e. not linked to any Version

yarikoptic:
@satra -- have a look - is that how we want to count/report the total size -- as an amount of unique "addressable" (from datasets) blobs?

satra:
@yarikoptic - looks good to me as long as `blob__id` is `dandi-etag` ?

satra:
or there is a one to one mapping.

yarikoptic:
good point, I would assume so since we deduplicate blobs.
Note: @dchiquito is doing his own version of a fix in #307

PR #302, opened 2021-05-21
First draft of publish process

dchiquito:
This is my current understanding of the publish process and associated machinery.
Most of the juicy bits are the description of `Assets` and the `Publish process` section.

### DIFF
--- a/doc/design/publish-1.md
+++ b/doc/design/publish-1.md
@@ -1,0 +1,139 @@
+# Moving Parts
+
+This is a summary of system components that are relevant to publish and how they relate to eachother.
+
+### Dandiset
+A DANDI dataset.
+Each dandiset is guaranteed to have a draft Version, and can also have any number of published Versions.
+
+### Version
+Versions consist of a JSON blob of metadata and any number of Assets.
+There are two kinds of Version: draft and published.
+Published Versions are immutable, only draft Versions are editable (except admins, who have superpowers).
+Versions also track their validation status.
+
+### AssetBlob
+This is basically just a pointer to a blob in S3.
+AssetBlobs are central to how we implement deduplication of blobs in S3.
+Multiple Assets can refer to the same AssetBlob, meaning we only need to store each blob in S3 once.
+
+Whenever a new AssetBlob is created, a background task runs to compute the sha256 checksum (a "digest") of the blob.
+This can take a long time for large blobs, as it must stream all of the data from S3.
+
+### Asset
+Assets consist of a JSON blob of metadata and a reference to an AssetBlob.
+Versions and Assets are many-to-many; a Version can have any number of Assets, and an Asset can belong to any number of Versions (even, hypothetically, versions in different Dandisets).
+Assets also track their validation status.
+
+Similarly to Versions, there are two kinds of Asset: draft and published.
+Draft Assets are only allowed to belong to draft Versions.
+Published Assets are basically the same as draft Assets, but have some extra metadata included that was defined during the publish process (`datePublished`, `publishedBy`, etc).
+
+Note that published Assets can belong to both draft and published Versions.
+
+Uploading new data or modifying the metadata of an Asset doesn't actually modify the Asset.
+Instead, it creates a new Asset and replaces the reference to that Asset in the Version.
+This ensures that other Versions which might refer to the original Asset are unmodified, while the Version containing the Asset being modified reflects the changes correctly.
+
+When a published Asset is modified in this way, the publish metadata is stripped out, effectively creating a new draft Asset.
+
+_In general_, Assets should be immutable; any modification will instead create a new draft Asset.
+* One exception to this is calculating the AssetBlob digest.
+When the digest completes, it will be injected directly into the metadata of any Assets referencing that AssetBlob.
+* The other exception is creating a published Asset during the publish process.
+This involves directly injecting the new metadata fields into the draft Asset, effectively turning it into a published Asset in place.
+
+### Metadata Schemas
+Version metadata and Asset metadata must conform to defined schemas.
+There are two sets of schemas.
+The first applies to both draft and published metadata, and is used by the web metadata editor.
+The second is more restrictive, and only applies to published metadata.
+It requires additional fields like `datePublished`, `publishedBy`, and `doi` that are only knowable at publish time.
+The second one is used to validate metadata prior to publish.
+
+The JSON schemas can be found at https://github.com/dandi/schema.
+
+The pydantic models can be found and imported from https://github.com/dandi/dandischema.
+dandi-api will use the pydantic models for validation.
+
+# Pre-publish
+
+## Uploading data / Modifying metadata
+Whenever changes are made to the draft Version of a Dandiset, metadata will almost always be modified.
+Several metadata fields are computed based on other properties of the Version or Asset, like `size` or`fileCount`.
+* Modifying Version metadata obviously alters the Version metadata.
+* Uploading a new Asset alters the `assetsSummary` of the Version metadata.
+* Uploading new data to an existing Asset (probably) alters the `contentSize` field of the Asset metadata.
+Note that the sha256 digest is included in the Asset metadata.
+* Modifying the Asset metadata obviously alters the Asset metadata, but also alters the `assetsSummary` of the Version metadata.
+Whenever a Version or Asset is saved, these computed fields are injected into the metadata.
+
+## Validation process
+Whenever metadata is modified, a validation task runs in the background to ensure that everything is ready for publish.
+There are separate tasks for version metadata validation and asset metadata validation which use different schemas.
+
+The validation tasks work by preparing the metadata as if to publish the Version or Asset, then checking if that new metadata conforms to the publish schema.
+
+Note that the asset metadata validation task actually requires both a Version and an Asset, since the `contentUrl` field of the Asset is different depending on which Version the Asset belongs to.
+
+Versions can have the following validation states:
+* `PENDING`
+* `VALIDATING`
+* `VALID`
+* `INVALID`
+* `PUBLISHED` (note that assets do not have this)
+
+The `PUBLISHED` state means that Versions are technically not `VALID` after they are published, so they cannot be publish repeatedly without changes.
+The next time a change is made to the metadata, the validation task will run on the version, which will change the state to `VALID`/`INVALID`, re-enabling publish.
+
+Assets can have the following validation states:
+* `PENDING`
+* `VALIDATING`
+* `VALID`
+* `INVALID`
+
+Because validation works with already computed metadata that is presumable of limited size, it is a very fast running task.
+
+# Publish
+Publishing a draft Version creates a new published Version that is a complete copy of the draft Version, but is immutable and permanent.
+
+Publish is started through an API call, presumably made by a user clicking the `Publish` button in the web GUI.
+Only dandiset owners have permission to use the API endpoint.
+* During development, only admins can publish.
+
+If a Version is not `VALID` or any of the Assets in the Version is not `VALID`, publish is not allowed.
+
+## Publish process
+1. The dandiset is locked so that no other publishes can happen simultaneously.
+
+2. A new published Version is created with the required publish metadata information.
+It is initialy empty.
+
+3. A new DOI is created that points to the new Version.
+This is an API call to an external service, https://datacite.org/.
+This involves digesting the Version metadata into something that can be used by Datacite.
+See the doi-generation-1.md design document for more details.
+
+4. For every draft Asset in the draft Version, the required publish metadata fields are injected, turning those Assets into published Assets.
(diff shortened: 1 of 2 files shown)


doc/design/publish-1.md
```diff
@@ -0,0 +1,139 @@
...
+This involves digesting the Version metadata into something that can be used by Datacite.
+See the doi-generation-1.md design document for more details.
+
+4. For every draft Asset in the draft Version, the required publish metadata fields are injected, turning those Assets into published Assets.
+Published Assets, e.g. Asssets that already have those fields from previous publishes, _are skipped_.
+Because they describe Assets that have already been published in a past version, there is no reason to re-publish them.
+
+5. Associate all of the publish Assets, new and old, with the new published Version.
+
+6. The dandiset is unlocked.
```
satra, line 123:
```suggestion
6. The dandiset is unlocked.

7. An email is sent to the owners that the Dandiset is published.
```

satra (approved):
thanks daniel. this looks great. as far as i can tell it covers everything we have discussed.

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+7. An email is sent to the owners that the Dandiset is published.
+
+7. Asynchronously, manifest files (`dandiset.yaml` and `assets.yaml`) are written out to S3 describing the newly published Version by a background task.
```
dchiquito, line 127:
```suggestion
8. Asynchronously, manifest files (`dandiset.yaml` and `assets.yaml`) are written out to S3 describing the newly published Version by a background task.
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
+# Moving Parts
+
+This is a summary of system components that are relevant to publish and how they relate to eachother.
```
yarikoptic, line 3:
```suggestion
This is a summary of system components that are relevant to publish and how they relate to each other.
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Dandiset
+A DANDI dataset.
+Each dandiset is guaranteed to have a draft Version, and can also have any number of published Versions.
```
yarikoptic, line 7: clarification: ATM "is guaranteed to have a draft Version" is true.  In the future we did think about possibility to archive dandisets, so no mutable `draft`  would be left behind, only published Versions.

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Version
+Versions consist of a JSON blob of metadata and any number of Assets.
```
yarikoptic, line 10:
```suggestion
Versions consist of a JSON blob of Dandiset metadata and any number of Assets.
```
right?

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+There are two kinds of Version: draft and published.
+Published Versions are immutable, only draft Versions are editable (except admins, who have superpowers).
+Versions also track their validation status.
```
yarikoptic, line 13: `### Version` is not under `### Dandiset` so a bit confusing here on which metadata (above suggestion) or validation we are talking about:  validation of dandiset record (metadata), of each asset, or a composite ( everything pertinent to Version: dandiset metadata + all assets metadata, etc).  If I got it right may be:
```suggestion
Versions also track their validation status (comprised of validation of dandiset and assets metadata validation, etc).
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+### AssetBlob
+This is basically just a pointer to a blob in S3.
```
yarikoptic, line 16:
```suggestion
This is basically just a pointer to a blob in `blobs/` on S3.
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+AssetBlobs are central to how we implement deduplication of blobs in S3.
+Multiple Assets can refer to the same AssetBlob, meaning we only need to store each blob in S3 once.
```
yarikoptic, line 18:
```suggestion
Multiple Assets can refer to the same AssetBlob, meaning we only need to store each blob in S3 once.
Each Asset could have a different metadata record for the same AssetBlob.
```
if we are to clarify, I think this might be valuable to add

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+Whenever a new AssetBlob is created, a background task runs to compute the sha256 checksum (a "digest") of the blob.
+This can take a long time for large blobs, as it must stream all of the data from S3.
```
yarikoptic, line 21:
```suggestion
This can take a long time for large blobs, as it must stream all of the data from S3.
Computed sha256 checksum is associated with the blob DB record (so no Asset metadata record is changed).
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+Similarly to Versions, there are two kinds of Asset: draft and published.
+Draft Assets are only allowed to belong to draft Versions.
+Published Assets are basically the same as draft Assets, but have some extra metadata included that was defined during the publish process (`datePublished`, `publishedBy`, etc).
+
+Note that published Assets can belong to both draft and published Versions.
+
+Uploading new data or modifying the metadata of an Asset doesn't actually modify the Asset.
+Instead, it creates a new Asset and replaces the reference to that Asset in the Version.
+This ensures that other Versions which might refer to the original Asset are unmodified, while the Version containing the Asset being modified reflects the changes correctly.
```
yarikoptic, line 36: I would reorder these two paragraphs to go from basic "Asset" behavior to IMHO yet to be finalized "published vs draft" Asset fresh discussion:
```suggestion
Uploading new data or modifying the metadata of an Asset doesn't actually modify the Asset.
Instead, it creates a new Asset and replaces the reference to that Asset in the Version.
This ensures that other Versions which might refer to the original Asset are unmodified, while the Version containing the Asset being modified reflects the changes correctly.

Similarly to Versions, there are two kinds of Asset: draft and published.
Draft Assets are only allowed to belong to draft Versions.
Published Assets are basically the same as draft Assets, but have some extra metadata included that was defined during the publish process (`datePublished`, `publishedBy`, etc).

Note that published Assets can belong to both draft and published Versions.
```

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+When a published Asset is modified in this way, the publish metadata is stripped out, effectively creating a new draft Asset.
+
+_In general_, Assets should be immutable; any modification will instead create a new draft Asset.
+* One exception to this is calculating the AssetBlob digest.
+When the digest completes, it will be injected directly into the metadata of any Assets referencing that AssetBlob.
+* The other exception is creating a published Asset during the publish process.
+This involves directly injecting the new metadata fields into the draft Asset, effectively turning it into a published Asset in place.
```
yarikoptic, line 44: per my question and brief discussion, I do not think we should bother with mutating Assets in DB, and can keep the notion of draft vs published asset metadata just implemented while returning asset record from an endpoint for dandiset_version.  If `version != "draft"` (thus published) -- we populate fields relevant to Published version.  If `draft` - we don't. Then we would stay consistent with the notion of immutable "Asset metadata as provided by user" in DB.
The same goes for 

> When the digest completes, it will be injected directly into the metadata of any Assets referencing that AssetBlob.

and we do not inject anything into actual asset metadata record but just populate it "on the fly" from blob DB record.
And we avoid the need for "the publish metadata is stripped out" which might get hairy IMHO.

If there is a desire to use `contentUrl` as from the first published version for an asset, could be achieved at DB query level by searching asset in Versions with minimal non-draft version.  Or we just KISS it and provide version specific contentUrl regardless.

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+There are two sets of schemas.
+The first applies to both draft and published metadata, and is used by the web metadata editor.
+The second is more restrictive, and only applies to published metadata.
+It requires additional fields like `datePublished`, `publishedBy`, and `doi` that are only knowable at publish time.
+The second one is used to validate metadata prior to publish.
+
+The JSON schemas can be found at https://github.com/dandi/schema.
+
+The pydantic models can be found and imported from https://github.com/dandi/dandischema.
+dandi-api will use the pydantic models for validation.
```
yarikoptic, line 57: At least ATM we have no provisioning for validating against different schema versions via pydantic and I do not see yet how we could in the long(er) run due to all possible dependencies on pydantic version/logic etc.  Multiple strategies could be explored (e.g. "upgrade provided metadata to the most recent pydantic model version and validate") but we do not yet have support for that even designed.  Filed https://github.com/dandi/dandischema/issues/5 .

In the short term, I guess we can go "pydantic model" way but would need to orchestrate reextraction (for assets) and upgrade (for dandisets) of all metadata records and make publish accept only the most recent one. Thus here we would need to abstract away IMHO into

```suggestion
dandi-api will use the `dandischema` validation API (ATM will not support older schema versions and thus require metadata re-extraction/upgrade).
```
or smth like that.

doc/design/publish-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+# Pre-publish
+
+## Uploading data / Modifying metadata
+Whenever changes are made to the draft Version of a Dandiset, metadata will almost always be modified.
```
yarikoptic, line 62: I guess I am confused by lack of explicit difference between "metadata JSON blob" (as in DB) and "metadata record returned by API" (JSOB blob + computed).

yarikoptic (changes requested):
First part of the review: yet to get through the end. I am still not sure we should mutate any metadata JSON blob: I see no benefit or need.

satra:
> I am still not sure we should mutate any metadata JSON blob: I see no benefit or need.

at least one benefit is not having to inject on every query. also we will not be injecting for published assets or dandisets. a return is much more efficient than an injection.

yarikoptic:
> > I am still not sure we should mutate any metadata JSON blob: I see no benefit or need.
> at least one benefit is not having to inject on every query. also we will not be injecting for published assets or dandisets. a return is much more efficient than an injection.

AFAIK injection is negligibly fast.  Any indication that I am wrong and optimization is needed? (there is that saying about "premature optimization" .... ;-))

satra:
unless there is a reason to be dynamic i say write once serve many. that's not premature optimization just good practice imo. is there a specific reason why being dynamic provides an advantage?

waxlamp:
> unless there is a reason to be dynamic i say write once serve many. that's not premature optimization just good practice imo. is there a specific reason why being dynamic provides an advantage?

The reason to inject rather than mutate is that it theoretically closes the window on a specific type of error we might otherwise be able to make. If the injected stuff is a pure function of the established metadata, then mutating the metadata record to include it opens us up to the risk that the mutated/added records could then be further mutated in the future, thus pushing it out of sync with the actually-mutable metadata.

If it is truly "negligibly fast" then I'm with @yarikoptic--we should just inject at return time and not have to worry about possible bit rot, etc.

doc/design/publish-1.md
dchiquito: Lets leave this for now, and clarify with a bigger, better design doc when we care about that feature.

doc/design/publish-1.md
dchiquito: As currently implemented, Versions only track the validation of their own metadata, not their Assets. You could have a `VALID` version with `INVALID` assets. When a publish is attempted, the `INVALID` assets are checked, so for publishing purposes the version is considered `INVALID`, despite having a column that is marked `VALID`.

```suggestion
Versions also track the validation status of their Dandiset metadata. Asset metadata validation is not recorded in the Version.
```

doc/design/publish-1.md
dchiquito: My understanding was that it was feasible to write a library that would return the correct pydantic models for a given version. 

```suggestion
dandi-api will look up the JSON schema for the specified `schemaVersion` and use it for validation. Hopefully at some point we can find a way to provide arbitrary pydantic model versions and use those instead.
```

doc/design/publish-1.md
dchiquito: In my draft, there is no difference. IMHO that is one of the benefits of not injecting information into the metadata.

dchiquito:
Since the only remaining discussion is regarding mutating metadata vs. injecting it during query, I'm going to merge this and write up a new PR describing Yarik's proposed injection scheme. We can continue the discussion there.

satra:
i think @yarikoptic is ok with the modify on write approach. he hasn't yet given me a use case that would work better for injection.

dchiquito:
https://github.com/dandi/dandi-api/pull/308 the saga continues

yarikoptic:
> i think @yarikoptic is ok with the modify on write approach. he hasn't yet given me a use case that would work better for injection.

neither I was given a use case where injection would be better (besides some performance aspect)... well -- I guess it might not take too long to be prove one way or another ;-)

PR #304, opened 2021-05-21
RF(DX): move tests deps under "dev" extra deps

yarikoptic:
Otherwise
- difficult to install all needed dependencies outside of tox
- README instructs for .[dev] for setup to "Develop natively" and testing
  is part of the development process

I am ok to split it apart into a dedicated tests  extra_requires
but IMHO there is little to no need for that really -- I do consider
testing being part of the development process.


dchiquito:
@yarikoptic The preferred way to run `pytest` is through tox:
```
# Just run tests
tox -e test
# Run the test_asset file
tox -e test -- -k test_asset
# Generate test coverage
tox -e test -- --cov=dandiapi --cov-report=html
```
This is all documented in `README.md`. Is there anything you specifically need pytest for?

yarikoptic:
I tend not to use `tox` and just use `pytest` directly (since then I can just `python -m pytest` on a project without remembering how any specific project decided to call their tox endpoint for running tests. `tox` is just a helper tool which actually runs `pytest`.  But regardless of "how to run the tests" -- why not to have testing developments declared in setup.* ?  It would also help IDEs to install needed dependencies for code/dependencies introspection etc.

waxlamp:
> `tox` is just a helper tool which actually runs `pytest`.

I feel like that "just" is unfairly diminishing what `tox` helps devs do 😸 

> But regardless of "how to run the tests" -- why not to have testing developments declared in setup.* ? It would also help IDEs to install needed dependencies for code/dependencies introspection etc.

I would be ok with adding these dependencies to the setup.py file (though I'm relying on others to explain the downsides of doing this). I am not ok with *removing* the dependencies from tox.ini, since indeed `tox` is the official way to run the tests (regardless of other particular arrangements you might want to make).

dchiquito (changes requested):
I would be happy with a `extras_require[test]` that you could use to install these dependencies locally if needed. I think tox should definitely be the officially supported way to run pytest, but adding an `extras_require` is an easy, nonintrusive way to support Yarik's preferred workflow.

I also admit that I have had to manually look up these dependencies before to `pip install` them so my IDE would autocomplete, so this would be useful to me. I would rather not have them in the `dev` dependencies because autocomplete is only a nice-to-have thing, not a strict requirement for a development environment.

yarikoptic:
ok, can move to `test`. Unless I populate `test` with all of the `dev` though, it would make little sense since would then require extra step (`pip install -e .[test]`) to setup a full fledged devel environment (with all the "nice to have features", although `tox` seems to be more than that since is part of `dev`).  Would need to think if there is any better solution for that: e.g. would you mind having `devel` (in addition to `dev` + `test` or just inplace of `test`) which would be the `dev` + "nice to have features"?

note: having `tox` in `dev` makes tox install `tox` (i.e. itself) also within its `test` environment, probably for no good reason.  FWIW, even could be a different version (on my laptop tox used is 3.15.1 and then tox installs within test 3.23.0).

waxlamp:
Having `dev` and `devel` seems like it will lead to confusion. Let's stick with `test` and let's update the README (and eventually the "developer docs" I am still thinking about) to explain why you would install `.[test]` in your environment.

dchiquito (approved):
I added `[test]` and added some docs, ready to ship

setup.py
```diff
@@ -71,7 +71,16 @@
...
+        ],
+        'test': [
+            'factory-boy',
+            'pytest',
+            'pytest-cov',
+            'pytest-django',
+            'pytest-factoryboy',
+            'pytest-mock',
+            'requests',
+        ],
```
yarikoptic, line 83: you can remove the list from tox.ini then and just add `test` to go along with `dev` as
```
diff --git a/tox.ini b/tox.ini
index b4e8fc6..8e6a363 100644
--- a/tox.ini
+++ b/tox.ini
@@ -48,14 +48,7 @@ passenv =
     DJANGO_DANDI_SCHEMA_VERSION
 extras =
     dev
-deps =
-    factory-boy
-    pytest
-    pytest-cov
-    pytest-django
-    pytest-factoryboy
-    pytest-mock
-    requests
+    test
 commands =
     pytest {posargs}
 
```

dchiquito (approved):

README.md
```diff
@@ -69,6 +69,8 @@ the appropriate `dev/.env.docker-compose*` file as a baseline for overrides.
 ### Initial Setup
 tox is used to execute all tests.
 tox is installed automatically with the `dev` package extra.
+To install tox's dependencies into your environment, run `pip install -e .[test]`.
```
brianhelba, line 72: > To install tox's dependencies

Strictly speaking, installing the `[test]` extra will only install the dependencies for the `test` Tox environment. Tox has other environments which are part of the development cycle, including `lint` and `format`.

brianhelba:
I'd agree with the consensus that Tox should remain the only officially supported way of executing all tests.

README.md
dchiquito:
```suggestion
To install the tox pytest dependencies into your environment, run `pip install -e .[test]`.
```

dchiquito (approved):

PR #307, opened 2021-05-24
Do not report orphaned assets in size stats

dchiquito:
Filter out any assets that do not belong to a version when collecting
total size statistics.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -147,4 +147,6 @@ class Asset(TimeStampedModel):
 
     @classmethod
     def total_size(cls):
-        return cls.objects.aggregate(size=models.Sum('blob__size'))['size'] or 0
+        return (
+            cls.objects.filter(versions__gt=0).aggregate(size=models.Sum('blob__size'))['size'] or 0
+        )
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -148,5 +148,8 @@ class Asset(TimeStampedModel):
     @classmethod
     def total_size(cls):
         return (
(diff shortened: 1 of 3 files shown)


AlmightyYakob:
Should this be merged into staging instead of master?

dandiapi/api/models/asset.py
```diff
@@ -147,4 +147,6 @@ def get_path(cls, path_prefix: str, qs: List[str]) -> Set:
 
     @classmethod
     def total_size(cls):
-        return cls.objects.aggregate(size=models.Sum('blob__size'))['size'] or 0
+        return (
+            cls.objects.filter(versions__gt=0).aggregate(size=models.Sum('blob__size'))['size'] or 0
```
yarikoptic, line 151: have you looked at #298 ?  this would not be sufficient AFAIK since would double count the same blobs across multiple versions/instances of dandisets and I believe we shouldn't do that

yarikoptic (changes requested):
would still double count blobs AFAIK

dchiquito:
@AlmightyYakob Eventually yes that is the plan, but for now `staging` is just a playground where we can push stuff for consequence free testing. Right now everything in staging is publish related, so I'd rather not complicate things by throwing this into the staging pile.

I can do staging instead, but this seems like a low risk, quick and easy thing, so I deliberately chose master.

dandiapi/api/models/asset.py
dchiquito: Good point, I have fixed this

yarikoptic (approved):
I think this should do it

satra:
@dchiquito - can this be merged into production? i'm not seeing the changes in size there.

PR #308, opened 2021-05-24
Inject metadata instead of computing it

dchiquito:
#302 Part Two

Do we precompute metadata every time it changes and store that in the DB, or store metadata as the user specifies it and inject computed fields when it is queried?

This is how I think injection would look.

### DIFF
--- a/doc/design/publish-1.md
+++ b/doc/design/publish-1.md
@@ -12,6 +12,11 @@ There are two kinds of Version: draft and published.
 Published Versions are immutable, only draft Versions are editable (except admins, who have superpowers).
 Versions also track the validation status of their Dandiset metadata. Asset metadata validation is not recorded in the Version.
 
+The metadata that is stored in the DB is strictly the user-specified metadata.
+Computed metadata, like `assetsSummary`, `size`, etc. are injected when the metadata is requested through the API.
+Published Versions also record some information about the publishing process (dates, current API version, doi, etc.)
+This information is used to generate the `publishedBy` and other publish-specific fields, which are likewise injected.
+
 ### AssetBlob
 This is basically just a pointer to a blob in `blobs/` on S3.
 AssetBlobs are central to how we implement deduplication of blobs in S3.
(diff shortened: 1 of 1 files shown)


doc/design/publish-1.md
```diff
@@ -20,29 +25,26 @@ Each Asset could have a different metadata record for the same AssetBlob.
...
+* The oldest published Version that the Asset belongs to is queried, and the `publishBy` and `datePublished` fields for that Version are injected into the Asset metadata.
+If an asset does not belong to any published Versions, `publishedBy` is not injected.
 
-_In general_, Assets should be immutable; any modification will instead create a new draft Asset.
-* One exception to this is calculating the AssetBlob digest.
-When the digest completes, it will be injected directly into the metadata of any Assets referencing that AssetBlob.
-* The other exception is creating a published Asset during the publish process.
-This involves directly injecting the new metadata fields into the draft Asset, effectively turning it into a published Asset in place.
+Because "modifying" an Asset actually creates a new Asset, Asset metadata will largely be stable despite having dynamically computed fields.
+`digests` will change when the digest is calculated, but everything else is immutable.
```
waxlamp, line 47: Why would a digest of an otherwise-immutable data collection change?

waxlamp (commented):
Looks ok to me. Withholding official approval to make way for people with deeper knowledge on the ground.

doc/design/publish-1.md
dchiquito: Immediately after a large new blob is upload and an asset is created with it, the digest is unknown because it hasn't been calculated. However, the metadata still exists and should be exposed in the API, so we have to provide it with no digest. When the digest calculation completes, it should be reported, so it is injected into the `digests` field of the metadata.

satra: just a point of clarification, it should have one digest immediately on asset creation the `dandi-etag`.

waxlamp: So (roughly speaking) `digests` is "mutable" just in the narrow sense that it will receive a new value (in addition to the dandi-etag) when the digest is computed, but it's not that it will change value at other times.

Thanks for clarifiying.

dchiquito:
To be clear, I wrote this while we were still actively debating the design. The code has already been updated to save the complete metadata as the design doc describes, so if we were to merge this PR it would involve some more rewriting.

satra:
in that case, i will close this, and label it as design documentation.

PR #312, opened 2021-05-25
Update asset paths endpoint to include more information

### LINKED ISSUE #292: Rework paths endpoint to include more data at once
Original discussion [here](https://github.com/dandi/dandi-api/pull/261).

Currently, we query the paths endpoint, and then make multiple requests for each entry in the response array. This is inefficient, and can be done in one go. The current implementation also doesn't include folder size or asset count. The list of properties should be something like:

Common
* path
* size
* is_item (flag)

Folder
* number_of_items

Item
* download_url
* delete_url

AlmightyYakob:
Fixes #292

~Draft PR because I'm still fixing the tests.~

This is largely similar to #261, but with some extra additions. @yarikoptic Could you PTAL and tell me if this is what you were looking for?

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import datetime
-from typing import Dict, List, Set
+from typing import Dict
 from urllib.parse import urlparse, urlunparse
 import uuid
 
@@ -209,27 +209,6 @@ class Asset(TimeStampedModel):
 
     def __str__(self) -> str:
         return self.path
-
-    @classmethod
-    def get_path(cls, path_prefix: str, qs: List[str]) -> Set:
-        """
-        Return the unique files/directories that directly reside under the specified path.
-
-        The specified path must be a folder (must end with a slash).
-        """
-        if not path_prefix:
-            path_prefix = '/'
-        prefix_parts = [part for part in path_prefix.split('/') if part]
-        paths = set()
-        for asset in qs:
-            path_parts = [part for part in asset['path'].split('/') if part]
-
-            # Pivot index is -1 (include all path parts) if prefix is '/'
-            pivot_index = path_parts.index(prefix_parts[-1]) if len(prefix_parts) else -1
-            base_path, *remainder = path_parts[pivot_index + 1 :]
-            paths.add(f'{base_path}/' if len(remainder) else base_path)
-
-        return sorted(paths)
 
     @classmethod
     def total_size(cls):
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -7,7 +7,8 @@ from guardian.shortcuts import assign_perm
 import pytest
 import requests
 
-from dandiapi.api.models import Asset, AssetBlob
+from dandiapi.api.models import Asset, AssetBlob, Version
+from dandiapi.api.views.serializers import AssetFolderSerializer, AssetSerializer
 
 from .fuzzy import HTTP_URL_RE, TIMESTAMP_RE, URN_RE, UUID_RE
 
@@ -16,22 +17,73 @@ from .fuzzy import HTTP_URL_RE, TIMESTAMP_RE, URN_RE, UUID_RE
 
 @pytest.mark.django_db
 @pytest.mark.parametrize(
-    'path,qs,expected',
+    'path,asset_paths,expected',
     [
-        ('', [{'path': '/foo'}, {'path': '/bar/baz'}], ['bar/', 'foo']),
-        ('/', [{'path': '/foo'}, {'path': '/bar/baz'}], ['bar/', 'foo']),
-        ('////', [{'path': '/foo'}, {'path': '/bar/baz'}], ['bar/', 'foo']),
-        ('a', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('a/', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('/a', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('/a/', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('a', [{'path': 'a/b'}, {'path': 'a/c/d'}], ['b', 'c/']),
-        ('a', [{'path': 'a/b/'}, {'path': 'a/c/d/'}], ['b', 'c/']),
-        ('a', [{'path': '/a/b/'}, {'path': '/a/c/d/'}], ['b', 'c/']),
+        ('', ['foo', 'bar/baz'], {'folders': ['bar'], 'files': ['foo']}),
+        # ('', ['/foo', '/bar/baz'], {'folders': ['bar'], 'files': ['foo']}),  # edge case
+        ('/', ['/foo', '/bar/baz'], {'folders': ['bar'], 'files': ['foo']}),
+        ('', ['foo/bar', 'foo/baz', 'foo/boo'], {'folders': ['foo'], 'files': []}),
+        ('/', ['foo/bar', 'foo/baz', 'foo/boo'], {'folders': [], 'files': []}),  # Negative test
+        ('////', ['/foo', '/bar/baz'], {'folders': [], 'files': []}),
+        ('a', ['a/b', 'a/c/d'], {'folders': ['c'], 'files': ['b']}),
+        ('a/', ['a/b', 'a/c/d'], {'folders': ['c'], 'files': ['b']}),
+        ('/a', ['/a/b', '/a/c/d'], {'folders': ['c'], 'files': ['b']}),
+        ('/a/', ['/a/b', '/a/c/d'], {'folders': ['c'], 'files': ['b']}),
     ],
 )
-def test_asset_get_path(path, qs, expected):
-    assert expected == Asset.get_path(path, qs)
+def test_asset_rest_path(
+    api_client,
+    draft_version_factory,
+    asset_factory,
+    asset_blob_factory,
+    path,
+    asset_paths,
+    expected,
+):
+    # Initialize version and contained assets
+    asset_blob = asset_blob_factory()
+    assets = [asset_factory(blob=asset_blob, path=p) for p in asset_paths]
+    version: Version = draft_version_factory()
+    for asset in assets:
+        version.assets.add(asset)
+
+    # Retrieve paths from endpointF
+    paths = api_client.get(
+        f'/api/dandisets/{version.dandiset.identifier}/'
+        f'versions/{version.version}/assets/paths/',
+        {'path_prefix': path},
+    ).data
+
+    # Ensure slash between path prefix and folders/files
+    query_prefix = path
+    if query_prefix and query_prefix[-1] != '/':
+        query_prefix = f'{query_prefix}/'
+
+    # Do folder assertions
+    for folder_path in expected['folders']:
+        assert folder_path in paths['folders']
+
+        folder_entry = paths['folders'][folder_path]
+        folder_assets = list(
+            Asset.objects.all().filter(path__startswith=f'{query_prefix}{folder_path}')
+        )
+        serialized_folder = AssetFolderSerializer(
+            {
+                'created': min(asset.created for asset in folder_assets),
+                'modified': max(asset.modified for asset in folder_assets),
+                'size': sum(asset.size for asset in folder_assets),
+                'num_files': len(folder_assets),
+            }
+        ).data
+
+        assert folder_entry == serialized_folder
+
+    # Do file assertions
+    for file_path in expected['files']:
+        assert file_path in paths['files']
+
+        asset: Asset = Asset.objects.get(path=f'{query_prefix}{file_path}')
+        assert paths['files'][file_path] == AssetSerializer(asset).data
 
 
 @pytest.mark.django_db
@@ -550,26 +602,3 @@ def test_asset_direct_download_head(api_client, storage, version, asset):
 
     with asset.blob.blob.file.open('rb') as reader:
         assert download.content == reader.read()
-
-
-@pytest.mark.django_db
-@pytest.mark.parametrize(
-    'path_prefix,results',
-    [
-        ('', ['foo/', 'no-root.nwb', 'root.nwb']),
-        ('/', ['foo/', 'root.nwb']),
-        ('/foo', ['bar/', 'baz.nwb']),
-        ('/foo/', ['bar/', 'baz.nwb']),
-    ],
-)
-def test_asset_rest_path_filter(api_client, version, asset_factory, path_prefix, results):
-    paths = ['/foo/bar/file.nwb', '/foo/baz.nwb', '/root.nwb', 'no-root.nwb']
-    for path in paths:
-        version.assets.add(asset_factory(path=path))
-    partial_path_assets = api_client.get(
-        f'/api/dandisets/{version.dandiset.identifier}/'
-        f'versions/{version.version}/assets/paths/',
-        {'path_prefix': path_prefix},
-    ).data
-
-    assert partial_path_assets == results
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -30,6 +30,7 @@ from dandiapi.api.tasks import validate_asset_metadata, validate_version_metadat
 from dandiapi.api.views.common import DandiPagination
 from dandiapi.api.views.serializers import (
     AssetDetailSerializer,
+    AssetPathsSerializer,
     AssetSerializer,
     AssetValidationSerializer,
 )
@@ -279,15 +280,10 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
         manual_parameters=[
             openapi.Parameter('path_prefix', openapi.IN_QUERY, type=openapi.TYPE_STRING)
         ],
-        responses={
-            200: openapi.Schema(
-                type=openapi.TYPE_ARRAY,
-                items=openapi.Schema(type=openapi.TYPE_STRING),
-            )
-        },
+        responses={200: AssetPathsSerializer()},
     )
     @action(detail=False, methods=['GET'])
-    def paths(self, request, **kwargs):
+    def paths(self, request, versions__dandiset__pk: str, versions__version: str, **kwargs):
         """
         Return the unique files/directories that directly reside under the specified path.
 
@@ -295,11 +291,38 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
         (to refer to the root folder) must be the empty string.
         """
         path_prefix: str = self.request.query_params.get('path_prefix') or ''
-        # Enforce trailing slash
-        if path_prefix and path_prefix[-1] != '/':
-            path_prefix = f'{path_prefix}/'
-        qs = self.get_queryset().filter(path__startswith=path_prefix).values()
+        qs = self.get_queryset().select_related('blob').filter(path__startswith=path_prefix)
 
-        return Response(Asset.get_path(path_prefix, qs))
+        folders = {}
+        files = {}
+
+        for asset in qs:
+            # Get the remainder of the path after path_prefix
+            base_path: str = asset.path[len(path_prefix) :].strip('/')
+
+            # Since we stripped slashes, any remaining slashes indicate a folder
+            folder_index = base_path.find('/')
+            is_folder = folder_index >= 0
+
+            if not is_folder:
+                files[base_path] = AssetSerializer(asset).data
+            else:
+                base_path = base_path[:folder_index]
+                entry = folders.get(base_path)
+                if entry is None:
+                    folders[base_path] = {
+                        'size': asset.size,
+                        'num_files': 1,
+                        'created': asset.created,
+                        'modified': asset.modified,
+                    }
+                else:
+                    entry['size'] += asset.size
+                    entry['num_files'] += 1
+                    entry['created'] = min(entry['created'], asset.created)  # earliest
+                    entry['modified'] = max(entry['modified'], asset.modified)  # latest
+
+        paths = AssetPathsSerializer({'folders': folders, 'files': files})
+        return Response(paths.data)
 
     # TODO: add create to forge an asset from a validation
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -113,3 +113,15 @@ class AssetDetailSerializer(AssetSerializer):
         fields = AssetSerializer.Meta.fields + ['metadata']
 
     metadata = serializers.SlugRelatedField(read_only=True, slug_field='metadata')
+
+
+class AssetFolderSerializer(serializers.Serializer):
+    size = serializers.IntegerField()
+    num_files = serializers.IntegerField()
+    created = serializers.DateTimeField()
+    modified = serializers.DateTimeField()
+
+
+class AssetPathsSerializer(serializers.Serializer):
+    folders = serializers.DictField(child=AssetFolderSerializer())
+    files = serializers.DictField(child=AssetSerializer())


yarikoptic:
Thank you @AlmightyYakob ! Will review tomorrow while mapping to the mighty #261 implementation to see if nothing is missed ;)

AlmightyYakob:
> Thank you @AlmightyYakob ! Will review tomorrow while mapping to the mighty #261 implementation to see if nothing is missed ;)

One difference you'll notice is that I didn't included `created` or `modified` on folders. I'm curious, what is the motivation behind including those properties on folders?

satra:
> what is the motivation behind including those properties on folders?

status checks for people uploading to specific folders. instead of going down to the asset level and checking every asset, they can immediately see that a folder containing the asset was modified at time X.

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,57 @@ def copy(cls, asset):
...
+
+                entry = paths['folders'][base_path]
+                entry['size'] += asset.size
+                entry['num_items'] += 1
+                entry['created'] = min(entry['created'], asset.modified)  # earlist
+                entry['modified'] = max(entry['modified'], asset.modified)  # latest
+            else:
+                kwargs['asset_id'] = asset.asset_id
+                asset_delete_url = base_api_url + reverse('asset-detail', kwargs=kwargs)
+                asset_download_url = base_api_url + reverse('asset-download', kwargs=kwargs)
```
yarikoptic, line 168: by the looks of it that `reverse` looks not that "speedy".  I think it would be well worth comparing against straight assignment of a URL based on our knowledge of what it should look like given asset information we have (or is anything lacking in the `Asset`?).

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,57 @@ def copy(cls, asset):
...
+                        'size': 0,
+                        'num_items': 0,
+                        'created': asset.created,
+                        'modified': asset.modified,
+                    }
+
+                entry = paths['folders'][base_path]
+                entry['size'] += asset.size
+                entry['num_items'] += 1
+                entry['created'] = min(entry['created'], asset.modified)  # earlist
```
yarikoptic, line 163:
```suggestion
                entry['created'] = min(entry['created'], asset.created)  # earliest
```

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,57 @@ def copy(cls, asset):
...
+                entry['modified'] = max(entry['modified'], asset.modified)  # latest
+            else:
+                kwargs['asset_id'] = asset.asset_id
+                asset_delete_url = base_api_url + reverse('asset-detail', kwargs=kwargs)
+                asset_download_url = base_api_url + reverse('asset-download', kwargs=kwargs)
+
+                paths['items'][base_path] = {
+                    'path': f'{base_path}/',
```
yarikoptic, line 171:
```suggestion
                    'path': base_path,
```
why asset should have a trailing `/`?

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,57 @@ def copy(cls, asset):
...
+        kwargs = {'versions__dandiset__pk': dandiset_pk, 'versions__version': version}
+
         for asset in qs:
-            path_parts = [part for part in asset['path'].split('/') if part]
+            path_parts = [part for part in asset.path.split('/') if part]
 
             # Pivot index is -1 (include all path parts) if prefix is '/'
             pivot_index = path_parts.index(prefix_parts[-1]) if len(prefix_parts) else -1
             base_path, *remainder = path_parts[pivot_index + 1 :]
-            paths.add(f'{base_path}/' if len(remainder) else base_path)
```
yarikoptic, line 144: BTW, in my PR I didn't want to touch any of the unrelated logic, but IMHO the above code does too much for what it needs to accomplish. If we assume that path_prefix is normalized to end with `/` (removed in this PR), code could be as simple as following IMHO easier to comprehend and likely faster (didn't do timing) code:
```python
    assert asset.path.startswith(path_prefix)  # just for paranoid me... not sure what happens in above if not fulfilled
    base_path = asset.path[len(path_prefix):]   # len is fast but could be even bound to a local var
    folder_idx = base_path.find('/')
    is_subfolder = folder_idx >= 0
    if is_subfolder:  # not sure actually if 0 can/should happen, might better just guard
        base_path = base_path[:folder_idx]
```

<details>
<summary>here is demo of it in work</summary> 

```python
class asset:
    path = '/d1/d2/123'

for path_prefix in '/', '/d1/', '/d1/d2/':
    assert asset.path.startswith(path_prefix)    
    base_path = asset.path[len(path_prefix):]   # len is fast but could be even bound to a local var
    folder_idx = base_path.find('/')
    is_subfolder = folder_idx >= 0
    if is_subfolder:  # not sure actually if 0 can/should happen, might better just guard
        base_path = base_path[:folder_idx]
    print(base_path, is_subfolder)
```
produces
```
d1 True
d2 True
123 False
```
</details>
curious ones could time-compare the two ;-)  or am I missing something above code does which "mine" doesn't?

yarikoptic (changes requested):
Sorry for the delay and thank you @AlmightyYakob for taking care about implementing this functionality.

Although premature optimization is the source of evil, whenever it is trivial and potentially makes code easier to read etc, I prefer it to be done instead of postponing. Thus I left my "invaluable" recommended feedback/changes .  May be some would be of use.
For now just reviewed the portion of `get_path` as it is what regained some memory of from #261 .

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,57 @@ def copy(cls, asset):
...
         The specified path must be a folder (must end with a slash).
         """
-        if not path_prefix:
-            path_prefix = '/'
         prefix_parts = [part for part in path_prefix.split('/') if part]
-        paths = set()
+        paths = {'folders': {}, 'items': {}}
+
+        # Declare base api url and kwargs for all asset reverse lookups
+        base_api_url = 'https://api.dandiarchive.org'
```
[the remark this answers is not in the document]
yarikoptic: FTR the issue is known: https://github.com/dandi/dandi-api/issues/286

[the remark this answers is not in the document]
AlmightyYakob: You're correct we'd like to address this, but this PR isn't a good place for it, so we'll do it separately.

dandiapi/api/models/asset.py
AlmightyYakob: This was more for display on the client, but I agree it's unnecessary, since the client can already tell that it's a folder from the parent key anyway. Should we even include this `path` field at all, since it's ultimately redundant?

dandiapi/api/models/asset.py
AlmightyYakob: The main idea is to prevent the need to change this URL, if we decide to change where the `asset-detail` and `asset-download` endpoints are mounted in the future. While that may not be likely to happen, it's good software practice.

I don't believe `reverse` is costly anyway, as it should just be little more than a dict lookup. I tested this locally by replacing the calls to `reverse` with a hardcoded string, and the results were identical (roughly ~180-200ms for a folder with 8 items on my local environment).

dandiapi/api/models/asset.py
AlmightyYakob: In 4b8bd5f I've removed the `path` field. If it's important that we keep that field included, I can revert the change.

dandiapi/api/models/asset.py
AlmightyYakob: I timed both, and it seems they are roughly equal, but I think yours is simpler, so I've changed it in 673a7a8. Also, your assertion at the beginning is always true, as these assets are initially found by querying the databse for assets that begin with `path_prefix`, so that part is always guaranteed.

yarikoptic: THANK YOU for checking the timing! 
re `assert` -- IMHO the purpose of having those in the code (if ever) is to ensure that assumption (valid at the time of adding an `assert`) on outside behavior still holds in the future and that nothing broken outside for some reason invalidates it. But feel welcome to remove it.

dandiapi/api/models/asset.py
```diff
@@ -125,25 +126,60 @@ def copy(cls, asset):
...
+                    folders[base_path] = {
+                        'size': asset.size,
+                        'num_files': 1,
+                        'created': asset.created,
+                        'modified': asset.modified,
+                    }
+                else:
+                    entry['size'] += asset.size
+                    entry['num_files'] += 1
+                    entry['created'] = min(entry['created'], asset.created)  # earlist
```
yarikoptic, line 164:
```suggestion
                    entry['created'] = min(entry['created'], asset.created)  # earliest
```
;-)

dandiapi/api/models/asset.py
yarikoptic: Thanks for timing! sounds oddly long though if it is milliseconds for just 8 items: most likely there is overall overhead bias and actual loop IMHO should take well under 1ms per asset to consider AFAIK.  So might be worth timing by explicit time printouts as diff in time() before/after the loop for adequate comparison.  If it was in µs (AKA us) - then ok ;)

AlmightyYakob: Sorry I gave the wrong number, the 200ms number was for all the assets in the dandiset (which is a local copy of dandiset `000006`, so 53 assets). I believe the number for the folder with 8 assets was ~35ms.

AlmightyYakob: Regarding the timing: I was timing the asset loop only, not the whole request, done with python's `time.time()`.

yarikoptic: Thank you! Still too high if can't get optimized, since if I do math right would be about 4 sec for one with 1000 assets.  https://gui.dandiarchive.org/#/dandiset/000026 has 1225 assets ATM .

yarikoptic: But I refuse to believe that python is that slow!!! There must be a gotcha. BTW py-spy is quite nice if you do get multi seconds runs and just py-spy top the process of even create a flame graph ;-)

AlmightyYakob: @yarikoptic Are you okay with merging this as is and working on optimization later (to preserve momentum)? I still have to make the client-side changes however, so this merge might still be delayed anyway.

yarikoptic: I will leave it at your discretion since it seems it might have detrimental effect on timings significantly effecting top level directories (typically with folders) and I would have first tried to figure out if such approach is pragmatically feasible.

Also -- Don't we auto-deploy master -- and then it would break web UI since if I got it right it changes the output for `/files` endpoint, and should instead be repositioned against `staging` anyways so then both dandi-api and client(s) could then be rolled out at the same time?

AlmightyYakob: You're correct, we'd have to coordinate the client changes and this one, which is why this change might still be delayed. Staging is in a bit of a weird spot right now, so I'm not sure this PR will be able to take advantage of it.

yarikoptic: That is-partially-why I was looking to add this extended listing as an option, without changing the default in #292 so it could have been merged/tried/benchmarked/improved (and possibly discarded ;))  without requiring any change to web UI.
Anyways -- I am still intrigued by the slow performance.  If your timings hold, I am not sure that this would be a viable solution worth pursuing since it would be of great negative effect on web ui navigation of majority of the dandisets.
Gimme an hour -- I want to see if it is indeed Python which inflicts itself upon us here or it is something of django so we are least get the lower bound.  My plan -- take assets records from e.g. 000026 and see how long it would take to get listing on top level (without talking to django to compose URLs).

satra: wouldn't this be only for files that are viewable at a level? so this should never be run on all assets ever right?

yarikoptic: it would loop through all assets if ran for top level prefix `/`, but indeed would not need to mint urls for them etc.

<details>
<summary>With this dummy script where I copy pasted get_path as is and run it across metadata records with no-op "reverse" to get URL</summary> 

```shell
from typing import Dict, List

class Asset(dict):
    def __getattr__(self, item):
        return self.__getitem__(item)


def reverse(*args, **kwargs):
    # noop for now
    return "TODO"


def get_path(dandiset_pk: str, version: str, path_prefix: str, qs: List[Asset]) -> Dict:
    """
    Return the unique files/directories that directly reside under the specified path.

    The specified path must be a folder (must end with a slash).
    """
    # Declare base api url and kwargs for all asset reverse lookups
    base_api_url = 'https://api.dandiarchive.org'
    kwargs = {'versions__dandiset__pk': dandiset_pk, 'versions__version': version}

    # Initialize dicts to store folders and files
    folders = {}
    files = {}

    for asset in qs:
        # Get the remainder of the path after path_prefix
        base_path: str = asset.path[len(path_prefix):].strip('/')

        # Since we stripped slashes, any remaining slashes indicate a folder
        folder_index = base_path.find('/')
        is_folder = folder_index >= 0

        if is_folder:
            base_path = base_path[:folder_index]
            entry = folders.get(base_path)
            if entry is None:
                folders[base_path] = {
                    'size': asset.size,
                    'num_files': 1,
                    'created': asset.created,
                    'modified': asset.modified,
                }
            else:
                entry['size'] += asset.size
                entry['num_files'] += 1
                entry['created'] = min(entry['created'], asset.created)  # earliest
                entry['modified'] = max(entry['modified'], asset.modified)  # latest
        else:
            kwargs['asset_id'] = asset.asset_id
            asset_delete_url = base_api_url + reverse('asset-detail', kwargs=kwargs)
            asset_download_url = base_api_url + reverse('asset-download', kwargs=kwargs)

            files[base_path] = {
                'size': asset.size,
                'created': asset.created,
                'modified': asset.modified,
                'delete_url': asset_delete_url,
                'download_url': asset_download_url,
            }

    return {
        'folders': folders,
        'files': files,
    }



import yaml
import time
# produced by dandi ls -r -f yaml DANDI:000026 > 000026-dandi-ls-r.json
# to not mess with pagination , see https://github.com/dandi/dandi-api/issues/252
# and overcome https://github.com/dandi/dandi-cli/issues/652 for -f json_pp
with open('000026-dandi-ls-r.yaml') as f:
    # wrap them into that odd Asset to provide `.__getattr__` so we could use original code as is
    assets = [Asset(_) for _ in yaml.load(f, Loader=yaml.SafeLoader) if 'asset_id' in _]

print(f"Read {len(assets)} assets")
"""noted a bug of eating first letter
*(Pdb) out['folders']
{'erivatives': {'size': 62098107735, 'num_files': 83, 'created': '2021-04-21T08:45:29.941149Z', 'modified': '2021-05-18T13:13:45.601444Z'}, 'awdata': {'size': 836971584508, 'num_files': 1142, 'created': '2021-04-21T08:49:14.803490Z', 'modified': '2021-05-18T13:15:42.881510Z'}}
"""
for prefix in ('/', 'derivatives/', 'rawdata/sub-I46/ses-SPIM/'):
    t0 = time.time()
    out = get_path("000026", "draft", prefix, assets)
    dt = time.time() - t0
    print(f"{prefix}: Got {len(out['folders'])} folders and {len(out['files'])} files in {dt:.3f} sec")
```
</details>

I got

1. timing -- Python is FAST enough! ;-) (despite my ugly `__getattr__`)
```
$> python time_api_get_path.py
Read 1225 assets
/: Got 2 folders and 0 in 0.002 sec
derivatives/: Got 9 folders and 0 in 0.002 sec
rawdata/sub-I46/ses-SPIM/: Got 4 folders and 1154 in 0.003 sec
```

so the loop/Python is not the issue but something about using `__getattr__` on assets or `reverse` which slows it down.

2. my recommended "simple" way logic might benefit from "guarding" the assumption that asset path might starts with `/` -- I do not think it would anyhow affect timing but would make cost more robust (talking about assumptions on what outside code provides), since otherwise it might  eat up leading character:
```
print(get_path("000026", "draft", "/", assets))
```
results in
```
{'folders': {'erivatives': {'size': 62098107735, 'num_files': 83, 'created': '2021-04-21T08:45:29.941149Z', 'modified': '2021-05-18T13:13:45.601444Z'}, 'awdata': {'size': 836971584508, 'num_files': 1142, 'created': '2021-04-21T08:49:14.803490Z', 'modified': '2021-05-18T13:15:42.881510Z'}}, 'files': {}}
```
since currently we do not prefix asset paths with `/` and we I do not think we need to revert that or make it special.
```
$> curl --silent -X GET "https://api.dandiarchive.org/api/dandisets/000026/versions/draft/assets/?page=1&page_size=100000" -H "accept: application/json" | jq '.' | grep path | head
      "path": "derivatives/EPIC/dataset_description.json",
      "path": "derivatives/EPIC/sub-I46/sub-I46_ses-MRI_flip-1_VFA.json",
      "path": "derivatives/EPIC/sub-I46/sub-I46_ses-MRI_flip-2_VFA.json",
      "path": "derivatives/EPIC/sub-I48/sub-I48_ses-MRI_flip-1_VFA.json",
```

AlmightyYakob: I did already test removing `reverse` and it seemed to have no effect at all. However, the main reason for the slowness may have been how we were accessing the asset blobs. Since we look at the blob to check the size, we need to get that from the database as well. With what you just tested, those were two separate calls. So we would make a call to the database to get all relevant assets, and the for _each_ individual asset, we'd make a call to the database to get that blob, just for the size.

In 3780cb2 I've fixed that, so it fetches the blobs at the same time as the asset. This seems to show a ~6-7x speed improvement.

yarikoptic: for my own education, that `self.get_queryset().select_related('blob')` in 3780cb2 is like a joint between assets and blobs? If that was the only reason -- surprised we are not getting even bigger boost.  
Well, may be now the `reverse` is the one which is taking most of the time and you would spot the effect (i.e. if before size was the major factor, it would have masked any effect of `reverse`)?

yarikoptic: note: for the top level of dandiset, since we do not have files there, `reverse` should have no effect.  So if it is still is taking `ms`nds "per each asset" -- there must be some additional culprit in how data is accessed which might need to be ironed out.

AlmightyYakob: > for my own education, that `self.get_queryset().select_related('blob')` in [3780cb2](https://github.com/dandi/dandi-api/commit/3780cb2073dfc04713c98a1b659ddb12a73c2ee1) is like a joint between assets and blobs?

Yes.


> Well, may be now the `reverse` is the one which is taking most of the time and you would spot the effect (i.e. if before size was the major factor, it would have masked any effect of `reverse`)?

I tried the same experiment again, and it still seems that removing `reverse` has no effect. On my local setup, it's taking ~100-120ms at the top level, with 53 assets. It appears to also take roughly the same time when in a folder with 8 assets, with or without the `reverse` calls.

AlmightyYakob: I should point out, the above times are for the entire request. If I just time the asset loop, it's pretty consistently `10ms` for all assets. So quite a reduction from the previous `200ms` for all assets.

yarikoptic: >  it's pretty consistently 10ms for all assets

regardless of the number of them (8 vs 53)? and that is done not for top level directory where there is no assets at that level (so `reverse` would not be even triggered), what about the folder with the assets?  

anyways -- it seems we would be shooting in the dark by simply rolling it out straight into deployment -- what was the problem with the staging besides flooding sentry? could we disconnect sentry ? then we could have populated some dummy datasets with some good number of assets (I could even mimic real ones, just make them with dummy small blobs so we do not torture that bucket) to assess the performance "first hand".

AlmightyYakob: I did some more investigation, and it seems that most of the time in that loop (the `10ms`) is spent on database overhead. If I cast the query to a list beforehand (to force the query to run before the loop), the loop takes roughly 0.2 milliseconds for all assets, and roughly _`1ms`_ for the folder with 8 assets. So it does seem that reverse is having an impact (testing without the reverse call, the loop takes `0.03ms`, which is expected.

If I time the casting of the queryset to a list, that call takes ~`10ms`. This would explain why we were seeing similar request times for 53 vs 8 assets. I'd guess the rest of the request time (total request taking a little over `100ms`) is spent handling the request. 

I still think it'd be good to test this in staging with large datasets, but with this new information I'm more confident that this implementation scales well to larger datasets. The calls to `reverse` are definitely taking up some time, but I think we should test out how it actually performs on large datasets before jumping the gun and doing away with it. If we test it out and see that the performance really is an issue, then I'd say we should use a different approach.

yarikoptic: Thanks for the investigation!  
- indeed sounds like we could proceed forward.  if `reverse` would become prohibitively expensive, we could provide ad-hoc composition of the urls as I thought about originally
- we probably shouldn't merge it straight into master at this point. With all the vacationing upon many, we agreed in today's brief standup call that bringing up staging so we could test + orchestrate joint releasing of dandi-api + web-ui would be the way to go forward (@waxlamp please attest that I am I not mis-interpreting our verbal decision).

So - may be this PR could be "rebased" against staging and merged meanwhile with accompanying changes to web-ui (dandiarchive) also merged into its own staging?

yarikoptic:
Staging setup seems to be coming along, so may be it would be a good time to resume this effort.

dandiapi/api/models/asset.py
mvandenburgh: I'm reviewing this PR now, so just to resolve some threads here -

> * indeed sounds like we could proceed forward.  if `reverse` would become prohibitively expensive, we could provide ad-hoc composition of the urls as I thought about originally

Agreed. Right now the slow loading time for assets of large datasets is an issue in the client, specifically the loading time is dominated by the fact that "n" HTTP requests need to be made to the server for "n" folders/files. Once the web client is updated to make use of the changes in this PR we can see if `reverse` is still a problem.

> * we probably shouldn't merge it straight into master at this point. With all the vacationing upon many, we agreed in today's brief standup call that bringing up staging so we could test + orchestrate joint releasing of dandi-api + web-ui would be the way to go forward (@waxlamp please attest that I am I not mis-interpreting our verbal decision).
So - may be this PR could be "rebased" against staging and merged meanwhile with accompanying changes to web-ui (dandiarchive) also merged into its own staging?

Once we have staging properly set up (should be soon), we can just merge this straight to master and test the changes using our staging infrastructure.

waxlamp: > Once we have staging properly set up (should be soon), we can just merge this straight to master and test the changes using our staging infrastructure.

Just to provide extra context: Mike and Daniel are working now to arrange the deployment infrastructure so that `master` autodeploys to staging, and creating a release causes a deployment to production. This lets us test all changes comprehensively in staging, and gives us fine control over when to push those changes to production. (I believe this is the approach outlined in a design doc.)

dandiapi/api/models/asset.py
```diff
@@ -205,25 +205,60 @@ def __str__(self) -> str:
...
+
+            if is_folder:
+                base_path = base_path[:folder_index]
+                entry = folders.get(base_path)
+                if entry is None:
+                    folders[base_path] = {
+                        'size': asset.size,
+                        'num_files': 1,
+                        'created': asset.created,
+                        'modified': asset.modified,
```
mvandenburgh, line 238: The frontend also needs `'asset_id': asset.asset_id` to be returned here.

dandiapi/api/models/asset.py
```diff
@@ -205,25 +205,60 @@ def __str__(self) -> str:
...
+                    }
+                else:
+                    entry['size'] += asset.size
+                    entry['num_files'] += 1
+                    entry['created'] = min(entry['created'], asset.created)  # earliest
+                    entry['modified'] = max(entry['modified'], asset.modified)  # latest
```
mvandenburgh, line 244: As with folders, the frontend also needs `'asset_id': asset.asset_id` to be returned here.

mvandenburgh (changes requested):
Looks good, one small change which would be useful to have for the frontend changes that go along with this PR. 

When integrated into the frontend these changes solve the slowdown bug in this issue https://github.com/dandi/dandiarchive/issues/600. it brings the loading time down from 20+ seconds to 1-2 seconds :partying_face:

dandiapi/api/models/asset.py
AlmightyYakob: Both of the places you commented on within the folder case, which doesn't have a single `asset_id` associated with it. I'm assuming where this is really needed is the file case below?

mvandenburgh: Yeah you're right, this is only needed for files.

dandiapi/api/models/asset.py
mvandenburgh: Disregard, folders don't need this.

mvandenburgh (approved):

dandiapi/api/models/asset.py
yarikoptic: confirming -- matches description in https://github.com/dandi/dandi-api/blob/master/doc/design/staging-1.md#release-procedure ... I guess eventually we might need to elaborate on having `maint` or alike (could be throw away release branches) branch for "hot releases" so we do not need to release "staging" `master` for some quick fix (e.g. like a "fix" for #384)

dandiapi/api/models/asset.py
```diff
@@ -211,25 +211,61 @@ def __str__(self) -> str:
...
         The specified path must be a folder (must end with a slash).
         """
-        if not path_prefix:
-            path_prefix = '/'
-        prefix_parts = [part for part in path_prefix.split('/') if part]
-        paths = set()
-        for asset in qs:
-            path_parts = [part for part in asset['path'].split('/') if part]
+        # Declare base api url and kwargs for all asset reverse lookups
+        base_api_url = 'https://api.dandiarchive.org'
```
yarikoptic, line 221: how would that work for staging??

dandiapi/api/tests/test_asset.py
```diff
@@ -7,31 +8,97 @@
...
-        ('////', [{'path': '/foo'}, {'path': '/bar/baz'}], ['bar/', 'foo']),
-        ('a', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('a/', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('/a', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('/a/', [{'path': '/a/b'}, {'path': '/a/c/d'}], ['b', 'c/']),
-        ('a', [{'path': 'a/b'}, {'path': 'a/c/d'}], ['b', 'c/']),
-        ('a', [{'path': 'a/b/'}, {'path': 'a/c/d/'}], ['b', 'c/']),
-        ('a', [{'path': '/a/b/'}, {'path': '/a/c/d/'}], ['b', 'c/']),
+        ('', ['foo', 'bar/baz'], {'folders': ['bar'], 'files': ['foo']}),
+        # ('', ['/foo', '/bar/baz'], {'folders': ['bar'], 'files': ['foo']}),  # edge case
```
yarikoptic, line 25: so this edge case shouldn't be tested/supported?

dandiapi/api/models/asset.py
mvandenburgh: Ah good catch - it wouldn't work for staging. Maybe the base api url can be set in the `HerokuProductionConfiguration` and `HerokuStagingConfiguration` here https://github.com/dandi/dandi-api/blob/292-paths-endpoint/dandiapi/settings.py#L94

yarikoptic: That would be great. Citing https://github.com/dandi/dandi-api/issues/286 from above Resolved discussion . I think it should be reasonably simple change to stop postponing fixing it

waxlamp: @mvandenburgh, it would be good to check with Brian to get help in evaluating a solution for this problem.

dandiapi/api/tests/test_asset.py
AlmightyYakob: This is commented out because while it ideally should be tested, it's difficult to achieve based on how this test is setup. I tried to fit this into what's currently there, but after a good deal of tinkering around, I was unsuccessful. The test itself does pass though, as I've done it manually, it's just hard to run alongside the other (more concrete and meaningful) tests.

I'm not really concerned about this, as this edge case shouldn't actually crop up in real use anyway, since we don't prefix asset paths with slashes. If we do care about this test case, I think we should track is separately from his PR, as I don't think it provides enough value to delay anything here.

dandiapi/api/models/asset.py
AlmightyYakob: @yarikoptic this point is moot with the changes I've recently pushed. We don't return any URLs in the current implementation, and while the point you raised is something we should address, it no longer makes sense to do so in this PR.

mvandenburgh:
LGTM and works fine locally for me. As discussed, this will break the client until my frontend PR is merged so @AlmightyYakob feel free to merge that when you merge/release this.

mvandenburgh (approved):
> LGTM and works fine locally for me. As discussed, this will break the client until my frontend PR is merged so @AlmightyYakob feel free to merge that when you merge/release this.

I meant to approve here, not comment :)

dandiapi/api/models/asset.py
yarikoptic: Great to see hardcoded urls being avoided.

yarikoptic:
Just to make sure: this PR is no longer intending to provide download urls and only extended attributes (size, dates), correct?

yarikoptic:
And to answer my own question: correct since client would just mint them based on known API  download end point for a given asset https://github.com/dandi/dandiarchive/pull/773/files#diff-7d09a6d75d4c857695273c3d27a9cbdf2f4fce4e1eaee4ee0dc125c2d81e528fR221

Sounds good, albeit spreading the logic across clients. Excited to see it in action

AlmightyYakob:
> And to answer my own question: correct since client would just mint them based on known API download end point for a given asset https://github.com/dandi/dandiarchive/pull/773/files#diff-7d09a6d75d4c857695273c3d27a9cbdf2f4fce4e1eaee4ee0dc125c2d81e528fR221
> 
> Sounds good, albeit spreading the logic across clients. Excited to see it in action

Exactly, we realized we has no reason to include those URLs in this endpoint. Addressing the hard coded URLs should still be addressed, but it no longer makes sense to do that here.

PR #313, opened 2021-05-27
make citation more like zenodo

satra:


dandiapi/api/models/version.py
```diff
@@ -86,10 +86,11 @@ def copy(cls, version):
     @classmethod
     def citation(cls, metadata):
         year = datetime.datetime.now().year
-        name = metadata['name']
+        name = metadata['name'].rstrip('.')
         url = metadata['url']
+        version = metadata['version']
         # If we can't find any contributors, use this citation format
-        citation = f'{name} ({year}). Online: {url}'
+        citation = f'{name} ({year}). (Version {version}) [Data set]. DANDI archive. {url}'
```
yarikoptic, line 93: just some context since I wondered why `Data set` and not `Dataset`? It seems that it is how Datacite renders it  as well even though uses `Dataset` to describe the entry (why oh why?!):
image

satra: oh why indeed . sometimes i just stop asking.

satra:
@dchiquito - i leave to you if this should be merged into master or staging. would be good to get this in. i'm assuming that this is one of the things that saved with metadata whenever it is altered/saved.

dchiquito (approved):
I think this should go into `staging`, so I rebased it.

PR #315, opened 2021-05-27
BF: swap method call names for announcement email renditions

yarikoptic:
ATM html is sent as text, and text as html

Just received an email which looked odd after html was processed by lynx for my glorious emacs.
Looked into the code and then confirmed in the email that indeed swapped.

Filed https://github.com/dandi/dandi-api/issues/314 as a supplement

edit 1:

since needed a fix in the test, PR became a little more than just a bugfix and introduced my opinionated changes to the email text body formatting and the test.  Do not hesitate to suggest improvements etc.

### DIFF
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -101,9 +101,12 @@ def registered_message(user, socialaccount):
 
 Welcome to DANDI.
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
eh, test made certain it is incorrect, uff...

waxlamp (approved):

yarikoptic:
Ok, no further comments, let's proceed

PR #316, opened 2021-05-29
Validate

dchiquito:
Bring validation+publish into compliance with the design doc.
This is a draft until I can come back and fix up the commit history for easy digestion. Feel free to review/merge if urgent.


dandiapi/api/models/version.py
```diff
@@ -153,14 +157,63 @@ def citation(cls, metadata):
...
+                if 'dataStandard' in metadata:
+                    data_standards = data_standards.union(set(metadata['dataStandard']))
+                if 'approach' in metadata:
+                    approaches = approaches.union(set(metadata['approach']))
+                if 'measurementTechnique' in metadata:
+                    measurement_techniques = measurement_techniques.union(
+                        set(metadata['measurementTechnique'])
+                    )
+                if 'species' in metadata:
+                    species = species.union(set(metadata['species']))
```
satra, line 201: unfortunately many of these properties are not at the top level.  also this will need an update of all asset metadata as proposed in this PR: https://github.com/dandi/dandi-cli/pull/644/files (one of the reasons why we need to coordinate the full 0.4.0 upgrade)

dandiapi/api/models/version.py
```diff
@@ -153,6 +153,22 @@ def citation(cls, metadata):
...
+        computed_fields = [
+            'name',
+            'identifier',
+            'version',
+            'id',
+            'url',
+            'citation',
+            'doi',
+            'datePublished',
+            'publishedBy',
```
dchiquito, line 172: Check that these are the only required fields.

dandiapi/api/models/asset.py
```diff
@@ -123,21 +124,65 @@ def sha256(self):
...
+            'digest': self.blob.digest,
+        }
+        return metadata
+
+    @classmethod
+    def published_asset(cls, asset: Asset):
+        """
+        Generate a published asset + metadata without saving it.
+
+        This is useful to validate version metadata without saving it.
```
dchiquito, line 149: validate asset metadata

dchiquito:
@satra I'm merging this into the staging branch so it doesn't grow any more out of control.

yarikoptic:
> @yarikoptic - what do you think. the CLI would take a list of asset metadata or ids and compute the assetsSummary. i could also put this in dandischema if there is some reason that should be preferred.

eventually we should indeed should add producing a summary (well -- there is also some but only at "visualization" level by pyout) for a list of assets. But at this point, not sure if worthwhile aiming for that at dandi-cli side - a big RF to dandi-cli is coming so might be a bit premature to do that now, in particular since function itself should be quite trivial and for now can just live in `dandi-api` - the place where it is needed most and can be done at "close to DB" level instead of going through full metadata records.

satra:
the main reason i was thinking cli/schema is because i would like the api not to really try to understand the schema, which unfortunately this aggregation requires. (like finding species, approaches, techniques) in the short term i'm fine with doing this in the api, but it would be nice to move it away from the api.

PR #317, opened 2021-06-01
Increase max page size to 1000

### LINKED ISSUE #252: page size greater than 100 has not effect on GET on assets
both the following snippets return 100 records. if number is less than 100, it returns the appropriate number.

```python
import requests
url = "https://api.dandiarchive.org/api/dandisets/000026/versions/draft/assets/"
querystring = {"path":"sub-I46/microscopy","page_size":"100"}
headers = {"Accept": "application/json"}
response = requests.request("GET", url, headers=headers, params=querystring)
len(response.json()["results"])
```

```python
querystring = {"path":"sub-I46/microscopy","page_size":"1000"}
response = requests.request("GET", url, headers=headers, params=querystring)
len(response.json()["results"])
```

waxlamp:
Closes #252

### DIFF
--- a/dandiapi/api/views/common.py
+++ b/dandiapi/api/views/common.py
@@ -3,5 +3,5 @@ from rest_framework.pagination import PageNumberPagination
 
 class DandiPagination(PageNumberPagination):
     page_size = 25
-    max_page_size = 100
+    max_page_size = 1000
     page_size_query_param = 'page_size'
--- a/dandiapi/api/views/common.py
+++ b/dandiapi/api/views/common.py
@@ -1,7 +1,13 @@
 from rest_framework.pagination import PageNumberPagination
 
+max_page_size = 1000
+
 
 class DandiPagination(PageNumberPagination):
     page_size = 25
-    max_page_size = 1000
+    max_page_size = max_page_size
     page_size_query_param = 'page_size'
+
+    page_size_query_description = (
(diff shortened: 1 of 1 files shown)


yarikoptic:
could this value be somehow displayed in a swagger interface along for `page_size` ? otherwise even with 1000 might be unexpected why I am not getting all my 10,000 in a single sweep .  So e.g. instead of `Number of results to return per page.` say `Number of results to return per page (limited to max: 1000).` or alike

waxlamp:
> could this value be somehow displayed in a swagger interface along for `page_size` ? otherwise even with 1000 might be unexpected why I am not getting all my 10,000 in a single sweep . So e.g. instead of `Number of results to return per page.` say `Number of results to return per page (limited to max: 1000).` or alike

@yarikoptic, I figured out how to do this. Try it out and let me know what you think.

yarikoptic:
> @yarikoptic, I figured out how to do this. Try it out and let me know what you think.

unfortunately my local devel dandi-api instance fails to install ATM - yet to figure out.  
But if works for you in a swagger interface -- I would trust you and just test whenever it gets merged (note: some tests are failing, might be unrelated)

waxlamp:
@yarikoptic, this has deployed to production, so you can see the updated documentation and increased max page size in action on the Swagger page. Let me know if it looks good to you.

yarikoptic:
I see 
image
and that looks great! Thank you @waxlamp !

PR #318, opened 2021-06-01
Add `--no-threading` to the Dockerfile

waxlamp:
Sometimes the database drops connections after a while, and this is the StackOverflow-sourced solution.

@brianhelba, can you advise to the downsides of doing it this way? Also, should we upstream this change to the Girder 4 cookiecutter?

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: ./dev/django.Dockerfile
-    command: ["./manage.py", "runserver", "0.0.0.0:8000"]
+    command: ["./manage.py", "runserver", "--nothreading", "0.0.0.0:8000"]
     # Log printing via Rich is enhanced by a TTY
     tty: true
     env_file: ./dev/.env.docker-compose


yarikoptic:
FWIW this issue came up before in `dandi-cli` testing https://github.com/dandi/dandi-api/issues/190, so we got it added in the https://github.com/dandi/dandi-cli/blob/cf8e0ea872b5c4d6d06109a2d9d9b3e7a9009fcf/dandi/tests/data/dandiarchive-docker/docker-compose.yml#L23 and I believe it was all better since then.

brianhelba (changes requested):
@waxlamp I do not understand why this would improve or fix anything. I generally understand Django's development server to be stable with the default CLI options. Can you post the StackOverflow post you're referencing?

waxlamp:
> @waxlamp I do not understand why this would improve or fix anything. I generally understand Django's development server to be stable with the default CLI options. Can you post the StackOverflow post you're referencing?

https://stackoverflow.com/questions/60843123/djangopostgres-fatal-sorry-too-many-clients-already/66270520#66270520

Note that it seems to be a postgres problem that happens to be solved by single-threading the Django dev server (well, that's my diffident understanding thereof).

brianhelba:
I'd recommend against this change as a general matter. A single threaded dev server would force all page resources to load serially and would slow down the developer experience. If something specific about the DANDI stack eventually requires this, it is not something which should be generalized to all Django projects.

If there is a problem with database connections, I hope we can investigate and solve that directly via a specific reported issue.

waxlamp:
Thanks for weighing in on this, @brianhelba. I'm going to close the PR and will open an issue if and when the problem recurs.

PR #321, opened 2021-06-08
First draft of staging design doc

dchiquito:
Staging will be important for testing the publish process, so this is how we are envisioning staging will be used.

### DIFF
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -16,6 +16,8 @@ This should exactly parallel production, except for the S3 bucket, which is priv
 Nothing is actually shared, everything is hosted adjacently.
 
 The staging server will contain some test data and some (smallish) real data that reflects the contents of production.
+Developers will manage this manually and upload/delete data as necessary for testing.
+All staging data is ephemeral and should never be relied on to continue existing.
 
 ## Release procedure
 The staging server will track the `master` branch in `dandi-api` and will automatically redeploy whenever `master` is changed.
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -1,0 +1,37 @@
+# Staging Environment and Procedures
+
+## Production Environment
+* Django API server, Postgres, and CloudAMQP running in Heroku
+* Sponsored S3 bucket in the sponsored account in AWS
+* Web GUI served through Netlify
+* Sentry instance
+
+## Staging Environment
+* Django API server, Postgres, and CloudAMQP running in Heroku
+* Private S3 bucket in the project account in AWS
+* Web GUI served through Netlify
+* Sentry instance
+
+This should exactly parallel production, except for the S3 bucket, which is private.
+Nothing is actually shared, everything is hosted adjacently.
+
+The staging server will contain some test data and some (smallish) real data that reflects the contents of production.
+
+## Release procedure
+The staging server will track the `master` branch in `dandi-api` and will automatically redeploy whenever `master` is changed.
+
+The production server will not track any branch. Instead, it will automatically redeploy whenever a release is tagged in GitHub.
+
+The staging web GUI will track the `staging` branch in `dandiarchive` and will automatically redeploy whenever `staging` is changed.
+
+The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
+
+* A developer has a feature they want to add that impacts both the backend and the frontend.
+* They checkout a new branch in `dandi-api` and `dandiarchive`, do the work, and submit related PRs in both repos.
+* Once approved, the `dandi-api` PR is merged into `master` and the `dandiarchive` PR is merged into `staging`.
+* The changes are tested in staging.
+* Once approved, a new release of `dandi-api` is tagged, pushing the changes to production.
+* Once approved, the developer manually merges `dandiarchive/staging` into `dandiarchive/master`.
+
+## TODO
+Figure out a less manual release procedure for `dandiarchive`.
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -1,15 +1,15 @@
 # Staging Environment and Procedures
 
 ## Production Environment
-* Django API server, Postgres, and CloudAMQP running in Heroku
+* Django API server (`api.dandiarchive.org`), Postgres, and CloudAMQP running in Heroku
 * Sponsored S3 bucket in the sponsored account in AWS
-* Web GUI served through Netlify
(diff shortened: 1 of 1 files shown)


doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+
+* A developer has a feature they want to add that impacts both the backend and the frontend.
+* They checkout a new branch in `dandi-api` and `dandiarchive`, do the work, and submit related PRs in both repos.
+* Once approved, the `dandi-api` PR is merged into `master` and the `dandiarchive` PR is merged into `staging`.
+* The changes are tested in staging.
+* Once approved, a new release of `dandi-api` is tagged, pushing the changes to production.
+* Once approved, the developer manually merges `dandiarchive/staging` into `dandiarchive/master`.
+
+## TODO
+Figure out a less manual release procedure for `dandiarchive`.
```
satra, line 37: perhaps once merged into staging, an action can autocreate a PR to merge staging into master.

satra:
this makes sense to me. is the notion of code respecting different api (production, staging, local development) a separate step or related to this design doc?

doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+Nothing is actually shared, everything is hosted adjacently.
+
+The staging server will contain some test data and some (smallish) real data that reflects the contents of production.
+
+## Release procedure
+The staging server will track the `master` branch in `dandi-api` and will automatically redeploy whenever `master` is changed.
+
+The production server will not track any branch. Instead, it will automatically redeploy whenever a release is tagged in GitHub.
+
+The staging web GUI will track the `staging` branch in `dandiarchive` and will automatically redeploy whenever `staging` is changed.
```
yarikoptic, line 25: why not to make it consistent between the two (dandi-api and webui) one way or another? 
E.g. the easiest might be to hav `staging` branch in both and keep `master` for auto-deployment (but be careful to not target development PRs against it).  `releases` could be "minted" on top of master at will and/or for each PR against `master`.  "place staging into production" could be a PR to have `staging` merged into `master`.

Alternative naming -- `master` (staging) + `maint` (or `releases`) branch.  One way or another -- consistency would only benefit IMHO.

doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+
+The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
+
+* A developer has a feature they want to add that impacts both the backend and the frontend.
+* They checkout a new branch in `dandi-api` and `dandiarchive`, do the work, and submit related PRs in both repos.
+* Once approved, the `dandi-api` PR is merged into `master` and the `dandiarchive` PR is merged into `staging`.
```
yarikoptic, line 31: would be simplified with symmetric workflow (see above)

doc/design/staging-1.md
yarikoptic: We use intuit's `auto` for dandi-cli, dandischema, and others. Could be used here as well.

yarikoptic (commented):
Initial comments.

doc/design/staging-1.md
dchiquito: I would like to have the deploy mechanism consistent and just tag a release in both, but the issue is that Netlify only supports automatic deploys from a branch. 

How about having development for `dandiarchive` happen in `master` and have staging track `master` as you suggest, then having a GitHub action that merges `master` into a `releases` branch whenever a release is tagged?

dchiquito:
> is the notion of code respecting different api (production, staging, local development) a separate step or related to this design doc?

That is separate. This is a description of the ideal world we would like to live in, not a description of the work required to get there.

doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Production Environment
+* Django API server, Postgres, and CloudAMQP running in Heroku
+* Sponsored S3 bucket in the sponsored account in AWS
+* Web GUI served through Netlify
+* Sentry instance
+
+## Staging Environment
+* Django API server, Postgres, and CloudAMQP running in Heroku
+* Private S3 bucket in the project account in AWS
```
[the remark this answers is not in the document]
dchiquito: I intended for this document to be a description of the development > staging > production process, not a comprehensive documentation of our environment. I think adding that information will only make the document expire quicker as we tweak things and forget to update it here.

It would be nice to have that information somewhere, but I don't think it should be here.

doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+* Web GUI served through Netlify
```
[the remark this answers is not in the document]
dchiquito: It will be `api-staging.dandiarchive.org`. I think that's relevant enough, I'll add it to the doc.

doc/design/staging-1.md
```diff
@@ -0,0 +1,37 @@
...
+
+This should exactly parallel production, except for the S3 bucket, which is private.
+Nothing is actually shared, everything is hosted adjacently.
+
+The staging server will contain some test data and some (smallish) real data that reflects the contents of production.
+
+## Release procedure
+The staging server will track the `master` branch in `dandi-api` and will automatically redeploy whenever `master` is changed.
+
+The production server will not track any branch. Instead, it will automatically redeploy whenever a release is tagged in GitHub.
```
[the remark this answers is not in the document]
dchiquito: Consistency sounds nice. I don't have strong feelings about the versioning scheme.

@waxlamp mentioned something about GitHub dispatching or something that would potentially make the release process cleaner?

doc/design/staging-1.md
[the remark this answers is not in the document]
waxlamp: I agree with @dchiquito on this: let's leave that decision to execution time, and then document somewhere that isn't this design doc.

doc/design/staging-1.md
waxlamp: As I mentioned above, I think leaving the dandiarchive workflow largely alone could also work, perhaps with a bit of CI that checks to see whether `netlify.toml` has been modified and disallows merge to master if it has.

doc/design/staging-1.md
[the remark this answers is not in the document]
yarikoptic: > GitHub does not create annotated tags for releases--why is that?

my guess so that you could adjust description without affecting the tag.  But if you push annotated tag, and just "Edit/Save" - it would produce a "proper" GitHub release based on that annotated tag. Have been doing that for many projects, never "minted" a release "from scratch" on github.

PR #322, opened 2021-06-09
Add migrate_version_metadata script

dchiquito:


satra (changes requested):
it may be good to start protecting our scripts against modifying published versions.

PR #323, opened 2021-06-09
Use schema version 0.4.1 in dockerfiles and CI config

mvandenburgh:
Updates the metadata schema version in the docker and CI config files. The new schema must be used in order to support the new version of VJSF in the client (https://github.com/dandi/dandiarchive/pull/730).


dchiquito:
@mvandenburgh could you rebase this onto the `staging` branch and change the merge target to `staging`? That's where we're putting everything publish related for now, and this is a publish blocker.

dchiquito:
Once this is merged, `master` will need to be merged into `staging` so this is also applied there, as this is a publish blocker.

dchiquito (changes requested):
Could you also update the `schemaVersion` fields in factories.py? I hardcoded them to 0.3.1 so there is a stable schema we use to test that metadata can validate.
In general I think it's not important that we track the latest version for metadata tests, but it's important that there are no problems with validating against the schema version we are using for publish.

dchiquito (approved):
Turns out bumping to 0.4.1 in the test factory breaks all the validation tests, as 0.4.1 has stricter requirements around metadata. That is outside the scope of this PR, so strike that request. LGTM

dchiquito:
The schema version has since moved on, we are now on 0.4.3. https://github.com/dandi/dandi-api/pull/345 upgrades the version already, so I'm closing this.

PR #325, opened 2021-06-10
Use dandischema to compute DOIs

dchiquito:
See https://github.com/dandi/dandischema/pull/33 for discussion of the approach taken here. That PR also blocks this one.

### DIFF
--- a/dandiapi/api/doi.py
+++ b/dandiapi/api/doi.py
@@ -18,26 +18,25 @@ def _generate_doi_data(version: Version):
 
 
 def create_doi(version: Version) -> str:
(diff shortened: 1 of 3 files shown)


mvandenburgh (approved):
LGTM

mvandenburgh:
Looks like some dandi-cli tests are failing, does something in that test need to be updated to account for DOI's now working outside of production?

dchiquito:
@mvandenburgh Those errors are from dandi-cli integration tests that are trying to publish with invalid data. They will need to be updated eventually before we merge `staging`.

PR #326, opened 2021-06-11
Clarify development procedures

dchiquito:
I forgot to clarify the development process in the original PR, here are those changes.

### DIFF
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -30,12 +30,22 @@ To test code changes to the web GUI, branch previews will also be pointed at the
 
 The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
 
-* A developer has a feature they want to add that impacts both the backend and the frontend.
+A developer has a feature they want to add that impacts both the backend and the frontend.
 * They checkout a new branch in `dandi-api` and `dandiarchive`, do the work, and submit related PRs in both repos.
-* Once approved, the `dandi-api` PR is merged into `master` and the `dandiarchive` PR is merged into `staging`.
-* The changes are tested in staging.
-* Once approved, a new release of `dandi-api` is tagged, pushing the changes to production.
-* Once approved, the developer manually merges `dandiarchive/staging` into `dandiarchive/master`.
+* Once approved, the `dandi-api` PR is merged into `master`.
+* The `dandiarchive` changes are tested in the branch preview, which points at the staging environment.
+* Once `dandiarchive` is approved, the developer merges the PR, and tags a new release of `dandi-api`, which is then automatically deployed to production.
+
+A developer has a feature they want to add that only impacts the frontend.
+* They checkout a new branch in `dandiarchive`, do the work, and submit a PR.
+* The changes are tested in the branch preview, which points at the staging environment.
+* Once approved, the developer merges the PR to `master`, which is then automatically deployed to production.
+
+A developer has a feature they want to add that only impacts the backend.
+* They checkout a new branch in `dandi-api`, do the work, and submit a PR.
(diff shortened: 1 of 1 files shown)


doc/design/staging-1.md
```diff
@@ -30,12 +30,22 @@ To test code changes to the web GUI, branch previews will also be pointed at the
 
 The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
 
-* A developer has a feature they want to add that impacts both the backend and the frontend.
+A developer has a feature they want to add that impacts both the backend and the frontend.
 * They checkout a new branch in `dandi-api` and `dandiarchive`, do the work, and submit related PRs in both repos.
```
yarikoptic, line 34: both against `master` branch? (it is better to state branches at this step since PR needs to know the base branch to start with, so better to state here and remove branches mentioning below for "PR is merged into ...")

doc/design/staging-1.md
```diff
@@ -30,12 +30,22 @@ To test code changes to the web GUI, branch previews will also be pointed at the
...
-* Once approved, the `dandi-api` PR is merged into `master` and the `dandiarchive` PR is merged into `staging`.
-* The changes are tested in staging.
-* Once approved, a new release of `dandi-api` is tagged, pushing the changes to production.
-* Once approved, the developer manually merges `dandiarchive/staging` into `dandiarchive/master`.
+* Once approved, the `dandi-api` PR is merged into `master`.
+* The `dandiarchive` changes are tested in the branch preview, which points at the staging environment.
+* Once `dandiarchive` is approved, the developer merges the PR, and tags a new release of `dandi-api`, which is then automatically deployed to production.
```
yarikoptic, line 37:
```suggestion
* Once `dandiarchive` PR is approved, the developer merges the PR, and tags a new release of `dandi-api`, which is then automatically deployed to production.
```

doc/design/staging-1.md
```diff
@@ -30,12 +30,22 @@ To test code changes to the web GUI, branch previews will also be pointed at the
 
 The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
 
-* A developer has a feature they want to add that impacts both the backend and the frontend.
+A developer has a feature they want to add that impacts both the backend and the frontend.
```
yarikoptic, line 33: in above text, not part of the PR, so can't comment there:
```
The staging web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
...
The production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
```
doesn't sound right since they look identical to me and thus in effect we do not have "staging" (beyond state of PR) for "web GUI", is that the intention now (sorry if I didn't chime in time on that discussion). Or is that the current plan?

yarikoptic: if that it the plan, might be better to have a single sentence like
```
Both staging and production web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
```

dchiquito: The next sentence for the staging GUI description clarifies:
```
The configuration variables for the staging Netlify deployment will point it at the staging server rather than the production server.
```
So there two permanent GUI servers that are running the exact same frontend code, but are pointing at different backends.

If you want to test frontend changes against staging, you would use a branch preview, which is also pointed at the staging API server.

doc/design/staging-1.md
dchiquito: I have removed any reference in this file to any other branch except `master`, so I think it would be implicit to anyone reading the document. If anyone is confused about where to base the PR when following these instructions, I don't think it's unreasonable for them to read one step ahead for the answer.

waxlamp (approved):

PR #327, opened 2021-06-11
Add additional sorting options for dandisets

mvandenburgh:
Server side changes required for https://github.com/dandi/dandiarchive/pull/732. 

As of right now, this PR determines a given dandiset's `modified` value using only the version metadata and not any of the asset metadata (see discussion [here](https://github.com/dandi/dandiarchive/pull/732#issuecomment-859722370)).

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -49,7 +49,8 @@ class DandisetFilterBackend(filters.OrderingFilter):
                 queryset = queryset.annotate(name=Subquery(latest_version.values('metadata__name')))
                 return queryset.order_by(ordering)
             elif ordering.endswith('modified'):
-                # name refers to the name of the most recent version, so a subquery is required
+                # modified refers to the modification timestamp of the most
+                # recent version, so a subquery is required
                 latest_version = Version.objects.filter(dandiset=OuterRef('pk')).order_by(
                     '-created'
                 )[:1]
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -53,12 +53,11 @@ class DandisetFilterBackend(filters.OrderingFilter):
                 latest_version = Version.objects.filter(dandiset=OuterRef('pk')).order_by(
                     '-created'
                 )[:1]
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -48,7 +48,16 @@ def filter_queryset(self, request, queryset, view):
...
+            elif ordering.endswith('modified'):
+                # name refers to the name of the most recent version, so a subquery is required
+                latest_version = Version.objects.filter(dandiset=OuterRef('pk')).order_by(
+                    '-created'
+                )[:1]
+                # get the `modified` field of the most recent version
+                queryset = queryset.annotate(
+                    modified_version=Subquery(latest_version.values('modified'))
+                )
+                return queryset.order_by(f'{ordering}_version')
```
dchiquito, line 60: Does Django allow this? If so I think it's cleaner. If not, maybe leave a note about why you can't just use `modified`.
```suggestion
                # Overwrite the `modified` field of the Dandiset with the most recent Version's
                queryset = queryset.annotate(modified=Subquery(latest_version.values('modified')))
                return queryset.order_by(ordering)
```

mvandenburgh: No, Django throws an error since `Dandiset` already has a `modified` field. I added a comment explaining it.

dchiquito (approved):

dandiapi/api/views/dandiset.py
```diff
@@ -48,7 +48,18 @@ def filter_queryset(self, request, queryset, view):
                 )[:1]
                 queryset = queryset.annotate(name=Subquery(latest_version.values('metadata__name')))
                 return queryset.order_by(ordering)
-
+            elif ordering.endswith('modified'):
+                # modified refers to the modification timestamp of the most
+                # recent version, so a subquery is required
+                latest_version = Version.objects.filter(dandiset=OuterRef('pk')).order_by(
+                    '-created'
+                )[:1]
```
brianhelba, line 56: @mvandenburgh FYI https://docs.djangoproject.com/en/3.2/ref/models/querysets/#first

PR #328, opened 2021-06-11
Documentation improvements

mvandenburgh:
* Add step to install docker/docker-compose
* More explicit install instructions for `psycopg2`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,21 @@ This configuration still uses Docker to run attached services in the background,
 but allows developers to run Python code on their native system.
 
 ### Initial Setup
-1. Run `docker-compose -f ./docker-compose.yml up -d`
-2. Install Python 3.8
-3. Install
-   [`psycopg2` build prerequisites](https://www.psycopg.org/docs/install.html#build-prerequisites)
-4. Create and activate a new Python virtualenv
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,15 +26,21 @@ This configuration still uses Docker to run attached services in the background,
...
+1. Install [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/)
+2. Run `docker-compose -f ./docker-compose.yml up -d`
+3. Install Python 3.8
+4. Install
+  [`psycopg2` build prerequisites](https://www.psycopg.org/docs/install.html#build-prerequisites):
+  ```
+  sudo apt install libpq-dev
+  export PATH=/usr/lib/postgresql/X.Y/bin/:$PATH
+  pip install psycopg2
+  ```
```
waxlamp, line 38: I think this is fine here, but in general it's probably too detailed for build instructions for a project of ours. In this case, you could even add a note saying something like "on Ubuntu <year.month>:" to make it clear that this is a specific example of how to do this.

waxlamp (approved):

waxlamp (approved):

PR #330, opened 2021-06-14
Wording tune up to TXT version of the registration email

yarikoptic:
In the attempt to establish a "human in the middle" approval of the accounts.

TODO:

- [ ] html version after TXT version agreed to be ok

### DIFF
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -106,7 +106,7 @@ You are now registered on the DANDI archive!
 It may take up to 24 hours for your account to be approved and activated.  During the approval
 process DANDI team might contact you with additional questions.
 
-Once your account is activated you would be able to create Dandisets and upload data.
+Once your account is activated you will be able to create Dandisets and upload data.
 Within 24 hours after approval, your account on Jupyterhub (https://hub.dandiarchive.org) for
 computing on dandisets in the cloud will be activated, and your email will be
 registered with our Slack workspace.
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -101,12 +101,15 @@ def registered_message(user, socialaccount):
 
 Welcome to DANDI.
 
-You are now registered on the DANDI archive. Registering allows you to create Dandisets and upload
-data right away. You can also use the Jupyterhub (https://hub.dandiarchive.org) for computing on
-dandisets in the cloud.
+You are now registered on the DANDI archive!
 
-It may take up to 24 hours for your hub account to be activated and for your email to be registered
-with our Slack workspace.
+It may take up to 24 hours for your account to be approved and activated.  During the approval
+process DANDI team might contact you with additional questions.
+
+Once your account is activated you would be able to create Dandisets and upload data.
+Within 24 hours after approval, your account on Jupyterhub (https://hub.dandiarchive.org) for
+computing on dandisets in the cloud will be activated, and your email will be
+registered with our Slack workspace.
 
 Please use the following links to post any questions or issues.
 


dandiapi/api/mail.py
```diff
@@ -101,12 +101,15 @@ def registered_message(user, socialaccount):
...
-data right away. You can also use the Jupyterhub (https://hub.dandiarchive.org) for computing on
-dandisets in the cloud.
+You are now registered on the DANDI archive!
 
-It may take up to 24 hours for your hub account to be activated and for your email to be registered
-with our Slack workspace.
+It may take up to 24 hours for your account to be approved and activated.  During the approval
+process DANDI team might contact you with additional questions.
+
+Once your account is activated you would be able to create Dandisets and upload data.
```
waxlamp, line 109:
```suggestion
Once your account is activated you will be able to create Dandisets and upload data.
```

dandiapi/api/mail.py
```diff
@@ -101,12 +101,15 @@ def registered_message(user, socialaccount):
...
 Welcome to DANDI.
 
-You are now registered on the DANDI archive. Registering allows you to create Dandisets and upload
-data right away. You can also use the Jupyterhub (https://hub.dandiarchive.org) for computing on
-dandisets in the cloud.
+You are now registered on the DANDI archive!
 
-It may take up to 24 hours for your hub account to be activated and for your email to be registered
-with our Slack workspace.
+It may take up to 24 hours for your account to be approved and activated.  During the approval
```
waxlamp, line 106: I used to put a double space after ending sentences but a few years ago I learned that it's no longer the prevailing practice.

```suggestion
It may take up to 24 hours for your account to be approved and activated. During the approval
```

waxlamp (changes requested):

dandiapi/api/mail.py
yarikoptic: "not prevailing" doesn't mean "not good" ;) Is it recommended for some specific reasons(s) to not do that?
Note: this is a text version of the email and IMHO double spacing helps to visually parse text.  HTML version will not have that

waxlamp: The problem is, because it's "not prevailing", most rendering engines won't treat them the way typewriters did (i.e., present an extra space to make visual processing faster). I've seen things like a the double space getting "broken" across the ends of lines, etc.

For stuff like this, I like to just go with the prevailing notion of correctness (sort of like using Python Black to format source files) but honestly I don't have very strong feelings about this particular instance. If you want to keep it, that's ok with me--just resolve this thread so it's out of sight/out of mind 😸

waxlamp:
@yarikoptic, could you finish this up and merge it so that #342 can be merged without causing a conflict here?

yarikoptic:
> @yarikoptic, could you finish this up and merge it so that #342 can be merged without causing a conflict here?

well - we should finish/merge/deploy this only if we switch default behavior to have disabled users upon creation.  Is that doable/agreed upon to proceed?

waxlamp:
> > @yarikoptic, could you finish this up and merge it so that #342 can be merged without causing a conflict here?
> 
> well - we should finish/merge/deploy this only if we switch default behavior to have disabled users upon creation. Is that doable/agreed upon to proceed?

In that case, let's hold off on this PR until after
1. @brianhelba merges his email template infrastructure changes, and
2. we can get more clarity on exactly how user registration should behave (including any other PRs necessary to implement what is reflected in this PR's explanatory changes).

If you are ok with that, could you change this PR's status to draft? After Brian's template changes land, it should be relatively easy to reconstitute your wording in the updated infrastructure.

yarikoptic:
it is draft now (should have been from the beginning, thanks @waxlamp )

dandiapi/api/mail.py
waxlamp: ping @yarikoptic

waxlamp:
This has been solved via the user questionnaire.

PR #335, opened 2021-06-15
Fix publish button

dchiquito:
Previously VersionDetailSerializer was simply serializing the Version
data, so it was listing a Version as `VALID` even though it contained
assets that were not `VALID`. This made the Publish button in the UI
available even when publish was not permitted.

Add some computed properties `publish_status` and
`publish_validation_error` that check both the Version status and the
status of all Assets in the Version.

Use those fields in the `VersionDetailSerializer`.


mvandenburgh (approved):
LGTM

PR #339, opened 2021-06-16
Use Django management commands for scripts

brianhelba:
Using custom management commands, along with `Django-Click`, is a simpler and more native way to provide scripts.


mvandenburgh (approved):
Looks good to me.

dchiquito (approved):

dandiapi/api/management/commands/refresh_metadata.py
```diff
@@ -0,0 +1,19 @@
...
+
+@click.command()
+def refresh_metadata():
+    """
+    Re-save all Versions and Assets to refresh their metadata.
+
+    This script should be run after modifying the computed metadata for either Versions or Assets.
+    Since the computed metadata is only computed when the model is saved, the models need to be
+    saved to trigger computation.
+    """
```
yarikoptic, line 14: I wonder what are exact "effects" of running this command since it talks in the terms of "model is saved" etc.  E.g. -- would it mint new assets?  would "published" (released) assets be reminted?

dchiquito: It does not mint new assets, but it might update the metadata of assets which have been published.

PR #340, opened 2021-06-16
Use logging instead of printing

brianhelba:


waxlamp (approved):
Does papertrail hook into Python's logging mechanism? That is, will this change cause these print statements to end up in the papertrail logs?

PR #342, opened 2021-06-16
Use Django templates to render email bodies

brianhelba:
This is significantly more maintainable than using inline strings.


brianhelba:
This might have conflicts with #330.

waxlamp (approved):
Looks like a good improvement. To test this, can we merge to `staging` rather than `master`?

PR #343, opened 2021-06-16
Validate version metadata triggers

dchiquito:
Modifying asset metadata (through upload/update) triggers a save of the Version, which also populates and potentially modifies the version metadata.

Trigger metadata validation whenever `version.save()` is called in Asset views.


mvandenburgh (approved):
Looks good to me.

PR #345, opened 2021-06-17
Dandischema validation

dchiquito:
The jsonschema is derived from the pydantic models in dandischema, so it is better to simply use those models to validate directly.

I was trying to publish locally and noticed that there were some pydantic model errors (I forget what) that weren't caught by the jsonschema, so I thought it was time for this update.

@satra FYI


satra:
awesome yes this looks good.

PR #351, opened 2021-06-21
Automatically add dandiset creator as contact person

mvandenburgh:
Closes https://github.com/dandi/dandiarchive/issues/425. The user creating a dandiset will be automatically added as a contributor with a role of `dandi:ContactPerson` when the dandiset is initially created. 

I had to hardcode certain assumptions about the schema to accomplish this, since the API server injects the logged-in user's info into the dandiset's metadata upon creation of it, meaning that a future schema change may break this.

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -110,6 +110,16 @@ class DandisetViewSet(ReadOnlyModelViewSet):
         # Strip away any computed fields
         metadata = Version.strip_metadata(metadata)
 
+        metadata['contributor'] = [
+            {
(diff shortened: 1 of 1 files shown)


dchiquito (commented):
I think it would be better to merge this into `staging` rather than `master` to avoid a merge conflict and to implement it with the more up to date schema version.

I know for a fact that the `roleName` will need to be changed to `dcite:ContactPerson`, and I think `schemaKey` may have been removed as well.

mvandenburgh:
Per our in person discussion yesterday, we can merge this into `staging` as-is and address the concerns in https://github.com/dandi/dandi-api/issues/352 later on.

PR #353, opened 2021-06-22
Allow wildcards in OAuth redirect URIs in test environment

mvandenburgh:
Allow Netlify branch deploys to login using OAuth in staging.


dandiapi/settings.py
```diff
@@ -83,3 +83,8 @@ class ProductionConfiguration(DandiMixin, ProductionBaseConfiguration):
 class HerokuProductionConfiguration(DandiMixin, HerokuProductionBaseConfiguration):
     # All login attempts in production should go straight to GitHub
     LOGIN_URL = '/accounts/github/login/'
+
+
+# use the same config for staging as production, but also allow regex for OAuth redirect URIs
+class HerokuStagingConfiguration(HerokuProductionBaseConfiguration):
```
mvandenburgh, line 89: The heroku config variable `DJANGO_CONFIGURATION` would need to be changed to `HerokuStagingConfiguration` to support this.

dandiapi/api/models/oauth.py
```diff
@@ -0,0 +1,20 @@
...
+        try:
+            super().clean()
+        except ValidationError as e:
+            # don't validate URLs so we can use regexes too
+            if 'Enter a valid URL.' not in str(e):
+                raise e
+
+    def redirect_uri_allowed(self, uri):
+        for allowed_uri in self.redirect_uris.split():
+            if re.fullmatch(uri, allowed_uri):
```
dchiquito, line 18: Let's see if we can use fnmatch instead https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch

For reference: https://github.com/girder/girder/blob/d55b2c60c678f7acc12f5c54b9d60f3838fcbab6/girder/api/rest.py#L754

dandiapi/api/models/oauth.py
```diff
@@ -0,0 +1,20 @@
+import re
+from django.core.exceptions import ValidationError
+from oauth2_provider.models import AbstractApplication
+
+
+class Application(AbstractApplication):
```
dchiquito, line 6: See if you can use `Application(Application)` instead to avoid the extra model/migration

dchiquito (commented):
Couple of comments

dandiapi/api/models/oauth.py
AlmightyYakob: FWIW reference use of `fnmatch` https://github.com/girder/girder/blob/d55b2c60c678f7acc12f5c54b9d60f3838fcbab6/girder/api/rest.py#L754

dandiapi/api/models/oauth.py
mvandenburgh: I tried this and seems a migration is unavoidable because `OAUTH2_PROVIDER_APPLICATION_MODEL` needs to be set to a custom model. I get
```
api.Application.application_ptr: (fields.E301) Field defines a relation with the model 'oauth2_provider.Application', which has been swapped out.
```

dandiapi/api/models/oauth.py
mvandenburgh: Done.

dandiapi/api/models/oauth.py
dchiquito: :(
This should be renamed to something more indicative of it's purpose, also docstrings

dchiquito (approved):

dandiapi/api/models/oauth.py
mvandenburgh: Done

PR #355, opened 2021-06-23
Fix order of `modified` sort

mvandenburgh:
Make the default `modified` sorting order go from most recently modified -> least recently modified. 

Closes https://github.com/dandi/dandiarchive/issues/758.


yarikoptic:
Well, I think icon arrow should also switch direction then (still waiting for preview, but seeing no relevant changes ) since IMHO later date is like a later character in the alphabet

Issue #349, opened 2021-06-18
metadata changes needed for publish


satra:
#347 - isolating these components in a separate issue:

for dandiset:
- [x] dandiset missing in url: `https://dandiarchive.org/000001/0.210618.0045` - this should be `https://dandiarchive.org/dandiset/000001/0.210618.0045` (this i believe is in the metadata save routine. i believe this is true even for draft metadata)
- [x] doi should add 'dandi' like this `10.80507/dandi.000001/0.210618.0045`
- [x] `manifestLocation` should only have assets.yaml i think
- [x] the edit metadata button is still active for the published dandiset in the gui. this should not be true even for admins i believe. just giving @mvandenburgh a heads up.
- [x] the published version also does not show proper assetsSummary in the metadata, so possibly something broke there.

for assets:
- [x] @context is missing
- [x] one of the contentUrl points to production server    "https://api.dandiarchive.org/api/assets/4ca8411a-c74a-4a1c-a909-c4523e71a468/download/",
- [x] `id` should be: `dandiasset:<id>`

also we should consider if we write these out as both yaml and json (@yarikoptic what do you think?)

satra:
adding one more:

in the software description in published by, `id` should become `identifier`. id's are intended to be unique and non-mutable, however, the same RRID could potentially be associated with different versions of the software.

hence this should look like `"identifier": "RRID:SCR_017571",`

```
    "publishedBy": {
        "id": "urn:uuid:b6077cb5-6ad7-46c0-a305-1daa5bd46f67",
        "name": "DANDI publish",
        "endDate": "2021-06-25T13:02:59.368907",
        "schemaKey": "PublishActivity",
        "startDate": "2021-06-25T13:02:59.368907",
        "wasAssociatedWith": [
            {
                "id": "RRID:SCR_017571",
                "name": "DANDI API",
                "version": "0.1.0",
                "schemaKey": "Software"
            }
        ]
    },
```

satra:
@dchiquito - i've checked off the ones that have been addressed.

dchiquito:
@satra The `contentUrl` field is currently hardcoded to always point to `api.dandiarchive.org`. This results in technically incorrect metadata in staging, but shouldn't break anything. Is that import enough to fix today?

satra:
perhaps we can file an issue for that and fix later.

dchiquito:
#286

dchiquito:
[...]

PR #364, opened 2021-06-25
make RRID reference an identifier and add uuid for id

satra:
addresses part of #349  with respect to identifiers for the publish activity


dchiquito:
This is superseded by #372

PR #366, opened 2021-06-25
ENH: run dandi-cli integration testing also for staging branch

yarikoptic:
Some changes in staging broke integration with dandi-cli (and I have missed an
announcement if it was made), and now it is tricky to identify which change
broke it in the list of all the builds.  By running also on staging it
would mitigate that.

Note: might need a merge into staging to trigger?

### DIFF
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - staging
   pull_request:
 
 jobs:


waxlamp (approved):
This looks good but I believe you should change the merge target to `staging` so that it will pick up the change.

Just FYI, after publish is complete, we plan to eliminate the staging branch and instead run staging off of `master`, with a manual release required to promote staging to production.

yarikoptic:
rright -- so this PR is no longer pertinent ;-)

PR #372, opened 2021-06-28
Metadata fixes

dchiquito:
* manifestLocation should only have assets.yaml i think
* the published version also does not show proper assetsSummary in the metadata, so possibly something broke there.
* https://github.com/dandi/dandi-api/pull/370/files#r659029609
* @context is missing


satra (approved):
the changes look good to me.

PR #382, opened 2021-06-30
Make metadata and blob_id optional for PUT asset

### LINKED ISSUE #392: Remove blob_id from asset PUT/POST
ATM to update metadata blob_id is needed, it shouldn't.  Even to create a new asset we do not really need blob_id as there must be etag digest in metadata record, dandi-api can identify the blob from the etag digest and thus completely avoid requiring blob_id in the API call.

While addressing this issue, please do not just remove blob_id from the end point but allow for some deprecation cycle so already released clients remain functionable.  Just needs to verify that blob_id provided matches the one detected I guess

PS from a discussion with @satra last week, but  I think  we forgot to file this isue.

dchiquito:
This should help to resolve https://github.com/dandi/dandi-cli/pull/676#discussion_r661705482

Closes #392.

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -358,6 +358,65 @@ def test_asset_rest_update(api_client, user, draft_version, asset, asset_blob):
 
 
 @pytest.mark.django_db
+def test_asset_rest_update_only_metadata(api_client, user, draft_version, asset):
+    assign_perm('owner', user, draft_version.dandiset)
+    api_client.force_authenticate(user=user)
+    draft_version.assets.add(asset)
+
+    new_path = 'test/asset/rest/update.txt'
+    new_metadata = {
+        'schemaVersion': '0.4.4',
+        'encodingFormat': 'application/x-nwb',
+        'path': new_path,
+        'foo': 'bar',
+        'num': 123,
+        'list': ['a', 'b', 'c'],
+    }
+
+    resp = api_client.put(
+        f'/api/dandisets/{draft_version.dandiset.identifier}/'
+        f'versions/{draft_version.version}/assets/{asset.asset_id}/',
+        {'metadata': new_metadata},
+        format='json',
+    ).data
+    new_asset = Asset.objects.get(asset_id=resp['asset_id'])
+    assert resp == {
+        'asset_id': UUID_RE,
+        'path': new_path,
+        'size': asset.blob.size,
+        'created': TIMESTAMP_RE,
+        'modified': TIMESTAMP_RE,
+        'metadata': new_asset.metadata.metadata,
+    }
+    for key in new_metadata:
+        assert resp['metadata'][key] == new_metadata[key]
+
+
+@pytest.mark.django_db
+def test_asset_rest_update_only_blob_id(api_client, user, draft_version, asset, asset_blob):
+    assign_perm('owner', user, draft_version.dandiset)
+    api_client.force_authenticate(user=user)
+    draft_version.assets.add(asset)
+
+    resp = api_client.put(
+        f'/api/dandisets/{draft_version.dandiset.identifier}/'
+        f'versions/{draft_version.version}/assets/{asset.asset_id}/',
+        {'blob_id': asset_blob.blob_id},
+        format='json',
+    ).data
+    new_asset = Asset.objects.get(asset_id=resp['asset_id'])
+    assert resp == {
+        'asset_id': UUID_RE,
+        'path': asset.path,
+        'size': asset_blob.size,
+        'created': TIMESTAMP_RE,
+        'modified': TIMESTAMP_RE,
+        'metadata': asset.metadata.metadata,
+    }
+    assert asset.blob.blob_id == str(new_asset.blob.blob_id)
+
+
+@pytest.mark.django_db
 def test_asset_rest_update_to_existing(api_client, user, draft_version, asset_factory):
     old_asset = asset_factory()
     existing_asset = asset_factory()
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -77,6 +77,11 @@ class AssetRequestSerializer(serializers.Serializer):
     blob_id = serializers.UUIDField()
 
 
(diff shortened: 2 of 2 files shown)


dandiapi/api/tests/test_asset.py
```diff
@@ -357,6 +357,65 @@ def test_asset_rest_update(api_client, user, draft_version, asset, asset_blob):
...
+
+    resp = api_client.put(
+        f'/api/dandisets/{draft_version.dandiset.identifier}/'
+        f'versions/{draft_version.version}/assets/{asset.asset_id}/',
+        {'metadata': new_metadata},
+        format='json',
+    ).data
+    new_asset = Asset.objects.get(asset_id=resp['asset_id'])
+    assert resp == {
+        'asset_id': UUID_RE,
```
yarikoptic, line 384: is there some magic which I am not aware here -- `==` would do `re.match`if it detects a value be a regex instance ???

dchiquito: Yep, it's magic. To make asserting large JSON blobs easier, we implemented [this class](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/tests/fuzzy.py) which evaluates as being equal to any string that matches the initialized regex. `UUID_RE` specifically is equal to any UUID string.

yarikoptic: interesting! thanks for the pointer!  attn @jwodder -- may be he would like to adopt this too

jwodder: I just recently wrote [a library of things like this](https://github.com/jwodder/anys) after people on IRC said it hadn't been done.

yarikoptic: awesome -- "Great minds think alike" ;)

dandiapi/api/views/asset.py
```diff
@@ -194,20 +199,26 @@ def update(self, request, versions__dandiset__pk, versions__version, **kwargs):
...
-        if 'path' not in metadata:
-            return Response('No path specified in metadata', status=404)
-        path = metadata['path']
-        # Strip away any computed fields
-        metadata = Asset.strip_metadata(metadata)
-        asset_metadata, created = AssetMetadata.objects.get_or_create(metadata=metadata)
-        if created:
-            asset_metadata.save()
+        if 'blob_id' in serializer.validated_data:
+            asset_blob = get_object_or_404(AssetBlob, blob_id=serializer.validated_data['blob_id'])
```
yarikoptic, line 206: If we decide to deprecate providing `blob_id` altogether (#392) we might better here test that it is the same asset here as the `old_asset.blob` instead of allowing for creating an asset for another blob here on PUT.  And if mismatch -- error out.

WDYT @satra?

dandiapi/api/views/asset.py
```diff
@@ -194,20 +199,26 @@ def update(self, request, versions__dandiset__pk, versions__version, **kwargs):
...
+            if 'path' not in metadata:
+                return Response('No path specified in metadata', status=404)
+            path = metadata['path']
+            # Strip away any computed fields
+            metadata = Asset.strip_metadata(metadata)
+            asset_metadata, created = AssetMetadata.objects.get_or_create(metadata=metadata)
+            if created:
+                asset_metadata.save()
+        else:
+            asset_metadata = old_asset.metadata
```
yarikoptic, line 221: I think we should just disallow PUT without metadata (in light of #392 and the fact that metadata might be completely wrong for the blob including the digests etc)

dandiapi/api/views/asset.py
```diff
@@ -194,20 +199,26 @@ def update(self, request, versions__dandiset__pk, versions__version, **kwargs):
...
+            asset_blob = get_object_or_404(AssetBlob, blob_id=serializer.validated_data['blob_id'])
+        else:
+            asset_blob = old_asset.blob
+
+        if 'metadata' in serializer.validated_data:
+            metadata = serializer.validated_data['metadata']
+            if 'path' not in metadata:
+                return Response('No path specified in metadata', status=404)
+            path = metadata['path']
+            # Strip away any computed fields
```
yarikoptic, line 215: before stripping - could we test that if dandi-etag is provided among digests, it matches the one of the `asset_blob`?

If not matches, it means that metadata is "wrong" in that it is not applicable to the blob/asset, and thus better not be allowed to be uploaded

yarikoptic (changes requested):
I think we better be "safe than sorry" and add additional guards right away to prevent inconsistent `blob+metadata == assset`s

dandiapi/api/views/asset.py
satra: PUT on an asset has three possibilities:
1. change path
2. change blob
3. change metadata (other than path or blob)

and in all cases unlink the previous asset and create a new asset. (everything could be seen as metadata update). so we are using PUT instead of DELETE + POST

so i would be hesitant to change what PUT does, unless we enforce those semantics consistently. i.e. 
- PUT does not change asset id and cannot be used to change blob
- replacement blob would require explicit DELETE + POST

dandiapi/api/views/asset.py
waxlamp: Since the billet of this PR (and #392, which it closes) is pretty narrow, just focusing on unblocking a specific problem of non-optional parameters, I'd like to defer this proposed functionality to a new issue. @yarikoptic, do you mind opening such an issue?

dandiapi/api/views/asset.py
waxlamp: If you open a new issue to address [the other thread](https://github.com/dandi/dandi-api/pull/382/files#r664671854) we can discuss this as part of that issue.

dandiapi/api/views/asset.py
waxlamp: It seems like this discussion should be part of that proposed new issue as well.

dchiquito:
@yarikoptic This PR was intended to unblock a feature in the CLI, but the CLI seems to have been working fine for the last few months. The comments here suggest pretty substantial changes to this PR. Since this PR is now so out of date, I'm closing this and moving discussion back to #382. Once we have requirements nailed down this PR can be resuscitated or rewritten, whatever's easier.

PR #387, opened 2021-07-01
Update staging design doc to be consistent with staging infrastructure

mvandenburgh:
We made a few minor changes to our staging plan after the staging design doc was written. This PR updates the design doc to be consistent with how we actually ended up setting up the staging environment.

### DIFF
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -46,6 +46,3 @@ A developer has a feature they want to add that only impacts the backend.
 * Once approved, the PR is merged to `master` and automatically deployed to `api-staging.dandiarchive.org`.
 * The changes are tested directly using `api-staging.dandiarchive.org`, or through the web GUI at `gui-staging.dandiarchive.org`.
 * Once everything is tested, tags a new release of `dandi-api`, which is then automatically deployed to production.
-
-## TODO
-Figure out a less manual release procedure for `dandi-api`.
--- a/doc/design/staging-1.md
+++ b/doc/design/staging-1.md
@@ -24,7 +24,7 @@ The staging server will track the `master` branch in `dandi-api` and will automa
 
 The production server will not track any branch. Instead, it will automatically redeploy whenever a release is tagged in GitHub.
 
-The staging web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
+The staging web GUI will track the `staging` branch in `dandiarchive` and automatically redeploy when the `staging` branch is changed. The `staging` branch will stay "synced" with `master` via a GitHub Action that will rebase `staging` on `master` whenever a change is pushed to `master`.
 The configuration variables for the staging Netlify deployment will point it at the staging server rather than the production server.
 To test code changes to the web GUI, branch previews will also be pointed at the staging server.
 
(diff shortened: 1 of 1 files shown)


doc/design/staging-1.md
```diff
@@ -48,4 +48,4 @@ A developer has a feature they want to add that only impacts the backend.
 * Once everything is tested, tags a new release of `dandi-api`, which is then automatically deployed to production.
 
 ## TODO
-Figure out a less manual release procedure for `dandiarchive`.
+Figure out a less manual release procedure for `dandi-api`.
```
waxlamp, line 51: We don't want a "less manual" procedure for `dandi-api`, do we?

mvandenburgh: I was correcting what I thought was a typo in the original design doc, since it is `dandi-api` that has a manual release procedure and not `dandiarchive`. I'm not sure if we still want a less manual procedure, @dchiquito is this statement outdated? FWIW I have no issues with the current release procedure.

doc/design/staging-1.md
```diff
@@ -24,7 +24,7 @@ The staging server will track the `master` branch in `dandi-api` and will automa
 
 The production server will not track any branch. Instead, it will automatically redeploy whenever a release is tagged in GitHub.
 
-The staging web GUI will track the `master` branch in `dandiarchive` and will automatically redeploy whenever `master` is changed.
+The staging web GUI will track the `staging` branch in `dandiarchive` and automatically redeploy when the `staging` branch is changed. The `staging` branch will stay "synced" with `master` via a GitHub Action that will rebase `staging` on `master` whenever a change is pushed to `master`.
```
yarikoptic, line 27: Why bother rebasing? Wouldn't it likely cause issues for PRs based on saying prior state of staging?

doc/design/staging-1.md
yarikoptic: By manual do you mean "mint a release"? That could be automated via PRs labeling if desired, like we do in dandi-cli and dandischema with auto

yarikoptic: But what I wonder - is it possible to unify and avoid staging branch for webui? IMHO such unique discrepancy (all other components by now rely on master + releases) is just prone to eventually put in production some intended for staging change

waxlamp: We are going to leave our staging/production setup as-is for now, until some problem with it arises in practice. (Though I like consistency in general, that's not reason enough for us to change how the frontend is deployed, due to some difficulties around how Netlify works.)

doc/design/staging-1.md
mvandenburgh: PRs should never be based off of the `staging` branch. In our current staging setup, the `staging` branch is basically a "read-only" mirror of `master` except it uses `api-staging.dandiarchive.org` as the server instead of `api.dandiarchive.org`.

waxlamp: The `staging` branch is always the same as `master`, with just the changes required to point the deployment at the staging API rather than the production API. Therefore, no PR should ever be targeted to `staging`, and rebasing this way is the appropriate way to set things up for the staging deploy.

doc/design/staging-1.md
dchiquito: If everyone is happy with how the release procedure currently works (I am), then I don't think there's anything left TODO and this comment can be deleted.

PR #386, opened 2021-07-01
Disable summary stats temporarily

waxlamp:
This is a workaround to speed up the bulk updating of asset metadata in dandisets with a large number of assets.

The way the code is currently written, each asset metadata update results in an O(N) computation across all the assets, making an update of all asset metadata an O(N^2) operation in total. This workaround drops that to O(N) at the cost of not updating the summary stats.

@AlmightyYakob and @mvandenburgh helped me review this as I wrote it, so I'm formally self-reviewing and merging to master so that it appears on staging; if it behaves well there, I will promote to production so that @satra is able to speed up his process.

attn: @dchiquito

Closes #384.

PR #389, opened 2021-07-03
Revert "Merge pull request #386 from dandi/workaround-no-summary"

yarikoptic:
This reverts commit 93faf5ff2942e800bf9093b4a709c7c566ad3b1d, reversing
changes made to da32e0e6c550100a6dd8529aa68484957a710e42.

AFAIK we (@satra and me) know -- all assets metadata were converted and we should resume normal operations.  I do not expect anything to break, so I might even proceed and merge/mint a release (to get it deployed) soonish unless someone stops me

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -274,9 +274,7 @@ class Version(TimeStampedModel):
             'numberOfBytes': 0,
             'numberOfFiles': 0,
         }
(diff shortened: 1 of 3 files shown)


yarikoptic:
@dchiquito do you see an easy quick way to trigger that offband generation of summaries for all dandisets?
may be we should expose some `/updateInternal` endpoint for each dandiset to facilitate such/similar updates in the future?

PR #395, opened 2021-07-06
Fix scripts

dchiquito:
* Move a script that was added after Brian's PR
* Don't refresh metadata for published versions

### DIFF
--- a/dandiapi/api/management/commands/refresh_metadata.py
+++ b/dandiapi/api/management/commands/refresh_metadata.py
@@ -15,5 +15,6 @@ def refresh_metadata():
     for asset in Asset.objects.all():
         asset.save()
 
-    for version in Version.objects.all():
+    # Only update the metadata for draft versions
+    for version in Version.objects.filter(version='draft'):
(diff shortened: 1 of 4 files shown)


dandiapi/api/management/commands/refresh_metadata.py
```diff
@@ -15,5 +15,6 @@ def refresh_metadata():
     for asset in Asset.objects.all():
         asset.save()
 
-    for version in Version.objects.all():
+    # Only update the metadata for draft versions
+    for version in Version.objects.filter(version='draft'):
```
yarikoptic, line 19: if that is the helper intended to be used to trigger update to `assetsSummary`, I do  not see a reason to go through all assets (and we do not even have GC yet, so it would be really all assets, right?) and mint new thousands of them if I get it right what this code does.

dchiquito: It would not mint new assets, but you are right about doing more work than necessary. I added some arguments.

PR #402, opened 2021-07-09
BF: Do not install in devel mode in requirements.txt

yarikoptic:
This might at least partially address a problem of incorrect version
being provided by versioneer (https://github.com/dandi/dandi-api/issues/222).
Because staging and deployment have different ways of deployment it is not
even necessary that this working on one would work on another.

More FTR: https://github.com/heroku/heroku-buildpack-python/issues/1222

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 # This file is only for use by Heroku, with the old pip resolver
 idna>=2.5,<3
--e .
+.


yarikoptic:
circle ci fail seems to be unrelated development: https://github.com/dandi/dandi-api/issues/403 . dandi-cli integration fails -- the known (but not yet attended to) https://github.com/dandi/dandi-api/issues/393

waxlamp:
> Because staging and deployment have different ways of deployment it is not
> even necessary that this working on one would work on another.

Not sure what you mean by this--could you clarify?

yarikoptic:
> > Because staging and deployment have different ways of deployment it is not
> > even necessary that this working on one would work on another.
> 
> Not sure what you mean by this--could you clarify?

IIRC staging relies on heroku's way to fetch the branch, whenever deployment relies on https://github.com/dandi/dandi-api/blob/master/.github/workflows/production-deploy.yml which is triggered only on releases.  IMHO should be harmonized to always be done by that github action but that is a separate issue I guess.

yarikoptic:
now it is with a magical number #404 ;)

waxlamp (approved):
Let's merge this and see what happens on the staging deployment. If it improves things for versioneer, then great! If it doesn't hurt anything, then perhaps we can simply leave it, following the advice from the linked Heroku issue thread.

PR #401, opened 2021-07-08
Add contact person to dandiset serializer

mvandenburgh:
Adds `contact_person` field to `DandisetSerializer` and `VersionDetailSerializer`.

Required for dandi/dandiarchive#783.

Closes #403.

PR #405, opened 2021-07-09
Add `email-validator` as an install-required package

waxlamp:
See this [Heroku build failure](https://dashboard.heroku.com/apps/dandi-api-staging/activity/builds/e0bbf36c-0a34-47f4-b099-94fc580b49fd) for details.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 # This file is only for use by Heroku, with the old pip resolver
 idna>=2.5,<3
+pydantic[email]
 .
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
 # This file is only for use by Heroku, with the old pip resolver
 idna>=2.5,<3
+# To address an odd case of email-validator not being installed
+# see https://github.com/dandi/dandi-api/pull/405
 pydantic[email]
 .
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,6 @@ setup(
         'djangorestframework-yaml',
         'drf-extensions',
         'drf-yasg',
-        'email-validator',
         'httpx',
         'jsonschema',
         # Production-only
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,7 @@ setup(
         'djangorestframework-yaml',
         'drf-extensions',
         'drf-yasg',
+        'email-validator',
         'httpx',
         'jsonschema',
         # Production-only


dchiquito (changes requested):
The build error recommends `pip install pydantic[email]` to fix it. Have you verified that this installs locally?

Also, this dependency is being pulled in from dandischema, so I think that's where we should be including this dependency. I think pinning `dandischema==0.2.9` might resolve it, as presumably this is an issue in `0.2.10`.

waxlamp:
> The build error recommends `pip install pydantic[email]` to fix it. Have you verified that this installs locally?

That looks like it works.

I was pulling in `email-validator` directly based on a random pydantic docs page I saw, but this is better.

> 
> Also, this dependency is being pulled in from dandischema, so I think that's where we should be including this dependency. I think pinning `dandischema==0.2.9` might resolve it, as presumably this is an issue in `0.2.10`.

I will shift this PR to do that instead.

Are you saying that a PR in dandischema to pull in `pydantic[email]` is the correct long-term fix?

mvandenburgh:
fyi CI is failing for this PR because I didn't update the tests properly in #401, I fixed them in #406.

dchiquito:
Yes, my suspicion is that dandischema should have a dependency on `pydantic[email]`, but doesn't. I think pinning to an earlier dandischema version would be a better solution for us for now, assuming that solves the problem

satra:
we should not pin to an earlier version. instead let's resolve the issue. also would be good to address why we are seeing this here and what is not being tested in dandischema. i see emails in the dandischema tests.

satra:
also this is clearly a dependency in dandischema: 

https://github.com/dandi/dandischema/blob/afd6055d7d8560b3969da41bacdd168117bee061/setup.cfg#L31

yarikoptic:
I wonder if 
```
       ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.
       
       We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
```
is anyhow related

yarikoptic:
FWIW pydantic v1.8.2 (as in git) does have
```
    extras_require={
        'email': ['email-validator>=1.0.3'],
        'dotenv': ['python-dotenv>=0.10.4'],
    },
```
so does declare `email` extra requires.  So it feels that it is indeed `pip` to blame after all

yarikoptic:
> Yes, my suspicion is that dandischema should have a dependency on `pydantic[email]`, but doesn't. I think pinning to an earlier dandischema version would be a better solution for us for now, assuming that solves the problem

I do not see why that would solve it . Most likely trigger for the failure is my https://github.com/dandi/dandi-api/pull/402 to switch away from `devel` mode of installation, and `pydantic[email]` dependency was there since dandischema inception.

waxlamp:
> > Yes, my suspicion is that dandischema should have a dependency on `pydantic[email]`, but doesn't. I think pinning to an earlier dandischema version would be a better solution for us for now, assuming that solves the problem
> 
> I do not see why that would solve it . Most likely trigger for the failure is my #402 to switch away from `devel` mode of installation, and `pydantic[email]` dependency was there since dandischema inception.

I do see that `dandischema` has `pydantic[email]` as an install dependency, but `email-validator` is not picked up for installation in the failing build log. I am not sure how to reconcile what is going on, but I agree that the change in the requirements file must be to blame.

yarikoptic:
I would've still proceeded with this PR (just an added comment why) and if that would be there only issue - dig deeper after

yarikoptic:
Just may be not in setup.py but requirements.txt which already has heroku specific listing

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # This file is only for use by Heroku, with the old pip resolver
 idna>=2.5,<3
+pydantic[email]
```
yarikoptic, line 3:
```suggestion
# To address an odd case of email-validator not being installed
# see https://github.com/dandi/dandi-api/pull/405
pydantic[email]
```

yarikoptic:
I guess the circle ci greenness is not depictive of success and we wouldn't know until we give it away to heroku... so let's do it!

yarikoptic:
"live news" from a log watcher:
```
       Collecting email-validator>=1.0.3; extra == "email"
         Downloading email_validator-1.1.3-py2.py3-none-any.whl (18 kB)
```

yarikoptic:
```
-----> Launching...
 !     Release command declared: this new release will not be available until the command succeeds.
       Released v76
       https://dandi-api-staging.herokuapp.com/ deployed to Heroku
```
and it is in now... unfortunately no magical remedy for the versioneer-inflicted version reporting: ` "version": "0+unknown"`

waxlamp:
Too bad--@yarikoptic, do you mind reopening your Heroku issue and see if they will help us figure out how to work with versioneer?

yarikoptic:
I don't mind at all, but feel that we might better unify deployment workflow first (https://github.com/dandi/dandi-api/issues/404) so that we don't need to chase two ghosts instead of one

yarikoptic:
And main deployment one would be ticket to chase down since we wouldn't have luxury of "can break" ;-)

PR #417, opened 2021-07-15
Improve validation error UX

mvandenburgh:
This modifies how version and asset validation errors are stored and retrieved in the database so the frontend can enumerate/display them. 

Instead of having a single string that contains all validation errors, the `Version` model now has separate fields for version metadata validation errors and asset metadata validation errors. In addition, the validation errors are now stored in arrays (via a `JSONField`) so they can be enumerated in the web UI. 

Required for https://github.com/dandi/dandiarchive/pull/796.


dandiapi/api/models/version.py
```diff
@@ -114,27 +114,25 @@ def publish_status(self) -> Version.Status:
...
+    def asset_validation_errors(self) -> list[str]:
         # Import here to avoid dependency cycle
         from .asset import Asset
 
         # Assets that are not VALID (could be pending, validating, or invalid)
-        invalid_assets = self.assets.filter(status=Asset.Status.INVALID).count()
-        if invalid_assets > 0:
-            return f'{invalid_assets} invalid asset metadatas'
+        invalid_assets = self.assets.filter(status=Asset.Status.INVALID)
+        errors = [invalid_asset.validation_error for invalid_asset in invalid_assets]
```
dchiquito, line 123:
```suggestion
        invalid_assets = self.assets.filter(status=Asset.Status.INVALID).values('validation_error')
        errors = [invalid_asset['validation_error'] for invalid_asset in invalid_assets]
```
This way the query only returns the `validation_error` fields. Iterating over every asset implicitly queries for everything else, including metadata.
The same optimization can be done on `unvalidated_assets` below.

dchiquito (changes requested):

dchiquito:
You will also need to update the tests, since you changed the shape of the response from the rest endpoint.

dandiapi/api/models/version.py
mvandenburgh: https://github.com/dandi/dandi-api/pull/417/commits/a1e1435dbe48a781c293b8dc917c8bc3bd0ff146

dchiquito (approved):

PR #418, opened 2021-07-16
Scheduled tasks design doc

dchiquito:
@satra I think a scheduled recurring job would be easier to work with than cancelling jobs whenever a new modification occurs.

### DIFF
--- a/doc/design/scheduled-tasks.md
+++ b/doc/design/scheduled-tasks.md
@@ -1,0 +1,55 @@
+# Asynchronous Tasks
+We are currently (7/16/2021) using celery to run these asynchronous tasks:
+
+ * Calculating blob sha256 digests
+ * Writing dandiset.yaml and assets.yaml after publishing
+ * Validating asset metadata
+ * Validating dandiset metadata
+
+These all work well at the usage levels that we are currently seeing, but may not scale well into the future.
+Specifically, every time an asset is uploaded, asset metadata and dandiset metadata are re-validated.
+When uploading large batches of assets, dandiset metadata is validated repeatedly and unnecessarily.
+
+## Motivation
+Currently (7/16/2021), every time a Version is saved, it recalculates the value of the `assetsSummary` metadata field.
+This requires performing a `SELECT *` on all Assets in the Version, which does not scale well for Versions with lots of Assets.
+Because uploading a new Asset to a Version involves saving that Version, uploading lots of new assets will eventually start causing request timeouts as the number of Assets increases.
+
+We would like to move this `assetsSummary` calculation into a celery task, but we also don't want to trigger a separate task for every new Asset, as that would only delay the load on the DB.
+We need a way to run this task only once per batch of uploads.
+
+## Proposal
+Instead of adding a new task to the queue whenever something needs to be done, we use a scheduled job that runs the task only when necessary.
+
+Since `assetsSummary` calculation and dandiset metadata validation both need to be triggered in the same circumstances, we'll consider them together.
+
+#### Request thread
+Whenever a Version is modified, a flag is set on the Version indicating that it has been modified (instead of queueing a task as it does now):
+```
+  # Things that modify the version, necessitating a validation/assetsSummary computation
+  ...
+  # validate_version_metadata.delay(version.id)
+  version.status = status.MODIFIED
+```
+
+#### Scheduled job
+Separately, a scheduled job that runs on a configurable fairly tight interval will check for any Versions that need to be updated.
+For each of those Versions, it also checks that the Version hasn't been modified in a certain amount of time. This prevents DB locks from happening while Assets are still being uploaded to the Version.
+
+The code would look something like this:
+```
+def scheduled_task():
+  versions = Version.objects.filter(status=status.MODIFIED).filter(modified__lt=timezone.now() - 5 minutes)
+  # SELECT ... FROM api_version WHERE api_version.status = "MODIFIED" AND now() - api_version.modified > 5 minutes;
+```
+
+For each of those Versions, the job would then calculate `assetsSummary`, inject it into the metadata, and perform the validation, marking the Version accordingly.
+
+## Changes to existing tasks
+We shouldn't need to change how calculating blob sha256 digests or how writing dandiset.yaml and assets.yaml work at all.
+They don't block the DB in any meaningful way, and so can continue working as they always have.
+
+Validating asset metadata could possibly be run with a job, but I would recommend not switching it yet.
+Uploading or updating Assets doesn't happen with the same frequency as with Versions, so it doesn't benefit from debouncing in the same way.
+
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
The technical aspects seem sound to me. There are tradeoffs with UX here that ought to be detailed; in particular, I suspect the frontend will need to report on the fact that certain updates haven't been done yet, etc. If information about the timing of such updates is available, it can be displayed in tooltips over the relevant elements in the UI.

doc/design/scheduled-tasks.md
```diff
@@ -0,0 +1,55 @@
+# Asynchronous Tasks
+We are currently (7/16/2021) using celery to run these asynchronous tasks:
+
+ * Calculating blob sha256 digests
+ * Writing dandiset.yaml and assets.yaml after publishing
+ * Validating asset metadata
+ * Validating dandiset metadata
```
satra, line 7: these processes should be more or less instantaneous. perhaps we do want to split out "assets Summary generation" as a non instantaneous task?

satra:
@dchiquito -  as long as the modified query across all dandisets is quick enough, which should be in the short term, this looks good to me.

doc/design/scheduled-tasks.md
dchiquito: I'm not following your suggestion. Tasks can be long-running, but there's no problem with having an instantaneous task. Why would we want to split `assetsSummary` generation into a separate task, and how would that change the implementation?

satra: i wasn't sure why these are celery tasks. they seem to be completable within the duration of a REST call and be done without too much load. perhaps it's the semantics again, but i'm seeing these as calls to dandischema.validate, which does not need to be asyncrhonous.

dchiquito: I see. I think that refactoring can be done later, outside of this design doc.

doc/design/scheduled-tasks.md
```diff
@@ -0,0 +1,55 @@
...
+#### Scheduled job
+Separately, a scheduled job that runs on a configurable fairly tight interval will check for any Versions that need to be updated.
+For each of those Versions, it also checks that the Version hasn't been modified in a certain amount of time. This prevents DB locks from happening while Assets are still being uploaded to the Version.
+
+The code would look something like this:
+```
+def scheduled_task():
+  versions = Version.objects.filter(status=status.MODIFIED).filter(modified__lt=timezone.now() - 5 minutes)
+  # SELECT ... FROM api_version WHERE api_version.status = "MODIFIED" AND now() - api_version.modified > 5 minutes;
+```
```
yarikoptic, line 44: how race condition around `version.status == status.MODIFIED` be resolved? i.e. when `version.status` would be set to not MODIFIED and couldn't interim another asset upload set it to `status.MODIFIED` but otherwise not be accounted in e.g. assetsSummary?
I guess it could be done by memoizing that "timezone.now() - 5 minutes" time point, and then not resetting `version.status` if its `modified` is later than that point (thus possibly something modified it interim)?  not sure if there wouldn't be some side-effects which would also cause `modified` to be   progressed.  May be some dedicated `.status_modified` could be added to disambiguate?

doc/design/scheduled-tasks.md
```diff
@@ -0,0 +1,55 @@
...
+## Motivation
+Currently (7/16/2021), every time a Version is saved, it recalculates the value of the `assetsSummary` metadata field.
+This requires performing a `SELECT *` on all Assets in the Version, which does not scale well for Versions with lots of Assets.
+Because uploading a new Asset to a Version involves saving that Version, uploading lots of new assets will eventually start causing request timeouts as the number of Assets increases.
+
+We would like to move this `assetsSummary` calculation into a celery task, but we also don't want to trigger a separate task for every new Asset, as that would only delay the load on the DB.
+We need a way to run this task only once per batch of uploads.
+
+## Proposal
+Instead of adding a new task to the queue whenever something needs to be done, we use a scheduled job that runs the task only when necessary.
```
yarikoptic, line 22: for the main effect, it would address https://github.com/dandi/dandi-api/issues/385

doc/design/scheduled-tasks.md
```diff
@@ -0,0 +1,55 @@
...
+
+Since `assetsSummary` calculation and dandiset metadata validation both need to be triggered in the same circumstances, we'll consider them together.
+
+#### Request thread
+Whenever a Version is modified, a flag is set on the Version indicating that it has been modified (instead of queueing a task as it does now):
+```
+  # Things that modify the version, necessitating a validation/assetsSummary computation
+  ...
+  # validate_version_metadata.delay(version.id)
+  version.status = status.MODIFIED
```
yarikoptic, line 32: there could be many notions of "modified" (well, there is `modified` time stamp which would flip on every change) and necessary actions to be taken.  I wonder if might better become more explicit, e.g.
```
version.assetsSummaryStatus = status.OUTDATED
```
or alike, which would make it cleared that scheduled "updateAssetsSummary" job would be the one to deal with it.

doc/design/scheduled-tasks.md
dchiquito: Good point. One way would be to make the modification tracking a boolean value `is_modified` instead of reusing the existing `status` enum.
Whenever a change is made to the version, `is_modified` is set to `True`.
At the start of the scheduled job, `is_modified` is immediately set to `False`. Any other modifications made while the job is still running may or may not be expressed in the final results, but that shouldn't matter, since the next running of the job will notice that `is_modified == True` and recompute.

doc/design/scheduled-tasks.md
```diff
@@ -24,26 +24,30 @@ Instead of adding a new task to the queue whenever something needs to be done, w
...
 ```
 def scheduled_task():
-  versions = Version.objects.filter(status=status.MODIFIED).filter(modified__lt=timezone.now() - 5 minutes)
-  # SELECT ... FROM api_version WHERE api_version.status = "MODIFIED" AND now() - api_version.modified > 5 minutes;
+  versions = Version.objects.filter(needs_recomputing=True).filter(modified__lt=timezone.now() - 5 minutes)
+  # SELECT ... FROM api_version WHERE api_version.needs_recomputing AND now() - api_version.modified > 5 minutes;
 ```
 
-For each of those Versions, the job would then calculate `assetsSummary`, inject it into the metadata, and perform the validation, marking the Version accordingly.
+The job would then immediately set `version.needs_recomputing = False` on all of the "dirty" Versions.
```
yarikoptic, line 46: what if the job fails to perform the "recomputing"?
what if `publish` query would be ran meanwhile causing publication before "recomputing" task finishes?

feels like a boolean flag wouldn't be sufficient, and there might be needed some "pending" state:
- set to `False` only on being "pending" -- would need to check if was not set to `True` while "computing"
- but then another attempt at `recomputing` should be forbidden if already  "pending"... so there is no two tasks "competing". not sure what behavior should be then.

overall -- I feel needs more thinking or we would end up with some assetSummaries not reflective of the actual state.

PR #423, opened 2021-07-20
Describe needs_recomputing flag more completely

dchiquito:
https://github.com/dandi/dandi-api/pull/418/files/880598145f44cf3568d9cb6e3f57957f444ca0cc..c09059f9379c805032d719962b09a30ff7df69dc#r672511632
>what if the job fails to perform the "recomputing"?
what if publish query would be ran meanwhile causing publication before "recomputing" task finishes?
>
>feels like a boolean flag wouldn't be sufficient, and there might be needed some "pending" state:
>
>set to False only on being "pending" -- would need to check if was not set to True while "computing"
but then another attempt at recomputing should be forbidden if already "pending"... so there is no two tasks "competing". not sure what behavior should be then.
overall -- I feel needs more thinking or we would end up with some assetSummaries not reflective of the actual state.

My thinking is that a boolean is all that is required for `needs_recomputing`, so I clarified my thoughts in the document.

### DIFF
--- a/doc/design/scheduled-tasks.md
+++ b/doc/design/scheduled-tasks.md
@@ -46,8 +46,15 @@ def scheduled_task():
 The job would then immediately set `version.needs_recomputing = False` on all of the "dirty" Versions.
 If any simultaneous requests happen to modify the Version while it is being recomputed/validated, those changes may or may not happen in time to be represented.
 However, they are guaranteed to trigger another recomputation in 5 minutes, so the system will eventually regain consistency.
+If the job takes too long to complete and the next job kicks off while the first is still running, this will also prevent any versions from being recomputed in different jobs.
 
 Finally, the actual task logic would be applied to each version: calculate `assetsSummary`, inject it into the metadata, and perform the validation, marking the Version accordingly.
+
+Recomputation should fail very infrequently, either because of SQL errors or because of introduced bugs in our code.
+In the event of an error, `needs_recomputing` should be set back to `True` so that it is retried, hopefully resolving intermittent errors.
+The error should then be thrown so that it shows up in Sentry.
+
+We shouldn't have to worry about publishes happening during recomputation. Any changes that would set `needs_recomputing = True` would also mark the Version's validation status as `PENDING`, and validation cannot happen until after recomputation is complete. Since publish is not allowed on non-`VALID` Versions, we don't have to worry about simultaneous publishes.
 
 ## Changes to existing tasks
 We shouldn't need to change how calculating blob sha256 digests or how writing dandiset.yaml and assets.yaml work at all.


doc/design/scheduled-tasks.md
```diff
@@ -46,9 +46,16 @@ def scheduled_task():
...
 However, they are guaranteed to trigger another recomputation in 5 minutes, so the system will eventually regain consistency.
+If the job takes too long to complete and the next job kicks off while the first is still running, this will also prevent any versions from being recomputed in different jobs.
 
 Finally, the actual task logic would be applied to each version: calculate `assetsSummary`, inject it into the metadata, and perform the validation, marking the Version accordingly.
 
+Recomputation should fail very infrequently, either because of SQL errors or because of introduced bugs in our code.
+In the event of an error, `needs_recomputing` should be set back to `True` so that it is retried, hopefully resolving intermittent errors.
+The error should then be thrown so that it shows up in Sentry.
+
+We shouldn't have to worry about publishes happening during recomputation. Any changes that would set `needs_recomputing = True` would also mark the Version's validation status as `PENDING`, and validation cannot happen until after recomputation is complete. Since publish is not allowed on non-`VALID` Versions, we don't have to worry about simultaneous publishes.
```
yarikoptic, line 57: >  and validation cannot happen until after recomputation is complete.

how would you know that recomputation is complete if "The job would then immediately set `version.needs_recomputing = False` "?

doc/design/scheduled-tasks.md
```diff
@@ -46,9 +46,16 @@ def scheduled_task():
 The job would then immediately set `version.needs_recomputing = False` on all of the "dirty" Versions.
 If any simultaneous requests happen to modify the Version while it is being recomputed/validated, those changes may or may not happen in time to be represented.
 However, they are guaranteed to trigger another recomputation in 5 minutes, so the system will eventually regain consistency.
+If the job takes too long to complete and the next job kicks off while the first is still running, this will also prevent any versions from being recomputed in different jobs.
```
yarikoptic, line 49: sorry, I am not sure what is `this` which "will also prevent any versions from being recomputed in different jobs", could you please elaborate? e.g. if we started computation at 5 min mark, it takes 7 min (very unlikely, but let's think hypothetically), what would prevent the next job at 10 min mark?

doc/design/scheduled-tasks.md
dchiquito: The validation function would be run in the big scheduled job, immediately after `assetsSummary` is computed. I added some pseudocode to better describe the order of events.

doc/design/scheduled-tasks.md
```diff
@@ -34,21 +34,46 @@ Whenever a Version is modified, a flag `needs_recomputing` is set on the Version
...
 #### Scheduled job
 Separately, a scheduled job that runs on a configurable fairly tight interval will check for any Versions that need to be updated.
-For each of those Versions, it also checks that the Version hasn't been modified in a certain amount of time (say 5 minutes). This prevents DB locks from happening while Assets are still being uploaded to the Version.
 
-The code would look something like this:
+The full code would look something like this:
 ```
 def scheduled_task():
   versions = Version.objects.filter(needs_recomputing=True).filter(modified__lt=timezone.now() - 5 minutes)
   # SELECT ... FROM api_version WHERE api_version.needs_recomputing AND now() - api_version.modified > 5 minutes;
```
yarikoptic, line 42: re `5 minutes` -- made a dedicated issue #431

doc/design/scheduled-tasks.md
yarikoptic: gotcha, got `this` -- thanks!

doc/design/scheduled-tasks.md
yarikoptic: thanks!

doc/design/scheduled-tasks.md
```diff
@@ -34,21 +34,46 @@ Whenever a Version is modified, a flag `needs_recomputing` is set on the Version
...
+    version.needs_recomputing = False
+    version.save()
+
+  for version in versions:
+    try:
+      # Calculate assetsSummary
+      version.metadata['assetsSummary'] = calculate_assetsSummary(version)
+
+      # Validate the version
+      validate_version(version)
```
yarikoptic, line 55: since we are not locking DB (`version`) from modification for the duration of this task, above code might be racy with possible users' modifications. Some race could be avoided if this `validate_version` would not set validation status `if version.needs_recomputing` -- i.e. that something was updated/uploaded before it completed.  Cons: leaks "needs_recomputing" into "validate_version" which should in principle not care about all this orchestration, so may be it should just return the value either it is valid or not and logic should stay here to assign (like for `assetsSummary`)

doc/design/scheduled-tasks.md
```diff
@@ -34,21 +34,46 @@ Whenever a Version is modified, a flag `needs_recomputing` is set on the Version
...
+    except:
+      # Something went wrong, mark the version as dirty again so it can be retried next task
+      version.needs_recomputing = True
+      version.save()
```
yarikoptic, line 59: will `validate_version` call `.save`?  see above to may be centralize all that logic here, and have `validate_version` to just return the result of validation

doc/design/scheduled-tasks.md
dchiquito: If those timeouts are configurable with environment variables, they could both be set to 1 second in the integration testing environment so the test only has to wait ~2 seconds for the job to run. You also get to test the actual functionality of the system instead of using a shortcut.

doc/design/scheduled-tasks.md
dchiquito: Good point, I changed the design to lock the DB for the duration of the task. In practice multiple Versions will very rarely be picked up at once, since a single user is generally only with a single version at a time.
If we ever run into excessive DB locks in the future, we can scale this behavior back.

doc/design/scheduled-tasks.md
yarikoptic: good to hear that we are on the same page .  might be worth adding in design doc defining some env var to control for time outs.

doc/design/scheduled-tasks.md
```diff
@@ -19,41 +19,66 @@ We would like to move this `assetsSummary` calculation into a celery task, but w
...
 Separately, a scheduled job that runs on a configurable fairly tight interval will check for any Versions that need to be updated.
-For each of those Versions, it also checks that the Version hasn't been modified in a certain amount of time (say 5 minutes). This prevents DB locks from happening while Assets are still being uploaded to the Version.
 
-The code would look something like this:
+The full code would look something like this:
 ```
 def scheduled_task():
-  versions = Version.objects.filter(needs_recomputing=True).filter(modified__lt=timezone.now() - 5 minutes)
-  # SELECT ... FROM api_version WHERE api_version.needs_recomputing AND now() - api_version.modified > 5 minutes;
+  with transaction.atomic():
```
yarikoptic, line 39: sorry , but can't grasp from the [doc](https://docs.djangoproject.com/en/3.2/topics/db/transactions/#controlling-transactions-explicitly) -- is `.atomic` blocking the entire DB from any writes or just not performing `DB writes` of changes until it is done?

if the former (total DB block) -- wouldn't we screw up interaction from other users/tasks with the DB?

if the later (no writes until out) -- wouldn't `version.needs_recomputing = False` and `version.status = PENDING` have no effect for other "tasks" which would check on them?

dchiquito: `atomic` is just a regular DB transaction.
`needs_recomputing` will remain `True` until the task commits the transaction successfully, so the Version cannot be published. This is the only place where `needs_recomputing` is set to `False`, so the Version is considered invalid for publishing until the task succeeds.

If the transaction were to fail, the next scheduled job would pick it up as if the first job had never run. Eventually the transaction should succeed.

PR #424, opened 2021-07-20
Lower staging celery concurrency to 2

dchiquito:


dandiapi/settings.py
```diff
@@ -93,3 +93,7 @@ class HerokuProductionConfiguration(DandiMixin, HerokuProductionBaseConfiguratio
 # the API server is running in (production/local or staging).
 class HerokuStagingConfiguration(HerokuProductionConfiguration):
     OAUTH2_PROVIDER_APPLICATION_MODEL = 'api.StagingApplication'
+
+    # We are using cheaper Heroku dynos for staging, so we need to artificially lower the number
+    # of concurrent tasks (default is 8) to keep memory usage down.
+    CELERY_WORKER_CONCURRENCY = 2
```
waxlamp, line 99: What's the total amount of memory available in staging and production?

dchiquito: Staging has a nominal capacity of 512MB, production has 1GB. In practice, it seems like you just get warnings when you reach the memory cap, then at 2x the cap your dyno is shut down immediately.

PR #435, opened 2021-07-21
Add migration to turn on skip_authorization

waxlamp:


brianhelba (approved):

dchiquito:
Is there a reason this is a migration instead of a manual administrative action? IIRC this is just a checkbox on the application.

dchiquito:
If anything I think migration 0006 should be modified to set the default value to True. This migration will just set everyone's `skip_authorization` to true regardless of environment/personal preference.

waxlamp:
> If anything I think migration 0006 should be modified to set the default value to True. This migration will just set everyone's `skip_authorization` to true regardless of environment/personal preference.

I discussed this with Brian while we were looking at ways to solve this.

I went with this migration rather than editing 0006 so that new deployments would be aligned with the current one. I actually have a different branch open that modified 0006 but this actually seemed more natural for what I want to express, which is that by default the skip authorization flag really should be set. Brian and I also realized that this doesn't have much practical effect at all since new deployments would have it turned off with both methods, and I chose this method largely by way of learning how to put a RunPython migration together.

To answer your first question, I had a slight preference for this approach because it specifically doesn't involve a manual action to check a checkbox.

If there are any major pitfalls with this approach, that would be cause to close this and operate on the other approach instead. But if not, I'd like to just get this one in, apply it, and move on. Let me know what you think, @dchiquito.

dchiquito:
@waxlamp It sounds to me like you want to turn on `skip_authorization` in production+staging. This migration will turn it on for every instance everywhere. If that's your intention, then :+1: 
Modifying 0006 and toggling prod+staging manually would make it so that new environments are all the same as prod+staging, without modifying any user configurations.

PR #440, opened 2021-07-22
A better migration for controlling `skip_authorization`

waxlamp:
This is a better approach to the broken migration I backed out via #439.

Thanks @dchiquito and @AlmightyYakob!

attn: @brianhelba


dandiapi/api/migrations/0006_default_oauth_application.py
```diff
@@ -21,7 +21,7 @@ def create_application(apps, schema_editor):
             client_secret='',
             name=name,
             # This can be turned off in production if appropriate.
-            skip_authorization=False,
+            skip_authorization=True,
```
AlmightyYakob, line 24: Should this still be modified here? I thought the point of modifying the `StagingApplication` class was to avoid this being changed here directly

AlmightyYakob: Nevermind, this was explained to me. This is the normal (non-staging) application, whereas the other migration is for the staging application.

dchiquito (approved):

PR #443, opened 2021-07-22
Do not allow deletion of dandisets with published versions

### LINKED ISSUE #288: Do not delete published dandisets
Currently any owner can delete a dandiset, even if it is published. We should only allow dandisets with published versions to be deleted if the user is an admin.

dchiquito:
We do not want normal users to be able to delete dandisets that have been published at least once, but we still want admins to have the power if necessary.

* Update old delete endpoint tests to test dandisets with draft versions.
* Add new tests for deleting dandisets with published verisons.

Fixes #288

### DIFF
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -507,25 +507,24 @@ def test_dandiset_rest_delete_not_an_owner(api_client, draft_version, user):
 
 
 @pytest.mark.django_db
-def test_dandiset_rest_delete_published(api_client, published_version, admin_user):
-    api_client.force_authenticate(user=admin_user)
-
-    response = api_client.delete(f'/api/dandisets/{published_version.dandiset.identifier}/')
-    assert response.status_code == 204
-
-    assert not Dandiset.objects.all()
-
-
-@pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


satra:
i would say we also don't allow admins to do so. so that i don't accidentally do a DELETE on a published dandiset with my API KEY. let's make this operation require some significant work and no programmatic route.

satra:
question: can we guarantee that a published asset cannot be deleted? if not we should ensure that no one can delete a published asset or blob.

dchiquito:
@satra `DELETE`ing an asset simply unlinks it from the version, the underlying Asset/AssetBlob is basically permanent.

satra:
> DELETEing an asset simply unlinks it from the version, the underlying Asset/AssetBlob is basically permanent.

but we should not be able to do so for a published dandiset. so `DELETE` should only apply when `version==draft`. is that currently true?

waxlamp (approved):
Looks good to me.

If you merge this as is, please follow up with a PR addressing Satra's question about DELETEing assets from published versions.

dchiquito:
@satra sorry, yes it is not permitted to delete assets from non-draft versions:
```
        if version.version != 'draft':
            return Response(
                'Only draft versions can be modified.',
                status=status.HTTP_405_METHOD_NOT_ALLOWED,
            )
```

PR #449, opened 2021-07-29
make dois published

satra:
https://support.datacite.org/docs/api-create-dois#create-a-findable-doi

this needs to be augmented such that we don't do this in tests but only for actual dandisets.

### DIFF
--- a/dandiapi/api/doi.py
+++ b/dandiapi/api/doi.py
@@ -26,6 +26,7 @@ def _generate_doi_data(version: Version):
 
 def create_doi(version: Version) -> str:
     doi, request_body = _generate_doi_data(version)
+    request_body["data"]["attributes"]["event"] = "publish"
     # If DOI isn't configured, skip the API call
     if (
         settings.DANDI_DOI_API_URL is not None


satra:
@dchiquito - the current dois from staging are set in draft mode, we should make them findable.  however, we don't really want this to happen for every doi. since a doi in registered/findable mode cannot be deleted. so this should only happen when we publish from the production server. so ideally some environment flag that says that code is being executed in a production environment.

yarikoptic:
AFAIK (see https://github.com/dandi/dandi-api/issues/181#issuecomment-877485448) new ones should be Findable (since we had some recent ones as Findable), but I could be wrong

yarikoptic:
or in other words: how did we end up with Finable DOIs in test. datacite without this change?

dchiquito:
I created a few findable ones manually when I was experimenting with the API before I fully understood the distinction.

IIRC dandischema needs to include `"event": "publish"` in the DOI data returned from to_datacite: https://support.datacite.org/docs/api-create-dois#create-a-findable-doi Probably an argument `publish = False` would include it if True and omit it if False.

Once that's there I can add a setting and call to_datacite with the correct arguments.

dchiquito:
I did the dandischema work in https://github.com/dandi/dandischema/pull/70

dchiquito:
I'm going to close this PR and move the discussion to https://github.com/dandi/dandischema/pull/70

satra:
the dandischema PR merged and released. but it would still require a change in the API server to actually publish only on the production server.

dchiquito:
https://github.com/dandi/dandi-api/pull/451 is that change. https://github.com/dandi/dandi-infrastructure/pull/87 turns that setting on in production only.

PR #450, opened 2021-07-29
Add top-level API endpoint for asset metadata

### LINKED ISSUE #436: extend top level /assets/ end-point to mirror dandisets_version ones
trigger for the request: https://github.com/dandi/dandi-cli/issues/723#issuecomment-884920772 i.e. inability to get asset metadata (to e.g. decide either it was modified or might be the same IIRC) to make `dandi download` handle top level `/assets/<asset_id>/download/` end points.

I think there should be
- `GET /assets/{asset_id}/` -- get metadata (of immediate need)
- `GET /assets/{asset_id}/validation/` -- forgot what it does exactly (triggers validation? returns validation status?) and swagger doesn't tell but I guess there is no reason to not have it at top level as well

waxlamp:
This adds a `GET /assets/{asset_id}/metadata` endpoint to the API, allowing for querying asset metadata without specifying a dandiset ID and version.

Closes #436.

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -1,3 +1,4 @@
+import json
 import os.path
 from uuid import uuid4
 
@@ -205,6 +206,11 @@ def test_asset_rest_retrieve(api_client, version, asset):
             f'/api/dandisets/{version.dandiset.identifier}/'
             f'versions/{version.version}/assets/{asset.asset_id}/'
         ).data
+        == asset.metadata.metadata
+    )
+
+    assert (
+        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/metadata/').content)
         == asset.metadata.metadata
     )
 
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -209,11 +209,6 @@ def test_asset_rest_retrieve(api_client, version, asset):
         == asset.metadata.metadata
     )
 
-    assert (
-        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/metadata/').content)
-        == asset.metadata.metadata
-    )
-
 
 @pytest.mark.django_db
 def test_asset_rest_retrieve_no_sha256(api_client, version, asset):
@@ -632,3 +627,11 @@ def test_asset_direct_download_head(api_client, storage, version, asset):
 
     with asset.blob.blob.file.open('rb') as reader:
         assert download.content == reader.read()
+
+
+@pytest.mark.django_db
+def test_asset_direct_metadata(api_client, asset):
+    assert (
+        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/metadata/').content)
+        == asset.metadata.metadata
+    )
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -632,6 +632,6 @@ def test_asset_direct_download_head(api_client, storage, version, asset):
 @pytest.mark.django_db
 def test_asset_direct_metadata(api_client, asset):
     assert (
-        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/metadata/').content)
+        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/').content)
         == asset.metadata.metadata
     )
--- a/dandiapi/urls.py
+++ b/dandiapi/urls.py
@@ -70,7 +70,7 @@ register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('api/', include(router.urls)),
     re_path(
-        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/',
+        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/$',
         asset_metadata_view,
         name='asset-direct-metadata',
     ),
--- a/dandiapi/urls.py
+++ b/dandiapi/urls.py
@@ -70,7 +70,7 @@ register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('api/', include(router.urls)),
     re_path(
-        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/metadata/',
+        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/',
         asset_metadata_view,
         name='asset-direct-metadata',
     ),
--- a/dandiapi/urls.py
+++ b/dandiapi/urls.py
@@ -11,6 +11,7 @@ from dandiapi.api.views import (
     DandisetViewSet,
     VersionViewSet,
     asset_download_view,
+    asset_metadata_view,
     auth_token_view,
     blob_read_view,
     info_view,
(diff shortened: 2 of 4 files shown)


yarikoptic:
a unit test? ideally testing that we are getting results identical to the `/dandiset/.../assets` endpoint

waxlamp:
> a unit test? ideally testing that we are getting results identical to the `/dandiset/.../assets` endpoint

See 13703e7.

dandiapi/api/tests/test_asset.py
```diff
@@ -208,6 +209,11 @@ def test_asset_rest_retrieve(api_client, version, asset):
         == asset.metadata.metadata
     )
 
+    assert (
+        json.loads(api_client.get(f'/api/assets/{asset.asset_id}/metadata/').content)
+        == asset.metadata.metadata
+    )
+
```
dchiquito, line 216: I would put this in a new test, `test_asset_direct_metadata` and put it at the bottom of the file. So far the pattern for these API tests is to test one endpoint per test.

dchiquito (changes requested):
one comment about the unit test, the view looks fine to me

dandiapi/urls.py
```diff
@@ -68,6 +69,11 @@ def to_url(self, value):
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('api/', include(router.urls)),
+    re_path(
+        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/metadata/',
```
satra, line 73: is there a reason to have `metadata` here or could this route be similar to the assets route?

`/assets/<id>` returns metadata
`/assets/<id>/download/` returns pre-signed url

waxlamp: I actually think that endpoint should also be `/assets/<id>/metadata`, while `/assets/<id>/` should ideally be reserved for the asset record itself.

If you feel strongly, I can shift mine to match what is there already, but this is definitely a violation of RESTful design principles.

satra: we aren't doing that at the other endpoint. i would keep parallel structure rather than different semantics.

waxlamp: I'll change this now to preserve the structure, but in the future we should consider changing both endpoints to my suggested design.

dandiapi/api/tests/test_asset.py
waxlamp: See 7f46190.

dandiapi/urls.py
waxlamp: See 55a00ff.

satra: we are being fairly consistent in our design. i know that it differs from some standard practice. but our returns for dandiset versions and assets is the metadata record. this was intentional and i don't think we should change that. and the posts themselves are also the metadata record.

dandiapi/urls.py
```diff
@@ -68,6 +69,11 @@ def to_url(self, value):
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('api/', include(router.urls)),
+    re_path(
+        r'api/assets/(?P<asset_id>[0-9a-f\-]{36})/',
```
satra, line 73: looks good to me. i'm assuming `assets/<id>` and `assets/<id>/` both will work here. and django will do the magic of adding the `/` if it's missing.

satra (approved):

dandiapi/urls.py
yarikoptic: IIRC that is not the case... yeap -- magic doesn't exist:
```
$> curl https://api.dandiarchive.org/api/dandisets/000027/versions/draft/assets/1c095f5f-d1e2-45db-b807-fdcfea08c6de/           
{"id":"dandiasset:1c095f5f-d1e2-45db-b807-fdcfea08c6de","path":"sub-RAT123/sub-RAT123.nwb","access":[{"status":"dandi:OpenAccess","schemaKey":"AccessRequirements"}],"digest":{"dandi:sha2-256":"1a765509384ea96b7b12136353d9c5b94f23d764ad0431e049197f7875eb352c","dandi:dandi-etag":"6ec084ca9d3be17ec194a8f700d65344-1"},"@context":"https://raw.githubusercontent.com/dandi/schema/master/releases/0.4.4/context.json","approach":[],"schemaKey":"Asset","contentUrl":["https://api.dandiarchive.org/api/assets/1c095f5f-d1e2-45db-b807-fdcfea08c6de/download/","https://dandiarchive.s3.amazonaws.com/blobs/2db/af0/2dbaf0fd-5003-4a0a-b4c0-bc8cdbdb3826"],"identifier":"1c095f5f-d1e2-45db-b807-fdcfea08c6de","repository":"https://dandiarchive.org/","contentSize":18792,"dateModified":"2021-06-29T18:59:29.491552-04:00","schemaVersion":"0.4.4","encodingFormat":"application/x-nwb","wasGeneratedBy":[{"id":"urn:uuid:e664a279-71dc-4290-9bb1-1c843e2b19cf","name":"Metadata generation","schemaKey":"Activity","description":"Metadata generated by DANDI cli","wasAssociatedWith":[{"url":"https://github.com/dandi/dandi-cli","name":"DANDI Command Line Interface","version":"0.21.0","schemaKey":"Software","identifier":"RRID:SCR_019009"}]}],"wasAttributedTo":[{"age":{"value":"P12M","unitText":"Years from birth","schemaKey":"PropertyValue"},"sex":{"name":"Male","schemaKey":"SexType","identifier":"http://purl.obolibrary.org/obo/PATO_0000384"},"species":{"name":"Brown rat","schemaKey":"SpeciesType","identifier":"http://purl.obolibrary.org/obo/NCBITaxon_10116"},"genotype":"WT","schemaKey":"Participant","identifier":"RAT123"}],"blobDateModified":"2020-10-21T10:10:35.457789-04:00","variableMeasured":[],"measurementTechnique":[]}%                                                                                            1 9731.....................................:Thu 29 Jul 2021 04:40:55 PM EDT:.
lena:~
$> curl https://api.dandiarchive.org/api/dandisets/000027/versions/draft/assets/1c095f5f-d1e2-45db-b807-fdcfea08c6de
1 9732.....................................:Thu 29 Jul 2021 04:40:56 PM EDT:.
lena:~
$> echo $?
0
```

yarikoptic (approved):
Looks LGTM to me. Thank you @waxlamp !

dandiapi/urls.py
satra: the server uses a redirect mechanism:

> curl -L https://api.dandiarchive.org/api/dandisets/000027/versions/draft/assets/1c095f5f-d1e2-45db-b807-fdcfea08c6de

will work.

waxlamp:
@dchiquito: there are a bunch of tests failing, where the API is giving a 200 instead of a 302. Any idea what's going on? Is it due to the changes I made?

dandiapi/urls.py
yarikoptic: I knew I should have just used wget ;) thanks @satra

dandiapi/urls.py
```diff
@@ -68,6 +69,11 @@ def to_url(self, value):
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('api/', include(router.urls)),
+    re_path(
```
yarikoptic, line 72: re 302 -> 200 fails, may be this one should come after the `/download/` below so it would get a chance to match that one first??? but then may be this (and the others?) one needs an anchor `$` at the end of the regex?

waxlamp: Oh wow, nice catch! Adding the `$` to the end of the regex prevents that one from matching the download URL as well. And indeed, before I added that the `/download/` endpoint magically transformed into another metadata endpoint.

See 6ce80df.

yarikoptic: cool. shouldn't all the regexes be anchored$ here to avoid some spurious "working" ones?

dchiquito: All the patterns that start with `re_path` probably should yes, although it is a moot point as long as there are no overlaps.

PR #452, opened 2021-07-30
Allow any dandiset owners to publish

dchiquito:
Publish is working in production, so it's almost time to let the general public have at it.


waxlamp (approved):
👏🏾

PR #454, opened 2021-08-03
Make users inactive by default, display "Pending Approval" message

### LINKED ISSUE #332: Create users inactive/disabled on deployed and staging instances
so information/accounts could be reviewed and enabled whenever deemed "kosher".  It should still be possible to create users "enabled" for the purpose of CI testing.

Related: #330 https://github.com/dandi/dandiarchive/issues/809

mvandenburgh:
Sets new users' `is_active` status to `False` in production and staging environments. DANDI team members can log into the admin console and check the `Active` checkbox to approve a new user. Note that this will not affect local dev/testing environments (i.e. users will still be `Active` by default in those environments). To test these changes locally, set this line to `True` https://github.com/dandi/dandi-api/blob/user-inactive-by-default/dandiapi/settings.py#L73

Immediately after signing up (and upon any attempt to sign in before being approved), users will see this screen:
Screenshot from 2021-08-03 15-25-43
 
If anyone has feedback or suggested changes for the wording, please share. In particular, I'm wasn't sure what amount of time we want to say users will be approved by (it's 1-2 business days ATM).

Closes #332 and https://github.com/dandi/dandiarchive/issues/809

### DIFF
--- a/dandiapi/api/templates/account/account_inactive.html
+++ b/dandiapi/api/templates/account/account_inactive.html
@@ -1,0 +1,20 @@
+{% extends "account/base.html" %}
+
+{% load i18n %}
+
+{% block head_title %}Account Pending Approval{% endblock %}
+
+{% block content %}
+<h1>Pending Approval</h1>
+<hr />
+
+<div class="pb-2 pt-5">
+  <p class="mb-5">This account has not been approved yet. Please allow 1-2 business days for approval.</p>
+  <p class="text-center">
+    <a href="https://dandiarchive.org" class="button inline-flex items-center is-primary">
+      Return to the DANDI Archive Homepage
+      <i class="ri-arrow-left-circle-line ml-3 text-xl"></i>
+    </a>
+  </p>
+</div>
+{% endblock %}
--- a/dandiapi/api/views/auth.py
+++ b/dandiapi/api/views/auth.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from allauth.account.signals import user_signed_up
 from django.conf import settings
 from django.db.models.signals import post_save
 from django.dispatch import receiver
@@ -18,6 +19,13 @@ from rest_framework.response import Response
 def create_auth_token(sender, instance=None, created=False, **kwargs):
     if created:
         Token.objects.create(user=instance)
+
+
+@receiver(user_signed_up)
+def set_user_inactive(sender, user, **kwargs):
+    if settings.SET_NEW_USER_INACTIVE:
+        user.is_active = False
+        user.save()
 
 
 @swagger_auto_schema(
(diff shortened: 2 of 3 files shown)


satra:
i think this should be coupled with the change in the UI to collect more info when registering.

yarikoptic:
> i think this should be coupled with the change in the UI to collect more info when registering.

and provide some admin-level web UI or API point so there could be a url to just to "go to" to enable.  Then an email to admins could be sent along for every new user, so we could just "click" on it.  Such email could include ready to click URL to user's github profile to ease "review".

Asking because it seems not that many people review new users' accounts, and I find it at least semi-tollerable to react post-fact on emails to review and typically not disable.  Needing to go through django interface for every new user would be painful.

dandiapi/api/templates/account/account_inactive.html
```diff
@@ -0,0 +1,20 @@
...
+<div class="pb-2 pt-5">
+  <p class="mb-5">This account has not been approved yet. Please allow 1-2 business days for approval.</p>
+  <p class="text-center">
+    <a href="https://dandiarchive.org" class="button inline-flex items-center is-primary">
+      Return to the DANDI Archive Homepage
+      <i class="ri-arrow-left-circle-line ml-3 text-xl"></i>
+    </a>
+  </p>
+</div>
+{% endblock %}
```
waxlamp, line 20: Nice SSR!

waxlamp (commented):
This looks great. Hold off on merging until we can figure out how to roll out the full user workflow.

waxlamp:
> > i think this should be coupled with the change in the UI to collect more info when registering.
> 
> and provide some admin-level web UI or API point so there could be a url to just to "go to" to enable. Then an email to admins could be sent along for every new user, so we could just "click" on it. Such email could include ready to click URL to user's github profile to ease "review".
> 
> Asking because it seems not that many people review new users' accounts, and I find it at least semi-tollerable to react post-fact on emails to review and typically not disable. Needing to go through django interface for every new user would be painful.

Just want to make sure that both of you (@satra and @yarikoptic) are saying to hold off on merging this until the new user survey and an admin-only "approve" API are both set up.

If that's correct, can we establish what you want to see in that user survey?

@mvandenburgh, can you work on adding a `NewUserForm.vue` type of component to the frontend for new users to fill out before they can be activated? (We can slot in the particulars of the form once Satra/Yarik say what they would like to see.)

Let's hold off for the moment on adding the admin-only endpoint because I think that one requires a bit more thought before we jump in with a design or implementation.

Thanks @mvandenburgh, this looks really nice!

yarikoptic:
> Just want to make sure that both of you (@satra and @yarikoptic) are saying to hold off on merging this until the new user survey and an admin-only "approve" API are both set up.

Given recent influx of users, I think it would be the best course of action to not introduce extra burden to our lives and frustration to users. Meanwhile it is easier to monitor and react to rare but appearing bad actors

satra:
in terms of the analogous info, i think simply asking the following through a quick entry form

- what do you plan to use DANDI for?
- affiliation

this should not take someone more than a minute to fill and simple enough to keep bad actors mostly away.

PR #456, opened 2021-08-05
Add revalidation script

dchiquito:
Add a script to revalidate all the things. It does not actually use `delay(...)`, so the validations will be done immediately in the script thread instead of being put into the task queue.

### DIFF
--- a/dandiapi/api/management/commands/revalidate.py
+++ b/dandiapi/api/management/commands/revalidate.py
@@ -1,0 +1,25 @@
+import djclick as click
+
+from dandiapi.api.models import Asset, Version
+from dandiapi.api.tasks import validate_asset_metadata, validate_version_metadata
+
+
+@click.command()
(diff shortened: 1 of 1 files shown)


dandiapi/api/management/commands/revalidate.py
```diff
@@ -0,0 +1,25 @@
+import djclick as click
+
+from dandiapi.api.models import Asset, Version
+from dandiapi.api.tasks import validate_asset_metadata, validate_version_metadata
+
+
+@click.command()
+@click.option('--assets', is_flag=True, default=False)
+@click.option('--versions', is_flag=True, default=False)
+def refresh_metadata(assets: bool, versions: bool):
```
mvandenburgh, line 10:
```suggestion
def revalidate(assets: bool, versions: bool):
```

mvandenburgh (approved):
Just a small typo, once that's fixed I approve :+1:

PR #457, opened 2021-08-05
Garbage collection

dchiquito:
The current script only deletes Assets, but we will eventually need to tidy up AssetBlobs, Uploads, and blobs in S3 as well, so there are hooks to add those later.

Things are considered as eligible for garbage collection after 7 days, but this can easily be adjusted later.


dandiapi/api/management/commands/collect_garbage.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if assets:
+        click.echo(f'Deleting {stale_assets().count()} assets...')
+        stale_assets().delete()
+    if assetblobs:
+        click.echo('Deleting AssetBlobs is not yet implemented')
+    if uploads:
+        click.echo('Deleting Uploads is not yet implemented')
+    if s3blobs:
+        click.echo('Deleting S3 Blobs is not yet implemented')
```
yarikoptic, line 33: Unless those other cleanups are intended to be added in this PR (it isn't draft), better comment out messages with noop

dandiapi/api/tests/test_garbage.py
```diff
@@ -0,0 +1,40 @@
...
+    # Only the stale orphan asset should be returned by stale_assets()
+    assert stale_assets().get() == stale_orphan_asset
+
+    # This is how assets will generally be deleted
+    stale_assets().delete()
+
+    # The stale asset should be gone
+    assert stale_assets().count() == 0
+    # The three other assets should still be present
+    assert Asset.objects.count() == 3
```
yarikoptic, line 40: Code and test look very neat and correct, but since it is a destructive operation - I wonder if we could/should add more tests (eg as for an asset to bring to multiple dandisets, had been published etc) trip sleep better

yarikoptic: May be test for exactly which assets weren't GCed? Not unseen to have bool reverted to see the same numbers of items as a result

dandiapi/api/management/commands/collect_garbage.py
waxlamp: I think the messages are fine (since this is a management script intended to be run only by us), but in these branches the script should exit with a nonzero code.

dandiapi/api/tests/test_garbage.py
waxlamp: > May be test for exactly which assets weren't GCed?

Good idea.

> Not unseen to have bool reverted to see the same numbers of items as a result

I'm not sure I know what you mean by this.

waxlamp (changes requested):
Can we add a report-and-confirm mechanism to this? Basically, add a `--force` or `--no-confirm` switch that would let us run this interactively or in the testing environment, but by default show the user which assets will be GCed and ask for confirmation before actually doing it.

Furthermore, I think we need to add some documentation explaining precisely how this is to be used. My vision for GC is a bit more expansive than a maintenance action: something that sends asset uploaders a report explaining that they have stale assets, and that they'll be deleted in such-and-such time, etc. It's a much bigger idea that we'd have to flesh out.

dandiapi/api/tests/test_garbage.py
yarikoptic: have 6 assets, remove 3 which should **not** (the `bool`) be removed instead of 3 which should, end up with 3 wrong ones and a passing test.

dchiquito:
@waxlamp I added a confirmation dialog. I'm merging this now, we can augment it later based on the GC design doc.

PR #460, opened 2021-08-06
Fix publish

### LINKED ISSUE #459: Fix publish process
The current publish process does not follow the procedure outlined in the design doc.

 https://github.com/dandi/dandi-api/blob/master/doc/design/publish-1.md#publish-process bullet 4 says that when publishing draft assets, they are modified in place to include the publish info.

https://github.com/dandi/dandi-api/blob/b55dd766d7aa28625f585bb49c8847f92feaf075/dandiapi/api/views/version.py#L139-L152 creates new published editions of the draft assets, and also does not replace the original draft assets. This means that publishing mints new Assets, which it definitely should not do.

dchiquito:
Fixes #459 

* Refactor the code that generated published Assets into code that mutates a draft asset into a published asset.
* Split `AssetFactory` into `DraftAssetFactory` and `PublishedAssetFactory`
* Add a new test for the publish endpoint that publishes a draft version with a draft asset and a published asset, and verifies that those assets are handled correctly.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -168,6 +168,7 @@ class Asset(TimeStampedModel):
         This is useful to validate asset metadata without saving it.
         To actually publish this Asset, simply save() after calling publish().
         """
+        # These fields need to be listed in the bulk_update() in VersionViewSet#publish.
         self.metadata = self.published_metadata()
         self.published = True
 
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -148,41 +148,28 @@ class Asset(TimeStampedModel):
             )
         return metadata
 
-    @classmethod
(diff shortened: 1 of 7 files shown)


dandiapi/api/models/asset.py
```diff
@@ -148,41 +148,28 @@ def _populate_metadata(self):
...
-        )
-        published_asset.metadata = published_metadata
+    def publish(self):
+        """
+        Modify the metadata of this asset as if it were being published.
 
-        return published_asset
+        This is useful to validate asset metadata without saving it.
+        To actually publish this Asset, simply save() after calling publish().
+        """
```
dchiquito, line 170: document the link to bulk_update

waxlamp (approved):
Daniel gave us a walkthrough of the code and everything looks good.

mvandenburgh (approved):

PR #467, opened 2021-08-10
DOI citations

dchiquito:
Use the DOI for the citation URL if it is present.

### DIFF
--- a/dandiapi/api/tests/factories.py
+++ b/dandiapi/api/tests/factories.py
@@ -1,3 +1,4 @@
+import datetime
 import hashlib
 
 from allauth.socialaccount.models import SocialAccount
@@ -75,6 +76,19 @@ class DraftVersionFactory(BaseVersionFactory):
 class PublishedVersionFactory(BaseVersionFactory):
     class Meta:
         model = Version
+
+    @classmethod
+    def _create(cls, *args, **kwargs):
+        version: Version = super()._create(*args, **kwargs)
+        version.doi = f'10.80507/dandi.{version.dandiset.identifier}/{version.version}'
+        now = datetime.datetime.utcnow()
+        version.metadata = {
+            **version.metadata,
+            'publishedBy': version.published_by(now),
+            'datePublished': now.isoformat(),
+        }
+        version.save()
+        return version
 
 
 class AssetBlobFactory(factory.django.DjangoModelFactory):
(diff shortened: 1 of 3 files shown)


dandiapi/api/tests/test_version.py
```diff
@@ -104,13 +105,24 @@ def test_published_version_metadata_computed(published_version):
...
+        draft_version.metadata['citation']
+        == f'{name} ({year}). (Version {draft_version.version}) [Data set]. DANDI archive. {url}'  # noqa: E501
+    )
+
+
+@pytest.mark.django_db
+def test_version_metadata_citation_published(published_version):
+    name = published_version.metadata['name'].rstrip('.')
+    year = datetime.now().year
+    url = f'https://doi.org/{published_version.doi}'
```
satra, line 122: are you actually calling the production server for this? or mocking it?

satra: i.e is `published_version` a fixture that spoofs a doi?

satra (approved):
the code looks good to me.

just left a question about `published_version` fixture.

dandiapi/api/tests/test_version.py
dchiquito: The DOI credentials are not set up in the test environment. When publishes occur in testing, a DOI string is generated, but no request is made to canonize that as a real DOI.

dandiapi/api/tests/factories.py
```diff
@@ -76,6 +77,19 @@ class PublishedVersionFactory(BaseVersionFactory):
     class Meta:
         model = Version
 
+    @classmethod
+    def _create(cls, *args, **kwargs):
+        version: Version = super()._create(*args, **kwargs)
+        version.doi = f'10.80507/dandi.{version.dandiset.identifier}/{version.version}'
```
waxlamp, line 83: Where does that hardcoded DOI prefix come from? Should it be parameterized somehow?

I realize this is a testing module, so maybe it doesn't matter here, but if this value corresponds to a production datacite instance or something like that, I wonder if it should be stored in and retrieved from an environment variable or something like that.

waxlamp (approved):

dandiapi/api/tests/factories.py
dchiquito: https://github.com/dandi/dandi-api/issues/472

PR #469, opened 2021-08-10
ENH: list supported digest algorithms in 400 error response

yarikoptic:
See https://github.com/dandi/dandi-api/issues/468 as a motivator for
introducing this change.  In general I really prefer informative error
messages which can help user to immediately resolve the "problem"
without RTFM etc.  Hence I think besides improving documentation
(no information on what algorithms are supported is shown), error
should be improved as well.

2nd commit adds that listing also to operation_description ~~the docstring... can't test locally ATM if works out as expected~~

### DIFF
--- a/dandiapi/api/tests/test_upload.py
+++ b/dandiapi/api/tests/test_upload.py
@@ -38,7 +38,7 @@ def test_blob_read_bad_algorithm(api_client, asset_blob):
         format='json',
     )
     assert resp.status_code == 400
-    assert resp.data == 'Unsupported Digest Algorithm'
+    assert resp.data.startswith('Unsupported Digest Algorithm. Supported:')
 
 
 @pytest.mark.django_db
--- a/dandiapi/api/tests/test_upload.py
+++ b/dandiapi/api/tests/test_upload.py
@@ -38,7 +38,7 @@ def test_blob_read_bad_algorithm(api_client, asset_blob):
         format='json',
     )
     assert resp.status_code == 400
-    assert resp.data.startswith('Unsupported Digest Algorithm. Supported:')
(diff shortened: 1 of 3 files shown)


yarikoptic:
finally it is all green -- first it was "bool" (forgot `not`) and then linter complained ;-)

dandiapi/api/views/upload.py
```diff
@@ -67,18 +69,22 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
 
 @swagger_auto_schema(
     method='POST',
+    operation_summary='Fetch an existing asset blob by digest, if it exists.',
+    operation_description=f'Known digest algorithms: {", ".join(supported_digests)}',
```
dchiquito, line 73:
```suggestion
    operation_description=f'Supported digest algorithms: {", ".join(supported_digests)}',
```

dandiapi/api/tests/test_upload.py
```diff
@@ -38,7 +38,7 @@ def test_blob_read_bad_algorithm(api_client, asset_blob):
         format='json',
     )
     assert resp.status_code == 400
-    assert resp.data == 'Unsupported Digest Algorithm'
+    assert resp.data.startswith('Unsupported Digest Algorithm. Supported:')
```
dchiquito, line 41:
```suggestion
    assert resp.data == 'Unsupported Digest Algorithm. Supported: dandi:dandi-etag'
```
We should be forced to update this test if we ever add new supported digest types.

dchiquito (changes requested):
Minor tweaks, otherwise looks good

yarikoptic:
> Minor tweaks, otherwise looks good

committed, thank you!

dchiquito (approved):

yarikoptic:
ok, approved, proceeding

PR #474, opened 2021-08-11
Fix migration 0016_remove_metadata_tables

dchiquito:
Migration 0016 ran perfectly in development and in staging, but was not
performant enough to run in production.

Update it to use subqueries rather than `.all()`.

This change only affects the `RunPython` step which copies the data, so
it should not affect any existing instances that have already run the
migration.


mvandenburgh (approved):

waxlamp (commented):
LGTM

Issue #437, opened 2021-07-22
Allow for upload/handling of a number/range of schema_version's


yarikoptic:
To some degree a continuation of https://github.com/dandi/dandi-api/issues/172 .

We have `schema_version` (as provided by `/info/` end-point). It is a singular version and requires very tight coordination between growing number of components (dandi-api, web ui, dandi-cli) to make everything "kosher".  E.g. to prevent users uploading assets with metadata in older versions, thus known to be invalid etc,  https://github.com/dandi/dandi-cli/pull/724 is introducing the check so no assets in a different from `schema_version` of dandi-api would be uploaded.

ATM `schema_version` signals at least 2 things within dandi-api context:
- a version of the schema dandi-api "understands" as where to find/place digests etc.
- the version of schema provided by dandischema and used by dandi-api for `to_datacite`, `aggregate_assets_summary`, `validate` (and management command `migrate`).

But in dandischema we
- already support migration of metadata between some versions of the schema (at least at dandiset level) (what `migrate` does)
- [WiP in dandischema/pull/60](https://github.com/dandi/dandischema/pull/60/files#diff-fca559c50b570e7d5a8bc3aace5e144a9cf0810d05f15879e7da7d4569bcedb5R10) with extra discussion in https://github.com/dandi/dandischema/issues/22#issuecomment-884907346 to support validation of multiple schemas .

So overall, AFAIK dandi-api isn't really the one which really cares about seeing a singular specific version of schema as long as that version remains consistent with placement of the few metadata fields which dandi-api relies on.  The rest (validation etc) falls under purview of dandischema, and as soon/long as it provides support for validating (or possibly even migrating?) metadata - dandi-api should just inform (in `/info/`) about versions of metadata schema it can allow for upload and expect them being validated ok (if they are valid in that version). And that information should be provided by `dandischema`.

So I am thinking that it should expose `allowed_schema_versions` as to what schema_versions to be allowed to be uploaded and be expected to be "valid".  We might decide to expose multiple variables or just make it into a dict to signal what versions are "natively" supported or dandi-api would attempt metadata migration to target `schema_version`, so could even be such a dict as

```json
"schema_version": "0.5.5",
"schema_versions": {
  "0.4.4": "migrate",
  "0.5.0": "native"
}
```
[...]

PR #475, opened 2021-08-12
Maintaining consistent schemas for draft metadata

satra:
This design is intended to keep the pieces of the archive in sync to the schema version deployed by the infrastructure.

### DIFF
--- a/doc/design/draft-metadata-migration.md
+++ b/doc/design/draft-metadata-migration.md
@@ -1,0 +1,13 @@
+Premise: Draft metadata should always be in the latest schema version.
+
+There are a few places in the API where draft metadata are altered (POST, PUT and served (GET). Note that the UI and 
+API use a shared schema version. All actions should thus assume that this is the only version available to the UI and
+server. Any other versions need to be migrated to this version for any display or editing.
+
+1. The API should call dandischema `migrate` on all metadata exchanges of draft metadata.
+2. The API can have the option (parameter `?migrate=true`) of providing migration on a `GET` on published metadata. 
+   This would be needed for the UI to display landing pages for published dandisets with older schemas.
+3. To determine the state of validity, the API should also revalidate all draft metadata on any schema version change 
+   in the infrastructure.
+
+


doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,13 @@
+Premise: Draft metadata should always be in the latest schema version.
+
+There are a few places in the API where draft metadata are altered (POST, PUT and served (GET). Note that the UI and 
+API use a shared schema version. All actions should thus assume that this is the only version available to the UI and
+server. Any other versions need to be migrated to this version for any display or editing.
+
+1. The API should call dandischema `migrate` on all metadata exchanges of draft metadata.
```
dchiquito, line 7: Does this mean that if a client tries to PUT/POST metadata with `schemaVersion=0.2.0` for example, then the server would automatically `migrate` it to whatever the current `schemaVersion` is before saving, and throw an error if migration fails?
Or does it mean that any draft metadata stored in the DB with an out of date `schemaVersion` will be automatically `migrate`d on every `GET` so that it appears to be up to date?

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,13 @@
...
+2. The API can have the option (parameter `?migrate=true`) of providing migration on a `GET` on published metadata. 
+   This would be needed for the UI to display landing pages for published dandisets with older schemas.
+3. To determine the state of validity, the API should also revalidate all draft metadata on any schema version change 
+   in the infrastructure.
```
dchiquito, line 11: Right now the validation task uses the `schemaVersion` specified in the metadata to validate. Revalidation is pointless unless the `schemaVersion` is changed in some way. Do you mean that all draft metadata should be migrated to the newest version whenever the version is bumped? Or should the validation task be using the server's `schemaVersion`, regardless of what `schemaVersion` is specified in the metadata?

satra: > Do you mean that all draft metadata should be migrated to the newest version whenever the version is bumped?

correct.

doc/design/draft-metadata-migration.md
dchiquito: I vote we only migrate on incoming data:
```suggestion
1. The API should call dandischema `migrate` on all incoming metadata draft metadata, and throw an error if migration fails. This ensures that the metadata in the DB will always have an up to date schemaVersion.
```
If we also migrate all drafts immediately after incrementing the schemaVersion, the DB is guaranteed to always be in a good state.

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,13 @@
...
+
+There are a few places in the API where draft metadata are altered (POST, PUT) and served (GET). Note that the UI and 
+API use a shared schema version. All actions should thus assume that this is the only version available to the UI and
+server. Any other versions need to be migrated to this version for any display or editing.
+
+1. The API should call dandischema `migrate` on all metadata exchanges of draft metadata.
+2. The API can have the option (parameter `?migrate=true`) of providing migration on a `GET` on published metadata. 
+   This would be needed for the UI to display landing pages for published dandisets with older schemas.
+3. To determine the state of validity, the API should also revalidate all draft metadata on any schema version change 
+   in the infrastructure.
```
dchiquito, line 11:
```suggestion
3. To determine the state of validity, the API should also migrate and  revalidate all draft metadata on any schema version change 
   in the infrastructure.
```

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,13 @@
...
+
```
dchiquito, line 12:
```suggestion
4. Publish should not be enabled for any dandisets with a metadata schemaVersion that is not up to date.

```

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,13 @@
+Premise: Draft metadata should always be in the latest schema version.
+
+There are a few places in the API where draft metadata are altered (POST, PUT) and served (GET). Note that the UI and 
+API use a shared schema version. All actions should thus assume that this is the only version available to the UI and
+server. Any other versions need to be migrated to this version for any display or editing.
+
```
dchiquito, line 6:
```suggestion

This proposal only applies to the `schemaVersion` in dandiset metadata, not asset metadata.

```

yarikoptic: good suggestion, I would have made it a bit more specific

```suggestion

This proposal only applies to the `schemaVersion` of dandiset metadata, not asset metadata.

```

since it is not just about the value but of the schema version that record would be expressed in.

doc/design/draft-metadata-migration.md
yarikoptic: this would not provide such a guarantee since if dandi-api starts to use newer dandischema (library/schema) then all dandiset metadata will be outdated in DB... that is why I would have preferred to

```suggestion
1. The API should provide `PUT /dandisets/{dandiset_pk}/migrate/` endpoint, which would return migrated version, and description of changes which happened, warnings.
1a. The API would refuse PUT/POST of incompatible metadata.
```

`1a` is similar to how `dandi-cli` refuses to upload an asset with non-matching metadata version ATM. More comprehensive support of non-exact versions upload is proposed in https://github.com/dandi/dandi-api/issues/437

doc/design/draft-metadata-migration.md
yarikoptic: note: With my [proposed above](https://github.com/dandi/dandi-api/pull/475/files#r688588814) `/migrate` endpoint `?migrate=true` would not be strictly necessary. Any client (e.g. web ui) could explicitly inform user that it is migrating the metadata from X to Y, list warnings etc, and proceed with the migrated version.  If we do `?migrate=true` client would still ideally need to get first without that flag to inform user that version is outdated, and then re-query with `?migrate=true` but then response will be just metadata record, without any fields to carry back information about migration procedure, warnings etc.

doc/design/draft-metadata-migration.md
yarikoptic: agree -- for stating validity, I would say it could even include that "Outdated schema" as a warning of its own

doc/design/draft-metadata-migration.md
satra: remember `?migrate=true` only applies to published dandisets (all draft dandisets auto migrate).

doc/design/draft-metadata-migration.md
dchiquito: I agree with Satra's philosophy:
> these are the kinds of things i would prefer to take care of behind the scenes. users should not have to think about this.

As long as we don't allow publishing with an out of date schema version, I don't see a problem with having schemaVersion being outdated briefly while the migration script is being run.

1. Someone upgrades the `schemaVersion` of the server. This triggers a redeploy of the server.
  Anyone who tries to publish a dandiset after the deployment will find the button and the endpoint disabled because the schema version is out of date.  
2. Whoever upgraded the `schemaVersion` runs the `migrate_version_metadata` script, which upgrades the schema version of all the draft dandisets.

yarikoptic: I would agree with auto- behind the scenes things, which are **guaranteed** not to fail/transform or loose (meta)data user entered etc. I do not think we have such a guarantee for metadata migrations. And for any migration support now we will need to pretty much grow the chain of transformations, which would become more and more "intrusive".  If all those happen behind the scene, and I as a user come and do not see metadata I have entered where I have entered it -- I might get confused etc.
Having said that DLP for read-only display could auto-migrate and add a notice that "Metadata was auto-migrated from version X for display". Needed anyways for all published dandisets since we would not be automigrating them, right? 
Then owner(s) upon publish would acknowledge the correctness of migration via edit/save and having been displayed relevant information.

> 2\. Whoever upgraded the `schemaVersion` runs the `migrate_version_metadata` script, which upgrades the schema version of all the draft dandisets.

and introspects correct migration for all 100s of dandisets, or just does not bother and thus possibly loosing data (I think it is not unprecedented to have bugs in the code) without a way to "redo it" (migrated metadata replaces original right?)?

yarikoptic: situation BTW is not unlike we are asking pynwb folks to address -- to be able to see data in the original (as in the file) version of schema, since all the automigration during opening without being able to load as it was originally saved did lead to surprising results and errors IIRC. Here also if we auto-migrate draft metadata during "long draft period", any user script which uses metadata for anything, would need to be constantly and immediately updated shortly after each migration, thus annoying the user/developer.

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,15 @@
...
+This proposal only applies to the `schemaVersion` of dandiset metadata, not asset metadata.
+
+1. The API should call dandischema `migrate` on all metadata exchanges of draft metadata.
+2. The API can have the option (parameter `?migrate=true`) of providing migration on a `GET` on published metadata.
```
yarikoptic, line 10:
```suggestion
2. The API can have the option (parameter `?migrate=<version>|bool`) of providing migration on a `GET` on published metadata. 
```
and if `/migrate` then more consistent `to_version`: we should expose `to_version` kwarg of `dandischema.migrate` so a user can consistently migrate to desired (later) schema -- e.g. to establish a meta-analysis script (relying on specific schemaVersion) across some already published dandisets.

doc/design/draft-metadata-migration.md
yarikoptic: then my `edit` in above [comment](https://github.com/dandi/dandi-api/pull/475/files#r688588814) was correct that `/migrate` should be at the level of versioned dandiset endpoint.

doc/design/draft-metadata-migration.md
satra: i would like to dissociate pynwb use case from this one. i do not think they are the same (dandischema would be). 

- from the API and  UI perspective both can use a single schema. 
- from a draft perspective, any usage of draft metadata by a user/script is not guaranteed. so if they are coding they should not rely on draft metadata being stable whether long or short
- in terms of migration, we would always be doing a one version (patch, minor, or release), it would always jump from the last schema, thus there should not be any process that should fail unless we decide something was ambiguous and cannot be transitioned. this would then be reflected in the schema validation message. "could not migrate, please contact help. etc.,."
 - for any already published dandiset, we can check what a new version of the schema will do for migration, and our migration code should account for that in dandischema.

doc/design/draft-metadata-migration.md
dchiquito: I think that is a feature we can add later if there is demand for it.

doc/design/draft-metadata-migration.md
dchiquito: To get the ball rolling again, how about this @yarikoptic ?
```suggestion
1. The API should provide `GET /dandisets/{dandiset_pk}/versions/{version}/migrated/` endpoint, which would return the migrated metadata of the version, and a description of changes and warnings.
   For up to date metadata, this is obviously a no-op.
   The meditor would use this endpoint to fetch the draft metadata whenever it is opened, which ensures that editing a draft with out of date schema version will implicitly migrate it.
   This endpoint can also be used to get the metadata of published versions as if they were migrated.
2. The API will not accept metadata with out of date `schemaVersion`.
   Draft metadata with out of date `schemaVersion` is considered invalid by the API (even if it would pass validation) and is not publishable.
```

Instead of having `PUT /dandisets/{dandiset_pk}/migrate/` do the migration and save it, `GET /dandisets/{dandiset_pk}/versions/{version}/migrated/` does the migration, which is then displayed in the meditor for the owner's awareness, and only saved if it's cool with them.

Now when we bump the schema version, we do not have to migrate anything behind the scenes. All drafts are immediately unpublishable. Migration would be done by the owners by opening the meditor and saving it. In the event of migration errors, the meditor would show those to the owner at that time.

waxlamp: Paging Dr. @yarikoptic ^^^

yarikoptic: Re 1. explicit `/migrated/` -- sounds good .

Re 2. "not accept metadata with out of date..." -- for *now* I think it is ok, but I think we might still end up needing to https://github.com/dandi/dandi-api/issues/437 to not demand some "metadata upgrade" whenever newer schema .patch version just fixes  something trivial and non-consequential (e.g. documentation etc).

dchiquito: When we implement #437 we can change "not out of date" to mean "within a minor version", or whatever makes sense at that time. I don't see any problem augmenting this design at a later date, so I'm going to apply this suggestion.

satra:
@yarikoptic - your suggestion would require enabling a `migrate` button to the UI, when most of it could be handled under the hood. also migrate of assets would require rereading the assets in many occasions. 

i suggest the following:
- draft dandiset metadata automigrate on version update
- asset metadata needs migration before publish is enabled, and the user can ask the archive to migrate for them or they can reupload with a newer CLI.
- published dandiset and asset metadata will be migrated under very controlled circumstances.

waxlamp:
> @yarikoptic - your suggestion would require enabling a `migrate` button to the UI, when most of it could be handled under the hood. also migrate of assets would require rereading the assets in many occasions.
> 
> i suggest the following:
> 
> * draft dandiset metadata automigrate on version update
> * asset metadata needs migration before publish is enabled, and the user can ask the archive to migrate for them or they can reupload with a newer CLI.
> * published dandiset and asset metadata will be migrated under very controlled circumstances.

I don't think the text of the proposed design doc matches all extant ideas from the discussion threads. Can we revisit the text, make new suggestions, and resolve existing open threads? I believe there's still quite a bit of variance in what we think is going on here, and I want to converge on a written document so that Daniel has something reliable to work from.

dchiquito:
@satra @yarikoptic I cleaned up the text of the proposal, added some more plot exposition, and cleaned up old discussions. I think this design satisfies the requirements I've heard about, and I'm ready to implement this.

Is there anything that this design does not do that it should do? If not, I will merge this and we can start prioritizing the work,

doc/design/draft-metadata-migration.md
```diff
@@ -0,0 +1,29 @@
+Premise: Draft metadata should always be in the latest schema version.
+
+There are a few places in the API where draft metadata are altered (POST, PUT) and served (GET). Note that the UI and
+API use a shared schema version. All actions should thus assume that this is the only version available to the UI and
+server. Any other versions need to be migrated to this version for any display or editing.
+
+This proposal only applies to the `schemaVersion` of dandiset metadata, not asset metadata.
+
+1. The API should provide `GET /dandisets/{dandiset_pk}/versions/{version}/migrated/` endpoint, which would return the migrated metadata of the version, and a description of changes and warnings.
```
satra, line 9: i'm assuming this always migrates to the API schema version that is deployed. there is no option to migrate to some other schema version. perhaps worth a clarification.

dchiquito: Correct. Supporting arbitrary versions would be possible, but significantly more work and it's not something we need right now. I will clarify.

satra:
LGTM

PR #479, opened 2021-08-13
CORS signal handler

### LINKED ISSUE #256: API has a cross-origin policy that does not allow other browser clients
mostly an fyi for now, but we should provide a way for accessing the API from other browser clients.

>Access to fetch at 'https://api.dandiarchive.org/api/dandisets/000043/versions/draft/' from origin 'https://cdpn.io' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

dchiquito:
Fixes #256 

* Add a signal handler that allows an CORS requests from any origin with method `GET`, `HEAD`, or `OPTIONS`.
* Move our other signal handlers into one location, `signals.py`.

### DIFF
--- a/dandiapi/api/signals.py
+++ b/dandiapi/api/signals.py
@@ -1,0 +1,9 @@
+from corsheaders.signals import check_request_enabled
+
+
+def cors_allow_anyone_read_only(sender, request, **kwargs):
+    """Allow any read-only request from any origin."""
+    return request.method in ('GET', 'HEAD', 'OPTIONS')
+
+
+check_request_enabled.connect(cors_allow_anyone_read_only)
--- a/dandiapi/api/signals.py
+++ b/dandiapi/api/signals.py
@@ -1,9 +1,18 @@
 from corsheaders.signals import check_request_enabled
+from django.conf import settings
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from rest_framework.authtoken.models import Token
 
 
+@receiver(check_request_enabled)
 def cors_allow_anyone_read_only(sender, request, **kwargs):
     """Allow any read-only request from any origin."""
     return request.method in ('GET', 'HEAD', 'OPTIONS')
 
 
-check_request_enabled.connect(cors_allow_anyone_read_only)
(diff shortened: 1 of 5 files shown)


dandiapi/api/signals.py
```diff
@@ -0,0 +1,28 @@
...
+from django.dispatch import receiver
+from rest_framework.authtoken.models import Token
+
+from dandiapi.api.mail import send_registered_notice_email
+
+
+@receiver(check_request_enabled, dispatch_uid='cors_allow_anyone_read_only')
+def cors_allow_anyone_read_only(sender, request, **kwargs):
+    """Allow any read-only request from any origin."""
+    return request.method in ('GET', 'HEAD', 'OPTIONS')
```
waxlamp, line 14: Does this cover our current need? If so, can we strip down our current list of allowed hosts in the Terraform setup? If not, we'll need to consult the list of hosts here to also grant access to the mutative operations if the request comes from one of those.

waxlamp (commented):
This is great, thank you @dchiquito!

A question about software design: is it better to put all the signals in one place, rather than distributed to the modules in which they are actually used? I know it doesn't matter to the functioning of the software, but from a legibility perspective, I wonder if it's good to keep related logic together in their "home" modules. Thoughts?

dandiapi/api/signals.py
dchiquito: This signal handler is basically a layer before the default CORS behavior. Before checking anything else, this will allow anyone to do any read-only operations. If a user is trying to do a POST/PUT/DELETE, this handler does not apply, so the server reverts to the default CORS behavior, which checks against the CORS_ORIGIN_WHITELIST.

@satra can better answer if there are any applications that are strictly read-only and can therefore be removed from CORS_ORIGIN_WHITELIST once this is merged.

satra: you could remove the codepen uris. we are not using it for anything other than `GET`.

waxlamp (approved):

PR #480, opened 2021-08-16
Allow admin users to delete published versions

### LINKED ISSUE #476: allow (with force=True, admin-only?) deletion of dandisets with published versions
ATM the need primarily for use of staging archive for testing of https://github.com/dandi/dandisets/pull/55 but eventually we might need this functionality to prune bogus/compromised/etc dandisets anyways.

jwodder:
Closes #476.

### DIFF
--- a/dandiapi/api/views/version.py
+++ b/dandiapi/api/views/version.py
@@ -153,3 +153,28 @@ class VersionViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelVie
 
         serializer = VersionSerializer(new_version)
         return Response(serializer.data, status=status.HTTP_200_OK)
+
+    @swagger_auto_schema(
+        manual_parameters=[DANDISET_PK_PARAM, VERSION_PARAM],
+    )
+    def destroy(self, request, **kwargs):
+        """
+        Delete a version.
+
+        Deletes a version. Only published versions can be deleted, and only by
+        admin users.
+        """
+        version: Version = self.get_object()
+        if version.version == 'draft':
+            return Response(
+                'Cannot delete draft versions',
+                status=status.HTTP_403_FORBIDDEN,
+            )
+        elif not request.user.is_superuser:
+            return Response(
+                'Cannot delete published versions',
+                status=status.HTTP_403_FORBIDDEN,
+            )
+        else:
+            version.delete()
+            return Response(None, status=status.HTTP_204_NO_CONTENT)
--- a/dandiapi/api/tests/test_version.py
+++ b/dandiapi/api/tests/test_version.py
@@ -798,3 +798,50 @@ def test_version_rest_publish_invalid_metadata(api_client, user, draft_version, 
     )
     assert resp.status_code == 400
     assert resp.data == 'Dandiset metadata or asset metadata is not valid'
+
+
+@pytest.mark.django_db
+def test_version_rest_delete_published_not_admin(api_client, user, published_version):
+    assign_perm('owner', user, published_version.dandiset)
+    api_client.force_authenticate(user=user)
+    response = api_client.delete(
+        f'/api/dandisets/{published_version.dandiset.identifier}'
+        f'/versions/{published_version.version}/'
+    )
+    assert response.status_code == 403
+    assert response.data == 'Cannot delete published versions'
+    assert published_version in Version.objects.all()
+
+
+@pytest.mark.django_db
+def test_version_rest_delete_published_admin(api_client, admin_user, published_version):
+    api_client.force_authenticate(user=admin_user)
+    response = api_client.delete(
+        f'/api/dandisets/{published_version.dandiset.identifier}'
+        f'/versions/{published_version.version}/'
+    )
+    assert response.status_code == 204
+    assert not Version.objects.all()
+
+
+@pytest.mark.django_db
+def test_version_rest_delete_draft_not_admin(api_client, user, draft_version):
+    assign_perm('owner', user, draft_version.dandiset)
+    api_client.force_authenticate(user=user)
+    response = api_client.delete(
+        f'/api/dandisets/{draft_version.dandiset.identifier}/versions/{draft_version.version}/'
+    )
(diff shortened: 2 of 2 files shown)


dandiapi/api/views/version.py
```diff
@@ -153,3 +153,28 @@ def publish(self, request, **kwargs):
...
+
+    @swagger_auto_schema(
+        manual_parameters=[DANDISET_PK_PARAM, VERSION_PARAM],
+    )
+    def destroy(self, request, **kwargs):
+        """
+        Delete a version.
+
+        Deletes a version. Only published versions can be deleted, and only by
+        admin users.
```
yarikoptic, line 165: I believe this would not be visible in the swagger, and even "Delete a version" will not be ... use `operation_*` args to `swagger_auto_schema` to provide this information, as e.g. in https://github.com/dandi/dandi-api/pull/469/files#diff-fce128f68807723a60c1a0ebd67b9c11634a49450b51c5dc9ecac7b2273cdcbcR72

dandiapi/api/views/version.py
```diff
@@ -153,3 +153,28 @@ def publish(self, request, **kwargs):
...
+        """
+        version: Version = self.get_object()
+        if version.version == 'draft':
+            return Response(
+                'Cannot delete draft versions',
+                status=status.HTTP_403_FORBIDDEN,
```
waxlamp, line 171: I am wondering if this (and the other instance below) should be `405` rather than `403`. But reading about the error codes, it seems like a toss-up. I'm ok with leaving this as is, but I wanted to mention my confusion over which is more appropriate as a discussion topic.

dandiapi/api/views/version.py
dchiquito: I think if those parameters are omitted from `@swagger_auto_schema`, swagger will use the docstring instead.

yarikoptic: You are 100% correct on that. I mixed up with the "opposite issue" that single line docstrings do not show up as the summary (so need to be clicked on to see it)

image

more of background on that: https://github.com/axnsan12/drf-yasg/issues/205#issuecomment-454331301 . So probably that should be addressed but unrelated to this PR.

yarikoptic: resolving by https://github.com/dandi/dandi-api/pull/482

dandiapi/api/views/version.py
dchiquito: I also vote 405, but the distinction is not important to me.

yarikoptic: from 
image
seems like `403` is a good one... the bible (https://httpwg.org/specs/rfc7231.html#status.403) confirms it:

- `403`: ... "The client SHOULD NOT automatically repeat the request with the same credentials. The client MAY repeat the request with new or different credentials. "
- `405`: also reminds that ". The origin server MUST generate an `Allow` header field in a 405 response containing a list of the target resource's currently supported methods.", which we do not want to mess here with, reconfirming that `403` is the right one to use

waxlamp: Yeah I think 403 is good here. But, your quote from the bible in the 403 section is not evidence for that: in our case, a `DELETE` to certain resources will always fail, so repeating the request with new or different credentials is not indicated.

In any case, let's just stick with 403. Thanks for the discussion.

yarikoptic: > a `DELETE` to certain resources will always fail

to some other "resources", not to the endpoint in question, right? ;)

waxlamp: Not sure what you mean, but in this case, doing a `DELETE` to any draft version of any dandiset will always result in failure, regardless of authentication or authorization level. The draft version is a `Version` resource, so that's what I meant in my last comment.

yarikoptic: ah, I thought that by "resources" [*] you meant some other endpoints (e.g. for a top level `/dandiset/{id}`) not the versions... 

[*] as the term used in in the [RFC, and identified by URL](https://httpwg.org/specs/rfc7231.html#resources) so means exactly what you had in mind -- "draft version of a dandiset"

indeed ! Now you convinced a blind me me in the opposite, that `405` should be the right one to use **here** (and keep 403 for admin-vs-mortal distinction), with `Allow` then listing `GET` and `PUT`? (is there programmatic way to get all known to this endpoint and just remove DELETE?)

waxlamp: > ah, I thought that by "resources" [*] you meant some other endpoints (e.g. for a top level `/dandiset/{id}`) not the versions...
> 
> [*] as the term used in in the [RFC, and identified by URL](https://httpwg.org/specs/rfc7231.html#resources) so means exactly what you had in mind -- "draft version of a dandiset"
> 
> indeed ! Now you convinced a blind me me in the opposite, that `405` should be the right one to use **here** (and keep 403 for admin-vs-mortal distinction), with `Allow` then listing `GET` and `PUT`?

The only trouble I can think of is that there is in fact one use case for `DELETE`ing a draft version--when we eventually get to "archive" or "protection" semantics for dandiset, we will indeed want to remove the draft version from consideration. So maybe 403 is, in fact, the right choice here.

> (is there programmatic way to get all known to this endpoint and just remove DELETE?)

I don't think so--the allowed methods list is specific knowledge about the semantics of a given endpoint. In this case, if we wanted to use 405, we would have to edit the headers of the response object and fill in the allowed method names.

But I think we should stick with 403 for now. 405 makes more sense when there are no conceivable semantics for a specific method, which I don't believe is the case here.

yarikoptic: I am fine either way.
If we go with 403 - anything else to do to get it merged?

dchiquito (approved):

dandiapi/api/views/version.py
waxlamp: Just Daniel's approval, which I see he just granted!

yarikoptic:
Great! Thank you @jwodder @dchiquito  and @waxlamp !!!

PR #482, opened 2021-08-17
ENH(UX): overload @swagger_auto_schema to show one-line docstrings as summary

yarikoptic:
A follow up to https://github.com/dandi/dandi-api/pull/480#discussion_r689903879

More information/references in the DANDISwaggerAutoSchema docstring.

Without such an overload, any oneline docstring is not used as an action
summary but rather as a description, so not visible until that action is
"clicked on" in the swagger interface.

compare 
image

to
image

### DIFF
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -10,6 +10,7 @@ from composed_configuration import (
     TestingBaseConfiguration,
 )
 from configurations import values
+from drf_yasg.inspectors import SwaggerAutoSchema
 
 
 class DandiMixin(ConfigMixin):
@@ -67,7 +68,9 @@ class DandiMixin(ConfigMixin):
     #
     # When Brian Helba is able to resolve this problem upstream (in
     # django-composed-configuration) we can remove this setting.
-    SWAGGER_SETTINGS = {}
+    SWAGGER_SETTINGS = {
+        'DEFAULT_AUTO_SCHEMA_CLASS': 'dandiapi.settings.DANDISwaggerAutoSchema',
+    }
 
 
 class DevelopmentConfiguration(DandiMixin, DevelopmentBaseConfiguration):
@@ -107,3 +110,19 @@ class HerokuStagingConfiguration(HerokuProductionConfiguration):
     # We are using cheaper Heroku dynos for staging, so we need to artificially lower the number
     # of concurrent tasks (default is 8) to keep memory usage down.
     CELERY_WORKER_CONCURRENCY = 2
+
+
+class DANDISwaggerAutoSchema(SwaggerAutoSchema):
+    """Custom class for @swagger_auto_schema to provide summary from one line docstrings.
+
+    See https://github.com/axnsan12/drf-yasg/issues/205 and
+    https://github.com/axnsan12/drf-yasg/issues/265 for more information on why it is not
+    a default behavior.
+    """
+
+    def split_summary_from_description(self, description):
+        summary, description = super().split_summary_from_description(description)
+        if summary is None:
+            return description, None
+        else:
+            return summary, description
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -10,7 +10,6 @@ from composed_configuration import (
     TestingBaseConfiguration,
 )
 from configurations import values
-from drf_yasg.inspectors import SwaggerAutoSchema
 
 
 class DandiMixin(ConfigMixin):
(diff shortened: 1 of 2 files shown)


waxlamp:
Could you post before and after screenshots of an opened Swagger block?

yarikoptic:
image

image

as expected... and looks identical for the one with multiline docstring

image

yarikoptic:
tests fail though, I wonder if due to something not resolved as commented on at https://github.com/dandi/dandi-api/pull/482/files#diff-d79ac6951b48e8506ebb640a3bbbd24f52cffc7206dc4a1bf7ad74fd7faeb145R69 and above?

waxlamp (approved):
This seems to work correctly, though I don't fully understand how returning `(description, None)` in case of a one-line docstring results in the full docstring being shown as both the summary and the description.

yarikoptic:
> This seems to work correctly, though I don't fully understand how returning `(description, None)` in case of a one-line docstring results in the full docstring being shown as both the summary and the description.

oh, should I may be add a comment?  The idea is to return both split `return summary, description` only in the case whenever "full docstring" (containing both summary and description) is present.  If it is not full, default logic of swagger makes `summary` to be None and then places everything into `description` -- I just literally just "swap" `summary` and `description` in such a case then, so that one liner becomes the `summary` instead of `description`

dchiquito:
Weird that the tests broke. Looking at that

waxlamp:
> > This seems to work correctly, though I don't fully understand how returning `(description, None)` in case of a one-line docstring results in the full docstring being shown as both the summary and the description.
> 
> oh, should I may be add a comment? The idea is to return both split `return summary, description` only in the case whenever "full docstring" (containing both summary and description) is present. If it is not full, default logic of swagger makes `summary` to be None and then places everything into `description` -- I just literally just "swap" `summary` and `description` in such a case then, so that one liner becomes the `summary` instead of `description`

That's what I figured, but I was nervous that returning `None` for `description` would cause the description itself to be blank. I guess this is not what happens. A quick comment in your code would be helpful IMO.

dandiapi/settings.py
```diff
@@ -107,3 +110,19 @@ class HerokuStagingConfiguration(HerokuProductionConfiguration):
...
+    """Custom class for @swagger_auto_schema to provide summary from one line docstrings.
+
+    See https://github.com/axnsan12/drf-yasg/issues/205 and
+    https://github.com/axnsan12/drf-yasg/issues/265 for more information on why it is not
+    a default behavior.
+    """
+
+    def split_summary_from_description(self, description):
+        summary, description = super().split_summary_from_description(description)
+        if summary is None:
```
yarikoptic, line 125:
```suggestion
        if summary is None:
            # If oneline docstring, drf_yasg will place it into "description" and not "summary".
            # We just swap it then ("description" could still be None) so oneliner serves as "summary".
```

yarikoptic:
smth like [this](https://github.com/dandi/dandi-api/pull/482/files#r691346909) ?

yarikoptic:
> Weird that the tests broke. Looking at that

my wild guess was that it has to do with
```
    # Clearing out the stock `SWAGGER_SETTINGS` variable causes a Django login
    # button to appear in Swagger, along with a spurious "authorize" button that
    # doesn't work. This at least enables us to authorize to the Swagger page on
    # the spot, which is quite useful.
    #
    # When Brian Helba is able to resolve this problem upstream (in
    # django-composed-configuration) we can remove this setting.
```
which now I populate with that single item - may be somehow it changes other behaviors as well then?

dchiquito:
Nope, adding `from drf_yasg.inspectors import SwaggerAutoSchema` somehow causes the error :confused: I'm going to dig a little deeper into that, that's very strange behavior

dchiquito:
Any submodule of `drf_yasg`, including `inspectors`, imports `drf_yasg.app_settings`, which tries to access the Django settings. Since you are adding an import of `drf_yasg.inspectors` in our `settings.py`, our settings aren't available yet. I have no idea why a more meaningful error isn't thrown, but if you move the definition of `DANDISwaggerAutoSchema` somewhere else it should fix the errors.

yarikoptic:
> I have no idea why a more meaningful error isn't thrown, but if you move the definition of `DANDISwaggerAutoSchema` somewhere else it should fix the errors.

"fun" indeed.  The question is -- where would you prefer that abomination to be moved... for now moved into a new dedicated `dandiapi/swagger_settings.py` .

dchiquito:
I think `dandiapi/swagger.py` would be fine.

yarikoptic:
> I think `dandiapi/swagger.py` would be fine.

done, let's wait for the tests, and if no surprises -- will merge later on

PR #483, opened 2021-08-19
Make /blobs/digest/ support SHA256

### LINKED ISSUE #468: blobs_digest_create (AKA POST /blobs/digest) - support 'dandi:sha2-256' algorithm
pretty much a replacement for a closed #135 , so let me answer @waxlamp question

> @yarikoptic can you clarify whether this is still needed? It sounds like you are asking for a content-addressable store for the blobs, but I'm not clear on why accessing a dandiset's asset blobs via asset IDs is a problem.

it is desired to provide user convenience and ability to integrate with external services (such as https://github.com/flatironinstitute/kachery) which operate based on a commonly used (unlike dandi-etag) sha256 checksum.

For some reason I thought that we already support fetching blobs by sha256 (well -- #135 was closed ;-)), so I desperately tried to get a blob while trying different "sha256" algorithm specifications, but then looked into the code: https://github.com/dandi/dandi-api/blob/master/dandiapi/api/views/upload.py#L80 

```python
    if request_serializer.validated_data['algorithm'] != 'dandi:dandi-etag':
        return Response('Unsupported Digest Algorithm', status=400)
```
to realize that it is not the case. It is not urgent, but I think because AFAIK we do have sha256 stored for blobs we should make it possible to request a blob by its sha256.

jwodder:
Closes #468.

### DIFF
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -70,7 +70,7 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
 @swagger_auto_schema(
     method='POST',
     operation_summary='Fetch an existing asset blob by digest, if it exists.',
-    operation_description=f'Supported digest algorithms: {", ".join(supported_digests.keys())}',
+    operation_description=f'Supported digest algorithms: {", ".join(supported_digests)}',
     request_body=DigestSerializer(),
     responses={200: AssetBlobSerializer()},
 )
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -83,7 +83,7 @@ def blob_read_view(request: Request) -> HttpResponseBase:
     algorithm = request_serializer.validated_data['algorithm']
     if algorithm not in supported_digests:
         return Response(
-            f'Unsupported Digest Algorithm. Supported: {", ".join(supported_digests.keys())}',
+            f'Unsupported Digest Algorithm. Supported: {", ".join(supported_digests)}',
             status=400,
         )
     digest = request_serializer.validated_data['value']
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -19,7 +19,7 @@ from dandiapi.api.models import AssetBlob, Upload
 from dandiapi.api.tasks import calculate_sha256
 from dandiapi.api.views.serializers import AssetBlobSerializer
 
-supported_digests = ('dandi:dandi-etag',)
+supported_digests = {'dandi:dandi-etag': 'etag', 'dandi:sha2-256': 'sha256'}
 
 
 class DigestSerializer(serializers.Serializer):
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/upload.py
```diff
@@ -70,7 +70,7 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
 @swagger_auto_schema(
     method='POST',
     operation_summary='Fetch an existing asset blob by digest, if it exists.',
-    operation_description=f'Supported digest algorithms: {", ".join(supported_digests)}',
+    operation_description=f'Supported digest algorithms: {", ".join(supported_digests.keys())}',
```
yarikoptic, line 73: AFAIK .keys() aren't needed here
```
$> python3 -c 'print(",".join({"1":2, "a":"b"}))'
1,a
```
which allows for a more "generic programming" (e.g. if `supported_digests` where just a tuple or a list of those supported).

```suggestion
    operation_description=f'Supported digest algorithms: {", ".join(supported_digests)}',
```

dandiapi/api/views/upload.py
```diff
@@ -80,14 +80,15 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
 def blob_read_view(request: Request) -> HttpResponseBase:
     request_serializer = DigestSerializer(data=request.data)
     request_serializer.is_valid(raise_exception=True)
-    if request_serializer.validated_data['algorithm'] not in supported_digests:
+    algorithm = request_serializer.validated_data['algorithm']
+    if algorithm not in supported_digests:
         return Response(
-            f'Unsupported Digest Algorithm. Supported: {", ".join(supported_digests)}',
+            f'Unsupported Digest Algorithm. Supported: {", ".join(supported_digests.keys())}',
```
yarikoptic, line 86:
```suggestion
            f'Unsupported Digest Algorithm. Supported: {", ".join(supported_digests)}',
```

yarikoptic (commented):
just a minor comment on AFAIK not needed portion of the diff

dandiapi/api/views/upload.py
```diff
@@ -80,14 +80,15 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
 def blob_read_view(request: Request) -> HttpResponseBase:
     request_serializer = DigestSerializer(data=request.data)
     request_serializer.is_valid(raise_exception=True)
-    if request_serializer.validated_data['algorithm'] not in supported_digests:
```
yarikoptic, line 83: per discussion in #483 let's support both `GET` and for just awhile "good old" `POST`. Depending on how that to be implemented "the best djangowy way" (@dchiquito ?) , `operation_summary` for `POST` can state that it is `DEPRECATED! Use GET instead`

dchiquito: I just tried to implement it and it turns out that GET requests do not have request bodies, which is what is being used here to send in the algorithm and value. Updating to use query parameters would be outside the scope of this PR, so I vote we file an issue and handle it later.

dchiquito (approved):

dandiapi/api/views/upload.py
yarikoptic: Isn't for GET it should just be provided in  request header?

dandiapi/api/views/upload.py
yarikoptic: @jwodder this was another similar point on keys

yarikoptic:
Ok, @dchiquito approved. Tests are green. Let's proceed. Thank you @jwodder @dchiquito

PR #484, opened 2021-08-19
Make `DEBUG` setting configurable w/ env var in development

mvandenburgh:
The Django debug toolbar interferes with the puppeteer e2e tests in the dandiarchive repo. This PR allows the `DEBUG` setting to be optionally configured via an environment variable. This only affects development.


mvandenburgh:
I think I was wrong about this, the tests looked like they were clicking the debug toolbar for me locally but they seem to be working fine in CI.

PR #487, opened 2021-08-20
Add create_dev_dandiset script

### LINKED ISSUE #287: Figure out dev dandiset data provisioning
Some options:

* Write dev script to create corresponding dandiset for an existing production dandiset, then pull in dandi-cli as dev dependency and upload it
* Write dev script without dandi-cli dev dependency (need to upload manually)
* Add dev upload option to dandi-cli

dchiquito:
I decided to manually create the test data because the only alternatives I could think of involved dandi-cli, which is a new dependency and is subject to API changes.

If we ever change the structure of anything, the test should catch it.

Fixes #287

### DIFF
--- a/dandiapi/api/management/commands/create_dev_dandiset.py
+++ b/dandiapi/api/management/commands/create_dev_dandiset.py
@@ -1,0 +1,69 @@
+from uuid import uuid4
+
+from django.conf import settings
+from django.contrib.auth.models import User
+from django.core.files.uploadedfile import SimpleUploadedFile
+import djclick as click
+
+from dandiapi.api.models import Asset, AssetBlob, Dandiset, Version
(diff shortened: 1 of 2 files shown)


mvandenburgh (commented):
Does `foo/bar.txt` appear correctly for you in the file browser after running this? the `foo/` folder shows up correctly for me, but it appears as empty:

Screenshot from 2021-08-20 12-50-50

dchiquito:
@mvandenburgh great catch, the asset path incorrectly started with `/` which breaks some assumptions the API makes.

There is now an error downloading the asset, but it looks like that is because of a change in the minio storage internals. I'll fix that in a later PR.

mvandenburgh (approved):
> There is now an error downloading the asset, but it looks like that is because of a change in the minio storage internals. I'll fix that in a later PR.

Sounds good, in that case this LGTM.

PR #488, opened 2021-08-20
Fix dev Minio environment

dchiquito:
Previously `base_url` was only set when `MINIO_STORAGE_MEDIA_URL` was
set, and canibilized the netloc from `MINIO_STORAGE_MEDIA_URL`. This
makes it impossible to sign asset download URLs in development where
`MINIO_STORAGE_MEDIA_URL` is not set and Minio is used.

I believe the latest major Minio release https://github.com/minio/minio/releases/tag/RELEASE.2021-07-08T01-15-01Z updated the minio console (which looks very snazzy now) to use a different port from the Minio API, so `docker-compose.yml` needs to be updated to use it.

I filed https://github.com/girder/cookiecutter-girder-4/issues/155 in the cookiecutter to fix it upstream.


mvandenburgh (approved):
This looks sane to me, and it does allow me to download files again in my local environment. 

I'm not well-versed in MinIO usage though so maybe you want to wait until @brianhelba takes a look (or maybe not, since this only affects dev environments). Either way I'll approve since it doesn't affect anything in production or staging.

PR #489, opened 2021-08-20
Fix user search in dev environments

mvandenburgh:
User search doesn't work in local deployments because it only considers `SocialAccounts` (which local environments don't use). This adds an additional query that searches all `User` objects too; the query is conditionalized so it only runs when the `SocialAccount` search returns no results (otherwise there would be duplicates between the `User` and `SocialAccount` objects).

This also allows the E2E tests to test adding an owner to a dandiset, which will hopefully prevent that feature from breaking again.

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.models import User
+from django.db.models import Q
 from django.http.response import HttpResponseBase
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import api_view, parser_classes, permission_classes
@@ -77,6 +78,17 @@ def users_search_view(request: Request) -> HttpResponseBase:
 
     social_accounts = SocialAccount.objects.filter(extra_data__icontains=username)[:10]
     users = [social_account_to_dict(social_account) for social_account in social_accounts]
+
+    # Try searching Django's regular `User`s if there aren't any results.
+    # This allows this feature to work in development without having to conditionalize
+    # code based on the type of deployment.
+    if not users:
+        users = [
+            user_to_dict(user)
+            for user in User.objects.filter(username__icontains=username).filter(
+                ~Q(username='AnonymousUser')
+            )[:10]
+        ]
 
     response_serializer = UserDetailSerializer(users, many=True)
     return Response(response_serializer.data)


dandiapi/api/views/users.py
```diff
@@ -78,5 +78,15 @@ def users_search_view(request: Request) -> HttpResponseBase:
     social_accounts = SocialAccount.objects.filter(extra_data__icontains=username)[:10]
     users = [social_account_to_dict(social_account) for social_account in social_accounts]
 
+    # Try searching Django's regular `User`s if there aren't any results.
+    # This allows this feature to work in development without having to conditionalize
+    # code based on the type of deployment.
+    if len(users) == 0:
```
dchiquito, line 84:
```suggestion
    if not users:
```
Apparently this is more pythonic https://stackoverflow.com/a/53522

dandiapi/api/views/users.py
```diff
@@ -78,5 +78,15 @@ def users_search_view(request: Request) -> HttpResponseBase:
...
+        users = [
+            user_to_dict(user)
+            for user in User.objects.filter(username__icontains=username)[:10]
+            if user.username != 'AnonymousUser'
```
dchiquito, line 88:
```suggestion
            for user in User.objects.filter(username__icontains=username).filter(~Q(username='AnonymousUser'))[:10]
```
You will also need `from django.db.models import Q`

dandiapi/api/views/users.py
mvandenburgh: Changed it

dandiapi/api/views/users.py
mvandenburgh: Changed it

dandiapi/api/views/users.py
```diff
@@ -78,5 +79,16 @@ def users_search_view(request: Request) -> HttpResponseBase:
     social_accounts = SocialAccount.objects.filter(extra_data__icontains=username)[:10]
     users = [social_account_to_dict(social_account) for social_account in social_accounts]
 
+    # Try searching Django's regular `User`s if there aren't any results.
+    # This allows this feature to work in development without having to conditionalize
+    # code based on the type of deployment.
+    if not users == 0:
```
dchiquito, line 85:
```suggestion
    if not users:
```

dchiquito (changes requested):

dandiapi/api/views/users.py
mvandenburgh: :facepalm:

dchiquito (approved):

waxlamp (commented):
Nice bugfix. I have a question: why not always search for `User` models? What's the benefit in looking at `SocialAccount` first, if the same results would show up in `User` as well?

mvandenburgh:
> Nice bugfix. I have a question: why not always search for `User` models? What's the benefit in looking at `SocialAccount` first, if the same results would show up in `User` as well?

Using `SocialAccount` allows searching by a user's github username.

PR #490, opened 2021-08-20
Use get_object_or_404 in views

### LINKED ISSUE #486: "Asset matching query does not exist" errors pop up in the logs
which leads to

```
1365203836443738138     2021-08-20 02:39:41     2021-08-20 02:39:41     6519188732      dandi-api       54.235.113.36   Local7  Info    app/web.1       [02:39:41] ERROR    Internal Server Error: /api/assets/f841afd9-e9d9- log.py:224\n
1365203836443738139     2021-08-20 02:39:41     2021-08-20 02:39:41     6519188732      dandi-api       54.235.113.36   Local7  Info    app/web.1                           408c-95b2-a9d31543c635/download/                            \n
```

Since GC is not yet in place, I wonder how that could happen... the only things I thought of
- staging instance web ui managing to provide URLs to main instance, but it is not the case -- all good there, urls point to api-staging and that one redirects to staging bucket urls.  
- is a mix up between staging and main instances, but that asset uuid is not on staging either.  
- it is not tests since such errors aren't anyhow regular
- Sure thing could be a user just trying random UUIDs 

so should we just assume it was some user trying random UUIDs or there could be more to the story?

dchiquito:
This results in 404 errors instead of 500 errors when an object cannot
be found in the DB.

Fixes #486

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -80,7 +80,7 @@ def _download_asset(asset: Asset):
 )
 @api_view(['GET', 'HEAD'])
 def asset_download_view(request, asset_id):
-    asset = Asset.objects.get(asset_id=asset_id)
+    asset = get_object_or_404(Asset, asset_id=asset_id)
     return _download_asset(asset)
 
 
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):

waxlamp (commented):
More "djangoic", excellent.

PR #492, opened 2021-08-20
Populate created date

### LINKED ISSUE #485: dandiset metadata dateCreated returns None
```python

with DandiAPIClient() as client:
    dandiset = next(client.get_dandisets())
    print(dandiset.created)
    metadata = dandiset.get_metadata()
    print(metadata.dateCreated)

2020-03-15 22:56:55.655000+00:00
None
```

dchiquito:
Fixes #485 

This also exposed a bug in our existing metadata dates. While we were correctly using UTC timestamps, there was no timezone info included (`+00:00`). Any metadata saved after this goes live will have the correct timezone info, but any existing metadata (i.e. any published dandisets) won't.
FYI @satra  @yarikoptic


yarikoptic:
what is "the correct timezone info"? I just assumed it was all in UTC as they had trailing `Z`
```
$> curl --silent https://api.dandiarchive.org/api/dandisets/000003/versions/ | jq -r '.results[] | "\(.created) \(.modified)"'
2020-03-15T22:56:55.655000Z 2021-08-12T14:48:45.203390Z
2021-08-12T14:48:45.032118Z 2021-08-12T14:48:45.186427Z
```
which is IMHO it is ok -- I do not care to see the time zone where server resides reflected there.

dchiquito:
When I was testing I noticed some cases where the `Z` was omitted, so I updated everything to explicitly use `+00:00`.

yarikoptic:
Attn @jwodder as it might have implications on dataladification and dandi-cli

PR #494, opened 2021-08-23
Consolidate DANDI_SCHEMA_VERSION

dchiquito:
Currently, we have the setting `DANDI_SCHEMA_VERSION` as an `environ_required`, which means that it must be specified as an environment variable during deployment, so the "real" schema version is updated in Terraform. This is not great, as several things in dandi-api could break if the schema changes.

Instead, set `DANDI_SCHEMA_VERSION` in settings.py (with an optional env var override). If we ever need to update it, then we will update it in settings.py and cut a new release, which will test the new schema version in CI and in staging.

See also https://github.com/dandi/dandi-infrastructure/pull/93


mvandenburgh (approved):

waxlamp (commented):
Does this mean Terraform should *stop* sending a schema version as well?

dchiquito:
Correct. If it did, it would override the setting specified here. I thought it would be nice to keep the functionality around just in case we needed to tweak the schema version temporarily for a specific deployment, but generally major schema updates could involve code updates, so we should update schema versions in the code that relies on it.

waxlamp:
We should probably have a docs section dealing with these variables and where/how to set them.

PR #496, opened 2021-08-23
New user questionnaire design doc

mvandenburgh:
A design doc describing the new approval-driven user registration workflow.

Link to markdown rendered doc https://github.com/dandi/dandi-api/blob/design/new-user-questionnaire/doc/design/new-user-questionnaire.md

### DIFF
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -45,7 +45,7 @@ user hasn't filled out the questionnaire yet, the button will say "COMPLETE SIGN
 
 ### Proposed admin approval process
 1) A user signs up as described in steps 1-4 in the previous section.
-2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
+2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc), their current `status`, as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
 3)
     a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED` and adds them to the `approved_users` group, granting them full access to DANDI.
     <br>
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -12,6 +12,7 @@ Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will b
 ### Changes to API
 - /users/me needs to include the user's current status
 - A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
+- Only approved users can create and edit dandisets.
 
 ### Changes to frontend
 - Configuration so the frontend knows where to find SSR pages
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -38,7 +38,7 @@ user hasn't filled out the questionnaire yet, the button will say "COMPLETE SIGN
 6) The DANDI admins will decide whether to approve or reject the user.
 7) a) If approved, the user will receive an email saying so and they will be able to logon and use DANDI as normal.
    <br>
-   b) If rejected, the user will receive an email containing the custom reason (if one was provided), as well as an email address to contact if they belieave
+   b) If rejected, the user will receive an email containing the custom reason (if one was provided), as well as an email address to contact if they believe
    they shouldn't have been rejected.
 
 
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -13,6 +13,7 @@ Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will b
 - /users/me needs to include the user's current status
 - A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
 - Only approved users can create and edit dandisets.
+- /users/search endpoint will only return users in the `approved_users` group.
 
 ### Changes to frontend
 - Configuration so the frontend knows where to find SSR pages
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -5,9 +5,16 @@ approval process to help prevent misuse of the archive.
 ## 1. Implementation Details
 ### Changes to User model
 Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will be introduced to the `User` model in dandi-api.
-- `status` is an enum that can be one of three values: `INCOMPLETE`, `PENDING`, `APPROVED`, or `REJECTED`.
+- `status` is an enum that can be one of four values: `INCOMPLETE`, `PENDING`, `APPROVED`, or `REJECTED`.
 - `questionnaire_form` is a JSON field. The keys are questions, and the values are the users' answers.
 - `rejection_reason` is an optional string that contains a message written by one of the admins describing why a user was denied approval (if applicable).
+
+### Changes to API
+- /users/me needs to include the user's current status
+
+### Changes to frontend
+- Configuration so the frontend knows where to find SSR pages
+- User avatar circley thing needs to show user's current status
 
 ### New views
 - SSR rendered pages for questionnaire form, pending approval page, rejected page, and admin approval page.
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -1,0 +1,48 @@
+# Premise
+Currently, anyone with a GitHub account can signup for DANDI and make full use of its functionality. The goal of this design document is to introduce an
+approval process to help prevent misuse of the archive.
+
+## 1. Implementation Details
+### Changes to User model
+Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will be introduced to the `User` model in dandi-api.
+- `status` is an enum that can be one of three values: `INCOMPLETE`, `PENDING`, `APPROVED`, or `REJECTED`.
+- `questionnaire_form` is a JSON field. The keys are questions, and the values are the users' answers.
+- `rejection_reason` is an optional string that contains a message written by one of the admins describing why a user was denied approval (if applicable).
+
+### New views
+- SSR rendered pages for questionnaire form, pending approval page, rejected page, and admin approval page.
+
+## 2. High-level Overview
+### Current user signup flow
+1) The user navigates to https://gui.dandiarchive.org and clicks the "LOG IN WITH GITHUB" button
+2) The user is redirected to GitHub, enters their credentials, and clicks "Sign in". This action creates a new DANDI Archive account for the user and gives
+them full capabilities of any other normal user (creating/editing dandisets, etc).
+3) The user is redirected back to https://gui.dandiarchive.org signed in to their new account.
+
+### Proposed user signup flow
+1) The user navigates to https://gui.dandiarchive.org and clicks the "LOG IN WITH GITHUB" button.
+2) The user is redirected to GitHub, enters their credentials, and clicks "Sign in". This action creates a new DANDI archive account for the user with the `status` field set to `INCOMPLETE`.
+3) The user is redirected to the SSR rendered questionnaire form and they fill it out and submit it. This populates the `questionnaire_form` field in the DB and sets the `status` field to `PENDING`.
+4) A message is displayed telling the user when they can expect approval.
+5) The user is redirected to the DANDI Archive homepage. The "LOG IN WITH GITHUB" button is now red and says something like "PENDING ACCOUNT APPROVAL". If the
+user hasn't filled out the questionnaire yet, the button will say "COMPLETE SIGNUP PROCESS" and will redirect the user to the questionnaire form when clicked.
+6) The DANDI admins will decide whether to approve or reject the user.
+7) a) If approved, the user will receive an email saying so and they will be able to logon and use DANDI as normal.
+   <br>
+   b) If rejected, the user will receive an email containing the custom reason (if one was provided), as well as an email address to contact if they belieave
+   they shouldn't have been rejected.
+
+
+### Proposed admin approval process
+1) A user signs up as described in steps 1-4 in the previous section.
+2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
+3)
+    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED`, granting them full access to
+    DANDI.
+    <br>
+    b) The admin decides to reject the user, optionally fills out the "Rejection reason" box, and clicks "REJECT". This sets the user's `status` field to `REJECTED` and `is_active` field to `false`.
+
+### Questionnaire
+(from https://github.com/dandi/dandi-api/pull/454#issuecomment-892218858)
+- What do you plan to use DANDI for?
+- Any affiliations (university, company, etc)
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -11,6 +11,7 @@ Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will b
 
 ### Changes to API
 - /users/me needs to include the user's current status
+- A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
 
 ### Changes to frontend
 - Configuration so the frontend knows where to find SSR pages
@@ -44,8 +45,7 @@ user hasn't filled out the questionnaire yet, the button will say "COMPLETE SIGN
 1) A user signs up as described in steps 1-4 in the previous section.
 2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
 3)
-    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED`, granting them full access to
-    DANDI.
+    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED` and adds them to the `approved_users` group, granting them full access to DANDI.
     <br>
     b) The admin decides to reject the user, optionally fills out the "Rejection reason" box, and clicks "REJECT". This sets the user's `status` field to `REJECTED` and `is_active` field to `false`.
 
--- a/doc/design/new-user-questionnaire.md
+++ b/doc/design/new-user-questionnaire.md
@@ -18,6 +18,8 @@ Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will b
 ### Changes to frontend
 - Configuration so the frontend knows where to find SSR pages
 - User avatar circley thing needs to show user's current status
+   -  This will come in the form of a red notification dot with a popover explaining what actions the user needs to take to be approved. 
+   -  If an unverified user is logged in and tries to do something that requires verification (edit metadata, upload a new dandiset, etc.) then the appropriate button would be grayed out, with a popover message saying they need to be verified, and to either fill out the questionnaire or email the admins (depending on what steps they've completed so far).
 
 ### New views
 - SSR rendered pages for questionnaire form, pending approval page, rejected page, and admin approval page.


doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,55 @@
...
+## 1. Implementation Details
+### Changes to User model
+Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will be introduced to the `User` model in dandi-api.
+- `status` is an enum that can be one of four values: `INCOMPLETE`, `PENDING`, `APPROVED`, or `REJECTED`.
+- `questionnaire_form` is a JSON field. The keys are questions, and the values are the users' answers.
+- `rejection_reason` is an optional string that contains a message written by one of the admins describing why a user was denied approval (if applicable).
+
+### Changes to API
+- /users/me needs to include the user's current status
+- A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
```
satra, line 14:
```suggestion
- A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
- Only approved users can create and edit dandisets.
```

satra (approved):
one small comment about un-approved users not being able to take any writeable action on the archive.

waxlamp:
@yarikoptic, are you good with this design doc?

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+### Current user signup flow
+1) The user navigates to https://gui.dandiarchive.org and clicks the "LOG IN WITH GITHUB" button
+2) The user is redirected to GitHub, enters their credentials, and clicks "Sign in". This action creates a new DANDI Archive account for the user and gives
+them full capabilities of any other normal user (creating/editing dandisets, etc).
+3) The user is redirected back to https://gui.dandiarchive.org signed in to their new account.
+
+### Proposed user signup flow
+1) The user navigates to https://gui.dandiarchive.org and clicks the "LOG IN WITH GITHUB" button.
+2) The user is redirected to GitHub, enters their credentials, and clicks "Sign in". This action creates a new DANDI archive account for the user with the `status` field set to `INCOMPLETE`.
+3) The user is redirected to the SSR rendered questionnaire form and they fill it out and submit it. This populates the `questionnaire_form` field in the DB and sets the `status` field to `PENDING`.
```
yarikoptic, line 34: so this step would require sending the filled out questionnaire to the API endpoint (e.g. `POST /users/questionnaire`) which would require being authenticated, which user would be.
`GET /users/questionnaire` would return for authorized user his/her filled out questionnaire.  That later could be used if we decide that a user needs to update questionnaire, so the same web UI end point for entering questionnair could be used and pre-populated.

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+4) A message is displayed telling the user when they can expect approval.
+5) The user is redirected to the DANDI Archive homepage. The "LOG IN WITH GITHUB" button is now red and says something like "PENDING ACCOUNT APPROVAL". If the
```
yarikoptic, line 36: I guess it relates to "User avatar circley thing needs to show user's current status" [above](https://github.com/dandi/dandi-api/pull/496/files#diff-48ab438eb0621a2dc7ee9fa4f116a6b492f0dbb555fde73b90bfb4a438fb1109R19) making it into a long squarish circley ;)

But that squarish circley should be like an active user circley then since we should still allow users to log out (they are already logged in via github).  So I wonder if it should just remain a circley with initials, but be just of different  color (not blue) to reflect status + that modal window which now shows API key instead providing details on the status (not the key) if clicked upon.  That would make UI more consistent IMHO, and also give place later possibly to add some "Contact us if waited for so long" button within that modal etc.

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+user hasn't filled out the questionnaire yet, the button will say "COMPLETE SIGNUP PROCESS" and will redirect the user to the questionnaire form when clicked.
+6) The DANDI admins will decide whether to approve or reject the user.
```
yarikoptic, line 38: how/where - django admin UI for now? will there be some "user status listener" which would react on change ?

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+7) a) If approved, the user will receive an email saying so and they will be able to logon and use DANDI as normal.
+   <br>
+   b) If rejected, the user will receive an email containing the custom reason (if one was provided), as well as an email address to contact if they belieave
```
yarikoptic, line 41:
```suggestion
   b) If rejected, the user will receive an email containing the custom reason (if one was provided), as well as an email address to contact if they believe
```

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+   they shouldn't have been rejected.
+
+
+### Proposed admin approval process
+1) A user signs up as described in steps 1-4 in the previous section.
+2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
```
yarikoptic, line 47: with aforementioned suggestion to avoid SSR -- would be web ui page which can bring up all information via API. but unfortunately ATM API is not designed toward having `/users/{username}`, so it might just need tobe able to provide `username` option to `GET /users/questionnaire`.

Alternative -- define `/users/all/{username}/` and `/users/all/{username}/questionnaire/` endpoints for which admins would be able to GET and POST/PUT (e.g. modify the `status` for the user).
That would allow to avoid SSR, keep user-management exposed via API as it is done for dandisets and assets etc. (also avoid `/search` + post-filtering to lookup user record)

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+3)
+    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED` and adds them to the `approved_users` group, granting them full access to DANDI.
+    <br>
+    b) The admin decides to reject the user, optionally fills out the "Rejection reason" box, and clicks "REJECT". This sets the user's `status` field to `REJECTED` and `is_active` field to `false`.
```
yarikoptic, line 51: `PUT /users/all/{username}` a new record which should just change `status` and dandi-api would do anything else needed upon change of status

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Questionnaire
+(from https://github.com/dandi/dandi-api/pull/454#issuecomment-892218858)
+- What do you plan to use DANDI for?
+- Any affiliations (university, company, etc)
```
yarikoptic, line 56: are you saying this should be that part of the webui we already have for affiliations in meditor (`dandischema`!)? ;)

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+1) A user signs up as described in steps 1-4 in the previous section.
+2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
+3)
+    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED` and adds them to the `approved_users` group, granting them full access to DANDI.
+    <br>
+    b) The admin decides to reject the user, optionally fills out the "Rejection reason" box, and clicks "REJECT". This sets the user's `status` field to `REJECTED` and `is_active` field to `false`.
+
+### Questionnaire
+(from https://github.com/dandi/dandi-api/pull/454#issuecomment-892218858)
+- What do you plan to use DANDI for?
```
yarikoptic, line 55: could be a free form text and/or (?) list of multiselect checkboxes

- [ ] upload data for publication
- [ ] collaboration
- [ ] DANDI Hub for computing

yarikoptic (changes requested):
I would prefer avoiding SSR if it is what I think it is ;) since that would bind API server and web UI and totally can be avoided. I do not see a reason (neither it seems was discussed) why it could not be full-fledged (albeit minimal) extension to API to provide user management available to the admin users + corresponding support on web UI to interact with it.
Not 100% certain yet if dandischema itself could/should provide the model for the questionnaire, but also do not see exactly why it could not.

doc/design/new-user-questionnaire.md
yarikoptic: Not sure if we should wake up a bear of GDPR here with a question either user is "from EU".

That questionnaire form should probably have a URL to our "Terms of Service" to which user should click "Agree" as well (not sure if then we should record here the version of the current Terms of Service since people might agree to one version and later we would issue a new one and might need to alert old users to its availability... brings point also, if possible, to include "or later version" in our Terms)

yarikoptic: also I wonder if among those 3 options I listed, we could add `download data`, and if only that one is selected, we could alert user that they do not even need to register to get access to the data? (although might be too late since we already added a record for them, right?)

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
+# Premise
+Currently, anyone with a GitHub account can signup for DANDI and make full use of its functionality. The goal of this design document is to introduce an
+approval process to help prevent misuse of the archive.
+
+## 1. Implementation Details
+### Changes to User model
+Three new columns, `status`, `questionnaire_form`, and `rejection_reason` will be introduced to the `User` model in dandi-api.
+- `status` is an enum that can be one of four values: `INCOMPLETE`, `PENDING`, `APPROVED`, or `REJECTED`.
+- `questionnaire_form` is a JSON field. The keys are questions, and the values are the users' answers.
```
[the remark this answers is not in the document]
waxlamp: I am hoping for a pretty basic usage here: we use the information to verify users, and if the questionnaire changes, we don't consider verified users to become unverified. For that reason I don't think a version field to audit who answered what, at what time, is really warranted.

As for where the questionnaire itself should be kept, I'm not sure why `dandischema` would be the right place. This is something that integrates tightly with how user accounts are created. It's not a part of the DANDI metadata schema or anything like that.

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+- `rejection_reason` is an optional string that contains a message written by one of the admins describing why a user was denied approval (if applicable).
+
+### Changes to API
+- /users/me needs to include the user's current status
+- A [Django group](https://docs.djangoproject.com/en/3.2/topics/auth/default/#groups), `approved_users`, will be created. Only users in this group will have access to protected endpoints.
+- Only approved users can create and edit dandisets.
```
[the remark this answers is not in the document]
waxlamp: I assume that we would tweak the search endpoint to only return verified users. Alternatively, we would identify how to redesign the user search endpoints to provide different data streams for different purposes, including one that only returns verified users. (This could be as simple as filtering the full search endpoint with a query argument of some kind.)

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Changes to frontend
+- Configuration so the frontend knows where to find SSR pages
+- User avatar circley thing needs to show user's current status
+
+### New views
+- SSR rendered pages for questionnaire form, pending approval page, rejected page, and admin approval page.
```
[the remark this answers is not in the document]
waxlamp: Not using SSR seems quite complex, when the user creation and verification procedures *already* "bind web ui and dandi-api".

@mvandenburgh can provide further detail on why SSR is needed here, but I don't think separation of concerns is a valid argument in this case. The users are being created and processed serverside, after all.

doc/design/new-user-questionnaire.md
waxlamp: I think we'd use something like the red status dot that Github and other services use. Color would be fine too. This is a question for @jtomeck to help us out with.

doc/design/new-user-questionnaire.md
waxlamp: For now it would be the Django admin console, yes. If that proves too cumbersome, we can think about a better solution (perhaps protected API endpoints for carrying out different actions for a user, linked into an email report sent to the admins with the questionnaire answers).

doc/design/new-user-questionnaire.md
waxlamp: > Not sure if we should wake up a bear of GDPR here with a question either user is "from EU".
> 
> That questionnaire form should probably have a URL to our "Terms of Service" to which user should click "Agree" as well (not sure if then we should record here the version of the current Terms of Service since people might agree to one version and later we would issue a new one and might need to alert old users to its availability... brings point also, if possible, to include "or later version" in our Terms)

Let's defer this. We are not lawyers, so binding people at signup time with a ToS doesn't seem like a good idea, especially if we don't know how to keep the door open for future updates.

On that particular topic, my own experience (which must be the same as everyone else's) is that services update their TsoS all the time, and the standard practice seems to be that if I don't like the new terms, I can delete my account, etc. I assume that's how we would handle it ourselves. (i.e., no need to include language that covers changes to the TsoS in the original ToS.)

waxlamp: > also I wonder if among those 3 options I listed, we could add `download data`, and if only that one is selected, we could alert user that they do not even need to register to get access to the data? (although might be too late since we already added a record for them, right?)

I don't think it's worthwhile for us to *discourage* people from making an account, even if they only want to download data.

It might be prudent to let people know up front that they can browse and download without an account, but yes by the time they agree to create an account, I think we should commit and not worry about this case where they just need to download things.

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
yarikoptic: Everything pretty much is being created and processed server side, while ui provides parameters etc.

Having more of such interactions going not through API and main web ui IMHO
- hinders UX (some custom web forms outside of the main web ui look/feel)
- might later manifest itself in other locations (testing etc) requiring more selenium style navigation instead of just a few API calls

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
mvandenburgh: > I assume that we would tweak the search endpoint to only return verified users.

I think this would be sufficient for the needs of the web app and hub. The only place where the user search endpoint is used in the web app is for adding owners to a dandiset, where we don't want non-`APPROVED` users showing up either.

doc/design/new-user-questionnaire.md
jtomeck: > I think we'd use something like the red status dot that Github and other services use. Color would be fine too. This is a question for @jtomeck to help us out with.

I agree with you, @waxlamp. The status dot is a nice, simple, yet highly effective UX choice.

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
waxlamp: I didn't really answer the original question: indeed, we do use SSR pages for the login flow. In particular, the "authorize DANDI?" page that we suppressed in deployment was an SSR page, which is appropriate because everything in the login flow happens very much serverside, without any participation from the client.

As far as UX, using something like Materialize CSS should easily match the style of the app. And as for testing, we would indeed use end-to-end testing for this, but since we're testing an extra component of the login flow, that would be an appropriate mode of testing.

In the end, doing this as an SSR page will simplify the development of this feature and avoid adding extra, unnecessary complexity to the API. It also matches the engineers' intuition about how this will work, so I think it's a win.

doc/design/new-user-questionnaire.md
waxlamp: I am strongly against a design that would require users to update their questionnaire in the future. The questionnaire is meant to help us verify who users are, and in using it we are shifting some burden onto users. I think it's totally reasonable to have them fill this out when they sign up, but anything beyond that burdens them excessively, IMO.

doc/design/new-user-questionnaire.md
waxlamp: So I think the regular "logged in" circle with initials would stay, plus the red notification dot (with a popover or something explaining why it's red). And if an unverified user is logged in and tries to do something that requires verification (edit metadata, upload a new dandiset, etc.) then the appropriate button would be grayed out, with a popover message saying they need to be verified, and to either fill out the questionnaire or email the admins (depending on what steps they've completed so far).

doc/design/new-user-questionnaire.md
waxlamp: To continue the discussion from above, I'm not sure what we get from squirming to avoid SSR like this. I want to avoid adding API endpoints just to help the admins verify users when SSR lets us cut right to the chase. The SSR approach solves the problem directly without the extra work of modifying the API.

I want to point out once more that our client and server are *already* intimately bound together. Even if your approach that avoids SSR, the login and verification flows simply won't work if the server's not there.

doc/design/new-user-questionnaire.md
waxlamp: Thinking about it again, I don't think a Terms of Service should appear alongside the questionnaire. If anything, a ToS should appear as soon as a user wishes to create an account, and the process should abort if they don't accept. But that's not something we're tackling in this workflow, so I'd like to open a new issue to discuss how a ToS can be presented to the user.

doc/design/new-user-questionnaire.md
yarikoptic: ok, so whatever clarifications users are to provide -- should be provided in the email correspondence and thus not captured in the questionnaire. I guess we can live with that

doc/design/new-user-questionnaire.md
yarikoptic: ok, let's go with SSR

doc/design/new-user-questionnaire.md
yarikoptic: ok, let's forget about reusing our existing dandischema/web-ui component for entering metadata here.

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
yarikoptic: ok, SSR it is

doc/design/new-user-questionnaire.md
```diff
@@ -0,0 +1,56 @@
...
+   they shouldn't have been rejected.
+
+
+### Proposed admin approval process
+1) A user signs up as described in steps 1-4 in the previous section.
+2) An email is sent to DANDI admins with a link to an SSR page that lists the user's info (name, email, GitHub account, etc) as well as their answers to the questionnaire. The page will contain two buttons - "APPROVE" or "REJECT" - and a textbox underneath the "REJECT" button entitled "Rejection Reason".
+3)
+    a) The admin decides to approve the user and clicks "APPROVE". This updates the user's `status` field to `APPROVED` and adds them to the `approved_users` group, granting them full access to DANDI.
+    <br>
+    b) The admin decides to reject the user, optionally fills out the "Rejection reason" box, and clicks "REJECT". This sets the user's `status` field to `REJECTED` and `is_active` field to `false`.
```
yarikoptic, line 51: what would be workflow if we REJECT someone, but they reach out and we decide to APPROVE.  Could that SSR page link be reused to then choose APPROVE?  So in other words -- could that page be "idempotent" as  be available for any user (which might have previously was APPROVEd or REJECTed but we need to change our decision)

doc/design/new-user-questionnaire.md
yarikoptic: ok, let's place ToS aside since we don't have any formulated yet.  But we should really get to it and indeed agreeing to ToS should be done while initiating account creation (well -- this questionnair is part of the process)

doc/design/new-user-questionnaire.md
yarikoptic: right, will be done via SSR page as described below.

doc/design/new-user-questionnaire.md
yarikoptic: above sounds good, I guess should be put into text

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
yarikoptic: should be verbalized ("only approved users returned"?) in text of the design doc.

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
yarikoptic: yeah, probably was a stretch, although might have allowed for consistent UI and also versioning of questionnairs etc.  Let's forget about it

doc/design/new-user-questionnaire.md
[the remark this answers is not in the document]
mvandenburgh: > should be verbalized ("only approved users returned"?) in text of the design doc.

https://github.com/dandi/dandi-api/commit/c8cde8bda07f8a069b98551068584ff311ade839

doc/design/new-user-questionnaire.md
mvandenburgh: Yes, the SSR page link would still work and could be used to change a user's status. I've updated the design doc to show the user's `status` on that SSR page https://github.com/dandi/dandi-api/pull/496/commits/9b1bfa1e67c8f93499521e8e1d20c4b9339d506a

doc/design/new-user-questionnaire.md
mvandenburgh: :+1:  Added to design doc https://github.com/dandi/dandi-api/commit/bebba7adadd193820bf24acd6eb368ed27cce68e

mvandenburgh:
@yarikoptic I believe I addressed all of your comments, PTAL.

doc/design/new-user-questionnaire.md
yarikoptic: ok.  I would have preferred some explicit wording that the page could be used later as well to change the status, but as long as we are on the same line, ok with me ;) thank you @mvandenburgh !

doc/design/new-user-questionnaire.md
yarikoptic: sounds good - thank you @mvandenburgh

yarikoptic (approved):
My comments were addressed, so IMHO ok to proceed. Thank you @mvandenburgh and @waxlamp !

PR #499, opened 2021-08-28
Fix next published version

### LINKED ISSUE #498: Published version numbers out of order on staging
Some recent tests of [dandi/dandisets](https://github.com/dandi/dandisets) were failing because, after publishing a Dandiset (number 000186, now deleted) on staging with version ID 0.210827.1436, the next published version had version ID 0.210827.1435.  Is this supposed to happen?  I thought version numbers were always assigned in increasing order.

dchiquito:
Fixes #498


jwodder:
This looks OK to me, but I don't think I'm familiar enough with the code base for my review to qualify for merge permission.

dandiapi/api/tests/test_version.py
```diff
@@ -12,25 +12,38 @@
...
+def test_version_next_published_version_simultaneous_save(
+    dandiset_factory,
+    published_version_factory,
+):
+    dandiset_1 = dandiset_factory()
+    dandiset_2 = dandiset_factory()
+    version_1 = published_version_factory(dandiset=dandiset_1)
+    version_2 = published_version_factory(dandiset=dandiset_2)
+    # Different dandisets published at the same time should have the same version string
+    assert version_1.version == version_2.version
```
yarikoptic, line 44: as version is time dependent, probably very unlikely to hit the "new minute" marker between two publications, but still possible, so might happen once in very while, causing need to review etc.  I wonder if it worth adding some explicit "redo" if we cross the minute-marker between two invocations?

mvandenburgh (approved):
LGTM

dandiapi/api/tests/test_version.py
```diff
@@ -12,25 +12,38 @@
...
 
 
+@pytest.mark.django_db
+def test_version_next_published_version_simultaneous_save(
+    dandiset_factory,
+    published_version_factory,
+):
+    dandiset_1 = dandiset_factory()
+    dandiset_2 = dandiset_factory()
+    version_1 = published_version_factory(dandiset=dandiset_1)
```
yarikoptic, line 41: BTW, does this trigger actual doi minting? If yes - does it explicitly delete that dandiset version upon completion of the fixture (thus triggering future #501 )?

dchiquito: it doesn't, pytest tests do not have the DOI credentials provided. The only environments that mint DOIs are production and staging.

PR #504, opened 2021-08-30
Delete DOI when deleting a version

### LINKED ISSUE #501: Delete/Cleanup Draft DOIs whenever removing versioned dandiset
Follow up to  PR #480 . I have noticed that we accumulated 371 DOIs on https://doi.test.datacite.org/repositories/dartlib.dandi/dois by running tests (ab)using 

- [ ] Deletion introduced in #480 should add an out-off-band job to remove a DOI found in that dandiset version metadata.  `Findable` DOIs cannot be removed, so we would not even try I guess, but all the `Draft` (which we populate in staging as far as I see it from those 371) could and should be removed 
- [ ] clean up already existing Draft DOIs on doi.test.datacite.org

ref: https://github.com/dandi/dandischema/issues/7

jwodder:
Closes #501.

### DIFF
--- a/dandiapi/api/tasks.py
+++ b/dandiapi/api/tasks.py
@@ -1,4 +1,3 @@
-import logging
 import os
 
 from celery import shared_task
@@ -6,9 +5,9 @@ from celery.utils.log import get_task_logger
 from django.conf import settings
 from django.db.transaction import atomic
 import jsonschema.exceptions
-import requests
 
 from dandiapi.api.checksum import calculate_sha256_checksum
+from dandiapi.api.doi import delete_doi
 from dandiapi.api.manifests import (
     write_assets_jsonld,
     write_assets_yaml,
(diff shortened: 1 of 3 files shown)


dandiapi/api/tasks.py
```diff
@@ -152,3 +154,33 @@ def validate_version_metadata(version_id: int) -> None:
     version.status = Version.Status.VALID
     version.validation_errors = []
     version.save()
+
+
+@shared_task
+def delete_doi(doi: str) -> None:
```
dchiquito, line 160: Can you move this method to `doi.py` so all the DOI code is in one place? This task can simply be a wrapper around it.

dchiquito (changes requested):
Sad that we can't really add a test.

dandiapi/api/tasks.py
jwodder: Done.

dchiquito (approved):

PR #506, opened 2021-08-30
Exclude disabled users from search results

### LINKED ISSUE #505: Disabled users show up in user search
The user search endpoint currently includes user accounts that have been disabled in its results. To fix, the user search endpoint should filter out users with `is_active` set to `False`.

mvandenburgh:
Closes #505.

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -76,17 +76,22 @@ def users_search_view(request: Request) -> HttpResponseBase:
     request_serializer.is_valid(raise_exception=True)
     username: str = request_serializer.validated_data['username']
 
-    social_accounts = SocialAccount.objects.filter(extra_data__icontains=username)[:10]
+    # Perform a search, excluding any inactive users
+    social_accounts = SocialAccount.objects.filter(
+        extra_data__icontains=username, user__is_active=True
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/users.py
```diff
@@ -76,17 +76,22 @@ def users_search_view(request: Request) -> HttpResponseBase:
...
     # This allows this feature to work in development without having to conditionalize
     # code based on the type of deployment.
     if not users:
+        # Perform a search, excluding any user accounts that
+        # are inactive + the default 'AnonymousUser' account
         users = [
             user_to_dict(user)
             for user in User.objects.filter(username__icontains=username).filter(
-                ~Q(username='AnonymousUser')
+                ~Q(username='AnonymousUser'), Q(is_active=True)
```
dchiquito, line 94:
```suggestion
                ~Q(username='AnonymousUser'),
                is_active=True,
```
I'm not sure what the linter-friendly ordering is, if any.

The Q object is only necessary here to provide the `~` negation, you can just use the normal kwarg form for `is_active`.

dchiquito (changes requested):

dchiquito (approved):

PR #508, opened 2021-09-06
DOI config check

### LINKED ISSUE #503: Settings check in `create_doi` uses wrong logic?
Shouldn't these `or`'s be `and`'s here?  I assume we only want to POST if *all* of the settings are set.

https://github.com/dandi/dandi-api/blob/77a30b2c0c3dcf057a1088512128ceccd97d26ca/dandiapi/api/doi.py#L31-L36

dchiquito:
Django has a facility to perform system checks, so this check should run whenever any server starts up.

Fixes #503

### DIFF
--- a/dandiapi/api/checks.py
+++ b/dandiapi/api/checks.py
@@ -1,12 +1,7 @@
 from django.conf import settings
 from django.core.checks import Error, register
 
-DANDI_DOI_SETTINGS = [
-    (settings.DANDI_DOI_API_URL, 'DANDI_DOI_API_URL'),
-    (settings.DANDI_DOI_API_URL, 'DANDI_DOI_API_USER'),
-    (settings.DANDI_DOI_API_PASSWORD, 'DANDI_DOI_API_PASSWORD'),
-    (settings.DANDI_DOI_API_PREFIX, 'DANDI_DOI_API_PREFIX'),
-]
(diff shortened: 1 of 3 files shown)


jwodder:
@dchiquito Should [this function](https://github.com/dandi/dandi-api/blob/2744de602e967c626eab303aa0c8807ef0132c77/dandiapi/api/doi.py#L28-L34) be updated as well?

dchiquito:
@jwodder I didn't think so, but I think you're right that it could use the same `DANDI_DOI_SETTINGS` list that the check uses. I will update it.

waxlamp (approved):

dandiapi/api/checks.py
```diff
@@ -0,0 +1,23 @@
...
+    any_doi_setting = any([setting is not None for setting, _ in DANDI_DOI_SETTINGS])
+    if not any_doi_setting:
+        # If no DOI settings are defined, DOIs will not be created on publish.
+        return []
+    errors = []
+    for setting, name in DANDI_DOI_SETTINGS:
+        if setting is None:
+            errors.append(
+                Error(
+                    f'Setting {name} is not specified, but other DOI settings were.',
```
yarikoptic, line 18: minor: just for a consistent tense
```suggestion
                    f'Setting {name} is not specified, but other DOI settings are.',
```

PR #510, opened 2021-09-07
Initial draft for Embargo (approach 1) design doc

yarikoptic:
Just pushed the state I got to so far, so possibly already some discussion could happen. Will keep working on it even if no immediate feedback would arrive, but if someone sees a *show stopper* please speak up.

### DIFF
--- a/doc/design/embargo-1.md
+++ b/doc/design/embargo-1.md
@@ -1,5 +1,29 @@
 # Embargo support
 
+## User experience
+
+### uploading embargoed data
+When a user creates a dandiset, they have the option to mark it as "open" or "embargoed". There is some help text that explains these options, either as a hover-hint or some other way. The text for "open" would indicate that the dandiset would be visible to anyone.
+
+If "embargoed" is selected, a page would be displayed that would explain the rules of embargoed data. The user is required to check "I agree" before proceeding. There is also a "back" option. On this page, the user is required to enter the grant or award number, and we would ideally validate that this is an active award. The message would indicate that this option is only available to BRAIN Initiative grants, and that the landing page of the dandiset would be visible to everyone but the underlying assets would not be. As per recent NIH policy, the user will be required to make the dandiset "open" upon the first publication of the data, or when the funding for the grant runs out, whichever comes first. We will be able to determine when the grant expires and we will automatically open the dandiset if it has not already been opened. Using the grant number, we determine the embargoedUntil date and add it automatically as metadata which is visible publicly on the dandiset landing page.
+
+It is expected that each grant have at least one of these embargoed dandisets. Since it is not always clear from the outset how the data from a grant will be divided into publications, it is expected that several groups may start with a single dandiset.
+
+For embargoed data the owner would have options to add user permissions, with an interface similar to Google Docs:
+
+* add read-only member. This would be used for collaborators with team members who do not need to submit data but only to read it. This would also be useful for reviewers of manuscripts who need to see the data as part of the review process. Read access allows a user to view available assets, and download them with the web API, the Python API, and reading data directly using s3 streaming. A DANDI Hub user will automatically use the credentials from the dandi hub account they are currently using.
+* add owner. This would give them read and write permissions and the ability to add users, including owners.
+
+Permissions are all by dandiset, not by asset. If a user needs to set different permissions for specific assets within the dandiset, they will need to create a new dandiset and put the assets in this separate place.
+When a there is a new publication, the user is expected to create a new dandiset for this publication. There should be an online interface for copying assets from an existing dandiset to this one.
+
+### requesting access
+
+When a user without access to an embargoed dandiset visits the landing page, they see a message that says that this
+dataset is currently under embargo until the embargo end date. They have a button to request read access. If they
+click this button, an email is sent to the contact person with a link that leads them to an interface where they can
+give this user access.
+
 ## Principles
 
 ### Use of key stores
--- a/doc/design/embargo-1.md
+++ b/doc/design/embargo-1.md
@@ -1,0 +1,95 @@
+# Embargo support
+
+## Principles
+
+### Use of key stores
+
+Rely on the fact that "stores" (`blobs/`, and future `zarr/` see #295) although "public" are not pragmatically "usable" for users, since the assets named based on their IDs, and provide no semantic meaning (not even a file extension).
+It would be an expensive procedure for someone to "mine" the store(s) for some relevant data they would like to "steal".
+
+### "Move" operation within/between buckets is "expensive"
+
+There is no AWS S3 API for quick move (renaming) of a key within a bucket.
+As a result, "move" operation can be **very** expensive in terms of traffic (thus possibly cost if not "sponsored"), time it would take for transfer, implementation e.g. handling of possible errors during transfer, etc
+
+### Publishing a dandiset is enabled only for Open dandisets
+
+### Use-cases to fulfill
+
+- UC-Collaboration: authorized users of emabrgoed dataset should be able to manipulate it as easily as if it was a public dataset.
+- UC-Review: ability to make embargoed dataset available to other users "read-only".
+- UC-Promotion (findability): make embargoed dataset discoverable by others, so that Corresponding contact could be contacted to provide access to the dandiset.
+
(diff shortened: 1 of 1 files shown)


doc/design/embargo-1.md
```diff
@@ -0,0 +1,95 @@
...
+
+There is no AWS S3 API for quick move (renaming) of a key within a bucket.
+As a result, "move" operation can be **very** expensive in terms of traffic (thus possibly cost if not "sponsored"), time it would take for transfer, implementation e.g. handling of possible errors during transfer, etc
+
+### Publishing a dandiset is enabled only for Open dandisets
+
+### Use-cases to fulfill
+
+- UC-Collaboration: authorized users of emabrgoed dataset should be able to manipulate it as easily as if it was a public dataset.
+- UC-Review: ability to make embargoed dataset available to other users "read-only".
```
bendichter, line 20: I agree that we should enable read-only access to datasets. I don't know if I would call it "Review." I think this could be also used in a user-case where one lab is collecting data and using DANDI to share it with a theorist in another lab.
```suggestion
- UC-Read-only: ability to make embargoed dandiset available to other users "read-only".
```

doc/design/embargo-1.md
```diff
@@ -0,0 +1,95 @@
+# Embargo support
+
+## Principles
+
```
bendichter, line 4: When a user creates a dandiset, they have the option to mark it as "public" or "embargoed". There is some help text that explains these options, either as a hover-hint or some other way. The text for "public" would indicate that the dandiset would be visible to anyone. 

If "embargoed" is selected, a page would be displayed that would explain the rules of embargoed data. The user is required to check "I agree" before proceeding. The message would indicate that this option is only available to BRAIN Initiative grants. The landing page of the dandiset would be visible to everyone but the underlying data would not be. As per recent NIH policy, the user will be required to open the dandiset upon the first publication of the data, or when the funding for the grant runs out, whichever comes first. The user would be required to enter the grant or award number, and we would ideally validate that this is an active award. We will be able to determine when the grant expires and we will automatically open the dandiset if it has not already been opened. It is expected that each grant have at least one of these embargoed dandisets. Since it is not always clear from the outset how the data from a grant will be divided into publications, it is expected that several groups may start with a single dandiset.

For embargoed data the owner would have options to add user permissions, with an interface similar to Google Docs:
* add read-only. This would be used for collaborations with team members who do not need to submit data but only to read it. This would also be useful for reviewers of manuscripts who need to see the data as part of the review process. Read access extends to downloading data with the python API, and reading data directly using s3 streaming. A DANDI Hub user will automatically use the credentials from the dandi hub account they are currently using.
* add collaborator. This would give them write permissions
* add owner. This would give them read and write permissions and the ability to add users, including owners.

Permissions are all my dandiset, not by asset. If a user needs to set different permissions for different assets within the dandiset, they will need to create a new dandiset and put the assets in this separate place.

When a there is a new publication, the user is expected to create a new dandiset for this publication. There should be an online interface for copying assets from an existing dandiset to this one.

doc/design/embargo-1.md
yarikoptic: A nice valid use case.  I just just trying to come up with Use Case "codenames" which are depicting the purpose (for your case would be something like `UC-Consumer` more than a "mechanism" (read-only).  I am not married to those names though, so can accept your suggestion or we could extend it to
```suggestion
- UC-Review, UC-Consumer: ability to make embargoed dataset available to other users "read-only".
```
or alike:

doc/design/embargo-1.md
yarikoptic: Thank you @bendichter !

Re use of "public". To some degree, with making those dandiset-level metadata available, they already become "public" under an obscure notion of "public" many archives use -- people could see them and request access to their data.  That is why I followed the terminology we have in the dandischema already, and called them `"open"`.

The other notes -- might better reside in more specific sections of the design (e.g. discussion of the web ui interactions) . 

"Permissions are all by dandiset, not by asset" - is a great one to go here in these high level principles.

Great point on separating "owner" from "collaborator", but somewhat orthogonal to "embargo" IMHO. ATM I believe we have no "owner"-vs-"collaborator" distinction. We could mix it in here, or postpone (to concentrate on embargo aspects) for a separate discussion/implementation.

doc/design/embargo-1.md
```diff
@@ -0,0 +1,119 @@
...
+TODO: ATM [dandischema](https://github.com/dandi/dandischema/blob/master/dandischema/model_types.py) defines AccessTypes: `Open`, `Embargoed`, `Restricted`.  I am not including `Restricted` (embargoed without temporal or publish restrictions?) aside.
+
+### dandi-api
+
+#### Dandiset
+
+- Create (`POST /dandisets/`) endpoint gains optional
+  - `accessType` field to enter, defaulting to `open`
+  - `embargoedUntil` field to be specified, only if `accessType == 'embargoed'` and error out if specified for `open`
+- DB for dandiset gains `accessType` column to avoid lookup in metadata
```
waxlamp, line 76: We don't actually need to include this in the design doc, since the Django model layer will take care of this for us.

doc/design/embargo-1.md
```diff
@@ -0,0 +1,119 @@
...
+
+### dandi-cli
+
+- no changes
+
+
+### DataLad dandisets (https://github.com/dandisets/)
+
+- embargoed dandisets are made "Private" on github
+-
```
waxlamp, line 113:
```suggestion
```

waxlamp (commented):
Looks like a good overall design to me.

doc/design/embargo-1.md
```diff
@@ -0,0 +1,119 @@
...
+
+If "embargoed" is selected, a page would be displayed that would explain the rules of embargoed data. The user is required to check "I agree" before proceeding. There is also a "back" option. On this page, the user is required to enter the grant or award number, and we would ideally validate that this is an active award. The message would indicate that this option is only available to BRAIN Initiative grants, and that the landing page of the dandiset would be visible to everyone but the underlying assets would not be. As per recent NIH policy, the user will be required to make the dandiset "open" upon the first publication of the data, or when the funding for the grant runs out, whichever comes first. We will be able to determine when the grant expires and we will automatically open the dandiset if it has not already been opened. Using the grant number, we determine the embargoedUntil date and add it automatically as metadata which is visible publicly on the dandiset landing page.
```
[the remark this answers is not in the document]
bendichter: I mean programatically. We'd have a list or an API where we can query active BRAIN award numbers. That way we could also get the end date of the award and know exactly when we would need to lift the embargo. This is just an idea, and not mission critical.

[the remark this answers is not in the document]
bendichter: My understanding is that the embargo goes with [this notice](https://grants.nih.gov/grants/guide/notice-files/NOT-MH-19-010.html), which applies specifically to BRAIN projects that applied after March 1, 2020. More specifically those are the projects that are required to be put on the archive prior to publication. Other NIH-funded work is not required to be uploaded to an archive prior to publication.

doc/design/embargo-1.md
```diff
@@ -0,0 +1,119 @@
...
+
+It is expected that each grant have at least one of these embargoed dandisets. Since it is not always clear from the outset how the data from a grant will be divided into publications, it is expected that several groups may start with a single dandiset.
+
+For embargoed data the owner would have options to add user permissions, with an interface similar to Google Docs:
+
+* add read-only member. This would be used for collaborators with team members who do not need to submit data but only to read it. This would also be useful for reviewers of manuscripts who need to see the data as part of the review process. Read access allows a user to view available assets, and download them with the web API, the Python API, and reading data directly using s3 streaming. A DANDI Hub user will automatically use the credentials from the dandi hub account they are currently using.
+* add owner. This would give them read and write permissions and the ability to add users, including owners.
```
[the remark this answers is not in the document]
bendichter: Yes, I had originally added a third mode that was read and write but not owner, but I was convinced that it was not necessary at this stage.

yarikoptic:
we now have
- https://github.com/dandi/dandi-archive//blob/HEAD/doc/design/embargo-mvp.md
- https://github.com/dandi/dandi-archive//blob/HEAD/doc/design/embargo-full.md
with WiP toward their implementation, so this PR could RiP

PR #511, opened 2021-09-08
add ux suggestions

bendichter:


yarikoptic:
I still think that some aspects should be split out into corresponding subsections etc, but for now I just cherrypicked this commit into the branch of #510 (could have one it via web ui I guess as well, heh), so it is there verbatim now.

waxlamp:
@yarikoptic, for future reference, it might have been better to simply merge this PR (since it was targeted to #510).

yarikoptic:
@bendichter remained the author since I cherry picked the commit. So that effect is exactly the same as if I hit here "Rebase and merge" or whatever that action is. Showing "Merged" indeed would have been nice.
For future reference, such additions should ideally be just suggestions within PRs, ideally (whenever possible) concise to facilitate review/tune up - not sure if we should breed PRs to PRs since even though I could suggest tune up to individual sentences here, I cannot anyways to do any "moving" etc in suggestions to non-diff inflicted part of the text.

waxlamp:
> @bendichter remained the author since I cherry picked the commit. So that effect is exactly the same as if I hit here "Rebase and merge" or whatever that action is. Showing "Merged" indeed would have been nice (and I thought that github would be smart enough to identify cherry-picked commit and mark this PR as "Merged" -- I was wrong).

Good point about cherry picking, thanks for setting me straight.

> For future reference, such additions should ideally be just suggestions within PRs, ideally (whenever possible) concise to facilitate review/tune up - not sure if we should breed PRs to PRs since even though I could suggest tune up to individual sentences here, I cannot anyways to do any "moving" etc in suggestions to non-diff inflicted part of the text.

Agreed, but if someone wants to make a bigger change, then a PR targeted to a PR is a fine mechanism for doing so. Generally, for small additions, etc., suggestions within a PR review work, but for a bigger thing like this, I see why Ben opted for a pull request.

PR #514, opened 2021-09-10
Use GH Actions instead of CircleCI for CI

mvandenburgh:
Swaps out the CircleCI config file for the girder4-cookiecutter GitHub Action yaml.

@dchiquito to my knowledge, CircleCI is only being used for legacy reasons. If there's some other reason, feel free to just close this.


waxlamp:
If these give the same results as the old tests, then I think it's actually better to be using GitHub Actions: they're better integrated with GH (obviously), and it's the standard choice for other greenfield Girder4 projects.

But we can wait to merge this until Daniel has a chance to look.

dchiquito (approved):

mvandenburgh:
@dchiquito I'm marking this as draft to prevent merging. Something weird is happening where the tests just freeze in the middle of execution with no error messages, and eventually the runner times out after 5+ hours.

mvandenburgh:
Well, removing the `DJANGO_CELERY_BROKER_URL` environment variable seems to have fixed it. That env var must have had the wrong url for rabbitmq.

PR #517, opened 2021-09-14
Management script to copy dandisets from another api deployment

### LINKED ISSUE #516: Create management script to copy dandisets from another deployment
A script like this would be helpful for viewing production dandisets in staging, for example. Only the metadata for each dandiset should be copied, since we don't want to be transferring terabytes of asset data into our staging bucket.

mvandenburgh:
Closes #516. This management script can be used to copy dandisets from any deployment of dandi-api into the environment it's being run from.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,10 +27,22 @@ This script is a simple way to get test data into your DB without having to use 
 
 ### import_dandisets
 ```
-python manage.py import_dandisets [API_URL]
+python manage.py import_dandisets [API_URL] --all
 ```
 
 This imports all dandisets (versions + metadata only, no assets) from the dandi-api deployment
 living at `API_URL`. For example, to import all dandisets from the production server into your
 local dev environment, run `python manage.py import_dandisets http://api.dandiarchive.org` from
 your local terminal.
+
+```
+python manage.py import_dandisets [API_URL] --identifier 000005
+```
+
+This imports dandiset 000005 from `API_URL` into your local dev environment.
+
+```
+python manage.py import_dandisets [API_URL] --identifier 000005 --replace 000001
+```
+
+This imports dandiset 000005 from `API_URL` and replaces 000001 with it in your local dev environment.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -24,3 +24,13 @@ python manage.py create_dev_dandiset --owner your.email@email.com --name My Dumm
 This creates a dummy dandiset with valid metadata and a single dummy asset.
 The dandiset should be valid and publishable out of the box.
 This script is a simple way to get test data into your DB without having to use dandi-cli.
+
+### import_dandisets
+```
+python manage.py import_dandisets [API_URL]
+```
+
+This imports all dandisets (versions + metadata only, no assets) from the dandi-api deployment
+living at `API_URL`. For example, to import all dandisets from the production server into your
+local dev environment, run `python manage.py import_dandisets http://api.dandiarchive.org` from
+your local terminal.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -33,25 +33,28 @@ python manage.py import_dandisets [API_URL] --all
 This imports all dandisets (versions + metadata only, no assets) from the dandi-api deployment
 living at `API_URL`. For example, to import all dandisets from the production server into your
 local dev environment, run `python manage.py import_dandisets http://api.dandiarchive.org` from
-your local terminal.
+your local terminal. Note that if a dandiset with the same identifier as the one being imported
+already exists, that dandiset will not be imported.
+
+```
+python manage.py import_dandisets [API_URL] --all --replace
+```
+
+Same as the previous example, except if a dandiset with the same identifier as the one being imported
+already exists, the existing dandiset will be replaced with the one being imported.
 
 ```
 python manage.py import_dandisets [API_URL] --all --offset 100000
 ```
 
 This imports all dandisets (versions + metadata only, no assets) from the dandi-api deployment
-living at `API_URL` and offsets their identifiers by 100000. In this case, if the most recent dandiset
-in your local deployment has an identifier of 0000020, the first imported dandiset would have an
-identifier of 100020, second would have an identifier of 100021, etc.
+living at `API_URL` and offsets their identifiers by 100000. This is helpful if you want to import
+a dandiset that has the same identifier as one already in your database.
 
 ```
 python manage.py import_dandisets [API_URL] --identifier 000005
 ```
 
-This imports dandiset 000005 from `API_URL` into your local dev environment.
-
-```
-python manage.py import_dandisets [API_URL] --identifier 000005 --replace 000001
-```
-
-This imports dandiset 000005 from `API_URL` and replaces 000001 with it in your local dev environment.
+This imports dandiset 000005 from `API_URL` into your local dev environment. Note that if there is already
+a dandiset with an identifier of 000005, nothing will happen. Use the --replace flag to have the script
+overwrite it instead if desired.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -36,6 +36,15 @@ local dev environment, run `python manage.py import_dandisets http://api.dandiar
 your local terminal.
 
 ```
+python manage.py import_dandisets [API_URL] --all --offset 100000
+```
+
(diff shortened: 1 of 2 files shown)


dandiapi/api/management/commands/import_dandisets.py
```diff
@@ -0,0 +1,82 @@
...
+@transaction.atomic
+def import_dandisets_from_response(api_url: str, dandiset_api_response: dict):
+    """Import dandisets given a response from /api/dandisets/."""
+    for result in dandiset_api_response['results']:
+
+        # Check if dandiset with this name already exists. If it does, skip it.
+        # I can't think of a better way to uniquely identify dandisets
+        # without being able to use identifiers.
+        existing_dandiset = Version.objects.filter(name=result['draft_version']['name'])
+        if existing_dandiset.first():
```
dchiquito, line 49:
```suggestion
        if existing_dandiset.exists():
```
`exists` is generally preferred to `first`, since postgres can run optimize it a bit more. However, in this case it probably doesn't make a difference, or might even slow it down, since using `first` will cache the value for later. No change necessary, just FYI

dchiquito (commented):
As I read it, this script would check against an API server and pull down any new dandisets with new names that don't exist locally. This is useful specifically for getting a lot of real metadata into a local environment quickly.

However, this script doesn't offer a lot of flexibility. I generally only care about a few dandisets, not the entire contents of an archive. I don't see a way to re-download a fresh copy of a dandiset, which is very plausibly something we would want since metadata gets updated/migrated pretty frequently. It assigns each dandiset a new identifier, which makes them harder to match up later. Finally, it doesn't offer a way to download assets (intentionally, I admit), which I have wanted in the past.

The meat of the download logic looks good, but I think as the command interface is written it only solves a small slice of the problems it could. These are things I would want from a download script, in roughly the order I care about:

- Downloading a single dandiset. If I'm trying to solve a problem with specific metadata, I only want to download the problem dandiset.
- An option to replace a local dandiset + ID with the remote data. Delete the dandiset that was using that ID and replace it with the remote contents from scratch. 
- An option to download some or all assets (default is off to prevent downloading too much accidentally)
- Download all the dandisets. When I'm initially setting up a new dev environment it would be easier to have one command to run rather than picking out some samples, but that's less important to me than downloading individually.

I'm OK with merging this script right now if it would be useful right away, but I think it could be substantially more generally useful than it is now. Prettying it up could be part of this PR or later work.

dandiapi/api/management/commands/import_dandisets.py
waxlamp: I would say favor clarity over possible optimization (that is, use `exists()` here).

mvandenburgh:
> * Downloading a single dandiset. If I'm trying to solve a problem with specific metadata, I only want to download the problem dandiset.
> * An option to replace a local dandiset + ID with the remote data. Delete the dandiset that was using that ID and replace it with the remote contents from scratch.
> * An option to download some or all assets (default is off to prevent downloading too much accidentally)
> * Download all the dandisets. When I'm initially setting up a new dev environment it would be easier to have one command to run rather than picking out some samples, but that's less important to me than downloading individually.
> 
> I'm OK with merging this script right now if it would be useful right away, but I think it could be substantially more generally useful than it is now. Prettying it up could be part of this PR or later work.

I implemented all of these except for asset downloading, can you take a lot @dchiquito?. Since this script will be immediately useful to get the prod data into staging to test the DLP redesign, I'd like to merge this as is and address asset download in a future PR.

dchiquito (approved):
lgtm

mvandenburgh:
To test the production => staging copy:

mvandenburgh:
To test this:
- Drop/reset your local DB
- Run `python manage.py import_dandisets http://api-staging.dandiarchive.org --all` to import all staging dandisets (i.e. to make your local environment a clone of staging)
- Run `ALTER SEQUENCE api_dandiset_id_seq RESTART WITH 100000;` on your local postgres terminal to configure the postgres counter like it is in staging now
- Run `python manage.py import_dandisets http://api.dandiarchive.org --all --offset 100000`. This will increment the dandiset identifier postgres counter by 100000 (so it will be 200000), import all the production data, and then reset the postgres counter back to what it was.

waxlamp:
> * Run `ALTER SEQUENCE api_dandiset_id_seq RESTART WITH 100000;` on your local postgres terminal to configure the postgres counter like it is in staging now

I think this should be wrapped in a maintenance script. While I think it's good for devs to understand how to use psql et al., I also think capturing this type of mutating action in the same set of maintenance scripts as the other tasks we do would go a long way towards a notion of strict reproducibility and avoiding one-off errors in issuing SQL commands directly.

DEVELOPMENT.md
```diff
@@ -24,3 +24,34 @@ python manage.py create_dev_dandiset --owner your.email@email.com --name My Dumm
...
 This script is a simple way to get test data into your DB without having to use dandi-cli.
+
+### import_dandisets
+```
+python manage.py import_dandisets [API_URL] --all
+```
+
+This imports all dandisets (versions + metadata only, no assets) from the dandi-api deployment
+living at `API_URL`. For example, to import all dandisets from the production server into your
+local dev environment, run `python manage.py import_dandisets http://api.dandiarchive.org` from
```
dchiquito, line 35:
```suggestion
local dev environment, run `python manage.py import_dandisets https://api.dandiarchive.org` from
```
Is this not HTTPS?

dandiapi/api/management/commands/import_dandisets.py
```diff
@@ -0,0 +1,137 @@
...
+@click.option('--all', is_flag=True, required=False, help='Download all dandisets.')
+@click.option('--identifier', required=False, help='The identifier of the dandiset to import.')
+@click.option('--replace', default=None, help='The dandiset to replace with the imported one.')
+@click.option('--offset', default=0, help="Offset to add to each new dandiset's identifier.")
+@transaction.atomic
+def import_dandisets(api_url: str, all: bool, identifier: str, replace: str, offset=0):
+    """Click command to import dandisets from another server deployment."""
+    # Save the current postgres sequence value and set it with the offset
+    original_sequence_value = get_sequence_value()
+    set_sequence_value(original_sequence_value + offset)
```
dchiquito, line 115: This is not something I thought we would be including in the script. I thought that `offset` would default to `200000`, and would be added individually to each dandiset identifier as it was imported. If you only import dandiset `000036` from production, this script would save that new dandiset at `200036`. 

Regardless, if I used this script to download a single dandiset metadata to reproduce something locally or whatever, I would certainly not expect it to reset my dandiset table sequence.

dchiquito (changes requested):
I would say that we don't need to mention or perform any sequence changes at all. We did it manually in staging once because we want a specific behavior in staging, I don't see any reason to automate that.

The only reason to put it in a script somewhere would be if we want dev's locally created dandisets to also start at 100000, in which case I agree that it should go in a setup script.

dandiapi/api/management/commands/import_dandisets.py
mvandenburgh: > This is not something I thought we would be including in the script. I thought that `offset` would default to `200000`, and would be added individually to each dandiset identifier as it was imported. If you only import dandiset `000036` from production, this script would save that new dandiset at `200036`.

I guess I was thinking about it differently, but your way seems more logical and way easier. I'll go ahead and implement it that way and remove those SQL queries

> Doesn't the script reset it to its original value at the end? Though your approach avoids the problem of the script crashing before it can do so, leaving the DB in the wrong state.

FWIW, yes it does reset it to the original value at the end. And there's no possibility that the DB would end up in a wrong state since the entire script is wrapped in a transaction (see `transaction.atomic` decorator).

mvandenburgh:
@dchiquito I reworked the script and updated the examples in `DEVELOPMENT.md` to show each combination of cli options/flags.

dandiapi/api/management/commands/import_dandisets.py
waxlamp: > > Doesn't the script reset it to its original value at the end? Though your approach avoids the problem of the script crashing before it can do so, leaving the DB in the wrong state.
> 
> FWIW, yes it does reset it to the original value at the end. And there's no possibility that the DB would end up in a wrong state since the entire script is wrapped in a transaction (see `transaction.atomic` decorator).

Ah nice touch, I had missed that!

dchiquito (approved):
LGTM

PR #520, opened 2021-09-15
Set `null=True` + `blank=True` for `validation_errors`

mvandenburgh:
Valid versions should be able to set their `validation_errors` field to null.

Setting `blank=True` fixes an issue where valid versions (i.e. versions with an empty `validation_errors` list) are unable to be saved in the Django admin without populating the `validation_errors` list first.


dchiquito (approved):

dchiquito (changes requested):
There is also a `validation_errors` on Asset. Can you make the same change there and roll that into the same migration?

dchiquito (approved):

PR #528, opened 2021-09-21
Add script to delete objects from versioned bucket

dchiquito:
This script is included at the top level instead of as a manage.py
command because it has no dependency on Django.

It's definitely on the slow side since it has to make a request for every single delete, but I could not find a better alternative.
I'm currently using the script to delete `largeuploads` and `girder-assetstore` and it's working great.


waxlamp (approved):
Instead of right at the top level, could we put this in a separate directory for this sort of out-of-band task? Maybe `maintenance/`?

dchiquito:
I moved it to `dev`, since I didn't want to make a new directory just for this script.

PR #529, opened 2021-09-22
Scheduled draft manifests

### LINKED ISSUE #522: Publish draft yaml files nightly

dchiquito:
* Add `scheduled_tasks.py` to house all our scheduled tasks
* Add a scheduled task that writes all draft manifests every day at 1:00 AM
* Use the `-B` option to start celery beat in the worker processes. The caveats to this are documented in commit history and in `Procfile`.
* Set `manifestLocation` to the real manifest location for draft metadata, instead of the workaround we were using while there was no manifest.

Closes #522.

### DIFF
--- a/dandiapi/api/manifests.py
+++ b/dandiapi/api/manifests.py
@@ -47,13 +47,6 @@ def s3_url(path: str):
 
 def manifest_location(version: Version):
     """Calculate the manifestLocation field for a Version."""
-    if version.version == 'draft':
-        return [
-            (
-                f'https://api.dandiarchive.org/api/dandisets/{version.dandiset.identifier}'
-                f'/versions/draft/assets/'
-            )
-        ]
     return [s3_url(assets_yaml_path(version))]
 
 
--- a/dandiapi/api/manifests.py
+++ b/dandiapi/api/manifests.py
@@ -47,6 +47,13 @@ def s3_url(path: str):
 
 def manifest_location(version: Version):
     """Calculate the manifestLocation field for a Version."""
+    if version.version == 'draft':
+        return [
+            (
+                f'https://api.dandiarchive.org/api/dandisets/{version.dandiset.identifier}'
+                f'/versions/draft/assets/'
+            )
+        ]
     return [s3_url(assets_yaml_path(version))]
 
 
--- a/dandiapi/api/scheduled_tasks.py
+++ b/dandiapi/api/scheduled_tasks.py
@@ -1,0 +1,32 @@
+"""
+Define and register any scheduled celery tasks.
+
+This module is imported from celery.py in a post-app-load hook.
+"""
+
+from celery import shared_task
+from celery.schedules import crontab
+from celery.utils.log import get_task_logger
+from django.db.transaction import atomic
+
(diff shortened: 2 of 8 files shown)


dandiapi/api/manifests.py
```diff
@@ -47,13 +47,6 @@ def s3_url(path: str):
 
 def manifest_location(version: Version):
     """Calculate the manifestLocation field for a Version."""
-    if version.version == 'draft':
-        return [
-            (
-                f'https://api.dandiarchive.org/api/dandisets/{version.dandiset.identifier}'
-                f'/versions/draft/assets/'
-            )
-        ]
```
waxlamp, line 56: What happens if you try to get manifest for a draft dandiset whose first 1am yamling hasn't happened yet?

dandiapi/api/tests/test_dandiset.py
```diff
@@ -231,10 +231,7 @@ def test_dandiset_rest_create(api_client, user):
     assert dandiset.draft_version.metadata == {
         **metadata,
         'manifestLocation': [
-            (
-                f'https://api.dandiarchive.org/api/dandisets/{dandiset.identifier}'
-                f'/versions/draft/assets/'
-            )
+            f'http://localhost:9000/test-dandiapi-dandisets/test-prefix/dandisets/{dandiset.identifier}/draft/assets.yaml'  # noqa: E501
```
waxlamp, line 234: Thank you. I find the breaking of long strings just to adhere to E501 to be something of a serious antipattern.

dandiapi/api/scheduled_tasks.py
```diff
@@ -0,0 +1,32 @@
...
+from dandiapi.api.tasks import write_manifest_files
+
+logger = get_task_logger(__name__)
+
+
+@shared_task
+@atomic
+def write_draft_manifest_files():
+    logger.info('Writing manifest files for all draft versions')
+    # TODO Optimize this if it causes too much load in production.
```
waxlamp, line 22: Have you tracked this TODO in your own systems? Out of an abundance of caution (having read the manual page you linked to) I think it would be better if we eventually didn't do this the "wrong" way, so I even think filing an issue just to make this concern global would be a good idea.

waxlamp (approved):
Looks good!

dandiapi/api/manifests.py
satra: also is there a reason not keep this path for this function (since this is what is injected into the landing page metadata), since this will always reflect current state of db, whereas the s3 url will reflect our periodic update.

satra: also the s3 location will always have the manifest written out as assets.yaml/jsonld, hence changing this function i don't think is necessary.

dchiquito: I recalled this being a hack to work around the fact that we didn't have draft assets. If it's actually the preferred value, I will just revert this change.

dandiapi/api/scheduled_tasks.py
dchiquito: https://github.com/dandi/dandi-api/issues/530

PR #535, opened 2021-09-30
Add separate dyno for celery beat

dchiquito:
This lets us scale the number of workers without inadvertently firing multiple beat events.

The downside is that we need a separate dyno just to run chron, but such is life.


mvandenburgh (approved):
The comment above needs to be updated too.

yarikoptic:
And it would take care about automagically scaling down by turning off unused workers?

dchiquito:
@yarikoptic scaling would be done manually in the heroku console as necessary. Right now we can't really scale at all without celery beat triggering an event for every active worker.

waxlamp (approved):

PR #538, opened 2021-09-30
Optimize celery tasks

dchiquito:
* `validate_asset_metadata` has proven to be consistently very quick, so we can simply do it in band and save the overhead of dispatching a task.
* Instead of revalidating Versions every time they are modified in any way (including adding, updating, removing Assets in the Version), we have a scheduled task that checks for any draft Versions with status `PENDING` and validates them then. This effectively debounces Version validation.
* Since we want to revalidate draft Versions basically any time they are changed, that is also a good time to dispatch the draft manifest task, which means we can do away with the 1:00 AM scheduled job and the associated inefficiencies.

This should also fix https://github.com/dandi/dandi-api/issues/537

### DIFF
--- a/dandiapi/api/scheduled_tasks.py
+++ b/dandiapi/api/scheduled_tasks.py
@@ -3,10 +3,11 @@ Define and register any scheduled celery tasks.
 
 This module is imported from celery.py in a post-app-load hook.
 """
+from datetime import timedelta
 
 from celery import shared_task
-from celery.schedules import crontab
 from celery.utils.log import get_task_logger
+from django.conf import settings
 from django.db.transaction import atomic
 
 from dandiapi.api.models import Version
@@ -35,4 +36,7 @@ def validate_draft_version_metadata():
 def register_scheduled_tasks(sender, **kwargs):
     """Register tasks with a celery beat schedule."""
     # Check for any draft versions that need validation every minute
-    sender.add_periodic_task(crontab(), validate_draft_version_metadata.s())
+    sender.add_periodic_task(
+        timedelta(seconds=settings.DANDI_VALIDATION_JOB_INTERVAL),
+        validate_draft_version_metadata.s(),
+    )
--- a/dandiapi/api/scheduled_tasks.py
+++ b/dandiapi/api/scheduled_tasks.py
@@ -17,17 +17,6 @@ logger = get_task_logger(__name__)
 
 @shared_task
 @atomic
-def write_draft_manifest_files():
-    logger.info('Writing manifest files for all draft versions')
-    # TODO Optimize this if it causes too much load in production.
-    # Rewriting every draft manifest every time is guaranteed not to miss any changes,
-    # so just do that for now to avoid the complexity involved with modification dates.
-    for draft_version in Version.objects.filter(version='draft').all():
-        write_manifest_files.delay(draft_version.id)
-
-
-@shared_task
-@atomic
 def validate_draft_version_metadata():
     logger.info('Checking for draft versions that need validation')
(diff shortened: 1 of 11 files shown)


dandiapi/api/scheduled_tasks.py
```diff
@@ -3,30 +3,40 @@
...
-    # Rewriting every draft manifest every time is guaranteed not to miss any changes,
-    # so just do that for now to avoid the complexity involved with modification dates.
-    for draft_version in Version.objects.filter(version='draft').all():
+def validate_draft_version_metadata():
+    logger.info('Checking for draft versions that need validation')
+    # Select only the id of draft versions that have status PENDING
+    pending_draft_versions = (
+        Version.objects.filter(status=Version.Status.PENDING).filter(version='draft').values('id')
+    )
+    logger.info('Found %s versions to validate', pending_draft_versions.count())
```
waxlamp, line 27: It might be good to list out the IDs found as well, perhaps at a more verbose log level.

dandiapi/settings.py
```diff
@@ -61,6 +61,8 @@ def before_binding(configuration: Type[ComposedConfiguration]):
     DANDI_DOI_API_PREFIX = values.Value(environ=True)
     DANDI_DOI_PUBLISH = values.BooleanValue(environ=True, default=False)
 
+    DANDI_VALIDATION_JOB_INTERVAL = values.IntegerValue(environ=True, default=60)
```
waxlamp, line 64: Is that default 60 seconds or 60 minutes?

waxlamp (commented):
This looks great! Leaving this as a comment because I only had very high level feedback.

dandiapi/api/scheduled_tasks.py
dchiquito: They will be logged when the task is picked up off the queue, I don't think it's that useful to log them here.

dandiapi/settings.py
dchiquito: Seconds. I added this so that the dandi-cli tests can use it to set the validation interval to something very short so they don't need to idle for a minute waiting for validation.

dchiquito:
I see no reason why the dandi-cli master branch failures are still there when all the other tests pass, so I'm going to merge this.

mvandenburgh (approved):
LGTM

dchiquito:
But the `master` dandi-cli test run still has 3 errors (`release` has 11), and I don't see anything that distinguishes those 3 tests from the rest of the tests that are now passing. A mystery

yarikoptic:
@jwodder please have a look at those https://github.com/dandi/dandi-api/runs/3760394541?check_suite_focus=true

jwodder:
@dchiquito The tests in question are all failing because the version validation is still "Pending" even after waiting 20 seconds.  How long is the new validation expected to take?

dchiquito:
The changes I made in https://github.com/dandi/dandi-cli/pull/778/files run the job scheduler (that's the `-B` option) and set the interval to 5 seconds. The scheduled job will run every 5 seconds and validating any draft version that is `Pending`, which should take ~1 second.
If the version is still `Pending` after 20 seconds, either the scheduled task isn't running or the queue is backed up. Is there any way to get the logs from the docker-compose fixtures to check?

jwodder:
@dchiquito This PR should let you see the logs: https://github.com/dandi/dandi-api/pull/543

jwodder:
@dchiquito I appear to have just run into this issue on staging while testing some modifications to a script.  Check out the logs for Dandisets 100343 and 100344 for something to investigate.

Issue #539, opened 2021-10-01
Admin pages not loading


dchiquito:
Some admin pages like https://api.dandiarchive.org/admin/api/version/556/change/ fail to load with a Heroku timeout error. I suspect the issue is that the Versions page is trying to show an unpaginated list of all the Assets in the Version, which is problematic for Versions with lots of Assets.

waxlamp:
This should be resolved by #545.

@mvandenburgh, if that is true, can you add a `Closes` statement to that PR?

mvandenburgh:
> I suspect the issue is that the Versions page is trying to show an unpaginated list of all the Assets in the Version, which is problematic for Versions with lots of Assets.

I don't think this is it. Even Versions with zero Assets fail to load, such as https://api.dandiarchive.org/admin/api/version/553/change/

waxlamp:
Hm. What does the server log show?

mvandenburgh:
This was closed in error #658

PR #547, opened 2021-10-04
Limit admin inlines to 10 entries

dchiquito:
Since inlines aren't paginated, inlines would try to load every asset in
a version, which would hit the Heroku timeout when trying to load the
admin page for a version.

Limit admin inlines to 10 entries to avoid this.

#539 it does not fix, there are problematic version with no assets

### DIFF
--- a/dandiapi/api/admin.py
+++ b/dandiapi/api/admin.py
@@ -1,11 +1,22 @@
 from django.contrib import admin
-from girder_utils.admin import ReadonlyTabularInline
+from django.contrib.admin.options import TabularInline
+from django.forms.models import BaseInlineFormSet
 from guardian.admin import GuardedModelAdmin
 
 from dandiapi.api.models import Asset, AssetBlob, Dandiset, Upload, Version
 
 
-class VersionInline(ReadonlyTabularInline):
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
Is there still a way to paginate all such resources?

dchiquito:
There is no convenient mixin/setting to do that. That was my first instinct, but doing it involves basically rolling our own template+form+query params into our own PaginatedAdminInline thing. I think the effort to do that is more than the benefit we get from browsing all objects from the parent page.

Eventually we should probably revisit the usability of the admin console, but so far I've had much more success using `shell_plus` to poke around the DB.

mvandenburgh:
> There is no convenient mixin/setting to do that. That was my first instinct, but doing it involves basically rolling our own template+form+query params into our own PaginatedAdminInline thing. I think the effort to do that is more than the benefit we get from browsing all objects from the parent page.

https://github.com/shinneider/django-admin-inline-paginator we could also look into using this.

waxlamp:
If this solves the immediate problem, then please go ahead and merge (after Mike gives his review). But let's file a new issue to investigate Mike's proposed library.

mvandenburgh (approved):

PR #551, opened 2021-10-05
Checksum queue

### LINKED ISSUE #542: Investigate Celery priority queues
Currently, we have one big queue that we throw all of our tasks into, and all of our workers pull from this queue. Hypothetically, we could get into a situation during a large upload where all of the workers are busy handling very large checksum calculations, leaving none of them free to deal with the scheduled version metadata validation tasks and manifest writing tasks that would pile up.

We could sort all the fast running bookkeeping tasks into one queue and have one worker dedicated to processing that, then put all the slow, heavy, non-time sensitive tasks (checksum calculation) into a different queue that has a separate stack of workers that can be scaled as necessary.

We could also implement some kind of priority queueing (rabbitmq supports this) so that lighter tasks are always done before heavy ones. This isn't guaranteed to work (workers could all be busy calculating the checksums of enormous files, the priority only matters when a worker is picking up a new task), but it would go a long way toward mitigating it I think.

See https://docs.celeryproject.org/en/stable/userguide/routing.html

dchiquito:
* SHA256 calculations now go to their own queue, `calculate_sha256`. Every other task still goes to the default queue, `celery`.
* The dev environment still pulls from both queues.
* In production there is one worker that handles checksums only, and one worker that handles every other task + running celery beat. The only tasks that eat up worker time are checksums, so we can now scale that separately without having to worry about scaling the high priority worker.

Fixes #542 

dandi-cli tests will break until https://github.com/dandi/dandi-cli/pull/787

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,10 @@
 release: ./manage.py migrate
 web: gunicorn --bind 0.0.0.0:$PORT dandiapi.wsgi
-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO
-celery-beat: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery beat --loglevel INFO
+# celery-beat: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery beat --loglevel INFO
+# Rather than using a dedicated worker for Celery Beat, we simply use the -B option on the priority task worker.
+# This means that we cannot safely scale up the number of priority-workers without Celery Beat triggering multiple events.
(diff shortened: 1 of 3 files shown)


mvandenburgh (approved):
LGTM

jwodder:
@dchiquito Is the lack of this PR likely to be the cause of the "Digest is missing dandi-etag or sha256 keys." errors I mentioned [here](https://github.com/dandi/dandi-api/issues/548)?  I just ran into a bunch of those errors [in this run](https://github.com/dandi/dandi-cli/runs/3816582731?check_suite_focus=true).

PR #558, opened 2021-10-07
Don't validate Assets with no sha256

dchiquito:
Previously, we validated Asset metadata immediately after
creating/updating an Asset, even if the sha256 checksum had not been
calculated yet for the blob. This resulted in a validation error related
to a missing digest, but that was shortly addressed when the checksum
finished calculation and revalidated the Asset.

However, this system had a race condition where the Asset creation
validation and the checksum completion validation would run
simultaneously. The checksum validation would save first with the valid
results, then the creation validation would overwrite those results with
erroneous validation errors.

Fix the race condition by not validating after creation/update when the
blob sha256 checksum is not yet calculated.

attn @satra

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -197,6 +197,7 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
             path=path,
             blob=asset_blob,
             metadata=metadata,
+            status=Asset.Status.PENDING,
         )
         asset.save()
         version.assets.add(asset)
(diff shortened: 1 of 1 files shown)


satra (commented):
in general looks good to me. 

does something else change the status and trigger validation ?

i would probably suggest an internal dashboard that lists the age of pending assets and if any request is over say 24 hours old, we get a ping.

also we should ensure that the cli validates after upload, so it can inform the user if something looked off.

satra:
> I don't think the CLI can "validate after upload",

assets can be validated with "Asset" instead of "PublishedAsset", and this is where validation can still be done by the CLI post upload to check for other issues.

PR #560, opened 2021-10-11
Add garbage collection design doc

### LINKED ISSUE #177: Garbage collection
I am adding to migration target since we better iron it out before switching to dandi-api:   bugs in GC can lead to data loss thus IMHO we first need make sure (as with extensive unit-testing and user-testing) that it works reliably before deploying it.

Initial design sketch on garbage collection is present within https://github.com/dandi/dandi-api/pull/150/files#diff-c96d4444d1714a52d5d08dd92d94919393a7db8ded038aa84f02ba1075d2c25eR37  but I think it is worth removing it from that PR and starting a new dedicated one.

I see following targets for GC

- uploads: not sure how if we could query S3 for some timestamp of the last activity for the upload urls
- blobs: any blob not linked to an asset is subject to GC 
  - GCing of a blob (blob_id) should also GC the actual key file in the store.  We have 1-to-1 relationship ATM
- assets: any asset not linked to any (released or draft) version of a dandiset

Additional aspects:
 
- Might need DB-level locking unless we `touch` any blob or asset upon "being queried", since otherwise we might GC a blob in the middle of an asset being "minted" for an existing blob; very less likely but I guess could happen for assets in GC of an asset is triggered while we are "modifying" it and creating a new asset
- For any of those we might wan to allow for some "expiry" duration, so even if not "linked", but newer than e.g. a week (or 24h) -- keep around;  If we add `touch` timestamping, this would allow to avoid locking (but would be costly for DB)
 - Bucket is versioned (yeap -- on my insistence), so there might be need for a 2nd stage GCing picking up elderly/stale keys in S3, unless removing of a key would explicitly take care about removing it (not just adding DeleteMarker)

### LINKED ISSUE #523: Write a design doc for garbage collection
This encompasses tools for performing GC manually (by the admins) and also policies for automated GC including warnings to owners of assets not associated to a dandiset, draft dandisets that have not been interacted with in a long time, etc.

See #457 for a start to this, plus some discussion that should be folded into the design doc.

dchiquito:
A design doc for garbage collection policies.

Fixes #523 
Fixes #177

### DIFF
--- a/doc/design/garbage-collection-1.md
+++ b/doc/design/garbage-collection-1.md
@@ -2,6 +2,7 @@
 There are several kinds of data that exist without being included in a dandiset, meaning they are simply taking up space in the system.
 These data should be cleaned up automatically after some time.
 
+# Data types
 ## Uploads
 As part of the multipart upload process, an `Upload` database record is created to track the status of the upload.
 `Upload`s can be orphaned when a client aborts the upload process before creating an `AssetBlob` from the `Upload`.
@@ -40,3 +41,12 @@ Identifying problematic objects requires iterating over every blob in the archiv
 Orphaned S3 objects should not be cleaned up automatically, as they are likely a symptom of a bug.
 Instead, they should be investigated manually as they occurr.
 There should be a scheduled job that runs and reports orphaned blobs somewhere, either throwing an exception to be reported in Sentry, or be saving some data somewhere to be displayed in the admin data dashboard.
+
+# Implementation
+All of these garbage collection operations should be added to the `collect_garbage` script.
+
+The actual implementation of these operations should be added to `dandiapi.api.garbage`.
+
+After thoroughly testing the script by hand in staging/production, a Celery Beat task should be created that checks for stale data nightly.
+
+All of these numbers should be available in the admin data dashboard so that we can independently verify that the scheduled task is working correctly.
--- a/doc/design/garbage-collection-1.md
+++ b/doc/design/garbage-collection-1.md
@@ -32,15 +32,15 @@ Unreferenced `Asset`s should be removed after 30 days, just in case a user accid
 
 ## S3 objects
 Hypothetically, we might at some point encounter a desynchronization between objects stored in S3 and references stored in the database.
-Specifically, we should clean up any blobs stored in S3 that do not have corresponding `AssetBlob`s.
-Dealing with `AssetBlob`s that don't have an S3 object is likely to be a case of data loss, and isn't really a garbage collection operation.
+We should identify up any objects stored in S3 that do not have corresponding `AssetBlob`s.
+We should also identify and `AssetBlob`s that do not have corresponding objects in S3.
 
 Deleting the blobs is an easy API call.
 Identifying problematic objects requires iterating over every blob in the archive, which would take a while.
 
-Orphaned S3 objects should not be cleaned up automatically, as they are likely a symptom of a bug.
-Instead, they should be investigated manually as they occurr.
-There should be a scheduled job that runs and reports orphaned blobs somewhere, either throwing an exception to be reported in Sentry, or be saving some data somewhere to be displayed in the admin data dashboard.
+Orphaned data should not be cleaned up automatically, as they are likely a symptom of a bug.
+Instead, they should be investigated manually as they occur.
+There should be a scheduled job that runs and reports orphaned data somewhere, either throwing an exception to be reported in Sentry, or be saving some data somewhere to be displayed in the admin data dashboard.
 
 # Implementation
 All of these garbage collection operations should be added to the `collect_garbage` script.
--- a/doc/design/garbage-collection-1.md
+++ b/doc/design/garbage-collection-1.md
@@ -1,0 +1,42 @@
+# Garbage collection
+There are several kinds of data that exist without being included in a dandiset, meaning they are simply taking up space in the system.
+These data should be cleaned up automatically after some time.
+
+## Uploads
+As part of the multipart upload process, an `Upload` database record is created to track the status of the upload.
+`Upload`s can be orphaned when a client aborts the upload process before creating an `AssetBlob` from the `Upload`.
+
+`Uploads` could be abandoned prior to any uploads, partially uploaded, or after upload is complete.
+The preferred way to deal with partial uploads in S3 is to set up a [bucket lifecycle rule](https://aws.amazon.com/blogs/aws/s3-lifecycle-management-update-support-for-multipart-uploads-and-delete-markers/).
+If the object is already uploaded, we can simply delete it from S3.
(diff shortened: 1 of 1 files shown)


doc/design/garbage-collection-1.md
```diff
@@ -0,0 +1,52 @@
...
+
+Unreferenced `AssetBlob`s should be removed after 1 day to prevent malicious actors from storing files in our S3 bucket for any significant length of time.
+
+## Assets
+An `Asset` is simply an `AssetBlob` with a path and some metadata.
+Because `Asset`s are immutable, "modifying" and `Asset` will actually create a new `Asset` object and replace the old `Asset`'s references with the new `Asset`. This results in a substantial number of orphaned `Asset` objects.
+
+Deleting an `Asset` is as simple as deleting the database record, although it might also result in an orphaned `AssetBlob`.
+
+Unreferenced `Asset`s should be removed after 30 days, just in case a user accidentally deletes some metadata that we need to recover.
```
satra, line 31: the reason i suggested 30 days, was that removing an unreferenced asset after 30 days would also mean removing an associated assetblob, unless the assetblob was linked to some other asset. i didn't want to remove the blob earlier. but 30 was just a suggestion, could be 7 days.

this is also partially related to the dandiarchive ui issue that it is currently very easy on the ui side to accidentally delete an asset. so once that is addressed, i expect unreferenced assets to mostly come from cli updates.

doc/design/garbage-collection-1.md
```diff
@@ -0,0 +1,52 @@
...
+
+## S3 objects
+Hypothetically, we might at some point encounter a desynchronization between objects stored in S3 and references stored in the database.
+Specifically, we should clean up any blobs stored in S3 that do not have corresponding `AssetBlob`s.
```
satra, line 35: under what circumstances would this happen. also seems like a data loss, just on the database side.

doc/design/garbage-collection-1.md
```diff
@@ -0,0 +1,52 @@
...
+Dealing with `AssetBlob`s that don't have an S3 object is likely to be a case of data loss, and isn't really a garbage collection operation.
+
+Deleting the blobs is an easy API call.
+Identifying problematic objects requires iterating over every blob in the archive, which would take a while.
```
satra, line 39: we are getting an inventory created by aws dumped into the archive. perhaps that can be leverged.

doc/design/garbage-collection-1.md
dchiquito: My thought was that a DB record with no blob is much more recoverable than a blob with no DB record, since a blob has no metadata about what it was associated with.
This would only even happen as the result of a bug, which is why I recommend not automating deletions here so that we can investigate the underlying bug.

dchiquito: Actually, since I'm recommending not deleting anything here, we might as well look for desynchronizations in both directions.

doc/design/garbage-collection-1.md
```diff
@@ -0,0 +1,52 @@
...
+`Uploads` could be abandoned prior to any uploads, partially uploaded, or after upload is complete.
+The preferred way to deal with partial uploads in S3 is to set up a [bucket lifecycle rule](https://aws.amazon.com/blogs/aws/s3-lifecycle-management-update-support-for-multipart-uploads-and-delete-markers/).
+If the object is already uploaded, we can simply delete it from S3.
+In all cases, we need to delete the database record.
+
+`Upload`s should be deleted whenever the presigned URLs expire, which is currently set in `multipary.py#DandiMultipartMixin` to 7 days.
+
+## AssetBlobs
+After an upload is complete, it is converted into an `AssetBlob`, which is then registered with an `Asset`.
+If a user skips this registration step, the file will be available in S3, but will not appear in any dandisets.
```
waxlamp, line 19: What's the usual step for registering an `AssetBlob`? Why do we allow this step to not be taken?

doc/design/garbage-collection-1.md
waxlamp: I think 30 days would be good because it's a round number that gives us a wider margin to help out users who make a mistake.

@satra, I am curious why faster deletion would be better as usage scales.

doc/design/garbage-collection-1.md
waxlamp: I would recommend not relying on an index that we ourselves create--that becomes a weak point in the GC system if e.g. there is a bug in the index generation procedure.

waxlamp (commented):
Good stuff, just left a couple of questions.

Fleshing out some of the mentions (e.g., lifecycle rules) could be a good idea, but we can also decide on those things at implementation time, or defer further discussion to a future design doc.

doc/design/garbage-collection-1.md
```diff
@@ -0,0 +1,52 @@
...
+## Uploads
+As part of the multipart upload process, an `Upload` database record is created to track the status of the upload.
+`Upload`s can be orphaned when a client aborts the upload process before creating an `AssetBlob` from the `Upload`.
+
+`Uploads` could be abandoned prior to any uploads, partially uploaded, or after upload is complete.
+The preferred way to deal with partial uploads in S3 is to set up a [bucket lifecycle rule](https://aws.amazon.com/blogs/aws/s3-lifecycle-management-update-support-for-multipart-uploads-and-delete-markers/).
+If the object is already uploaded, we can simply delete it from S3.
+In all cases, we need to delete the database record.
+
+`Upload`s should be deleted whenever the presigned URLs expire, which is currently set in `multipary.py#DandiMultipartMixin` to 7 days.
```
yarikoptic, line 15: typo

```suggestion
`Upload`s should be deleted whenever the presigned URLs expire, which is currently set in `multipart.py#DandiMultipartMixin` to 7 days.
```

yarikoptic: minor comment: "7 days" duration should be specified in some constant to be reused in multipart.py and the `gc`

doc/design/garbage-collection-1.md
yarikoptic, line 31: `X days`  after what time(stamp)?  I think we have no time stamp for "asset is now detected orphaned".

dchiquito:
@yarikoptic correct, no more separate metadata tables.

doc/design/garbage-collection-1.md
dchiquito: As far as I can tell the `manifest.json` files in `dandiarchive/dandiarchive/dandiarchive` are only indexing `dandiarchive/dandiarchive/dandiarchive/data`, not `blobs`.

satra: correct. however the info in data is indexing all the files in the archive. all files in there are generated automatically by aws inventry service, which is a logger for changes happening in the bucket at some periodicity (i think set as daily).

PR #568, opened 2021-10-18
User signup questionnaire

mvandenburgh:
There are some slight deviations from the design doc, most of which are technical issues that came up while I was implementing this. I've updated the design doc to account for these changes https://github.com/dandi/dandi-api/commit/8f3078a4f1a8895cfeb57a033992572c882b6c7c


dandiapi/api/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+from dandiapi.api.models.user import UserMetadata
+
+
+class IsApproved(IsAuthenticated):
+    def has_permission(self, request, view):
+        return bool(
+            super().has_permission(request, view)
+            and (
+                request.user.is_superuser
```
mvandenburgh, line 11: Note that this will result in an extra SQL query for every request that hits an endpoint protected by these permission classes. It should be possible to override the current auth backend to use `select_related` to mitigate this. This PR is already massive enough though, so this can postponed to a future PR if we want to do it.

dandiapi/api/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+                or request.user.metadata.status == UserMetadata.Status.APPROVED
```
mvandenburgh, line 12: Note that this will result in an extra SQL query for every request that hits an endpoint protected by these permission classes. It should be possible to override the current auth backend to use `select_related` to mitigate this. This PR is already massive enough though, so this can postponed to a future PR if we want to do it.

dandiapi/api/templates/api/account/base.html
```diff
@@ -0,0 +1,53 @@
...
+    }
+  </style>
+</head>
+
+<body>
+  <nav style="background-color: #f5f5f5;">
+    <div class="nav-wrapper logo-container">
+      <a href="https://dandiarchive.org" class="brand-logo">
+        {% load static %}
+        <img src="{% static 'api/dandi_logo.svg' %}" class="logo">
```
dchiquito, line 42: Could we not link in the logo from https://github.com/dandi/dandi.github.io/blob/master/assets/dandi_logo.svg instead?

PR #571, opened 2021-10-19
Fix failing tests when running tox with docker

mvandenburgh:
Some tests assume minio is running at `localhost:9000`, but this is only true when running tox natively. If running it via docker, minio will be running at `minio:9000`. To fix this, I had the tests retrieve the minio url from the `DJANGO_MINIO_STORAGE_ENDPOINT` environment variable.

### DIFF
--- a/dandiapi/api/tests/test_version.py
+++ b/dandiapi/api/tests/test_version.py
@@ -1,5 +1,4 @@
 from datetime import datetime
-import os
 
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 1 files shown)


dandiapi/api/tests/test_version.py
```diff
@@ -90,8 +91,9 @@ def test_published_version_metadata_computed(published_version: Version):
         **original_metadata,
         'manifestLocation': [
             (
-                f'http://localhost:9000/test-dandiapi-dandisets/test-prefix/dandisets'
-                f'/{published_version.dandiset.identifier}/{published_version.version}/assets.yaml'
+                f'http://{os.getenv("DJANGO_MINIO_STORAGE_ENDPOINT")}/test-dandiapi-dandisets'
```
dchiquito, line 94: This should be available from settings:
```suggestion
                f'http://{settings.MINIO_STORAGE_ENDPOINT}/test-dandiapi-dandisets'
```

dchiquito (changes requested):

dchiquito (approved):
Awesome, I've had trouble reproducing this so I have not been able to solve it yet. Thanks!

PR #572, opened 2021-10-19
Support requesting sorted assets

### LINKED ISSUE #563: Give dandisets_versions_assets_list query parameters for controlling the sort order
Feature request: The `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/` endpoint should take one or more query parameters for controlling the order in which the assets are returned.  To borrow from the GitHub API, the parameters could be `sort=<field name>` and `direction=<"asc" or "desc">`.  Fields that users would likely want to sort on include "created", "modified", and "path".

jwodder:
I don't know how to write a test for this.

Closes #563.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -109,8 +109,6 @@ class AssetFilter(filters.FilterSet):
 
 
 class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewSet):
-    queryset = Asset.objects.all().order_by('created')
-
     permission_classes = [IsAuthenticatedOrReadOnly]
     serializer_class = AssetSerializer
     serializer_detail_class = AssetDetailSerializer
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/asset.py
```diff
@@ -122,6 +120,14 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
     filter_backends = [filters.DjangoFilterBackend]
     filterset_class = AssetFilter
 
+    def get_queryset(self):
+        sortfield = self.request.query_params.get('sort', 'created')
+        if sortfield not in ('created', 'modified', 'path'):
+            sortfield = 'created'
```
waxlamp, line 126: I think this should be an error rather than falling back to `created`.

dandiapi/api/views/asset.py
```diff
@@ -122,6 +120,14 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
...
+        if self.request.query_params.get('direction', None) == 'desc':
```
waxlamp, line 127: Similarly, we should enforce that `direction` is either `asc` or `desc`.

dandiapi/api/views/asset.py
jwodder: Done.

dandiapi/api/views/asset.py
jwodder: Done.

PR #573, opened 2021-10-19
Support requesting sorted assets (take 2)

### LINKED ISSUE #563: Give dandisets_versions_assets_list query parameters for controlling the sort order

jwodder:
This seems to be the standard way to implement this.  With this, sort order can be specified in API requests with the `ordering` query parameter, e.g., `ordering=created` (or `ordering=-created` for reverse order).

Closes #563.

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -206,6 +206,40 @@ def test_asset_rest_list(api_client, version, asset):
     }
 
 
+@pytest.mark.parametrize(
+    'order_param,ordering',
+    [
+        ('created', ['b', 'a', 'c']),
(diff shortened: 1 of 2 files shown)


jwodder:
So this doesn't work.  Should I go back to #572?  Should I follow [these instructions](https://www.django-rest-framework.org/api-guide/filtering/#custom-generic-filtering) and implement a `filter_queryset()`?

PR #574, opened 2021-10-20
3rd draft of zarr design doc

dchiquito:
Everything required for zarr, except publishing.

I've already started implementing this as it is written with no issues.

Outstanding requests from the last PR:
* calling the `dandiarchive/zarr/...` directory in S3 something else, to reflect that we can store any kind of folder structure in it. I think that we are building this for zarr files, and we are only planning on putting zarr files in it. If there are other formats that we want to explicitly include, we can always create new directories for those formats.
* Keeping `.checksum` files in `dandiarchive/zarr/ZARR_ID.checksum/...` instead of `dandiarchive/zarr_checksums/ZARR_ID/`. I like having two directories for two kinds of files, rather than having `dandiarchive/zarr` contain pairs of directories, but I don't feel that strongly.

### DIFF
--- a/doc/design/zarr-support-3.md
+++ b/doc/design/zarr-support-3.md
@@ -21,9 +21,9 @@ The intended use of Zarr archives is primarily to support use cases where data a
 ## API endpoints
 * **POST /api/zarr/**
 
-  Create a new zarr archive.
+  Create a new zarr archive and mint a UUID for it which would become `zarr_id`
 
-  Returns a zarr ID
+  Returns the `zarr_id`
 
 * **GET /api/zarr/{zarr_id}/**
 
@@ -46,6 +46,7 @@ The intended use of Zarr archives is primarily to support use cases where data a
 
   Requires a list of file paths and ETags (md5 checksums).
   The number of files being uploaded must be less than some experimentally defined limit (say ~500).
+  If the limit is exceeded, return response 400.
   This limit should be chosen so that no upload requests can conceivably exceed the Heroku request timeout (30s).
   The file paths may include already uploaded files; this is how updates are done.
 
--- a/doc/design/zarr-support-3.md
+++ b/doc/design/zarr-support-3.md
@@ -73,7 +73,7 @@ The intended use of Zarr archives is primarily to support use cases where data a
   Return an asset ID
 
 When added to a dandiset, zarr archives will appear as a normal `Asset` in all the asset API endpoints.
-There will be a flag `zarr: true` indicating that the asset is a zarr archive.
+The asset metadata will contain the information required to determine if an asset is a normal file blob or a zarr file.
 
 ## Storage implementation
 A zarr archive with a `zarr_id` `c1223302-aff4-44aa-bd4b-952aed997c78` is stored in the public S3 bucket: `dandiarchive/zarr/c1223302-aff4-44aa-bd4b-952aed997c78/...`
--- a/doc/design/zarr-support-3.md
+++ b/doc/design/zarr-support-3.md
@@ -3,14 +3,14 @@ A zarr archive contains directories and zarr files.
 Zarr archives are stored in their entirety in S3.
 Zarr archives are represented by a single Asset through the API.
 A tree hash is computed as the zarr archive is uploaded.
-Each node in the tree hash in the hash is stored in S3 to avoid bloating the DB.
-The intended use of Zarr archives is primarily to support use cases where data are too big to be copied. However, Zarr archives can be small as well.
+The tree hash details (all nodes) are stored on S3 to avoid bloating the DB.
+The intended use of Zarr archives is primarily to support use cases where data are too big to be copied or to be stored in a single file due to available storage limits. However, Zarr archives can be small as well.
 
 # Requirements
 1. Zarr archives are stored in a "directory" in S3.
 1. Each zarr archive corresponds to a single Asset.
 1. The CLI uses some kind of tree hashing scheme to compute a checksum for the entire zarr archive.
-1. The API verifies the checksum _immediately_ after upload.
+1. The API verifies the checksum immediately after upload, so the validation status is available in the response to the upload REST request.
 1. The system can theoretically handle zarr files with ~1 million subfiles, each of size 64 * 64 * 64 bytes ~= 262 kilobytes.
 1. Zarr metadata must be mutable for zarr files in draft dandisets.
 1. Zarr archives must not be copied, as they are too large to reasonably store multiple copies.
@@ -45,6 +45,8 @@ The intended use of Zarr archives is primarily to support use cases where data a
   No more batch uploads are permitted until this one is completed or aborted.
 
   Requires a list of file paths and ETags (md5 checksums).
+  The number of files being uploaded must be less than some experimentally defined limit (say ~500).
+  This limit should be chosen so that no upload requests can conceivably exceed the Heroku request timeout (30s).
   The file paths may include already uploaded files; this is how updates are done.
 
   Returns a list of presigned upload URLs
@@ -64,7 +66,8 @@ The intended use of Zarr archives is primarily to support use cases where data a
 * **DELETE /api/zarr/{zarr_id}/files/**
 
   Deletes zarr files from S3, and updates the tree hash accordingly.
-  Requires a list of file paths
+  Requires a list of file paths.
+  All paths must be present in S3, otherwise return 404 without deleting anything.
 
 * **POST /api/dandisets/{...}/versions/{...}/assets/**
 
--- a/doc/design/zarr-support-3.md
+++ b/doc/design/zarr-support-3.md
@@ -6,10 +6,6 @@ A tree hash is computed as the zarr archive is uploaded.
 Each node in the tree hash in the hash is stored in S3 to avoid bloating the DB.
 Zarr archives as a whole are too big to be copied.
 
-## Publishing support
-We have not yet finalized a plan for publishing zarr archives, so this design is a stopgap measure to support uploading in the meantime.
-Until the plan is finished and implemented, publishing dandisets with zarr archives will simply not be permitted.
-
 # Requirements
 1. Zarr archives are stored in a "directory" in S3.
 1. Each zarr archive corresponds to a single Asset.
@@ -18,7 +14,7 @@ Until the plan is finished and implemented, publishing dandisets with zarr archi
 1. The system can theoretically handle zarr files with ~1 million subfiles, each of size 64 * 64 * 64 bytes ~= 262 kilobytes.
 1. Zarr metadata must be mutable for zarr files in draft dandisets.
 1. Zarr archives must not be copied, as they are too large to reasonably store multiple copies.
-1. Publishing dandisets with zarr archives is not a requirement at this time.
+1. Zarr archives are immutable as long as they are contained by a published dandiset.
 
 # Implementation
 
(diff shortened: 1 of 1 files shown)


doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
+# Zarr Support
+A zarr archive contains directories and zarr files. 
+Zarr archives are stored in their entirety in S3.
+Zarr archives are represented by a single Asset through the API.
+A tree hash is computed as the zarr archive is uploaded.
+Each node in the tree hash in the hash is stored in S3 to avoid bloating the DB.
+Zarr archives as a whole are too big to be copied.
```
satra, line 7:
```suggestion
The intended use of Zarr archives is primarily to support use cases where data are too big to be copied. However, Zarr archives can be small as well.
```

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,142 @@
...
+  Deletes zarr files from S3, and updates the tree hash accordingly.
+  Requires a list of file paths
+
+* **POST /api/dandisets/{...}/versions/{...}/assets/zarr/{zarr_id}/**
+
+  Creates a new Asset that points to the zarr archive.
+  Return an asset ID
+
+When added to a dandiset, zarr archives will appear as a normal `Asset` in all the asset API endpoints.
+There will be a flag `zarr: true` indicating that the asset is a zarr archive.
```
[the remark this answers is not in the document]
satra: there should be info in the content type field in the metadata, which is mandatory for an asset. let's just use that instead of adding another flag. also i would be more concerned in general about what happens, if anything, when there is a mismatch, someone uploads zarr and says its a text file. i don't think there is anything that checks.

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+  Requires a `zarr_id` in the request body instead of a `blob_id`.
+  Return an asset ID
+
+When added to a dandiset, zarr archives will appear as a normal `Asset` in all the asset API endpoints.
+There will be a flag `zarr: true` indicating that the asset is a zarr archive.
+
+## Storage implementation
+A zarr archive with a `zarr_id` `c1223302-aff4-44aa-bd4b-952aed997c78` is stored in the public S3 bucket: `dandiarchive/zarr/c1223302-aff4-44aa-bd4b-952aed997c78/...`
+
+Paths for zarr files should never use a `/` prefix.
```
satra, line 81:
```suggestion
Paths for zarr files, or any assets in the archive, should never use a `/` prefix.
```

doc/design/zarr-support-3.md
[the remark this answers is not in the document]
dchiquito:
```suggestion
When added to a dandiset, zarr archives will appear as a normal `Asset` in all the asset API endpoints.
```

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+  Returns 404 if the path is not present in S3.
+
+* **GET /api/zarr/{zarr_id}/upload/**
+
+  Returns a 204 if an upload is in progress, or a 404 if an upload is not in progress.
+
+* **POST /api/zarr/{zarr_id}/upload/**
+
+  Ask to upload a batch of zarr files as part of the zarr archive.
+  No more batch uploads are permitted until this one is completed or aborted.
```
yarikoptic, line 45: "no more ... permitted" -- what will be the response code to state that?  
can I "reupload" the same batch?

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+* **POST /api/zarr/{zarr_id}/upload/complete/**
+
+  Completes an upload of a batch of zarr files.
+  Fail if any of the checksums do not match.
+  Also, recursively update the tree hash for every parent node for each child.
+
+* **DELETE /api/zarr/{zarr_id}/upload/**
+
+  Cancels a batch upload.
+  Any files already uploaded are deleted.
```
yarikoptic, line 61: might take awhile I guess

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+  A new batch upload can then be started.
```
yarikoptic, line 62: how would they know that already uploaded are already deleted?  DELETE would not return until all deleted?

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+* **GET /api/zarr/{zarr_id}/upload/**
+
+  Returns a 204 if an upload is in progress, or a 404 if an upload is not in progress.
+
+* **POST /api/zarr/{zarr_id}/upload/**
+
+  Ask to upload a batch of zarr files as part of the zarr archive.
+  No more batch uploads are permitted until this one is completed or aborted.
+
+  Requires a list of file paths and ETags (md5 checksums).
```
yarikoptic, line 47: is there any limit on POST size?  do we impose limit on how many files could be in a batch?

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+* **DELETE /api/zarr/{zarr_id}/upload/**
+
+  Cancels a batch upload.
+  Any files already uploaded are deleted.
+  A new batch upload can then be started.
+
+* **DELETE /api/zarr/{zarr_id}/files/**
+
+  Deletes zarr files from S3, and updates the tree hash accordingly.
+  Requires a list of file paths
```
yarikoptic, line 67: if some paths are not found - do we 404 immediately or delete all which we find and then 404? how do we inform which were deleted and which not?

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+
+* **POST /api/dandisets/{...}/versions/{...}/assets/**
+
+  Augments the existing endpoint to create a new Asset that points to the zarr archive.
+  Requires a `zarr_id` in the request body instead of a `blob_id`.
+  Return an asset ID
+
+When added to a dandiset, zarr archives will appear as a normal `Asset` in all the asset API endpoints.
```
yarikoptic, line 75: what would happen on `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/{asset_id}/download/` endpoint for an asset which is zarr file and not blob?

yarikoptic (commented):
some comments/questions on the initial part (until storage implementation)

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+Zarr archives are represented by a single Asset through the API.
+A tree hash is computed as the zarr archive is uploaded.
+Each node in the tree hash in the hash is stored in S3 to avoid bloating the DB.
+The intended use of Zarr archives is primarily to support use cases where data are too big to be copied. However, Zarr archives can be small as well.
+
+# Requirements
+1. Zarr archives are stored in a "directory" in S3.
+1. Each zarr archive corresponds to a single Asset.
+1. The CLI uses some kind of tree hashing scheme to compute a checksum for the entire zarr archive.
+1. The API verifies the checksum _immediately_ after upload.
```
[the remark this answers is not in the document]
dchiquito:
```suggestion
1. The API verifies the checksum immediately after upload, so the validation status is available in the response to the upload REST request.
```
The alternative would be validating asynchronously, which I was told was not an option. The scheme described here requires a request to S3 for each directory in the paths being updated (`a/b/c/d/e/f` = 5 requests), which should be nowhere near the 30s Heroku request timeout.

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+1. Zarr archives are immutable as long as they are contained by a published dandiset.
+
+# Implementation
+
+## API endpoints
+* **POST /api/zarr/**
+
+  Create a new zarr archive.
+
+  Returns a zarr ID
```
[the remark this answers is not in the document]
dchiquito: UUID

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,146 @@
...
+
+* **GET /api/zarr/{zarr_id}/**
+
+  Returns some information about the zarr archive, like name, S3 path+url, and checksum
+
+* **GET /api/zarr/{zarr_id}/{path}**
+
+  When the path exists in S3, returns some information about the zarr file, like name, S3 path+url, and checksum.
+  When the path is a directory in S3, returns a paginated list of files and directories.
```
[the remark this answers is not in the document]
dchiquito: Good point. This could leverage the .checksum file, but we would need to distinguish between files and directories.

doc/design/zarr-support-3.md
dchiquito: 403 sounds right.
You can start a batch upload, complete it, then start exactly the same batch again, and complete that. That would be a no-op, although you would reupload all the data again.
You may not start a batch, then start the same batch again. Any other batch starts will return 403 until the current batch is completed or aborted.

doc/design/zarr-support-3.md
dchiquito:
```suggestion
  Requires a list of file paths and ETags (md5 checksums).
  The number of files being uploaded must be less than some experimentally defined limit (say ~500).
  This limit should be chosen so that no upload requests can conceivably exceed the Heroku request timeout (30s).
```

doc/design/zarr-support-3.md
dchiquito: I think it would be pretty quick, and if worse comes to worse we can do it asynchronously and just freeze the zarr file from new uploads until everything is deleted.
This is one of the factors that needs to be considered when choosing the batch size limit.

doc/design/zarr-support-3.md
dchiquito: That's what I was picturing, yes. If that's not feasible they can simply retry starting the next batch until the asynchronous delete is completed.

doc/design/zarr-support-3.md
dchiquito: 404 immediately would be my assumption.
```suggestion
  Requires a list of file paths.
  All paths must be present in S3, otherwise return 404 without deleting anything.
```

doc/design/zarr-support-3.md
dchiquito: A 400 error, or something to that effect.

doc/design/zarr-support-3.md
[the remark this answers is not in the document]
yarikoptic: Is that implemented so I could check the code? I am still a bit lost since upload is a batch of files which could come for various directories, so it could in principle (e.g. a file per directory) require  traversal of even more "directories" (up the hierarchy) than files.

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,149 @@
...
+1. Zarr archives are immutable as long as they are contained by a published dandiset.
+
+# Implementation
+
+## API endpoints
+* **POST /api/zarr/**
+
+  Create a new zarr archive.
+
+  Returns a zarr ID
```
yarikoptic, line 26: may be then

```suggestion
  Create a new zarr archive and mint a UUID for it which would become `zarr_id`

  Returns the `zarr_id`
```
following [above](https://github.com/dandi/dandi-api/pull/574/files#r758787229)

doc/design/zarr-support-3.md
```diff
@@ -0,0 +1,149 @@
...
+
+  Returns a 204 if an upload is in progress, or a 404 if an upload is not in progress.
+
+* **POST /api/zarr/{zarr_id}/upload/**
+
+  Ask to upload a batch of zarr files as part of the zarr archive.
+  No more batch uploads are permitted until this one is completed or aborted.
+
+  Requires a list of file paths and ETags (md5 checksums).
+  The number of files being uploaded must be less than some experimentally defined limit (say ~500).
```
yarikoptic, line 48: should we provision some return error code to instruct client to "try less" if for some reason `500` would still be "too many"?

dchiquito:
```suggestion
  The number of files being uploaded must be less than some experimentally defined limit (say ~500).
  If the limit is exceeded, return response 400.
```

dchiquito:
We need to get moving on implementing this stuff so I'm going to go ahead and merge this. If there's any redesigning that needs to happen that can go in a new PR.

PR #588, opened 2021-11-03
Revert "Update deprecated `before_binding` to `mutate_configuration`"

mvandenburgh:
Reverts dandi/dandi-api#585

Staging deploys are failing ever since this change was introduced. We should still ultimately make the change to `mutate_configuration` since `before_binding` is deprecated, but merging this will tell us if that is the cause of the deploy failures or something else entirely.


mvandenburgh:
I forgot you can trigger a manual deploy of a branch in the Heroku web console, there's no need to create a PR just to trigger a deploy. I'll close this.

PR #590, opened 2021-11-04
Collect first and last names from users who don't have them

### LINKED ISSUE #570: Collect full names for nameless users
This needs to be done after #564 is completed, since solving that issue will mean all new users come with names.

One way to accomplish this is to check for blank name fields for any existing user, and redirect them to a form that is similar to the questionnaire of #564, but that just asks for their name.

mvandenburgh:
This is one potential solution to #570.

This adds a check to the `oauth/authorize` endpoint that is called whenever a user logs in. If the user doesn't have a first or last name, they are redirected to a form that they must fill out before being allowed to log in. This would need to be accompanied by a manual operation to revoke all session cookies from users without first or last names.

Closes #570

### DIFF
--- a/dandiapi/api/templates/api/account/questionnaire_form.html
+++ b/dandiapi/api/templates/api/account/questionnaire_form.html
@@ -23,6 +23,9 @@
       </div>
     </div>
   </form>
+  <div style="margin-top: 4%">
+    Contact us at <a href="mailto:help@dandiarchive.org">help@dandiarchive.org</a> if you have any questions.
+  </div>
 
   <script>
     // Length validation
--- a/dandiapi/api/views/auth.py
+++ b/dandiapi/api/views/auth.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
+import json
+from json.decoder import JSONDecodeError
+
 from django.contrib.auth.models import User
 from django.http import HttpRequest, HttpResponse
-from django.http.response import HttpResponseBase, HttpResponseRedirect
+from django.http.response import Http404, HttpResponseBase, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
 from django.views.decorators.http import require_http_methods
@@ -34,6 +37,14 @@ def auth_token_view(request: Request) -> HttpResponseBase:
     return Response(token.key)
 
 
+QUESTIONS = [
+    {'question': 'First Name', 'max_length': 100},
+    {'question': 'Last Name', 'max_length': 100},
+    {'question': 'What do you plan to use DANDI for?', 'max_length': 1000},
+    {'question': 'Please list any affiliations you have.', 'max_length': 1000},
+]
+
+
 @require_http_methods(['GET'])
 def authorize_view(request: HttpRequest) -> HttpResponse:
     """Override authorization endpoint to handle user questionnaire."""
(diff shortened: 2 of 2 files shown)


waxlamp (approved):
This is a good approach. I would do the following before merging:

- [x] Add an "escape hatch" for users to send us an email (`Contact us if you have any questions` would be good text for that part)
- [x] File follow-on issues to
  - write a maintenance script to revoke logins for name-free users
  - cause validation to fail for empty-named contact person or contributors

dchiquito (commented):
Is there any reason not to just use the full questionnaire? I guess it would be slightly more convenient for the users to only fill out their names, but it's more plumbing for us, and we would probably want to have a full questionnaire for old users eventually.

mvandenburgh:
> Is there any reason not to just use the full questionnaire? I guess it would be slightly more convenient for the users to only fill out their names, but it's more plumbing for us, and we would probably want to have a full questionnaire for old users eventually.

Yeah it was just a convenience thing. I'm on board with using the full questionnaire if that's not an issue, it would certainly make this code less complicated.

waxlamp:
> > Is there any reason not to just use the full questionnaire? I guess it would be slightly more convenient for the users to only fill out their names, but it's more plumbing for us, and we would probably want to have a full questionnaire for old users eventually.
> 
> Yeah it was just a convenience thing. I'm on board with using the full questionnaire if that's not an issue, it would certainly make this code less complicated.

This is a question for @satra and @yarikoptic. My own opinion is to grandfather in users who are already approved, which is why we're going out of our way to just collect the names in this case. I agree it would be less work for us to throw the full questionnaire at these users, but the ultimate goal is to make this system useful and easy for users (which may sometimes require us to do more work).

satra:
i think the eventual solution is a user profile panel + richer user data model with links to github, orcid, etc.,. for now grandfathering and using the full questionnaire sounds good.

waxlamp:
To be clear: we can use the full questionnaire, but we won't change any user's approved status based on what they supply (to me, that's what "grandfathered" means in this context). Is that right?

This way we can reduce the burden on our devs while still collecting the full name information we need.

satra:
The only place where this will get used is the user icon and in any subsequent emails to users. There are no other places where this information is currently used. since this is only for user registration, anyone who has already registered can be grandfathered in.

waxlamp:
> The only place where this will get used is the user icon and in any subsequent emails to users. There are no other places where this information is currently used. since this is only for user registration, anyone who has already registered can be grandfathered in.

The driving reason we're adding it is to prevent empty Contact Person fields in new dandisets created by users without a name set. This is why we need names from approved users who don't have one set, and also why I'm worrying about presenting a full questionnaire to approved users when we only need their name.

satra:
Contact person doesn't come from a user's/owner credentials. It comes from
the contactperson set up in the contributor list. Hence this should be more
of a dandiset creation thing than a user registration thing. Contact person
may be someone other than the owner it any owners.

waxlamp:
> Contact person doesn't come from a user's/owner credentials. It comes from the contactperson set up in the contributor list. Hence this should be more of a dandiset creation thing than a user registration thing. Contact person may be someone other than the owner it any owners.

Agreed, but by default the creator of the dandiset is listed as the initial contact person, which leaves the dandiset broken for nameless users. I just filed an issue to disallow empty contact person fields as part of validation, but it would still be good to have the users' names (for this purpose, but also the others you noted).

In the large, this discussion still leaves open my original question, which is whether we can simply ask approved users for their names, rather than showing them the full questionnaire.

satra:
> I just filed an issue to disallow empty contact person fields as part of validation,

this is already true and checked by the validator.

i'm fine with asking users for their names when they log in.

waxlamp:
> > I just filed an issue to disallow empty contact person fields as part of validation,
> 
> this is already true and checked by the validator.
> 

Looking at [164](https://gui.dandiarchive.org/#/dandiset/000164), notice that the contact person is blank, but the validation errors do not point that out. Technically, you are right: the contact person here is listed as a single comma (`,`), which the regex currently recognizes as non-empty, but I'm suggesting that a contact person should have non-empty first and last names.

> i'm fine with asking users for their names when they log in.

Excellent, thanks!

satra:
indeed this is a validation issue, the regex is not write. this will require a schema update, which means the metadata update issue comes into play.

dchiquito (approved):

PR #591, opened 2021-11-05
Fix partial filename from paths_prefix

AlmightyYakob:
I was using the asset paths code for [OASIS](https://github.com/ResonantGeoData/RD-OASIS) project, and I noticed a bug: If you provide part of an asset's path in `path_prefix`, the resulting path in `files` will be the remainder, not the entire filename, as you'd expect.

For example, given an asset with the name `a/b/c.zip`, providing `a/b/c` to `path_prefix` yields an entry in `files` with `.zip`.

This PR just fixes the name if it's a file. The behavior is still valid if it's a folder, so it doesn't touch that branch.

While that endpoint does have the following description

>The specified path must be a folder; it either must end in a slash or
(to refer to the root folder) must be the empty string.

I don't see a reason not to handle this case  :man_shrugging:

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -440,6 +440,8 @@ class AssetViewSet(NestedViewSetMixin, DetailSerializerMixin, ReadOnlyModelViewS
             is_folder = folder_index >= 0
 
             if not is_folder:
+                # Ensure base_path is entire filename
+                base_path = asset.path[asset.path.rfind('/') + 1 :]
                 files[base_path] = asset
             else:
                 base_path = base_path[:folder_index]


dandiapi/api/views/asset.py
```diff
@@ -440,6 +440,8 @@ def paths(self, request, versions__dandiset__pk: str, versions__version: str, **
             is_folder = folder_index >= 0
 
             if not is_folder:
+                # Ensure base_path is entire filename
+                base_path = asset.path[asset.path.rfind('/') + 1 :]
```
yarikoptic, line 444: could you please add a unittest for this case?

AlmightyYakob: Apologies for never providing an update: I attempted to add a quick test for this, but ran into logistical issues. Since at the time I wasn't active on this project, and this enhancement came from work on another repository, I had to abandon this to continue my other obligations. Maybe now that I'm back on the project I can finish this up.

yarikoptic: Welcome back @AlmightyYakob !

yarikoptic: @AlmightyYakob please update on adding a unittest -- this is the eldest (6 mo?) PR ATM. We better decide on its destiny asap

yarikoptic: @AlmightyYakob ping on this PR.

AlmightyYakob: Adding a unit test here would be nice, but requires a refactor of our testing setup, and doesn't provide much value, so I'll just merge this.

Issue #286, opened 2021-05-17
Use Sites instead of hard-coding URLs


AlmightyYakob:
_Originally posted by @dchiquito in https://github.com/dandi/dandi-api/pull/279#discussion_r631981359_

AlmightyYakob:
[Here](https://docs.djangoproject.com/en/3.2/ref/contrib/sites/#hooking-into-the-current-site-from-views) is documentation on how to retrieve the current site with and without a request context.

yarikoptic:
note: needed for correct operation of `staging` (or any other) version of the dandi archive.

AlmightyYakob:
It looks like we'll want to address this using `Site.objects.get_current().domain`, and likely restricting deployments to only have one `Site` object.

AlmightyYakob:
This also may require us to remove pre-pended protocols where we generate these URLs, to maintain consistency between deploys. Hopefully this shouldn't be an issue though.

waxlamp:
I think this has been solved. @dchiquito and/or @AlmightyYakob, please re-open if not.

AlmightyYakob:
> I think this has been solved. @dchiquito and/or @AlmightyYakob, please re-open if not.

Where was this resolved? I think we're still hardcoding some URLs, but it's possible I just missed some changes.

yarikoptic:
it was probably resolved in some spots but not all yet:
```
(git)lena:~/proj/dandi/dandi-api[master]git
$> git describe
fatal: No annotated tags can describe '0373d72886426199fa97bea5f1dc0daf50b4cd58'.
However, there were unannotated tags: try --tags.
(dev3) 1 22732 ->128.....................................:Tue 20 Jul 2021 02:28:57 PM EDT:.
(git)lena:~/proj/dandi/dandi-api[master]git
$> git describe --tags
v0.1.3-15-g0373d72
(dev3) 1 22733.....................................:Tue 20 Jul 2021 02:28:59 PM EDT:.
(git)lena:~/proj/dandi/dandi-api[master]git
$> git grep api.dandiar
dandiapi/api/models/asset.py:        download_url = 'https://api.dandiarchive.org' + reverse(
dandiapi/api/models/version.py:                    f'https://api.dandiarchive.org/api/dandisets/{self.dandiset.identifier}'
```
and I just ran into the use case 
[...]

PR #594, opened 2021-11-08
Send daily email listing pending users

### LINKED ISSUE #576: Send daily email that lists pending users
https://github.com/dandi/dandi-api/pull/568#discussion_r734640814

Can be a celery beat task that checks for any pending users and sends an email that lists each user + the URL to approve them.

mvandenburgh:
Closes #576

### DIFF
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -1,9 +1,10 @@
 import logging
-from typing import List
+from typing import List, Union
 
 from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.models import User
 from django.core import mail
+from django.db.models.query import QuerySet
 from django.template.loader import render_to_string
 
 logger = logging.getLogger(__name__)
@@ -151,5 +152,22 @@ def build_rejected_user_message(user: User, socialaccount: SocialAccount = None)
 def send_rejected_user_message(user: User, socialaccount: SocialAccount):
     logger.info(f'Sending rejected user message to {user}')
     messages = [build_rejected_user_message(user, socialaccount)]
+    with mail.get_connection() as connection:
+        connection.send_messages(messages)
+
+
+def build_pending_users_message(users: Union[QuerySet, List[User]]):
+    render_context = {'users': users}
+    return build_message(
+        subject='DANDI: new user registrations to review',
+        message=render_to_string('api/mail/pending_users_message.txt', render_context),
+        to=['dandi@mit.edu'],
+        html_message=render_to_string('api/mail/pending_users_message.html', render_context),
+    )
+
+
+def send_pending_users_message(users: Union[QuerySet, List[User]]):
+    logger.info(f'Sending pending users message to admins at dandi@mit.edu')
+    messages = [build_pending_users_message(users)]
     with mail.get_connection() as connection:
         connection.send_messages(messages)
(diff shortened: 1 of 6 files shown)


dandiapi/api/tasks.py
```diff
@@ -161,3 +163,12 @@ def validate_version_metadata(version_id: int) -> None:
 @shared_task
 def delete_doi_task(doi: str) -> None:
     delete_doi(doi)
+
+
+@shared_task
+def send_pending_users_email() -> None:
+    """Send an email to admins listing users with status set to PENDING"""
```
dchiquito, line 170:
```suggestion
    """Send an email to admins listing users with status set to PENDING."""
```

dandiapi/api/mail.py
```diff
@@ -153,3 +154,20 @@ def send_rejected_user_message(user: User, socialaccount: SocialAccount):
...
     with mail.get_connection() as connection:
         connection.send_messages(messages)
+
+
+def build_pending_users_message(users: Union[QuerySet, List[User]]):
+    render_context = {'users': users}
+    return build_message(
+        subject='DANDI: new user registrations to review',
+        message=render_to_string('api/mail/pending_users_message.txt', render_context),
+        to=['dandi@mit.edu'],
```
dchiquito, line 164: Probably this should be a constant/setting at some point

dandiapi/api/tasks.py
```diff
@@ -161,3 +163,12 @@ def validate_version_metadata(version_id: int) -> None:
...
+    pending_users = User.objects.select_related('metadata').filter(
```
dchiquito, line 171: IIRC `select_related` is only useful if you're going to use `user.metadata` somewhere, as it saves the DB from performing a separate `SELECT`. I don't think you actually do that anywhere though, all I see are accesses of `user.username` and `user.date_joined` in the template. Are those aliases for metadata fields?

satra:
can we use `info@dandiarchive.org` instead of `dandi@mit.edu` to make it more generic?

dandiapi/api/tasks.py
mvandenburgh: Oops, I think I was accessing `user.metadata` somewhere but then removed it. And `username` and `date_joined` are both fields directly on the `User` model, so yeah select_related isn't needed here

mvandenburgh:
> can we use `info@dandiarchive.org` instead of `dandi@mit.edu` to make it more generic?

Maybe the admin email should be a Django setting/env var that is set via Terraform. @dchiquito what do you think?

dchiquito:
Assuming that there is one email that it is appropriate to send all admin traffic to, then yes it should definitely be a setting. Since we generally want to keep the repository divorced from our deployment, this is a task like https://github.com/dandi/dandi-api/issues/286 that would be a good idea to clean up at some point.

mvandenburgh:
> Assuming that there is one email that it is appropriate to send all admin traffic to, then yes it should definitely be a setting. Since we generally want to keep the repository divorced from our deployment, this is a task like #286 that would be a good idea to clean up at some point.

:+1: I filed #605 to track this so we can keep this PR well-scoped. In the meantime, I've hardcoded `info@dandiarchive.org` as the email for DANDI admins.

dchiquito (approved):

PR #595, opened 2021-11-10
Management script to restore missing schema keys

mvandenburgh:
Following up on https://github.com/dandi/dandiarchive/issues/954#issuecomment-965295820

<details>
<summary>The following dandisets were affected in production:</summary>
<br>

```
{'field': '', 'message': "'schemaKey' is a required property"}
000146

{'field': 'access.0', 'message': "'schemaKey' is a required property"}
000161

{'field': '', 'message': "'schemaKey' is a required property"}
000144

{'field': '', 'message': "'schemaKey' is a required property"}
000145

{'field': 'access.0', 'message': "'schemaKey' is a required property"}
000119

{'field': 'relatedResource.0', 'message': "'schemaKey' is a required property"}
000165

{'field': 'relatedResource.1', 'message': "'schemaKey' is a required property"}
000165

{'field': 'access.0', 'message': "'schemaKey' is a required property"}
000156

{'field': 'assetsSummary', 'message': "'schemaKey' is a required property"}
000025
```
</details>

This PR adds a one-off management script that will restore the schema key field to those dandisets.

### DIFF
--- a/dandiapi/api/management/commands/repair_schema_keys.py
+++ b/dandiapi/api/management/commands/repair_schema_keys.py
@@ -1,0 +1,31 @@
+import djclick as click
+
+from dandiapi.api.models import Version
+
+# maps name of metadata field to its correct value for "schemaKey"
+SCHEMA_KEYS = {
+    '': 'Dandiset',
+    'access': 'AccessRequirements',
+    'relatedResource': 'Resource',
+    'assetsSummary': 'AssetsSummary',
+}
+
+
+@click.command()
+def repair_schema_keys():
+    versions = Version.objects.filter(validation_errors__isnull=False)
+    for version in versions:
+        for error in version.validation_errors:
+            if "'schemaKey' is a required property" in error['message']:
+                error_split: list[str] = error['field'].split('.')
+                field = error_split[0]
+
+                if field == '':
+                    version.metadata['schemaKey'] = SCHEMA_KEYS[field]
+                elif len(error_split) == 1:
+                    version.metadata[field]['schemaKey'] = SCHEMA_KEYS[field]
+                else:
+                    # handles cases where error is like "relatedResource.0"
+                    index = int(error_split[1])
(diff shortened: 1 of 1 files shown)


dchiquito:
I would actually say that this is something that could be done manually in the django shell rather than having it's own script. I find it unlikely that we will need to do this exact task again in the future, so better not to clutter our script collection with more stuff that makes it harder to find useful scripts in the future.

I'm open to debate, as always. Just my 2¢

mvandenburgh:
> I would actually say that this is something that could be done manually in the django shell rather than having it's own script. I find it unlikely that we will need to do this exact task again in the future, so better not to clutter our script collection with more stuff that makes it harder to find useful scripts in the future.
> 
> I'm open to debate, as always. Just my 2¢

I agree, and I completely forgot this is doable with the `shell_plus` command. Can you review the code here, and then if it looks okay to you I'll just run the script directly in the shell and close this PR.

dandiapi/api/management/commands/repair_schema_keys.py
```diff
@@ -0,0 +1,31 @@
...
+                field = error_split[0]
+
+                if field == '':
+                    version.metadata['schemaKey'] = SCHEMA_KEYS[field]
+                elif len(error_split) == 1:
+                    version.metadata[field]['schemaKey'] = SCHEMA_KEYS[field]
+                else:
+                    # handles cases where error is like "relatedResource.0"
+                    index = int(error_split[1])
+                    version.metadata[field][index]['schemaKey'] = SCHEMA_KEYS[field]
```
dchiquito, line 30: This will cause an error if the field isn't in `SCHEMA_KEYS`. That's fine for our current batch of errors, but might not be for the next time we whip this script out.

dchiquito (approved):
If you're going to merge this, can you add an entry to the readme documenting all our scripts, and note that it only repairs Dandiset, AccessRequirements, Resource, and AssetsSummary?

PR #601, opened 2021-11-11
Add a linebreak to the end of the message text

waxlamp:


dandiapi/api/templates/api/mail/new_user_message.html
```diff
@@ -1,3 +1,3 @@
 <p>A new user ({{ username }}) has signed up for DANDI.</p>
 <p>Please proceed to the following URL to approve or reject their account:</p>
-<a href="https://api.dandiarchive.org/dashboard/user/{{ username }}">https://api.dandiarchive.org/dashboard/{{ username }}</a>
+<a href="https://api.dandiarchive.org/dashboard/user/{{ username }}">https://api.dandiarchive.org/dashboard/{{ username }}</a><br>
```
mvandenburgh, line 3:
```suggestion
<p><a href="https://api.dandiarchive.org/dashboard/user/{{ username }}">https://api.dandiarchive.org/dashboard/{{ username }}</a></p>
```
I think this has the same effect and is consistent with how the rest of the emails are formatted using the `p` tag.

mvandenburgh (approved):

dandiapi/api/templates/api/mail/new_user_message.html
waxlamp: I think this would work. What we were seeing in the email with the horizontal divider that starts the email signature starting at the end of the last line of the message text. The lack of a terminating `</p>` is probably what was causing the issue. Thanks!

PR #606, opened 2021-11-17
Add design doc to handle "test" dandisets

waxlamp:


doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,102 @@
...
+
+Part of the problem may be that people simply cannot delete their own abanonded
+dandisets. Offering a delete button with the proper semantics could help here.
+
+The archive does not allow for deletion of *published* dandisets, by design. But
+dandisets that are draft-only should be deleteable.
+
+And, as en extension to that idea, dandisets that are published should be able
+to have their draft dandiset deleted (essentially "archiving" that dandiset, and
+closing it off to further updates).
```
satra, line 75: would there be a way to reopen this?

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,102 @@
...
+way to (1) discourage the creation and abandonment of exploratory dandisets; (2)
+mark dev dandisets in some way to prevent them from appearing in the main
+listing; and (3) clean up the archive from such abandoned dandisets.
+
+## Proposals
+
+### "Test" flag
+
+This proposes to add a `testing` flag to all dandisets. At creation time, a
+dandiset can optionally have this flag set to prevent it from showing up in
```
satra, line 20: I would also suggest a pre-test flag step that simply points users to the staging archive.

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,102 @@
...
+pre-emptive approval process. If the owner explains that indeed they have
+collaborators, they are collecting data, etc., but it will take a while, the
+admins could put a timer on the dandiset for when to check in again, etc.
+
+This is just a sketch of how this feature might work. It is extremely complex,
+requiring policy, careful communication, and buy-in from the community. It would
+also take focused labor from the admin team, and there are risks associated with
+non-communicative dandiset owners who may become frustrated with having their
+dandisets deleted, etc. But on the plus side, it would also bring richness and
+value to the dandisets hosted on the archive.
```
mvandenburgh, line 102: Maybe instead of deleting invalid dandisets, they can simply be "archived". This would remove the dandiset from all public facing pages, and it would only be viewable by the owners in their "My Dandisets" page. Upon a dandiset being archived, all owners of the dandiset would receive an email alerting them of it. The only thing an owner needs to do to unarchive it is to login to dandi and click "unarchive" on the dandiset.

This also has the benefit of being completely automatable.

doc/design/test-dandisets-1.md
bendichter: I like the idea of deleting, but I think we should make it a soft delete, where we truly just de-list the dandiset for a set amount of time (say 2 weeks) before truly deleting it, and the owner would have the ability to get it back. For now this can be done by email, but in the future it would be great to have a recovery process integrated into the UI.

bendichter:
I like all of these ideas!

doc/design/test-dandisets-1.md
bendichter: I agree though I think the language might be confusing, just because DANDI calls itself an "archive." By that definition, wouldn't all dandisets be "archived?" It might not be clear what "archiving" would do (though I understand it would be similar to how gmail uses the term)

mvandenburgh: Ah good point, yes I agree "delist" would be better than "archive".

> On the other hand, this might be abused to use DANDI resources to create private dandisets on the sly..

If a dandiset has assets it wouldn't be considered an "invalid" dandiset and thus couldn't be delisted. And a "private" dandiset without any assets doesn't have much use to people looking to abuse the system. 

We could also make delisted dandisets readonly until they are relisted by an owner.

bendichter: oh, ok. Sure, we can only allow assetless dandisets to be delisted. Or we could allow any dandiset to be delisted, and allow users to view delisted dandisets in advanced settings.

doc/design/test-dandisets-1.md
waxlamp: If we don't call it "deleting", and instead use something like "frozen" or "finalized" or something, then definitely it would be a reversible operation.

doc/design/test-dandisets-1.md
waxlamp: I love the delisting idea, including making delisted dandisets read-only (to prevent abuse).

I also think there could be a reason to delist dandisets with assets in them--specifically, draft-only dandisets with, e.g., incomplete assets aren't useful to the community, so at some point it's gotta be "publish or perish" so to speak.

dchiquito: delisting sounds like the `"Test" flag` section, but with a different label. I think it would make sense for that to be a boolean setting available for all draft dandisets that can be set during creation and updated later on by owners or admins.

doc/design/test-dandisets-1.md
waxlamp: See b63eca8f2b3f2dd51484676270a1db25b0b23aae.

doc/design/test-dandisets-1.md
waxlamp: Good idea. Check out cc13895fc8eda7eb970ba5983d2199b80914f095.

I am, however, starting to worry about the UX here. For one thing, calling it a "staging" server may start to be troublesome if the intent is for end users to see it as a playground or testbed of sorts. We've already had one confused user ask if stuff they upload there is actually being "staged" for publishing to the production server. A change in nomenclature may be in order if we want to push this aspect of our staging server.

Otherwise, perhaps marking a dandiset as "test" should simply prevent the user from publishing it, and would forcibly delete/GC that dandiset after a set time (7 days)? Then we could retain the staging server in the devops meaning of the term.

doc/design/test-dandisets-1.md
waxlamp: I toyed [above](https://github.com/dandi/dandi-api/pull/606#discussion_r779652437) with the idea of making the "test" flag autodelete the dandiset after 7 days, so that if a user just wants to try something out they can sort of fire and forget. Then, the "delisting" concept would apply more broadly to non-test dandisets as discussed in this thread.

waxlamp: I significantly reworked this section to focus on delisting instead; see 5b8ed7112466876c9992e4d41f6720d6500cb9ce. There's a lot to discuss here, so please give me feedback.

(In particular, the time ranges I suggested for delisting criteria are arbitrary and probably too short.)

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,105 @@
...
+The anticipation date question would delay considering the dandiset abandoned
+until after that date.
+
+### Enable deletion of dandisets
+
+Part of the problem may be that people simply cannot delete their own abanonded
+dandisets. Offering a delete button with the proper semantics could help here.
+
+The archive does not allow for deletion of *published* dandisets, by design. But
+dandisets that are draft-only should be deleteable.
```
yarikoptic, line 72:
```suggestion
dandisets that are draft-only should be deleteable, and admins can delete published versions
of the dandisets.
```

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,105 @@
...
+And, as en extension to that idea, a published dandiset should be able to be
+"sealed" by having its draft dandiset deleted, thus closing it off to further
+updates. This would serve as a signal that the dandiset is in its final state
+and has become a permanent member of the archive. Owners and admins will be able
+to "unseal" sealed dandisets in case further modification is needed.
+
+### Delisting of inactive dandisets
+
+Dandisets in a certain set of conditions will be automatically "delisted",
+meaning they will not appear in listings or searches other than to the owners
```
yarikoptic, line 83: listed/delisted is yet another flag to implement/manage.  Why not just to remove?

Also there is a notion of "revert 'draft' changes" which we might introduce, similarly to OpenNeuro.
FWIW, here is a recently announced OpenNeuro data retention policy sent out as email:



Dear OpenNeuro user,
We are writing to inform you that OpenNeuro's data retention policy has gone into effect as of July 18, 2022. This policy affects datasets that are in draft status, which are datasets that have had no snapshots made, or have had changes made to them since the most recent snapshot. Public and private datasets may be in draft status, and making a snapshot does not automatically make a dataset public.Starting on August 15, 2022, if a dataset has been in draft status for at least 28 days, then it MAY be reverted to the latest snapshot at any time. If no snapshot is found, this means that the dataset will be deleted.This policy is being set for two reasons: 1) Dataset owners may make changes to their datasets and unknowingly fail to snapshot these changes, making them inaccessible to other users. 2) Unsaved changes to datasets impose a higher maintenance burden on OpenNeuro, limiting our ability to make changes that can benefit all users of the site.To determine which of your datasets may be affected, you may find a full list at https://openneuro.org/search?mydatasets. On each dataset, navigate to the "Draft" version and locate the banner at the top. If the banner is green, no further action is needed. If the banner is not green, please make sure to snapshot your dataset prior to August 15, 2022. If you are reluctant to snapshot because the dataset is not in a "final" state, it is possible to release a new version and mark previous versions as "deprecated", indicating that users should update to the latest version.We will be sending additional emails to all dataset owners with affected datasets.If you have any questions regarding this policy, please feel free to submit a support ticket. Thank you,The OpenNeuro team | Dear OpenNeuro user,We are writing to inform you that OpenNeuro's data retention policy has gone into effect as of July 18, 2022. This policy affects datasets that are in draft status, which are datasets that have had no snapshots made, or have had changes made to them since the most recent snapshot. Public and private datasets may be in draft status, and making a snapshot does not automatically make a dataset public.Starting on August 15, 2022, if a dataset has been in draft status for at least 28 days, then it MAY be reverted to the latest snapshot at any time. If no snapshot is found, this means that the dataset will be deleted.This policy is being set for two reasons: 1) Dataset owners may make changes to their datasets and unknowingly fail to snapshot these changes, making them inaccessible to other users. 2) Unsaved changes to datasets impose a higher maintenance burden on OpenNeuro, limiting our ability to make changes that can benefit all users of the site.To determine which of your datasets may be affected, you may find a full list at https://openneuro.org/search?mydatasets. On each dataset, navigate to the "Draft" version and locate the banner at the top. If the banner is green, no further action is needed. If the banner is not green, please make sure to snapshot your dataset prior to August 15, 2022. If you are reluctant to snapshot because the dataset is not in a "final" state, it is possible to release a new version and mark previous versions as "deprecated", indicating that users should update to the latest version.We will be sending additional emails to all dataset owners with affected datasets.If you have any questions regarding this policy, please feel free to submit a support ticket. Thank you,The OpenNeuro team

yarikoptic (commented):
just some comments to possibly bring attention back to this PR

doc/design/test-dandisets-1.md
waxlamp: I know this is true in a strict sense, but I can't remember if we talked about making this a simple or easy operation for admins through the UI (i.e., perhaps it is possible only through a command line operation, which makes it harder to accidentally delete published versions, etc.).

I will update the language to reflect that is is *possible* in principle for admins to delete published versions, but not a usual use case for the archive.

doc/design/test-dandisets-1.md
waxlamp: > listed/delisted is yet another flag to implement/manage. Why not just to remove?

This is a great point, but it represents a more aggressive stance on usage of the archive. If that's where we want to go, then great: it indeed makes developing functionality and performing maintenance much easier, but at a cost to the user's experience of using the archive. My background sense is that it's already hard enough to convince people to put their data in a public archive without also subjecting them to strict deletion policies for dandisets that have, e.g., not seen activity in a while. The problem here is (as usual for us lately) one of scale: users that have uploaded very large data such as Zarrs of sufficient size will lose all trust in the archive if we do things like delete after a period of inactivity.

All that being said, I would like to establish what our intended "terms of use" really are. That will help me tighten up the language in this design doc, and enable us to write proper documentation for users of the archive so they know what it's meant for and what to expect from it.

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,105 @@
...
+public searches.", with "staging server" linked thereto).
+
+### Hide draft-only dandisets
+
+Similarly to hiding explicitly marked test dandisets, we could, by default, also
+hide dandisets that have no published versions (as above, these dandisets would
+still appear for the owners). As the public may *want* to see draft-only
+dandisets, everyone would be able to toggle this mode with a setting.
+
+### Hide dandisets with no assets
```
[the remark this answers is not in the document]
waxlamp: Thank you--see 27c911a932c42eedf420dbdb46d60fc6daaefde3.

doc/design/test-dandisets-1.md
waxlamp: See b8f90a27e334b2553874b06374046a73eb2b8f16.

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,111 @@
...
+- When do you anticipate posting data to the dandiset?
+
+may help to stem the tide of abandoned dandisets.
+
+The collaborators question, answered positively, would add a badge to the
+dandiset's landing page and search result display indicating the call for
+collaboration.
+
+The anticipation date question would delay considering the dandiset abandoned
+until after that date.
```
waxlamp, line 68: This seems like an awkward and indirect way to prevent the problem we're addressing here. I move to strike this section from the design.

doc/design/test-dandisets-1.md
waxlamp: The other idea here is to delist with a notification to the owner, and *then* delete after another grace period. The delisting then is simply a way to keep the archive "clean" without also permanently destroying possibly valuable data.

doc/design/test-dandisets-1.md
```diff
@@ -0,0 +1,111 @@
...
+
+Dandisets in a certain set of conditions will be automatically "delisted",
+meaning they will not appear in listings or searches other than to the owners
+themselves. Additionally, a delisted dandiset will become read-only so that only
+actively listed dandisets can be used by their owners to store data.
+
+Conditions for delisting can include the following:
+- draft-only and assetless for more than 7 days
+- draft-only and no activity for more than 14 days
+- published with no activity for more than 30 days (also see "sealing" above)
```
satra, line 96: we would refine these conditions, but the sentiment holds. i would add not just delisting, but deletion as well as a condition.

satra:
this is also related to our terms/policies doc in handbook. so we should have a corresponding PR there to update those in relation to this. 

more generally this design doc is related to lifecycle policies of objects in dandi. while i'm fine with accepting this, we need to work on lifecycles of all objects, not just the dandiset.

satra:
one area that's missing from consideration here are embargoed dandisets. do they receive special status that's different from these consideration. i'm fine with leaving them off for now, but there should be a section that says these policies do not apply to embargoed dandisets at present, but only to "visible"/publicly listed dandisets.

waxlamp:
After some internal discussion, we have decided to drop this initiative.

PR #607, opened 2021-11-17
Empty dandiset filter

dchiquito:
required by https://github.com/dandi/dandiarchive/pull/977

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -5,7 +5,7 @@ from dandischema.models import Dandiset as PydanticDandiset
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import OuterRef, Q, Subquery
+from django.db.models import Count, OuterRef, Q, Subquery
 from django.db.utils import IntegrityError
 from django.http import Http404
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -82,9 +82,34 @@ def get_queryset(self):
...
+            show_empty = self.request.query_params.get('empty', 'false') == 'true'
+
+            if not show_draft:
+                # Only include dandisets that have more than one version, i.e. published dandisets.
+                queryset = queryset.annotate(version_count=Count('versions')).filter(
+                    version_count__gt=1
+                )
+            if not show_empty:
+                # Only include dandisets that have assets.
+                # It is enough to check the draft version, since any published version is required
```
mvandenburgh, line 102: Wouldn't this exclude a dandiset with a draft version that has no assets, but a published version that does?

In other words, if the user uploaded assets to a dandiset, published it, then deleted them from the draft version, that dandiset would be regarded as an "empty" dandiset here.

dchiquito: Good point

mvandenburgh (approved):
LGTM

yarikoptic:
commit "Add empty and draft filters to dandiset list" and PR title says "filter", which to me means that by default API should return all, and if filter applied -- then filter them out.  But I see dandi-cli release test failing with
```
______________________________ test_get_dandisets ______________________________

text_dandiset = {'client': <dandi.dandiapi.DandiAPIClient object at 0x7f5886e1bd00>, 'dandiset': <dandi.dandiapi.RemoteDandiset object at 0x7f5886db0160>, 'dandiset_id': '000006', 'dspath': PosixPath('/tmp/pytest-of-runner/pytest-0/text_dandiset1'), ...}

    def test_get_dandisets(text_dandiset):
        dandisets = list(text_dandiset["client"].get_dandisets())
>       assert (
            sum(1 for d in dandisets if d.identifier == text_dandiset["dandiset_id"]) == 1
        )
E       assert 0 == 1
E         +0
E         -1

/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/dandi/tests/test_dandiapi.py:367: AssertionError
```

meaning that by default everything possible filter is applied and then optionally disabled (looking at the test confirms it).  I think that such approach might quickly shown to be problematic.  E.g. what if we decide to add another filter (e.g. hypothetical "published_this_year") -- we would filter everything else by default and require addition of options...

So I think logic should be reverted that no filters are applied by default!  Web UI can provide `empty=false` for its view

PR #608, opened 2021-11-18
Use odysseus in tests with names

dchiquito:
Previously we used `jane` as a generic name for user search tests. This
worked most of the time, but would break whenever the User factory
happened to generate a use with the name `Jane`.
Switch the generic name to `Odysseus`, which is presumably not in
Fakers name pool.

Take 2: explicitly order the users in the search endpoint.
Also clean up the SocialAccountFactory a bit to tie it more tightly to the base User object.


waxlamp (approved):

dchiquito:
Well the CI test is still failing, so I guess my diagnosis was wrong :(

mvandenburgh (approved):

PR #610, opened 2021-11-19
allowing for ordering by size

djarecka:
I'm trying to add sort by size, as requested in https://github.com/dandi/dandiarchive/issues/943

However, I'm having issues with getting the proper size of Dandisets. I can see that `Version` has `size` as a property, but I was not able to use `size` to query. I've decided to use `metadata__assetsSummary__numberOfBytes`, but this always returns 0, as it is set [here](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/models/version.py#L223).

I've tried also to add `size` to `metadata` in `model.Version`, but I don't see any effect. Not sure how this should be done. 

Closes https://github.com/dandi/dandiarchive/issues/943.

**edit**
changes can be seen with https://github.com/dandi/dandiarchive/pull/992


djarecka:
@dchiquito @mvandenburgh - I will need some help with it. Thank you

dandiapi/api/views/dandiset.py
```diff
@@ -64,6 +64,14 @@ def filter_queryset(self, request, queryset, view):
                     modified_version=Subquery(latest_version.values('modified'))
                 )
                 return queryset.order_by(f'{ordering}_version')
+            elif ordering.endswith('size'):
+                latest_version = Version.objects.filter(dandiset=OuterRef('pk')).order_by(
+                    '-created'
+                )[:1]
+                queryset = queryset.annotate(
+                    size=Subquery(latest_version.values('metadata__assetsSummary__numberOfBytes'))
```
mvandenburgh, line 72: > However, I'm having issues with getting the proper size of Dandisets. I can see that Version has size as a property, but I was not able to use size to query. I've decided to use metadata__assetsSummary__numberOfBytes, but this always returns 0, as it is set here.

The `size` of a dandiset is equivalent to the sum of the sizes of the most recent `Version`'s `Asset`'s, which we can indeed query for instead of using the `assetsSummary` field.

The `size` property on `Version` is a convenience method when writing Python code, but it's not visible at the SQL level. Instead, you'll need to write a new query that will be similar to the one inside that property method. It won't be exactly the same, but I think you'll still want to use the [`Sum` queryset method](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.Sum).

One important thing to note is that the `Version.size` property is operating on `Assets`, but here we're dealing with `Versions`. So, instead of `blob__size` you will need to use `assets__blob__size` (which should work because of Django's magic).

I'm happy to meet and go through together if this isn't clear and/or you think that would be beneficial.

dchiquito (approved):
We may see some poor performance here since postgres needs to add up all of the Asset sizes in the Version, and then do that for every Version so it can order the Versions correctly. This involves touching every AssetBlob that is currently part of any dandiset, which is a whole lot.

However, I just tried out the query in production and it was basically instantaneous, so I guess it'll work fine at our current data scale. Maybe it will degrade later if we have a few million more assets, maybe not.

djarecka:
should I just merge it?

mvandenburgh:
> should I just merge it?

Yes, whenever you're ready feel free to merge.

Issue #612, opened 2021-11-23
are staging "new user registration to review" addresses "correct"?


yarikoptic:
We were just pinged by @vikramvijayan that his account was not activated on staging server.  I see two pairs of emails for his registration 

```
   1 O   Nov 22 Vikram Vijayan   [Dandi] Dandi staging access
   2     Nov 18 admin@tropical-  [Dandi] DANDI: New user registered: XXX@gmail.com
   3     Nov 18 admin@tropical-  [Dandi] DANDI: Review new user: XXX@gmail.com
   4     Nov 18 admin@developme  [Dandi] DANDI: New user registered: XXX@gmail.com
   5     Nov 18 admin@developme  [Dandi] DANDI: New user registration to review: XXX@gmail.com
```
with the `5` looking like the one for staging (note different subject -- didn't check git history which one is "newer").   But URL in all of the emails points to https://api.dandiarchive.org/ and not staging.

mvandenburgh:
Ah yes those emails have https://api.dandiarchive.org hardcoded  in them. I guess it is time to implement #286.

yarikoptic:
> Ah yes those emails have https://api.dandiarchive.org hardcoded in them. I guess it is time to implement #286.

yes, please.

waxlamp:
> > Ah yes those emails have https://api.dandiarchive.org hardcoded in them. I guess it is time to implement #286.
> 
> yes, please.

https://github.com/dandi/dandi-api/pull/613 will fix this, but that in turn requires a review for https://github.com/dandi/dandischema/pull/98 along with some operations stuff (new schema release, etc.).

satra:
we cannot do a full split at this point as i noted in dandi/dandischema#98, so we either do the full decoupling, or a temporary fix as needed to address the current issue. 
 
 full decoupling will require:
 - instance specific prefix support (DANDI has a very specific meaning and maps on in specific places in the model) support in API and CLI
 - doi prefix (we are currently using prod/test datacite to parallel our needs, but technically this is like saying add a DOI prefix)
 - merging dandiarchive.org and gui.dandiarchive.org (so we are not dependent on a redirector, but we would need any of those functionalities somewhere -  a client side web app doesn't provide all those things - for example HEAD requests)
 - bucket location
 - schemas that are now dependent on these variables, and therefore separate context files generated by prefix and picked up by the api server
 - make gui.dandiarchive.org still responsive and redirecting in the short term.
 
[...]

PR #613, opened 2021-11-23
Use env vars instead of hardcoding URLs

### LINKED ISSUE #286: Use Sites instead of hard-coding URLs

mvandenburgh:
Closes #286 and #612.

~~Requires https://github.com/dandi/dandischema/pull/98 and a new schema release before merging.~~ Edit: we're leaving any schema changes for later, for now just leave the hardcoded URLs in the metadata

Requires https://github.com/dandi/dandi-cli/pull/845 and https://github.com/dandi/dandi-infrastructure/pull/104 (plus a Terraform deploy)

The original issue proposed using Django's built in Sites framework, but there's a few issues with that:

1) Assuming I am reading the docs correctly (https://docs.djangoproject.com/en/3.2/ref/contrib/sites/), Django Sites is intended for the use case where you have a single Django server deployment that is pointed at by multiple URLs. Our use case is slightly different, in that we have multiple URLs that each point to a different Django server deployment
2) Django Sites is designed for websites that are entirely powered by Django, and is only applicable to Django deployments. So, using Sites would let us remove all references to api.dandiarchive.org/api-staging.dandiarchive.org but wouldn't cover the URLs to the frontend Vue app (gui.dandiarchive.org / gui-staging.dandiarchive.org).
3) Django Sites is entirely configured through the database. So, we'd either have to write a custom migration to update the sites that somehow runs differently on production vs staging (also not ideal since we'd be hardcoding deployment details), or we'd need to manually update the `Sites` page in the Django admin. 

So, I decided to instead make the web app and Django server URLs environment variables. We'd have to do this anyways for the web app even if we used Sites for the Django server, plus using environment variables allows us to set everything through Terraform and not have to touch the DB. 

If everyone is happy with these changes, I'll open a PR in dandi-infrastructure to set these variables.


waxlamp (commented):
What would be our production deployment values for `DANDI_WEB_APP_URL` and `DANDI_API_URL`?

dandiapi/api/mail.py
```diff
@@ -1,14 +1,21 @@
...
 from django.core import mail
-from django.template.loader import render_to_string
+from django.template.loader import render_to_string as _render_to_string
 
 logger = logging.getLogger(__name__)
 
 
+def render_to_string(template: str, render_context: Dict[str, Any]):
+    render_context['dandi_api_url'] = settings.DANDI_API_URL
+    render_context['dandi_gui_url'] = settings.DANDI_WEB_APP_URL
```
waxlamp, line 15: What is this `render_context` object used for, and why does it have a different name for what I would expect to be `dandi_web_app_url`?

mvandenburgh: `render_context` is a dictionary of variables passed to the Django templating engine that provides the variables the templates use. Anytime you see `{{ <variable_name> }}` in a template, it's getting that variable from the render context object.

mvandenburgh: > What is this `render_context` object used for, and why does it have a different name for what I would expect to be `dandi_web_app_url`?

Yeah it should be `dandi_web_app_url`, I originally had the Django setting called `DANDI_GUI_URL` but I renamed it and forgot to update the context variables. It's fixed now.

mvandenburgh:
> What would be our production deployment values for `DANDI_WEB_APP_URL` and `DANDI_API_URL`?

For production we would have `DANDI_WEB_APP_URL` set to `https://gui.dandiarchive.org` and `DANDI_API_URL` set to `https://api.dandiarchive.org`

mvandenburgh:
CI is failing because the metadata validation expects a hardcoded `dandiarchive.org` URL to be in the metadata, but in the test environment it's set to `localhost:8085`. I'll open a PR upstream in dandischema.

waxlamp:
> > What would be our production deployment values for `DANDI_WEB_APP_URL` and `DANDI_API_URL`?
> 
> For production we would have `DANDI_WEB_APP_URL` set to `https://gui.dandiarchive.org` and `DANDI_API_URL` set to `https://api.dandiarchive.org`

This poses a slight problem. We're trying to get away from the `gui` subdomain, so it would be counterproductive to sprinkle those URLs through the help text and in other places. Will it work properly to set that to `https://dandiarchive.org` instead?

mvandenburgh:
> > > What would be our production deployment values for `DANDI_WEB_APP_URL` and `DANDI_API_URL`?
> > 
> > 
> > For production we would have `DANDI_WEB_APP_URL` set to `https://gui.dandiarchive.org` and `DANDI_API_URL` set to `https://api.dandiarchive.org`
> 
> This poses a slight problem. We're trying to get away from the `gui` subdomain, so it would be counterproductive to sprinkle those URLs through the help text and in other places. Will it work properly to set that to `https://dandiarchive.org` instead?

Actually I misspoke, `DANDI_WEB_APP_URL` would be `https://dandiarchive.org/` (no www) since that is the hardcoded URL currently being used in production.

waxlamp (approved):

dandiapi/api/mail.py
```diff
@@ -1,14 +1,21 @@
...
-from django.template.loader import render_to_string
+from django.template.loader import render_to_string as _render_to_string
 
 logger = logging.getLogger(__name__)
 
 
+def render_to_string(template: str, render_context: Dict[str, Any]):
+    render_context['dandi_api_url'] = settings.DANDI_API_URL
+    render_context['dandi_web_app_url'] = settings.DANDI_WEB_APP_URL
+    return _render_to_string(template, render_context)
```
dchiquito, line 16:
```suggestion
from django.template.loader import render_to_string

logger = logging.getLogger(__name__)


BASE_RENDER_CONTEXT = {
    'dandi_api_url': settings.DANDI_API_URL,
    'dandi_web_app_url': settings.DANDI_WEB_APP_URL,
}
```

Then you can just do something like this whenever you need to customize the render context:
```
render_context = {
    **BASE_RENDER_CONTEXT,
    'foo': 'bar',
}
```
That reads easier to me than doing function composition, but the difference is purely cosmetic, so your call.

dandiapi/api/views/auth.py
```diff
@@ -96,5 +97,9 @@ def user_questionnaire_form_view(request: HttpRequest) -> HttpResponse:
     return render(
         request,
         'api/account/questionnaire_form.html',
-        {'questions': QUESTIONS, 'query_params': request.GET.dict()},
+        {
+            'questions': QUESTIONS,
+            'query_params': request.GET.dict(),
+            'dandi_web_app_url': settings.DANDI_WEB_APP_URL,
```
dchiquito, line 103: Why is this included now?

dchiquito (approved):
Looks good to me! I had thought that the sites framework was the "correct" way to record domains, but you make a compelling argument. Just a couple of cosmetic questions.

dchiquito:
> CI is failing because the metadata validation expects a hardcoded dandiarchive.org URL to be in the metadata, but in the test environment it's set to localhost:8085. I'll open a PR upstream in dandischema.

There's some magic already that should have dealt with that: https://github.com/dandi/dandischema/blob/master/dandischema/models.py#L41-L45

dchiquito:
I think you're going to have to add those environment variables to the github action config and in dandi-cli.

dandiapi/api/views/auth.py
mvandenburgh: [The base template that this one extends uses it](https://github.com/dandi/dandi-api/blob/use-django-sites/dandiapi/api/templates/api/account/base.html)

PR #615, opened 2021-11-30
Return names for users with no social account

mvandenburgh:
This lets the frontend properly display owners' names when an owner does not have an associated `SocialAccount` (which is always true in local dev and E2E browser testing environments).

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -298,5 +298,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
                 owners.append(owner_dict)
             except SocialAccount.DoesNotExist:
                 # Just in case some users aren't using social accounts, have a fallback
-                owners.append({'username': owner.username})
+                owners.append(
+                    {'username': owner.username, 'name': f'{owner.first_name} {owner.last_name}'}
+                )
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -298,5 +298,7 @@ def get_user_or_400(username):
                 owners.append(owner_dict)
             except SocialAccount.DoesNotExist:
                 # Just in case some users aren't using social accounts, have a fallback
-                owners.append({'username': owner.username})
+                owners.append(
+                    {'username': owner.username, 'name': f'{owner.first_name} {owner.last_name}'}
```
dchiquito, line 302: won't this just be `' '` if the user doesn't have a first/last name? I think test users always will, but I think I had to set my dev user first/last name manually. Perhaps a check that first_name or last_name are defined first?

dchiquito (changes requested):

dandiapi/api/views/dandiset.py
mvandenburgh: Since #590 users are prompted to enter their first/last name if there isn't one from github, so they will always have a name. This would include any user created when running in a dev environment.

dchiquito (approved):

PR #619, opened 2021-12-07
Embargo design docs

dchiquito:
There are two documents here, one for the minimum viable product we are rushing out, and another that describes the additional features that will eventually be required.

There a couple of zarr related TODOs in the MVP because the zarr design doc isn't quite final yet, but that will be patched up soon.

### DIFF
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -2,7 +2,6 @@
 This document takes the Embargo MVP as a predicate and describes the remaining features required for full Embargo support.
 
 ## Non-private embargoed dandisets
-
 In the MVP, all embargoed dandisets are invisible to non-owners.
 There should be an additional embargo mode, "Visible".
 Visible embargoed dandisets are searchable, viewable, and their metadata is public to everyone, but their asset data and metadata is not.
@@ -16,3 +15,23 @@ The asset endpoints will not permitted for anonymous users on embargoed assets t
 ### TODO: can users switch the embargo mode between private and visible?
 It would be easy to do, since it's just a change `embargo_status`, no data needs to be copied.
 We would need a place in the UI to set that.
+
+
+## Read-only access
+In the MVP, the only role a user can have on a dandiset is "Owner".
+There should be a new role, "Reviewer".
+
+Reviewers are allowed to access asset data on private and visible embargoed dandisets.
+Reviewers are allowed to see private embargoed dandisets.
+Reviewers are not allowed to make any changes to embargoed dandisets.
+
+If a user is a reviewer on an embargoed dandiset, they do not have any permissions on other embargoed dandisets unless explicitly added to those dandisets as well.
+
+The web UI will show reviewer selection directly under owner selection.
+
+Reviewers are not acknowledged for unembargoed dandisets and will not show up in the web UI.
+
+This will be implemented as a new django-guardian role.
+All read endpoints for assets are permitted for users who are reviewers of the embargoed dandiset.
+All read endpoints for dandisets and versions will treat reviewers the same as owners.
+All write and upload endpoints are unchanged.
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -1,3 +1,18 @@
 # Embargo
 This document takes the Embargo MVP as a predicate and describes the remaining features required for full Embargo support.
 
+## Non-private embargoed dandisets
+
+In the MVP, all embargoed dandisets are invisible to non-owners.
+There should be an additional embargo mode, "Visible".
+Visible embargoed dandisets are searchable, viewable, and their metadata is public to everyone, but their asset data and metadata is not.
+
+Users will select their desired embargo mode using radio buttons when creating a dandiset.
+
+In addition to the dandiset `embargo_status` values from the MVP (`PRIVATE`, `RELEASING`, and `PUBLIC`), there is a new value, `VISIBLE`.
+All of the dandiset and version endpoints will only hide `PRIVATE` dandisets from appearing.
+The asset endpoints will not permitted for anonymous users on embargoed assets that are `PRIVATE` or `VISIBLE`.
+
+### TODO: can users switch the embargo mode between private and visible?
+It would be easy to do, since it's just a change `embargo_status`, no data needs to be copied.
+We would need a place in the UI to set that.
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -3,16 +3,16 @@ This document takes the Embargo MVP as a predicate and describes the remaining f
 
 ## Non-private embargoed dandisets
 In the MVP, all embargoed dandisets are invisible to non-owners.
-There should be an additional embargo mode, "Visible".
-Visible embargoed dandisets are searchable, viewable, and their metadata is public to everyone, but their asset data and metadata is not.
+There should be an additional embargo mode, "Restricted".
+Restricted embargoed dandisets are searchable, viewable, and their metadata is public to everyone, but their asset data and metadata is not.
 
 Users will select their desired embargo mode using radio buttons when creating a dandiset.
 
-In addition to the dandiset `embargo_status` values from the MVP (`PRIVATE`, `RELEASING`, and `PUBLIC`), there is a new value, `VISIBLE`.
-All of the dandiset and version endpoints will only hide `PRIVATE` dandisets from appearing.
-The asset endpoints will not permitted for anonymous users on embargoed assets that are `PRIVATE` or `VISIBLE`.
+In addition to the dandiset `embargo_status` values from the MVP (`EMBARGOED`, `RELEASING`, and `OPEN`), there is a new value, `RESTRICTED`.
+All of the dandiset and version endpoints will only hide `EMBARGOED` dandisets from appearing.
+The asset endpoints will not permitted for anonymous users on embargoed assets that are `EMBARGOED` or `RESTRICTED`.
 
-### TODO: can users switch the embargo mode between private and visible?
+### TODO: can users switch the embargo mode between embargoed and restricted?
 It would be easy to do, since it's just a change `embargo_status`, no data needs to be copied.
 We would need a place in the UI to set that.
 
@@ -21,7 +21,7 @@ We would need a place in the UI to set that.
 In the MVP, the only role a user can have on a dandiset is "Owner".
 There should be a new role, "Reviewer".
 
-Reviewers are allowed to access asset data on private and visible embargoed dandisets.
+Reviewers are allowed to access asset data on private and restricted embargoed dandisets.
 Reviewers are allowed to see private embargoed dandisets.
 Reviewers are not allowed to make any changes to embargoed dandisets.
 
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -66,3 +66,9 @@ Admins will be at liberty to manually delete or release embargoed dandisets with
 Admins will be at liberty to manually release embargoed dandisets that have expired.
 
 ### TODO policy details
+
+## TODO miscellaneous
+- email the owners of an embargoed dandiset (and admins) when it has been successfully released
+- on the stats page, display the full amount of data stored in the archive, and indicate how much of it is under embargo
+  - e.g., `300 TB total data size (100 TB currently under embargo)`
+  - link the word "embargo" to docs/explanation thereof
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -35,3 +35,15 @@ This will be implemented as a new django-guardian role.
 All read endpoints for assets are permitted for users who are reviewers of the embargoed dandiset.
 All read endpoints for dandisets and versions will treat reviewers the same as owners.
 All write and upload endpoints are unchanged.
+
+## Anonymous read-only access
+Embargoed dandisets need to be shared anonymously with reviewers.
+There are two modes to support: browsing the embargoed dandiset anonymously, and downloading the asset data anonymously with the CLI.
+
+Browsing the embargoed dandiset anonymously can be achieved with a secret URL parameter unique to the dandiset.
+When the web UI detects the secret URL parameter, it passes it as an authorization token header/URL parameter to all API requests.
+The API will treat any request with the correct secret as coming from a reviewer, unless the user is already an owner.
+
+A link to the dandiset with this secret URL parameter is included somewhere on the embargoed dandiset page as a way to share access anonymously.
+
+### TODO how to download anonymously with the CLI?
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -76,6 +76,10 @@ If this becomes an issue, we could:
   This could be as simple as adding a new django app and a new Heroku dyno, or as complex as a Lambda@Edge+CloudFront service.
 * Dynamically provision IAM users with permission to access prefixes in the embargo bucket and distribute access keys to users.
   This would require the API server to manage IAM directly, which is a lot of complexity to manage.
+* Make the embargo bucket publically readable, but not listable.
+  If anyone knows the full S3 object key they have the ability to download the data, but they will not have the ability to search for or scan the bucket for new content.
+  We would then distribute the zarr_id to anyone who needs to access an embargoed zarr archive, giving them permanent read access to it.
+  The downside is that access is not revocable, since we cannot take back the zarr ID from the user or efficiently change the location of the zarr archive.
 
 ## TODO miscellaneous
 - email the owners of an embargoed dandiset (and admins) when it has been successfully unembargoed
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -46,4 +46,23 @@ The API will treat any request with the correct secret as coming from a reviewer
 
 A link to the dandiset with this secret URL parameter is included somewhere on the embargoed dandiset page as a way to share access anonymously.
 
+### TODO implementation details
 ### TODO how to download anonymously with the CLI?
+
+## Embargo period enforcement
+NIH embargoes (and embargos in general) will have an end date to ensure that the data is not secret forever.
+We will enforce that an end date be specified for every new embargoed dandiset, and forcibly release embargoed dandisets that expire.
+
+The MVP collects the NIH award number and stores it in the metadata.
+We can use the NIH API to determine the required release date for the award (?).
+### TODO should we gather the end date during creation time, as we do award number?
+
+We should add scheduled jobs (in the manner of garbage collection) that:
+* Notify admins and owners of any embargoed dandisets without end dates
+* Notify admins and owners of any embargoed dandisets that are approaching their end dates (daily emails for the last week?)
+* Notify admins and owners of any embargoed dandisets that have past their end dates
+
+Admins will be at liberty to manually delete or release embargoed dandisets without end dates.
+Admins will be at liberty to manually release embargoed dandisets that have expired.
+
+### TODO policy details
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -8,7 +8,7 @@ Restricted embargoed dandisets are searchable, viewable, and their metadata is p
 
 Users will select their desired embargo mode using radio buttons when creating a dandiset.
 
-In addition to the dandiset `embargo_status` values from the MVP (`EMBARGOED`, `RELEASING`, and `OPEN`), there is a new value, `RESTRICTED`.
+In addition to the dandiset `embargo_status` values from the MVP (`EMBARGOED`, `UNEMBARGOING`, and `OPEN`), there is a new value, `RESTRICTED`.
 All of the dandiset and version endpoints will only hide `EMBARGOED` dandisets from appearing.
 The asset endpoints will not permitted for anonymous users on embargoed assets that are `EMBARGOED` or `RESTRICTED`.
 
@@ -62,8 +62,8 @@ We should add scheduled jobs (in the manner of garbage collection) that:
 * Notify admins and owners of any embargoed dandisets that are approaching their end dates (daily emails for the last week?)
 * Notify admins and owners of any embargoed dandisets that have past their end dates
 
-Admins will be at liberty to manually delete or release embargoed dandisets without end dates.
-Admins will be at liberty to manually release embargoed dandisets that have expired.
+Admins will be at liberty to manually delete or unembargo embargoed dandisets without end dates.
+Admins will be at liberty to manually unembargo embargoed dandisets that have expired.
 
 ### TODO policy details
 
@@ -78,7 +78,7 @@ If this becomes an issue, we could:
   This would require the API server to manage IAM directly, which is a lot of complexity to manage.
 
 ## TODO miscellaneous
-- email the owners of an embargoed dandiset (and admins) when it has been successfully released
+- email the owners of an embargoed dandiset (and admins) when it has been successfully unembargoed
 - on the stats page, display the full amount of data stored in the archive, and indicate how much of it is under embargo
   - e.g., `300 TB total data size (100 TB currently under embargo)`
   - link the word "embargo" to docs/explanation thereof
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -67,6 +67,16 @@ Admins will be at liberty to manually release embargoed dandisets that have expi
 
 ### TODO policy details
 
+
+## Zarr download
+The MVP embargo zarr download design involves a request to our API server for every file being downloaded from the zarr archive.
+Since zarr archives can contain hundreds of thousands of files, there would be a corresponding number of requests, which might degrade performance of the API server.
+If this becomes an issue, we could:
+* Spin off the zarr download functionality into a separate microservice that can be scaled separately, and would isolate any accidental DDOS effects.
+  This could be as simple as adding a new django app and a new Heroku dyno, or as complex as a Lambda@Edge+CloudFront service.
+* Dynamically provision IAM users with permission to access prefixes in the embargo bucket and distribute access keys to users.
+  This would require the API server to manage IAM directly, which is a lot of complexity to manage.
+
 ## TODO miscellaneous
 - email the owners of an embargoed dandiset (and admins) when it has been successfully released
 - on the stats page, display the full amount of data stored in the archive, and indicate how much of it is under embargo
--- a/doc/design/embargo-full.md
+++ b/doc/design/embargo-full.md
@@ -1,0 +1,3 @@
+# Embargo
+This document takes the Embargo MVP as a predicate and describes the remaining features required for full Embargo support.
+
--- a/doc/design/embargo-mvp.md
+++ b/doc/design/embargo-mvp.md
@@ -75,7 +75,7 @@ If specified, the uploaded data will be sent to the embargoed bucket instead of 
 
 * Get/List dandiset endpoint
 
-  The DandisetViewSet queryset will filter out dandisets with `embargo_status != PUBLIC` that are also not owned by the current user.
+  The DandisetViewSet queryset will filter out dandisets with `embargo_status == PRIVATE` that are also not owned by the current user.
   This will prevent them from showing up in the listing or fetching endpoints.
 
   The `DandisetSerializer` will also be updated to include the `embargo_status` field so that the web client can render the embargoed dandiset appropriately.
--- a/doc/design/embargo-mvp.md
+++ b/doc/design/embargo-mvp.md
@@ -9,7 +9,6 @@ A subsequent document will describe the full suite of features.
 ## Web
 When creating a new dandiset, users will have checkbox to choose whether or not the dandiset should be embargoed.
 If so, the user must also specify an NIH award number (https://era.nih.gov/files/Deciphering_NIH_Application.pdf).
-The award number is an optional argument that is passed to the create dandiset endpoint.
 The award number will be added as a new `Funder` on an automatically generated Contributor Organization in the metadata.
 
 Instead of the normal `Publish` button and version browser, the DLP will instead show a `Release` button.
--- a/doc/design/embargo-mvp.md
+++ b/doc/design/embargo-mvp.md
@@ -77,8 +77,8 @@ A new `EmbargoedAssetBlob` model will be added.
 This behaves the same as a normal `AssetBlob`, but points to the embargo bucket rather than the public bucket.
 It also contains a reference to the `Dandiset` it belongs to.
 
-`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob`.
-An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob`.
+`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob` or `EmbargoedZarrArchive` or `ZarrArchive`.
+An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob` or `EmbargoedZarrArchive`.
 An embargoed `Asset` can only belong to the dandiset named in the `EmbargoedAssetBlob`.
 The automatically populated `Asset` metadata will also set the `access` field to the appropriate value depending on whether or not the asset is embargoed.
 
--- a/doc/design/embargo-mvp.md
+++ b/doc/design/embargo-mvp.md
@@ -77,9 +77,10 @@ If specified, the uploaded data will be sent to the embargoed bucket instead of 
 ## API
 * Create dandiset endpoint
 
-  Add a new optional `award_number` field.
-  If set, the `access` metadata field will be set to `EmbargoedAccess` and the award number will be recorded as a `Funder` on a new Contributor Organization.
+  Add an `embargo` URL parameter to the endpoint which is normally absent.
+  If `embargo` is present (` POST .../dandisets/?embargo`), the `access` metadata field will be set to `EmbargoedAccess`.
   The `embargo_status` of the dandiset will also be set to `EMBARGOED`.
+  The expectation is that the web UI will have requested an award number from the user and recorded it as a `Funder` on a new Contributor Organization in the metadata.
 
 * Get/List dandiset endpoint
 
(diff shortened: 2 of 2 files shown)


doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+The award number is an optional argument that is passed to the create dandiset endpoint.
+The award number will be added as a new `Funder` on an automatically generated Contributor Organization in the metadata.
+
+Instead of the normal `Publish` button and version browser, the DLP will instead show a `Release` button.
+Clicking it will open a confirmation modal informing the user that they are unembargoing, all their data will be publicized, there is no undoing, and that it can take some time for large dandisets.
+Confirming will lock the dandiset for the duration of the release.
+Once the release finishes, the dandiset will be like any other unpublished draft dandiset.
+
+Once created, an embargoed dandiset is only visible or searchable to owners.
+Instead of the `draft` chip, an `embargoed` chip should be used instead.
```
satra, line 21: when you say chip here, i'm assuming you mean ui. will the api endpoint version parameter say `embargo` instead of `draft` or will it remain as `draft` with simply the metadata containing access info. i think i would prefer to continue using "draft" as a version identifier for this.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+## CLI
+The CLI experience will be basically unchanged.
+Uploads to an embargoed dandiset will function exactly the same from an API perspective, the data simply goes to a different location.
+
+### TODO: how to browse/download embargoed zarr files?
+Credentials need to be generated, distributed, and used by the client.
+
+
+# Data storage
+Embargoed assets will be stored in a separate S3 bucket.
```
satra, line 32: separate bucket + separate prefix within bucket for each dandiset?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+This bucket is private and not browseable by the general public.
+
+The API server will use the embargo bucket in exactly the same way it uses the public bucket, but it will only store embargoed data.
+
+When releasing an embargoed dandiset, all asset data for that dandiset is copied to the public bucket.
+
+When uploading a new asset to an embargoed dandiset, the server will first check if that blob has already been uploaded publically.
+If so, the public blob will be used instead of uploading the data again to the embargo bucket.
```
satra, line 40: in case anyone who has questions about privacy based on hash checking, i would say that should not be a concern. the data in dandi are way too specific for bitwise things to match or for someone to try generating data that matches a hash.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+# Django changes
+
+## Models
+The `Dandiset` model will have an `embargo_status` field that is one of `PRIVATE`, `RELEASING`, or `PUBLIC`.
+* `PUBLIC` means that the Dandiset is publically accessible and publishable.
+  This is the state all Dandisets currently have.
+* `PRIVATE` means that the Dandiset is embargoed.
```
satra, line 49: can we semantically match this to the access restrictions, `open`, `embargoed`, `restricted`. the reason i bring this up is that we will eventually have restricted datasets as well, which will be accessible to only those that have access, but are "public".

also we may need to have sub-qualifiers for round 2 of implementation. would that be just extended states, for example: `embargoed_openmetadata`, `restricted_openmetadata`, etc. ?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+This behaves the same as a normal `AssetBlob`, but points to the embargo bucket rather than the public bucket.
+It also contains a reference to the `Dandiset` it belongs to.
+
+`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob`.
+An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob`.
+An embargoed `Asset` can only belong to the dandiset named in the `EmbargoedAssetBlob`.
+The automatically populated `Asset` metadata will also set the `access` field to the appropriate value depending on whether or not the asset is embargoed.
+
+The `Upload` model will have a new optional field `embargoed_dandiset` that points to an embargoed dandiset.
+If specified, the uploaded data will be sent to the embargoed bucket instead of the public bucket, and the `Upload` will create an `EmbargoedAssetBlob` when finalized.
```
satra, line 67: i've thought about one generic change related to this. that upload should require the dandiset id to be sent as a parameter when initiating. this would let allow us to use that information and would not allow an upload without creating a dandiset (since the token authorization must give the uploader ownership on the dandiset).

satra: this would require a breaking change to the CLI, but i think this is an easy change on both the api and cli side.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+## API
+* Create dandiset endpoint
+
+  Add a new optional `award_number` field.
```
satra, line 72: we have to be careful with this, since a dandiset can potentially be the result of multiple awards. thus doing it with the metadata is better. i'm not sure moving from a one to one relationship to a one to many relationship is going to be easy. so the alternate thought here is adding an `award` model and then connecting the dandiset to one  or more awards.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+  Release an embargoed dandiset.
+  
+  Only permitted for owners and admins. If the `embargo_status` is `PUBLIC` or `RELEASING`, return 400.
+
+  Set the `embargo_status` to `RELEASING`, then dispatch the release task.
+
+* Release task
+
+  For every `Asset` with an `EmbargoedAssetBlob` in the dandiset, convert the `EmbargoedAssetBlob` into an `AssetBlob` by moving the data from the embargo bucket to the public bucket.
+  These could be >5GB, so the [multipart copy API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html) must be used.
```
satra, line 102: https://aws.amazon.com/premiumsupport/knowledge-center/s3-large-transfer-between-buckets/

the main thing is to ensure we use something that doesn't change the etag.

satra (commented):
went through the MVP one, since that is more urgent.

doc/design/embargo-mvp.md
mvandenburgh: We could just have an additional `embargo` chip next to the draft one.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+### TODO: how to browse/download embargoed zarr files?
+Credentials need to be generated, distributed, and used by the client.
+
+
+# Data storage
+Embargoed assets will be stored in a separate S3 bucket.
+This bucket is private and not browseable by the general public.
+
+The API server will use the embargo bucket in exactly the same way it uses the public bucket, but it will only store embargoed data.
```
dchiquito, line 35: Technically it's different because of the dandiarchive prefixes

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+  This is the state all Dandisets currently have.
+* `PRIVATE` means that the Dandiset is embargoed.
+  It is not searchable or viewable to non-owners.
+  It is not publishable.
+  Manifest YAML/JSON files will be written to the embargo bucket rather than the public bucket.
+* `RELEASING` means that the Dandiset is currently transitioning from embargoed to public.
+  All modification operations will return a 400 error while the release is in progress.
+  This includes metadata changes and uploads.
+
+A new `EmbargoedAssetBlob` model will be added.
```
dchiquito, line 57: We might not need to do this

doc/design/embargo-mvp.md
dchiquito: This is solely so that we can guarantee that an award is specified when creating a new embargoed dandiset. It will be recorded in the metadata. If you had multiple awards, you would pick one to specify initially, then modify the metadata later to include the rest. That might be a subpar UX, but this is the MVP, so I think it's fine for now, and easy to swap out later.

If it's really important to have multiple awards, we could eventually add the ability to specify multiple when creating the dandiset, or we could remove that requirement completely and make users set all of their awards through the meditor.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+
+### TODO: how to browse/download embargoed zarr files?
+Credentials need to be generated, distributed, and used by the client.
+
+
+# Data storage
+Embargoed assets will be stored in a separate S3 bucket.
+This bucket is private and not browseable by the general public.
+
+Everything in the embargoed bucket will be prefixed with a dandiset identifier.
```
waxlamp, line 36: Trying to resolve some possible ambiguity in this sentence:

```suggestion
Each dandiset stored in the embargoed bucket will be prefixed with a dandiset identifier.
```

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+This document is dedicated to the minimum viable product.
+A subsequent document will describe the full suite of features.
+
+
+# User experience
+
+## Web
+When creating a new dandiset, users will have checkbox to choose whether or not the dandiset should be embargoed.
+If so, the user must also specify an NIH award number (https://era.nih.gov/files/Deciphering_NIH_Application.pdf).
+The award number is an optional argument that is passed to the create dandiset endpoint.
```
waxlamp, line 12: The award number is optional in terms of the endpoint, but required if a user wishes to create an embargoed dataset. The optionality is tied directly to whether or not the endpoint creates an embargoed or open dandiset, right?

That is, you never have to supply an award number, but you get an embargoed dandiset if and only if you supply one. Is that correct?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+The award number will be added as a new `Funder` on an automatically generated Contributor Organization in the metadata.
+
+Instead of the normal `Publish` button and version browser, the DLP will instead show a `Release` button.
+Clicking it will open a confirmation modal informing the user that they are unembargoing, all their data will be publicized, there is no undoing, and that it can take some time for large dandisets.
+Confirming will lock the dandiset for the duration of the release.
+Once the release finishes, the dandiset will be like any other unpublished draft dandiset.
```
waxlamp, line 18: Need to make sure the release is all or nothing, which might be a little bit tricky since it will generally involve moving a lot of data from bucket to bucket.

One approach may be to perform the full copy, validate it, and only then delete the embargoed bucket assets and finish "releasing" the dandiset.

waxlamp: We should maybe also email the owner/admins when the release is done?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+Once created, an embargoed dandiset is only visible or searchable to owners.
+Instead of the `draft` chip, an `embargoed` chip should be used on the dandiset listing pages.
+These dandisets+chips will only appear in the listings to owners of the embargoed dandisets.
+
+## CLI
+The CLI experience will be basically unchanged.
+Uploads to an embargoed dandiset will function exactly the same from an API perspective, the data simply goes to a different location.
+
+### TODO: how to browse/download embargoed zarr files?
+Credentials need to be generated, distributed, and used by the client.
```
waxlamp, line 29: Is this going to be part of the MVP?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+  `EmbargoedAssetBlob`s should also be checked for deduplication, but only within an embargoed dandiset.
+  An embargoed dandiset should use the same `EmbargoedAssetBlob` if the same file appears in multiple places, but two embargoed dandisets should upload the same data twice if they both contain the same file.
+
+  Return error 400 if `dandiset.embargo_status == RELEASING`.
+
+* TODO zarr upload
+
+* stats_view
+  
+  The total size value should include the size of `EmbargoedAssetBlob`s as well as `AssetBlob`s.
```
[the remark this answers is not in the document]
waxlamp: For the MVP, let's not worry about this. I think in the full version, it would be nice to see something like "300 TB published (of which 100 TB currently under embargo)" with `embargo` linked to an explanation.

waxlamp (commented):
Just some questions/suggestions.

doc/design/embargo-mvp.md
satra: sounds about right.

doc/design/embargo-mvp.md
satra: since we would essentially copy from blob store to blob store, there are several s3 ways of doing that. the key is to preserve the s3 generated etag, which should also be possible. this would mean validation is simply checking the etag post copy.

doc/design/embargo-mvp.md
satra: i think credentials are part of mvp almost independent of zarr, but essential for zarr. since people should be able to access the endpoint without always going through pre-signed urls right? for example to load an hdf5 file via streaming one would need the s3 path. a presigned url doesn't work for s3 read access for all tools.

doc/design/embargo-mvp.md
satra: as long as embargoed means `authorization required` then the restricted explanation makes sense. so the alternative would be to change:
`completely locked down` --> `authorization required to access`
`embargoed asset data` --> `authorization required to access asset data`

doc/design/embargo-mvp.md
satra: if we are simply using the metadata model instead of adding a field, the multiple award thing works already. the only addition we would add would be in the validator that the embargoed dandiset must have an award associated with it. (ideally we would even check that the nih api returns a valid signal for the award). and the ui could also be updated to simply interface with the nih api. (i would say this isn't essential for MVP). but whether a database model field is created or simply the metadata is used should be decided and my preference would be for the metadata.

doc/design/embargo-mvp.md
satra: didn't quite see this around ~115. my suggestion was simply that this would be a generic endpoint change to require dandiset as a parameter to the upload initialize endpoint.

doc/design/embargo-mvp.md
waxlamp, line 12: Suggest cutting this sentence to avoid implementation confusion later.

```suggestion
```

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+  
+  Only permitted for owners and admins. If the `embargo_status` is `PUBLIC` or `RELEASING`, return 400.
+
+  Set the `embargo_status` to `RELEASING`, then dispatch the release task.
+
+* Release task
+
+  For every `Asset` with an `EmbargoedAssetBlob` in the dandiset, convert the `EmbargoedAssetBlob` into an `AssetBlob` by moving the data from the embargo bucket to the public bucket.
+  These could be >5GB, so the [multipart copy API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html) must be used.
+  The ETag and checksum must remain undisturbed; the only change should be where the data is stored.
```
waxlamp, line 112:
```suggestion
  The ETag and checksum must remain undisturbed; the only change should be where the data is stored.
  Verify that the resulting unembargoed assets match one-for-one (in the database) with the embargoed assets that were copied.
```

doc/design/embargo-full.md
```diff
@@ -0,0 +1,68 @@
...
+
+We should add scheduled jobs (in the manner of garbage collection) that:
+* Notify admins and owners of any embargoed dandisets without end dates
+* Notify admins and owners of any embargoed dandisets that are approaching their end dates (daily emails for the last week?)
+* Notify admins and owners of any embargoed dandisets that have past their end dates
+
+Admins will be at liberty to manually delete or release embargoed dandisets without end dates.
+Admins will be at liberty to manually release embargoed dandisets that have expired.
+
+### TODO policy details
```
waxlamp, line 68:
```suggestion
### TODO policy details

## TODO miscellaneous
- email the owners of an embargoed dandiset (and admins) when it has been successfully released
```

doc/design/embargo-mvp.md
waxlamp: Added some text to other parts of the doc to cover all this, resolving.

doc/design/embargo-mvp.md
waxlamp: Daniel added text below (line 36) to cover this.

doc/design/embargo-mvp.md
waxlamp: Sounds good. Resolving (unresolve if there's more to discuss).

doc/design/embargo-mvp.md
waxlamp: After discussing with Daniel, it seems like having a separate emabargoed asset blob model is worth it.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,141 @@
...
+* upload_initialize_view
+
+  An optional field `embargoed_dandiset` will be available on the serializer.
+  If specified, it will be passed to the `Upload` object.
+  This will mean the final upload will result in an `EmbargoedAssetBlob`.
+
+  Even if `embargoed_dandiset` is specified, if an `AssetBlob` with a matching checksum already exists, return it instead of uploading embargoed data.
+  This means that an embargoed dandiset can contain both embargoed and unembargoed assets.
+
+  `EmbargoedAssetBlob`s should also be checked for deduplication, but only within an embargoed dandiset.
```
waxlamp, line 132:
```suggestion
  `EmbargoedAssetBlob`s should also be checked for deduplication, but only within an embargoed dandiset.
  This is to keep the permission model clean for owners of different embargoed dandisets that might contain the same asset.
```

doc/design/embargo-mvp.md
waxlamp: The relevant text can be seen [here](https://github.com/dandi/dandi-api/blob/ad8ac167cd27004de0119b54f9833e607cebf52c/doc/design/embargo-mvp.md?plain=1#L123-L133). This proposal keeps the parameter optional; if we want to make it required that would be out of scope for this MVP effort. I have filed #628 to track this.

waxlamp: (Resolving this thread; please unresolve if you want to continue the discussion, or feel free to discuss further at #628.)

doc/design/embargo-mvp.md
waxlamp: We agree largely with what you wrote, @satra.

To be explicit: we will design the UI to have a checkbox for creating an embargoed dataset; that will in turn make a required form field for an NIH award number appear. The client will craft a metadata blob that includes the award, and send all this information to the creation endpoint. The creation endpoint will do all the nitty gritty stuff it needs to do thereafter (as described in this document).

If it is imperative to support multiple NIH awards from the jump, it should not be too difficult to design the UI to require one award number, and show a plus button to add more. Then, the "craft a metadata blob" idea from above will simply iterate through a list of award numbers. Otherwise, we can expand to this capability when we implement the full design.

(In the full design, we will also consider the idea of validating an embargoed dandiset--the idea is that an owner could put a valid NIH award number in to get past the initial hurdle, then remove it later in order to have an indefinitely embargoed secret dandiset; by "validating" for qualities such as "does this dandiset have a valid award number", etc., we could simply "freeze" any invalid dandisets, thus preventing use until they become valid. This shouldn't be necessary for the MVP.)

satra: just to clarify i don't think it's necessary for the create dandiset ui to have support for multiple awards, just one is sufficient there, and may be a note that additional awards can be added through the metadata editor after the dandiset is created. my comment was more specific to adding a db field that supports only one award.

waxlamp: I think there was some confusion in the original language: there was never an intent to track a single award number in the DB, separately from the metadata (but rather, a serializer field in the endpoint for dandiset creation). I think we're all on the same page now.

doc/design/embargo-mvp.md
dchiquito: I renamed the fields accordingly.

doc/design/embargo-full.md
```diff
@@ -0,0 +1,68 @@
+# Embargo
+This document takes the Embargo MVP as a predicate and describes the remaining features required for full Embargo support.
+
+## Non-private embargoed dandisets
+In the MVP, all embargoed dandisets are invisible to non-owners.
+There should be an additional embargo mode, "Visible".
+Visible embargoed dandisets are searchable, viewable, and their metadata is public to everyone, but their asset data and metadata is not.
```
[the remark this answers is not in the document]
dchiquito: This sounds like what I described for restricted dandisets. Do you mean that restricted dandisets should be searchable and visible publicly, but their metadata should be restricted? And the `openmetadata` suffix means that the metadata is public?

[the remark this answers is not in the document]
satra: restricted dandisets could come in two flavors: metadata part of search, metadata not part of search (only limited metadata like title part of search). typically restricted == authorization required (i.e. these would likely stay in the embargoed bucket).

doc/design/embargo-mvp.md
dchiquito: I updated the design of the create dandiset endpoint, it now takes the award information via the metadata as requested.
Resolving

doc/design/embargo-full.md
```diff
@@ -67,6 +67,16 @@ Admins will be at liberty to manually release embargoed dandisets that have expi
 
 ### TODO policy details
 
+
+## Zarr download
+The MVP embargo zarr download design involves a request to our API server for every file being downloaded from the zarr archive.
+Since zarr archives can contain hundreds of thousands of files, there would be a corresponding number of requests, which might degrade performance of the API server.
+If this becomes an issue, we could:
+* Spin off the zarr download functionality into a separate microservice that can be scaled separately, and would isolate any accidental DDOS effects.
+  This could be as simple as adding a new django app and a new Heroku dyno, or as complex as a Lambda@Edge+CloudFront service.
```
yarikoptic, line 76: well, you know my opinion - overhead from such per-file operations would be too large and will result not only in necessity to 'scale up' but also in a significant performance impairment on access to such zarr file.

But it is all speculation at this point since I have not done any benchmarking. Ideally we should compare performance (e.g. to even just download or perform some prototypical operation which would need some "sparse access" to zarr) accessing a prototypical zarr "file" between
- access on public bucket
- access on private bucket with a specific IAM
- access on private bucket through "mint a URL per file" (suggested here)

My wild uninformed bet is that a "full download" of zarr would be at least x10 less efficient in the latter than former case even in case of "minting service" to not be overloaded.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+  Manifest YAML/JSON files will be written to the embargo bucket rather than the public bucket.
+* `RELEASING` means that the Dandiset is currently transitioning from embargoed to public.
+  All modification operations will return a 400 error while the release is in progress.
+  This includes metadata changes and uploads.
+
+A new `EmbargoedAssetBlob` model will be added.
+This behaves the same as a normal `AssetBlob`, but points to the embargo bucket rather than the public bucket.
+It also contains a reference to the `Dandiset` it belongs to.
+
+`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob`.
```
yarikoptic, line 79:
```suggestion
`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob` or `EmbargoedZarrArchive` or `ZarrArchive`.
```
? (combinatorics start to creep in)

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob`.
+An embargoed `Asset` can only belong to the dandiset named in the `EmbargoedAssetBlob`.
+The automatically populated `Asset` metadata will also set the `access` field to the appropriate value depending on whether or not the asset is embargoed.
+
+A new `EmbargoedUpload` model will be added.
+This is just like the normal `Upload` model, but the uploaded data will be sent to the embargoed bucket instead of the public bucket, and the `Upload` will create an `EmbargoedAssetBlob` when finalized.
+It will also contain a reference to the `Dandiset` it belongs.
+
+A new `EmbargoedZarrArchive` model will be added in the same vein as `EmbargoedAssetBlob`.
```
yarikoptic, line 88: should we use `*ZarrFile` terms instead of `*ZarrArchive` to stay inline with below `EmbargoedZarrUploadFile` etc?

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+* `RELEASING` means that the Dandiset is currently transitioning from embargoed to public.
+  All modification operations will return a 400 error while the release is in progress.
+  This includes metadata changes and uploads.
+
+A new `EmbargoedAssetBlob` model will be added.
+This behaves the same as a normal `AssetBlob`, but points to the embargo bucket rather than the public bucket.
+It also contains a reference to the `Dandiset` it belongs to.
+
+`Asset`s will point to exactly one `AssetBlob` or `EmbargoedAssetBlob`.
+An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob`.
```
yarikoptic, line 80:
```suggestion
An `Asset` will be considered "embargoed" if it has an `EmbargoedAssetBlob` or `EmbargoedZarrArchive`.
```
?

yarikoptic (commented):
Some old comments I found not "finish"ed into a review + some new ones regarding zarr (also on slack I just asked a question about linkeage of Zarr to Assets which is also reflected in the comments here)

doc/design/embargo-full.md
satra: i would very much say "mint a url per file" is not going to be efficient for any downstream application like neuroglancer or xarray-based operations which do their own lookup and caching based on awareness of a zarr dataset. i'm going to suggest that this is perhaps not an acceptable MVP, and just like accessing any other s3 blob we rely on s3 services for zarr.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+This document is dedicated to the minimum viable product.
+A subsequent document will describe the full suite of features.
+
+
+# User experience
+
+## Web
+When creating a new dandiset, users will have checkbox to choose whether or not the dandiset should be embargoed.
+If so, the user must also specify an NIH award number (https://era.nih.gov/files/Deciphering_NIH_Application.pdf).
+The award number is an optional argument that is passed to the create dandiset endpoint.
```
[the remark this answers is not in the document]
waxlamp: It's optional for the endpoint--if it's not passed, the endpoint will create a non-embargoed dandiset.

For the MVP, we're focusing on NIH awardees. I assume later we will have other institutions as well, but to get this going--just NIH.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+The award number will be added as a new `Funder` on an automatically generated Contributor Organization in the metadata.
+
+Instead of the normal `Publish` button and version browser, the DLP will instead show a `Release` button.
+Clicking it will open a confirmation modal informing the user that they are unembargoing, all their data will be publicized, there is no undoing, and that it can take some time for large dandisets.
```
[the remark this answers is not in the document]
waxlamp: We probably need to revisit all our terminology comprehensively at some point. I agree that "release" is a bit too general purpose. Perhaps we can use the word "unembargo" here--it would be quite clear in context.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+Credentials need to be generated, distributed, and used by the client.
+
+
+# Data storage
+Embargoed assets will be stored in a separate S3 bucket.
+This bucket is private and not browseable by the general public.
+
+The API server will use the embargo bucket in exactly the same way it uses the public bucket, but it will only store embargoed data.
+
+When releasing an embargoed dandiset, all asset data for that dandiset is copied to the public bucket.
```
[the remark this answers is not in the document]
waxlamp: The idea is that garbage collection will vacuum it up on its next pass.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+# Django changes
+
+## Models
+The `Dandiset` model will have an `embargo_status` field that is one of `PRIVATE`, `RELEASING`, or `PUBLIC`.
+* `PUBLIC` means that the Dandiset is publically accessible and publishable.
+  This is the state all Dandisets currently have.
+* `PRIVATE` means that the Dandiset is embargoed.
+  It is not searchable or viewable to non-owners.
```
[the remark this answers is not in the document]
waxlamp: Searchable and visible to owners, [yes](https://en.wikipedia.org/wiki/Exception_that_proves_the_rule#Proving_the_existence_of_the_rule).

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+  It is not publishable.
+  Manifest YAML/JSON files will be written to the embargo bucket rather than the public bucket.
+* `RELEASING` means that the Dandiset is currently transitioning from embargoed to public.
```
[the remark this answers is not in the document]
waxlamp: Or `UNEMBARGOING`.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+The `Upload` model will have a new optional field `embargoed_dandiset` that points to an embargoed dandiset.
+If specified, the uploaded data will be sent to the embargoed bucket instead of the public bucket, and the `Upload` will create an `EmbargoedAssetBlob` when finalized.
+
+## API
+* Create dandiset endpoint
+
+  Add a new optional `award_number` field.
+  If set, the `access` metadata field will be set to `EmbargoedAccess` and the award number will be recorded as a `Funder` on a new Contributor Organization.
+  The `embargo_status` of the dandiset will also be set to `PRIVATE`.
```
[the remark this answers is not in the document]
waxlamp: I don't think this is a good idea. Naming this field `access` will cause almost certainly cause confusion for us later, since e.g. `UNEMBARGOING`/`MAKING_PUBLIC` is *not* a metadata access level.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+* create asset, update metadata, and any other dandiset/version modification endpoints:
+
+  Return error 400 if `dandiset.embargo_status == RELEASING`.
+
+* New endpoint: `POST /api/dandisets/{dandiset_id}/release`
+
+  Release an embargoed dandiset.
+  
+  Only permitted for owners and admins. If the `embargo_status` is `PUBLIC` or `RELEASING`, return 400.
```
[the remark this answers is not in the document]
waxlamp: Technically, 405 refers to the *HTTP* method not being allowed, so let's keep this at 400 (it is a client error to request this operation without the proper authorization). 403 might be appropriate here as well.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+* Get/List dandiset endpoint
+
+  The DandisetViewSet queryset will filter out dandisets with `embargo_status == PRIVATE` that are also not owned by the current user.
+  This will prevent them from showing up in the listing or fetching endpoints.
+
+  The `DandisetSerializer` will also be updated to include the `embargo_status` field so that the web client can render the embargoed dandiset appropriately.
+
+* publish version endpoint
+
+  Return error 400 if `dandiset.embargo_status != PUBLIC`.
```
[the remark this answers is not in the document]
waxlamp: Same as above, 400 is fine, or perhaps 403.

doc/design/embargo-mvp.md
```diff
@@ -0,0 +1,131 @@
...
+
+* create asset, update metadata, and any other dandiset/version modification endpoints:
+
+  Return error 400 if `dandiset.embargo_status == RELEASING`.
```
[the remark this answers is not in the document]
waxlamp: 503 is almost certainly the wrong thing--that indicates that the whole server is (temporarily) borked. I don't think we need to send a Retry-After here; we can instead send a reason in the response body and allow the client to try again after a while.

doc/design/embargo-mvp.md
waxlamp: I think there's a meaningful distinction: the "archive" is the full asset, while the "file" refers to a single file of the zarr (as it is uploaded). @dchiquito?

doc/design/embargo-full.md
waxlamp: Perhaps we can discuss this one live, but for me the current position is: we need data. That is, for the MVP effort we may be able to try out the straightforward "proxy service" and see how it does. Brian Helba has already developed some intuition that some simple caching techniques could probably provide a speedup over the base case. But again... data needed.

yarikoptic: We used to have a sample zarr under s3://dandiarchive/largeuploads/blobs/150slab.zarr/ but then we cleaned it up.  @satra - do you still have it somewhere?  would be also a good test for `upload` functionality

doc/design/embargo-mvp.md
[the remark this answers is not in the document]
yarikoptic: might be worth explicit wording?

doc/design/embargo-mvp.md
yarikoptic: eh... I always disliked referring to `.zarr/` directory as "zarr file" but it grew on me I guess.
Also as we refer to "files" as "blobs", may be then it still could be `ZarrFile` (the "archive") and `ZarrBlob` (an individual file within Zarr)?  or should we may be check how zarr calls itself and individual files? ;)

doc/design/embargo-mvp.md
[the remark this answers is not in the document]
yarikoptic: indeed -- 503 is no good, thank you @waxlamp for clarifying.  
I still think that some explicit, server provided "Retry-After" is better than some "a while" (a second? minute? ...) which might cause some eager clients retry way too often.  But it seems there is no other suitable response code with Retry-After (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After). And since we will not be monitoring individual connections etc, we can't use [429 "Too many requests"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) really

doc/design/embargo-mvp.md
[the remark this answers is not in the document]
dchiquito: I renamed "Release" to "Unembargo"

doc/design/embargo-mvp.md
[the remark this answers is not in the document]
dchiquito: I added some some explicit wording

doc/design/embargo-mvp.md
dchiquito: >  referring to .zarr/ directory as "zarr file"
That's the opposite of what we're doing, the entire zarr directory is the "zarr archive", the files it contains are "zarr files".

Looks like zarr might refer to the component files as "chunks", but I'm not sure if that might also refer to subdirectories.

PR #620, opened 2021-12-09
User admin dashboard page

### LINKED ISSUE #575: Dashboard page that lists pending users
https://github.com/dandi/dandi-api/pull/568#discussion_r734640814

mvandenburgh:
We already have a list of *approved* users in the dashboard, so to address #575 I moved that list to a separate dashboard page and added a filter for user `status`. I also added a new `DashboardMixin` that can be subclassed by all dashboard pages to ensure only admins are allowed to view them. I think this is a cleaner approach to authentication over raising an exception in `get_context_data` now that we have multiple dashboard views.

Closes #575.

### DIFF
--- a/dandiapi/api/dashboard.py
+++ b/dandiapi/api/dashboard.py
@@ -1,4 +1,5 @@
 from allauth.socialaccount.models import SocialAccount
+from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied
 from django.db.models import Exists, OuterRef, Q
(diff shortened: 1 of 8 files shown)


dandiapi/api/dashboard.py
```diff
@@ -1,22 +1,28 @@
...
+from django_filters.views import FilterView
 
+from dandiapi.api.filters import UserStatusFilter
 from dandiapi.api.mail import send_approved_user_message, send_rejected_user_message
 from dandiapi.api.models import Asset, AssetBlob, Upload, UserMetadata, Version
 from dandiapi.api.views.users import social_account_to_dict
 
 
-class DashboardView(TemplateView):
+class DashboardMixin(LoginRequiredMixin, UserPassesTestMixin):
```
dchiquito, line 17: Use `UserPassesTestMixin` will redirect to the login page if the user doesn't pass the test instead of returning an unauthorized status code. Was that the intention?

dchiquito (approved):

dandiapi/api/dashboard.py
mvandenburgh: Yes, at leas that's the behavior of the existing dashboard view (https://github.com/dandi/dandi-api/blob/master/dandiapi/api/dashboard.py#L18-L19 causes the redirect).

PR #621, opened 2021-12-09
Temporarily pin django~=3.2

mvandenburgh:
This will make CI pass again. Once https://github.com/jazzband/django-oauth-toolkit/issues/1037 is fixed we can revert this and upgrade to Django 4.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,9 @@ setup(
     install_requires=[
         'celery',
         'dandischema==0.4.3',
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
I'll add that I think all such versions should be specified with `~=`, and we should only upgrade major versions if there's a positive reason to do so. In Django's case, there probably is (e.g., new database interaction features), but the risks in upgrading need to be considered as well.

PR #623, opened 2021-12-09
Fix user approval page styling when browser is in dark mode

mvandenburgh:
I believe @satra mentioned this before.

The user approval admin page is difficult to read when using the "dark mode" browser setting. This is because the CSS framework we're using for this view doesn't support this mode (yet?). To fix this, I added a CSS media query to darken the background color when the user is in dark mode. The page remains the same for users not using dark mode.

Current page in dark mode:
before

New page in dark mode:

after


waxlamp (commented):
👍🏾

PR #629, opened 2021-12-15
Remove extra character displayed on user dashboard page

mvandenburgh:


waxlamp (approved):
Hah, where did that extra `d` even come from?

mvandenburgh:
> Hah, where did that extra `d` even come from?

Actually the automatic HTML tag renaming in VS code has been a little wonky for me recently, and since that particular line ends with a `td` maybe that is what caused the extra `d`. :man_shrugging:

PR #640, opened 2021-12-20
Embargo models

### LINKED ISSUE #634: Update models to implement embargoed dandiset concept
[design doc](https://github.com/dandi/dandi-api/blob/96c85ca5b3c1932b450cb00d4c375f0a176e2261/doc/design/embargo-mvp.md?plain=1#L53-L78)

dchiquito:
Resolves #634

* Move the storage generator function to `storage.py` instead of having everything import it from models/asset.py
  * This involved editing some migrations, but I think it's safe since the function implementation didn't change
* Setting for setting the embargo bucket name/prefix
* Add `EmbargoedAssetBlob` and `EmbargoedUpload` models
* Add `embargo_status` field to `Dandiset`
* Migration and fixtures

I opted to create abstract base classes `BaseAssetBlob` and `BaseUpload` that have most of the model logic. Their children basically just have a `blob` field that points to the correct storage, and any non-general code.

I also skipped zarr models for now, my first objective is embargoing normal assets since zarr isn't even fully adopted yet.

Tests will fail until https://github.com/dandi/dandi-cli/pull/854 is merged

Requires https://github.com/dandi/dandi-infrastructure/pull/105 before merging, since a new setting must be set by terraform


dandiapi/api/migrations/0001_initial.py
```diff
@@ -74,8 +75,8 @@ class Migration(migrations.Migration):
                     'blob',
                     models.FileField(
                         blank=True,
-                        storage=dandiapi.api.models.asset.get_asset_blob_storage,
-                        upload_to=dandiapi.api.models.asset.get_asset_blob_prefix,
+                        storage=dandiapi.api.storage.get_storage,
+                        upload_to=dandiapi.api.storage.get_storage_prefix,
```
mvandenburgh, line 79: Why modify a past migration instead of making a new one? Will this cause problems with our heroku deployment?

dandiapi/api/storage.py
```diff
@@ -101,3 +102,19 @@ def create_s3_storage(bucket_name: str) -> Storage:
...
+def get_storage() -> Storage:
+    return create_s3_storage(settings.DANDI_DANDISETS_BUCKET_NAME)
+
+
+def get_storage_prefix(instance: Any, filename: str) -> str:
+    return f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}{filename}'
+
+
+def get_embargo_storage() -> Storage:
+    return create_s3_storage(settings.DANDI_DANDISETS_BUCKET_NAME)
```
mvandenburgh, line 116:
```suggestion
    return create_s3_storage(settings.DANDI_DANDISETS_EMBARGO_BUCKET_NAME)
```

dandiapi/api/storage.py
```diff
@@ -101,3 +102,19 @@ def create_s3_storage(bucket_name: str) -> Storage:
...
+
+
+def get_embargo_storage_prefix(instance: Any, filename: str) -> str:
+    return f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}{filename}'
```
mvandenburgh, line 120:
```suggestion
    return f'{settings.DANDI_DANDISETS_EMBARGO_BUCKET_PREFIX}{filename}'
```

mvandenburgh (commented):
Looks like a great start!

dandiapi/api/migrations/0001_initial.py
mvandenburgh: Oh I missed your comment on https://github.com/dandi/dandi-api/pull/640/commits/99464659b8d2f1601e2faf1f781faa7b6222ceb0, I think that answers my question.

dchiquito: All the migration files are run in order when Django loads, so if they have import errors, the app crashes. Generally you shouldn't import anything in migrations to avoid exactly this problem. We ran into it because we wanted a handy helper function that generates storages.

dandiapi/api/storage.py
dchiquito: :fearful:

Issue #628, opened 2021-12-15
Require dandiset parameter for asset upload endpoint


waxlamp:
> i've thought about one generic change related to this. that upload should require the dandiset id to be sent as a parameter when initiating. this would let allow us to use that information and would not allow an upload without creating a dandiset (since the token authorization must give the uploader ownership on the dandiset).

_Originally posted by @satra in https://github.com/dandi/dandi-api/pull/619#r766782896_

waxlamp:
This is completed.

PR #647, opened 2021-12-23
Embargo upload asset endpoints

### LINKED ISSUE #644: Implement embargoed asset upload
https://github.com/dandi/dandi-api/blob/65dc262ea565d03b089cd435c8a0d912638a95ae/doc/design/embargo-mvp.md?plain=1#L137-L155

> upload_initialize_view
>
> An optional field `embargoed_dandiset` will be available on the serializer.
  If specified, it will be passed to the `Upload` object.
  This will mean the final upload will result in an `EmbargoedAssetBlob`.
>
>  Even if `embargoed_dandiset` is specified, if an `AssetBlob` with a matching checksum already exists, return it instead of uploading embargoed data.
  This means that an embargoed dandiset can contain both embargoed and unembargoed assets.
>
>  `EmbargoedAssetBlob`s should also be checked for deduplication, but only within an embargoed dandiset.
  This is to keep the permission model clean for owners of different embargoed dandisets that might contain the same asset.
  An embargoed dandiset should use the same `EmbargoedAssetBlob` if the same file appears in multiple places, but two embargoed dandisets should upload the same data twice if they both contain the same file.
>
>  Return error 400 if `dandiset.embargo_status == UNEMBARGOING`.

Also ensure that embargoed assets don't show up in listings for non-owners

dchiquito:
* Add a new required `dandiset` option to the upload initialize endpoint. This must now always be specified. If the dandiset is embargoed, an `EmbargoedUpload` will be created instead of an `Upload`.
* Use `EmbargoedUploads` in the upload complete and validate endpoints in addition to `Uploads`.
* Make embargoed uploads go to the correct prefix in the correct bucket, with the dandiset identifier prefixed.

Fixes #644


satra:
@dchiquito - i thought we would implement the suggestion in #628 as opposed to anything that requires specific options for embargo?

dandiapi/api/models/upload.py
```diff
@@ -50,18 +51,14 @@ class Meta:
...
-    @classmethod
-    def object_key(cls, upload_id):
-        upload_id = str(upload_id)
-        return (
-            f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}'
-            f'blobs/{upload_id[0:3]}/{upload_id[3:6]}/{upload_id}'
-        )
+    @abstractclassmethod
+    def object_key(cls, upload_id, **kwargs):  # noqa: N805
+        ...
```
mvandenburgh, line 56: We use `pass` elsewhere consistently in the codebase
```suggestion
        pass
```

dandiapi/api/views/upload.py
```diff
@@ -162,15 +187,20 @@ def upload_complete_view(request: Request, upload_id: str) -> HttpResponseBase:
     request_serializer.is_valid(raise_exception=True)
     parts: List[TransferredPart] = request_serializer.save()
 
-    upload = get_object_or_404(Upload, upload_id=upload_id)
+    try:
+        upload = Upload.objects.get(upload_id=upload_id)
+    except Upload.DoesNotExist:
+        upload = get_object_or_404(EmbargoedUpload, upload_id=upload_id)
+        if not request.user.has_perm('owner', upload.dandiset):
+            raise Http404()
```
mvandenburgh, line 195:
```suggestion
      upload = get_object_or_404(BaseUpload, upload_id=upload_id)
```

Would this work? Maybe you already thought of this and Django doesn't do polymorphism like that

dandiapi/api/views/upload.py
```diff
@@ -199,7 +229,12 @@ def upload_validate_view(request: Request, upload_id: str) -> HttpResponseBase:
 
     Also starts the asynchronous checksum calculation process.
     """
-    upload = get_object_or_404(Upload, upload_id=upload_id)
+    try:
+        upload = Upload.objects.get(upload_id=upload_id)
+    except Upload.DoesNotExist:
+        upload = get_object_or_404(EmbargoedUpload, upload_id=upload_id)
+        if not request.user.has_perm('owner', upload.dandiset):
+            raise Http404()
```
mvandenburgh, line 237:
```suggestion
      upload = get_object_or_404(BaseUpload, upload_id=upload_id)
```
Same question as above

mvandenburgh (changes requested):

dandiapi/api/views/upload.py
mvandenburgh: Actually, I guess this doesn't make sense to do because of that `if` statement checking if the user is an owner on the embargoed dandiset.

dandiapi/api/models/upload.py
dchiquito: All the docs on the `@abstractmethod` page use `...` so I thought it was some magically syntax lol, I didn't know it was synonymous with `pass`.

dandiapi/api/models/upload.py
```diff
@@ -126,6 +135,17 @@ class EmbargoedUpload(BaseUpload):
         Dandiset, related_name='embargoed_uploads', on_delete=models.CASCADE
     )
 
+    @classmethod
+    def object_key(cls, upload_id, **kwargs):
+        upload_id = str(upload_id)
+        dandiset_identifier = kwargs.get('dandiset')
+        dandiset_prefix = f'{dandiset_identifier}/' if dandiset_identifier is not None else ''
```
mvandenburgh, line 142: Isn't a `dandiset_identifier` **required** for an `EmbargoedUpload`?

https://github.com/dandi/dandi-api/blob/embargo-design-doc/doc/design/embargo-mvp.md#models

dchiquito: Good point.
I think I'm actually going to make `dandiset` a field on `Upload` as well as `EmbargoedUpload`, since as of https://github.com/dandi/dandi-api/pull/647/commits/3bbc1e871103d2130dff798d66798c58058cc06f it is required on all upload requests. e might as well track the dandiset for all uploads since we have the information.

mvandenburgh (approved):
The netlify build failures are occurring because this branch was created before the `web/` directory existed - they can be safely ignored.

dandiapi/api/models/upload.py
yarikoptic: As asked elsewhere - if that would render already released/used dandi clients unable to upload to noon embargoed datasets, then let's not break API here

dchiquito: That seems to me to be contradictory to https://github.com/dandi/dandi-api/issues/628.

Also, if we were to do that, what would happen if a user with an old CLI version naively tried to upload data to an embargoed dandiset? The upload would create a valid public AssetBlob, since the API has no way of knowing that the client intends for it to be embargoed. The CLI would then try to add that AssetBlob to the embargoed dandiset, which would also succeed, because public data is allowed to be included in embargoed dandisets. The result would be all the data uploaded publicly to the embargoed dandiset.

ping @yarikoptic @satra @waxlamp

satra: @yarikoptic - i think this simple change would also ensure that no person could upload a file without creating a dandiset first and having ownership of this dandiset (@dchiquito - is that checked). 

i think we should introduce this breaking change. i would be fine with an interim process, which suspends this requirement for non-embargoed dandisets, but is there really a reason why we could not deprecate former clients?

satra: i think @dchiquito comment is even more relevant as to why this should be introduced.

dchiquito: (ownership is required to upload)

yarikoptic: yes, @dchiquito 's example is a good one.
But then how such migration would be implemented (besides clicking "release" simultaneously and somehow enforcing)?
- would current dandi-api (didn't try) allow (just ignore) `dandiset` parameter and not error out?
- if `yes` (no error) -- we should rush out dandi-cli release which would provide dandiset id, so whenever this is released, already released/installed client would be ready

dchiquito: The server will currently ignore the `dandiset` parameter, so the CLI should be released first.

yarikoptic: ok, @jwodder already furnished a PR https://github.com/dandi/dandi-cli/pull/863 .  The only question is to either declare all prior versions "not good" or wait a little more (e.g. to may be discover other gotchas needing similar action) if deployment of embargo support is not imminent yet.

PR #649, opened 2021-12-23
Include request body in zarr file delete swagger entry

dchiquito:


dandiapi/api/views/zarr.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from django.db import transaction
+from django.db.models.fields.related import ManyToManyField
```
waxlamp, line 4: This doesn't look like it's needed for the change set. Was it missing before or something?

waxlamp (approved):
Looks fine.

dandiapi/api/views/zarr.py
dchiquito: I was trying to type `many=True` into the serializer and it accidentally auto-import `ManyToManyField`

PR #650, opened 2021-12-27
Fix zarr s3 upload

dchiquito:
* Specify ACL=`bucket-owner-full-control`. I'm not sure why it is the way it is, but AWS requires that the client declare who will be the owner of the uploaded object. The staging (and also production) buckets have a rule that only requests with ACL=`bucket-owner-full-control` are permitted, which effectively makes it so that uploaded objects are always owned by the bucket owner. The downside of AWS doing it this way is that clients need to explicitly do this or their requests are not permitted, which is awful UX.

* Fix a bug where minio presigned zarr upload urls pointed to `http://minio:9000/...`, the name of the `minio` docker container.

### DIFF
--- a/dandiapi/api/models/zarr.py
+++ b/dandiapi/api/models/zarr.py
@@ -73,6 +73,7 @@ class ZarrUploadFile(TimeStampedModel):
                     Params={
                         'Bucket': storage.bucket_name,
                         'Key': self.blob.name,
+                        'ACL': 'bucket-owner-full-control',
                     },
                     ExpiresIn=60,  # TODO proper expiration
                 )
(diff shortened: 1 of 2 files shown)


waxlamp (approved):
Looks fine to me. Just a couple of questions:
- does/will `dandi-cli` need a similar change in its upload routines?
- you said this is bad UX, but we're fully abstracting over the difficulty, so that end users won't have to deal with this, right?

dchiquito:
* this is only a change to the zarr file upload, which isn't working yet in dandi-cli. Hopefully this unblocks it.
* I'm mostly griping about AWS design because I am the user and I am having a poor experience :'(

PR #651, opened 2021-12-28
Don't leak spurious 400 error to client

waxlamp:
When the input to user search doesn't validate, it leaks a 400 error through to the browser (specifically, in the user search form that appears in the "manage ownership" dialog) but doesn't break user functionality in any way. In fact, the user typing bad input into this field *shouldn't* be considered an error, and should instead just return no results.

Closes https://github.com/dandi/dandiarchive/issues/708.

Asking @dchiquito for a direct review, and @mvandenburgh to verify that this doesn't negatively impact frontend behavior.

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -6,6 +6,7 @@ from django.db.models import Q
 from django.http.response import HttpResponseBase
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import api_view, parser_classes, permission_classes
+from rest_framework.exceptions import ValidationError
 from rest_framework.parsers import JSONParser
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.request import Request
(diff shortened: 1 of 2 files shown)


mvandenburgh (commented):
Confirming that this will not break anything in the frontend.

dandiapi/api/views/users.py
```diff
@@ -80,7 +81,14 @@ def users_me_view(request: Request) -> HttpResponseBase:
...
     """Search for a user."""
     request_serializer = UserSerializer(data=request.query_params)
-    request_serializer.is_valid(raise_exception=True)
+
+    # Swallow validation errors in the input string, and just send back null
+    # results.
+    try:
+        request_serializer.is_valid(raise_exception=True)
+    except ValidationError as e:
+        return Response([])
```
dchiquito, line 90:
```suggestion
    if not request_serializer.is_valid(raise_exception=False)
        return Response([])
```

dchiquito (changes requested):

dchiquito (approved):

PR #658, opened 2021-12-30
Fix slow loading time/crash for `Version` admin list page

### LINKED ISSUE #539: Admin pages not loading

mvandenburgh:
The Django admin listing page for `Version` is issuing a SQL query for every single `Version` on the page due to the use of the `asset_count` property. This PR modifies the behavior to issue a single SQL query to retrieve the asset count of every `Version`.

This partially fixes #539; the `Version` detail page still needs optimizing.


dchiquito (approved):
Looks like 100 versions are shown on the page, so this would cut the DB requests down from O(100) to O(1). I'm suspicious that there might still be performance problems because the same number of assets still need to be counted, but this is definitely a step in the right direction.

waxlamp:
@mvandenburgh, you didn't mean to close #539 with this PR right? Your "partially fixes" phrase did cause it to be closed.

mvandenburgh:
Oh, no that wasn't intentional. I reopened it.

PR #659, opened 2021-12-30
Codify `skip_authorization` policy in OAuth2 model setup

### LINKED ISSUE #300: Skip authorization prompt when authenticating
> One simple way to avoid this confusion would be to configure Django Oauth Toolkit [to skip authorization](https://django-oauth-toolkit.readthedocs.io/en/latest/advanced_topics.html#skip-authorization-completely-for-trusted-applications) for the DANDI GUI specifically.

_Originally posted by @brianhelba in https://github.com/dandi/dandi-api/issues/280#issuecomment-845279383_

waxlamp:
Closes #300.

I believe in order to deploy this, we'd need to go through some migrational gymnastics. @dchiquito, you accomplished this once before on staging, and may have documented what you did then. In your opinion, how disruptive would it be to go through that procedure again for production?

If it's a lot, or it puts data at risk, I'm completely fine with closing both this PR, and #300, and documenting the "runtime" method of setting the `skip_authorization` policy we want.


dandiapi/api/migrations/0021_productionapplication.py
```diff
@@ -0,0 +1,37 @@
...
+import django.db.models.deletion
+import oauth2_provider.generators
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('api', '0020_zarr'),
+    ]
```
dchiquito, line 14: This should look like https://github.com/dandi/dandi-api/blob/74d80f6855b34c09e338b61dc3fc657d07e0d632/dandiapi/api/migrations/0011_stagingapplication.py#L13-L15 instead so that it is available for oauth-toolkit, then you would need another "dummy" migration that has this migration as a dependency (see migration 0012).

dchiquito (changes requested):
The shenanigans involved making migrations [11](https://github.com/dandi/dandi-api/blob/74d80f6855b34c09e338b61dc3fc657d07e0d632/dandiapi/api/migrations/0011_stagingapplication.py) and [12](https://github.com/dandi/dandi-api/blob/74d80f6855b34c09e338b61dc3fc657d07e0d632/dandiapi/api/migrations/0012_merge_20210630_1354.py). Note that migration 11 is set to apply before the first migration in oauth-toolkit, and migration 12 makes 11 a dependency for subsequent dandi-api migrations.

This is all well and good, but modifying the migration history like that involves manual intervention on every instance, which right now would be staging, production, and every dev's environment. A non-trivial undertaking.

My recommendation is that rather than merging this now and complicating embargo with this migration stuff, we wait until embargo is shipped. At that time we should consider simply discarding our migration history entirely and starting with a fresh migration 0001. The only advantage to maintaining this migration history is if we need to roll back, so as long as a manual intervention is required anyway we might as well take the time to clean things up a bit.

dandiapi/api/migrations/0021_productionapplication.py
waxlamp: Ok, I'll need your help in getting that right, but let's wait until after embargo lands, as you said.

waxlamp:
Going to close this is low-value. We can always revive this concept later if we really think it matters.

PR #660, opened 2021-12-30
Add content at the root of the API server

### LINKED ISSUE #320: api server landing page
continuation to #119  (and may be some other recently closed issue I failed to find ATM)

ATM both https://api.dandiarchive.org/ and https://api.dandiarchive.org/api/ stimply 404. IMHO it is suboptimal. I would have at least populated them with some basic information about API server (URL to swagger, may be this repository) or at least redirected somewhere (https://api.dandiarchive.org/swagger/ might not be the worst choice).

waxlamp:
Closes #320.

### DIFF
--- a/dandiapi/api/views/__init__.py
+++ b/dandiapi/api/views/__init__.py
@@ -3,6 +3,7 @@ from .auth import auth_token_view, authorize_view, user_questionnaire_form_view
 from .dandiset import DandisetViewSet
 from .info import info_view
 from .stats import stats_view
+from .root import root_content_view
 from .upload import (
     blob_read_view,
     upload_complete_view,
@@ -31,4 +32,5 @@ __all__ = [
     'users_search_view',
     'stats_view',
     'info_view',
+    'root_content_view',
 ]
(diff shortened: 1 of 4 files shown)


mvandenburgh (approved):
LGTM, linting errors will need to be fixed before merging.

dandiapi/api/views/__init__.py
```diff
@@ -31,4 +32,5 @@
     'users_search_view',
     'stats_view',
     'info_view',
+    'root_content_view',
```
yarikoptic, line 35: FWIW, someone eventually should sort all those entries IMHO

yarikoptic (commented):
Looks lovely.  Doesn't address 404 of `/api` though. Even if we are expecting it to serve something later on, for now it could probably just provide the same page  as for the `/`?

waxlamp:
> Looks lovely. Doesn't address 404 of `/api` though. Even if we are expecting it to serve something later on, for now it could probably just provide the same page as for the `/`?

I'd rather not put something there--I really don't see what value it would bring, and it would likely confuse any tool that did try to make a request to that location. Right now, there's literally nothing to be found there, so I think it would be harmful to put even friendly HTML there, since that provides a signal to downstreams that they found something valuable. I am not too worried about humans hitting that path either, since we have plenty of resources that are more easily found (including the new content at the root of the Django site) that will guide them to the more appropriate places to go.

PR #668, opened 2022-01-05
Hide embargoed versions assets

### LINKED ISSUE #667: Protect Version and Asset endpoints from embargoed dandisets
`DandisetViewSet` will behave correctly and hide any embargoed dandisets that a user does not have permission to view, but the `VersionViewSet` and `AssetViewSet` do not. They should also use query set filtering to hide Versions and Assets that belong to embargoed dandisets.

dchiquito:
I missed a spot when concealing the `DandisetViewSet` from returning data on embargoed Dandisets: Versions and Assets would still dispense data happily even when associated with embargoed Dandisets. This fixes that, and adds tests.

Technically, this implementation leaks the fact that an embargoed dandiset exists with that identifier, since querying `/api/dandisets/{embargoed_dandiset_id}/versions/` will return an empty pagination instead of a 404. I think that's fine, since embargoed dandisets will conspicuously occupy gaps in the dandiset identifier sequence, and there isn't anything a bad actor can do with an identifier. @waxlamp can you advise?

Fixes #667

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -10,7 +10,7 @@ import pytest
 import requests
 from rest_framework.test import APIClient
 
-from dandiapi.api.models import Asset, AssetBlob, Version
+from dandiapi.api.models import Asset, AssetBlob, Dandiset, Version
 from dandiapi.api.views.serializers import AssetFolderSerializer, AssetSerializer
 
 from .fuzzy import HTTP_URL_RE, TIMESTAMP_RE, URN_RE, UTC_ISO_TIMESTAMP_RE, UUID_RE
@@ -913,3 +913,35 @@ def test_asset_direct_download_head(api_client, storage, version, asset):
 @pytest.mark.django_db
 def test_asset_direct_metadata(api_client, asset):
     assert json.loads(api_client.get(f'/api/assets/{asset.asset_id}/').content) == asset.metadata
+
+
+@pytest.mark.parametrize(
+    ('embargo_status'),
+    [
+        Dandiset.EmbargoStatus.EMBARGOED,
+        Dandiset.EmbargoStatus.UNEMBARGOING,
+    ],
+)
+@pytest.mark.django_db
+def test_asset_embargoed_visibility(
+    api_client, dandiset_factory, draft_version_factory, asset_factory, embargo_status
+):
+    dandiset = dandiset_factory(embargo_status=embargo_status)
+    version = draft_version_factory(dandiset=dandiset)
+    asset = asset_factory()
+    version.assets.add(asset)
+
+    # The version should be hidden because the dandiset it belongs to is embargoed
+    response = api_client.get(
+        f'/api/dandisets/{dandiset.identifier}/versions/{version.version}/assets/'
+    )
+    assert response.json() == {
+        'count': 0,
+        'next': None,
+        'previous': None,
+        'results': [],
+    }
+    response = api_client.get(
+        f'/api/dandisets/{dandiset.identifier}/versions/{version.version}/assets/{asset.asset_id}/'
+    )
+    assert response.status_code == 404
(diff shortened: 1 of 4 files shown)


yarikoptic:
FWIW, leaking knowledge about embargoed identifier existence is IMHO ok. And for that reason I even wonder if we should step away from GitHub inspired 404 into a proper 401 and 403 responses depending on either session is already authenticated or not.

But also I wondered if there is a way in Django to provide handling for that `/api/dandisets/{embargoed_dandiset_id}/` prefix, so that it is always invoked first for any end point starting with it? If so, then all embargo authentication related handling at dandiset level could be conveniently centralized

yarikoptic:
BTW, forgot to ping @satra on above

satra:
identifier existence is ok.

waxlamp:
> FWIW, leaking knowledge about embargoed identifier existence is IMHO ok. And for that reason I even wonder if we should step away from GitHub inspired 404 into a proper 401 and 403 responses depending on either session is already authenticated or not.
> 

I agree with this. See #669. I'm deferring to @yarikoptic and @satra on the issue of "leaking' the existence of embargoed dandisets--my own $0.02 is that because this is an open scientific archive, there's nothing to be lost and perhaps something small to be gained by subtly broadcasting the existence of these dandisets.

I would possibly go even further and imagine an eventual feature that would enable information about embargoed data to be published, without publishing the data itself. But this is just me thinking out loud.

> But also I wondered if there is a way in Django to provide handling for that `/api/dandisets/{embargoed_dandiset_id}/` prefix, so that it is always invoked first for any end point starting with it? If so, then all embargo authentication related handling at dandiset level could be conveniently centralized

I don't think this is how it would be done, but rather with the right permission-related decorators on all relevant endpoints. I'll leave the details of this to @dchiquito.

yarikoptic:
> > But also I wondered if there is a way in Django to provide handling for that `/api/dandisets/{embargoed_dandiset_id}/` prefix, so that it is always invoked first for any end point starting with it? If so, then all embargo authentication related handling at dandiset level could be conveniently centralized
> 
> I don't think this is how it would be done, but rather with the right permission-related decorators on all relevant endpoints. I'll leave the details of this to @dchiquito.

I do get such an approach.  The question was more of satisfying my curiosity on either a more robust (e.g. so a forgotten decorator on a new endpoint doesn't leak it to public) way possible.

waxlamp:
> > > But also I wondered if there is a way in Django to provide handling for that `/api/dandisets/{embargoed_dandiset_id}/` prefix, so that it is always invoked first for any end point starting with it? If so, then all embargo authentication related handling at dandiset level could be conveniently centralized
> > 
> > 
> > I don't think this is how it would be done, but rather with the right permission-related decorators on all relevant endpoints. I'll leave the details of this to @dchiquito.
> 
> I do get such an approach. The question was more of satisfying my curiosity on either a more robust (e.g. so a forgotten decorator on a new endpoint doesn't leak it to public) way possible.

There is probably some way to do what you're suggesting, but I don't think it would be the "django way", but rather a more generic Python approach to refactoring some gating logic out or something. It is a good question though, about triggering common logic because of the common endpoint prefix.

dchiquito:
I implemented it in the most generic method I know how.

To give an overview, we are using a [library](https://www.django-rest-framework.org/) that lets us define a class that represents a family of endpoints. On that class, we define a `queryset` or a `get_queryset` method that returns a Django QuerySet, which is directly translatable into SQL by the Django ORM. I modified the `queryset` on `VersionViewSet` and `AssetViewSet` to filter out dandisets which are not visible to the current user.
The issue with implementing it in a more generic way is that the construction of the `queryset` directly impacts the SQL performance. In general I think it's better to be explicit when defining a `queryset` so that it's easier to read and debug. We are also unlikely to be adding any new models like `Version` or `Asset` that might belong to embargoed dandisets that are similar enough to be able to use some kind of generic parent class.

mvandenburgh (approved):

PR #678, opened 2022-01-07
Fix zarr metadata

### LINKED ISSUE #676: Set encodingFormat in metadata for zarr assets
The metadata for zarr assets should enforce that encodingFormat always be set appropriately. Non-zarr assets will not set an encodingFormat, that is left to the uploader.

https://github.com/dandi/dandi-cli/issues/852#issuecomment-1007601714
>  let's use encodingFormat: "application/x-zarr" for now and we can adjust if the zarr folks say otherwise.

### LINKED ISSUE #677: Set contentUrl correctly for zarr assets
Currently zarr assets still use the `contentUrl` used by normal assets, which does not work. We should modify the normal asset download endpoint to redirect to the zarr download API endpoint: `/api/zarr/{zarr_id}.zarr/` when the asset is a zarr file.

dchiquito:
* Fixes #676 
* Fixes #677 
* Include a full S3 URL as the second `contentUrl` metadata field for zarr archives.

Triggered by discussion on https://github.com/dandi/dandi-cli/issues/852#issuecomment-1007758461

### DIFF
--- a/dandiapi/api/models/zarr.py
+++ b/dandiapi/api/models/zarr.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from datetime import timedelta
 from pathlib import Path
+from urllib.parse import urlparse, urlunparse
 from uuid import uuid4
 
 from django.conf import settings
@@ -168,6 +169,13 @@ class ZarrArchive(TimeStampedModel):
     def s3_path(self, zarr_path: str | Path):
         """Generate a full S3 object path from a path in this zarr_archive."""
         return f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}{settings.DANDI_ZARR_PREFIX_NAME}/{self.zarr_id}/{str(zarr_path)}'  # noqa: E501
+
+    @property
+    def s3_url(self):
+        signed_url = ZarrUploadFile.blob.field.storage.url(self.s3_path(''))
+        parsed = urlparse(signed_url)
+        s3_url = urlunparse((parsed[0], parsed[1], parsed[2], '', '', ''))
+        return s3_url
 
     def get_checksum(self, path: str | Path = ''):
         listing = ZarrChecksumFileUpdater(self, path).read_checksum_file()
--- a/dandiapi/api/models/zarr.py
+++ b/dandiapi/api/models/zarr.py
@@ -173,6 +173,7 @@ class ZarrArchive(TimeStampedModel):
     @property
     def s3_url(self):
         signed_url = ZarrUploadFile.blob.field.storage.url(self.s3_path(''))
+        # Strip off the query parameters from the presigning, as they are different every time
         parsed = urlparse(signed_url)
         s3_url = urlunparse((parsed[0], parsed[1], parsed[2], '', '', ''))
         return s3_url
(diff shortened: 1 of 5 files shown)


satra:
@dchiquito - the blob (including zarr) url is always the s3 path. 

the download url should remain consistent across assets, what it does internally can change by content type. is there a specific reason for the asset `download` endpoint to be different ?

dchiquito:
@satra I believe that is what I implemented here. `contentUrl` will always take the form `[downloadUrl, s3Url]`.
* `downloadUrl` still points to `/assets/{asset_id}/download/`. That endpoint has been modified internally to redirect to the zarr download API for zarr archives, otherwise it redirects to a presigned S3 URL for the blob as it always has.
* `s3Url` now points to the base of the zarr archive in the S3 bucket for zarr archives, or the blob in the S3 bucket for non-zarr assets.

satra:
thank you @dchiquito

dandiapi/api/models/zarr.py
```diff
@@ -169,6 +170,13 @@ def s3_path(self, zarr_path: str | Path):
         """Generate a full S3 object path from a path in this zarr_archive."""
         return f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}{settings.DANDI_ZARR_PREFIX_NAME}/{self.zarr_id}/{str(zarr_path)}'  # noqa: E501
 
+    @property
+    def s3_url(self):
+        signed_url = ZarrUploadFile.blob.field.storage.url(self.s3_path(''))
+        parsed = urlparse(signed_url)
+        s3_url = urlunparse((parsed[0], parsed[1], parsed[2], '', '', ''))
```
mvandenburgh, line 177: Is the purpose of this just to remove the query string from the `signed_url`? If so, can you add a comment specifying that? It's not immediately obvious to me what is happening here.

Also, [this solution](https://stackoverflow.com/a/32214777) removes the query string in one line and avoids explicit list indexing.

dandiapi/api/tests/test_asset.py
```diff
@@ -253,6 +253,35 @@ def test_asset_populate_metadata(draft_asset_factory):
...
+    }
+
+    # This should trigger _populate_metadata to inject all the computed metadata fields
+    asset = draft_asset_factory(metadata=raw_metadata, blob=None, zarr=zarr_archive)
+
+    download_url = 'https://api.dandiarchive.org' + reverse(
+        'asset-direct-download',
+        kwargs={'asset_id': str(asset.asset_id)},
+    )
+    s3_url = f'http://localhost:9000/test-dandiapi-dandisets/test-prefix/test-zarr/{zarr_archive.zarr_id}/'  # noqa: E501
```
mvandenburgh, line 270: We'll want to use the `DJANGO_MINIO_STORAGE_ENDPOINT` setting here like in https://github.com/dandi/dandi-api/blob/master/dandiapi/api/tests/test_version.py#L94, otherwise this test will fail when run in docker-compose where `DJANGO_MINIO_STORAGE_ENDPOINT` is `http://minio/....`

mvandenburgh (changes requested):

dandiapi/api/models/zarr.py
dchiquito: I just tried that and `urlparse('http://my.domain/foo/bar').path` only returns `/foo/bar`, no protocol or domain. Sadly I think this is the most complete solution short of doing a string split on `?`.

I will add a comment.

dandiapi/api/tests/test_asset.py
dchiquito: Good point

mvandenburgh (approved):

PR #681, opened 2022-01-10
Asset info

### LINKED ISSUE #673: Include meta-metadata in individual asset endpoint responses
When listing assets via the `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/` endpoint, the following fields are returned:

* `asset_id`
* `blob`
* `zarr`
* `path`
* `size`
* `created`
* `modified`

However, when getting information about an individual asset via the `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/{asset_id}/` or `/assets/{asset_id}/` endpoint, only the asset metadata is returned.  This metadata includes values for the `asset_id`, `path`, and `size` fields, but it does not include a `created`, `blob`, or `zarr` field, and neither of the metadata fields `dateModified` and `blobDateModified` corresponds to `modified`.  This makes getting all of the information about an asset more awkward than it should have to be.

I would therefore like to request that the `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/{asset_id}/` endpoint (and `/assets/{asset_id}/`, if possible) return the same structure as returned when listing assets but with an added "metadata" field containing the current response structure.

dchiquito:
* Add `/assets/{asset_id}/info/`
* Add `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/{asset_id}/info/`

Both return the same thing, the django serialization of the asset.
Resolves #673


dandiapi/api/tests/test_asset.py
```diff
@@ -343,6 +343,25 @@ def test_asset_rest_retrieve_no_sha256(api_client, version, asset):
     )
 
 
+@pytest.mark.django_db
+def test_asset_rest_info(api_client, version, asset):
+    version.assets.add(asset)
```
mvandenburgh, line 348: Maybe consider adding a second asset here to replicate the bug fixed by #683.

mvandenburgh (approved):

PR #683, opened 2022-01-10
Correctly invoke super().get_queryset()

dchiquito:
Turns out `NestedViewSetMixin` already implemented `get_queryset`, so when I overwrote it I also overwrote the code that filters versions/assets. This restores it.


mvandenburgh (approved):
LGTM

AlmightyYakob (approved):
LGTM, never seen request filtering done this way. Seems pretty clean.

Issue #674, opened 2022-01-07
zarr issues from cli discussion


satra:
- [x] dandiset size doesn't take zarr asset size into account
- [x] zarr content url should point to s3 location (and zarr proxy api end point when that is ready). but the s3 prefix should also be there like other assets.
- [x] move code for zarr checksum compute to dandischema
- [x] needs API (endpoint) to list individual components of a zarr archive to simplify/robustify download; ideally those records should come with minimal metadata (path, size, mtime; mtime could be the time when upload finalized for that file I guess) (added by @yarikoptic channeling https://github.com/dandi/dandi-cli/issues/852#issuecomment-1011348765).

dchiquito:
The size field definitely should be taking the zarr assets into account, I will look into that.

The content URL should be resolved by https://github.com/dandi/dandi-api/pull/678.

checksum compute move is coming soon.

dchiquito:
@yarikoptic The `/zarr/{zarr_id}.zarr/{path}` endpoint should suffice for zarr file browsing. For directories, it returns something like:
```
{
  "directories": [
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/0",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/1",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/2",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/3",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/4",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/5"
  ],
  "files": [
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/.zattrs",
    "https://api-staging.dandiarchive.org/api/zarr/79bb04a0-3f1b-46ae-8821-1071fc69ed6e.zarr/.zgroup"
  ]
}
```

For files, it returns a redirect to the object in S3.

[...]

PR #802, opened 2022-01-14
Add checksum fields to zarr explore API

dchiquito:
* Add a `checksum` field to the zarr directory listing endpoint containing the checksum for the directory
* Add a `checksums` field to the zarr directory listing endpoint containing the checksums for all subfiles and subdirectories of the directory

Addresses API deficiencies brought up in #674


mvandenburgh (approved):
LGTM after linting is fixed

PR #804, opened 2022-01-14
Replace user dashboard with customized django admin page

mvandenburgh:
The Django admin can replicate everything we do on the user dashboard page with a few extra lines of code to the `UserAdmin` class. In addition to being able to remove all of the custom user template code we have, we also get things like user search and pagination for free with this approach.


satra:
@mvandenburgh - how do we sort that page by date joined? and can we copy just a column of email addresses?

mvandenburgh:
> @mvandenburgh - how do we sort that page by date joined?

I've added sorting by date joined - clicking the column header will do the sorting.

> and can we copy just a column of email addresses?

I believe the Django admin has some mechanism that would support this, I can look into it.

dchiquito (approved):

yarikoptic:
do we store github login in DB? if yes -- could it be displayed in the table as well?

mvandenburgh:
> overall looks good to me. if there is a way to export email addresses only or better still copy a set of them without any additional stuff that would be ideal.

I added an action to export user emails to CSV -
email



> do we store github login in DB? if yes -- could it be displayed in the table as well?
Yes we do - I added it as a column.

PR #805, opened 2022-01-14
Add HEAD support to Zarr API

### LINKED ISSUE #746: Support HEAD requests to `/zarr/{zarr_id}.zarr/{path}`
It would be nice if dandi-api responded to HEAD requests to `/zarr/{zarr_id}.zarr/{path}` with a redirect to an S3 URL suitable for performing a second HEAD request.  Currently, making a HEAD request to `/zarr/{zarr_id}.zarr/{path}` results in "405 Method Not Allowed", and making a GET request to the endpoint returns a redirect to an S3 URL whose signature makes it only valid for GET requests.  Being able to make a HEAD request in this way would simplify retrieving the size & "last modified" date of a file in a Zarr asset.

dchiquito:
Fixes #746 

Also break up the monolithic zarr API test into smaller chunks.

### DIFF
--- a/dandiapi/api/tests/test_zarr.py
+++ b/dandiapi/api/tests/test_zarr.py
@@ -157,36 +157,25 @@ def test_zarr_rest_delete_upload_in_progress(
 
 
 @pytest.mark.parametrize(
-    ('path', 'status', 'directories', 'files'),
+    ('path', 'directories', 'files'),
     [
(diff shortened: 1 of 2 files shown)


mvandenburgh (approved):
I don't have a zarr file on hand to test this, but your usage of the S3/Minio APIs look correct to me. If it works for you locally i'm good with it

PR #811, opened 2022-01-17
Create/update assets with embargoed blobs

### LINKED ISSUE #810: Cannot register embargoed blobs as new Assets
All the machinery is in place to upload embargoed asset blobs, but the code that creates new assets still only checks for normal blobs, not embargoed.

Sentry Issue: [DANDI-API-57](https://sentry.io/organizations/dandiarchive/issues/2937180817/?referrer=github_integration)

```
AssetBlob.DoesNotExist: AssetBlob matching query does not exist.
  File "contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "dandiapi/api/tasks.py", line 37, in calculate_sha256
    asset_blob = AssetBlob.objects.get(blob_id=blob_id)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
```

dchiquito:
Fixes #810 

I added all the necessary API code to upload embargoed data, but I forgot about the part that creates/updates assets to point to that embargoed data.

Also I added a new serializer field that is used to "combine" two fields (like `blob` and `embargoed_blob`) into a single field for serialization purposes. Now the `blob` field returned from the API call will refer to either `blob` or `embargoed_blob` in the DB as appropriate for that Asset.

### DIFF
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -130,6 +130,26 @@ class AssetValidationSerializer(serializers.ModelSerializer):
         fields = ['status', 'validation_errors']
 
 
+class EmbargoedSlugRelatedField(serializers.SlugRelatedField):
+    """
+    A Field for cleanly serializing embargoed model fields.
+
+    Embargoed fields are paired with their non-embargoed equivalents, like "blob" and
+    "embargoed_blob", or "zarr" and "embargoed_zarr". There are DB constraints in place to ensure
+    that only one field is defined at a time. When serializing one of those pairs, we would like to
+    conceal the fact that the field might be embargoed by silently using the embargoed model field
+    in place of the normal field if it is defined.
+    """
+
+    def get_attribute(self, instance: Asset):
+        attr = super().get_attribute(instance)
+        if attr is None:
+            # The normal field was not defined on the model, try the embargoed_ variant instead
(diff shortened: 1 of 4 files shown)


dandiapi/api/views/serializers.py
```diff
@@ -130,6 +130,26 @@ class Meta:
...
+    in place of the normal field if it is defined.
+    """
+
+    def get_attribute(self, instance: Asset):
+        attr = super().get_attribute(instance)
+        if attr is None:
+            # The normal field was not defined on the model, try the embargoed_ variant instead
+            embargoed_source = f'embargoed_{self.source}'
+            attr = getattr(instance, embargoed_source, None)
+        return attr
```
mvandenburgh, line 150: This is cool, didn't know it was possible to do something like this :open_mouth:

mvandenburgh (approved):
LGTM

dandiapi/api/views/serializers.py
dchiquito: Me neither until just now ¯\\_(ツ)\_/¯ took some digging to work out, and it only serializes since we never use it to deserialize.

PR #813, opened 2022-01-17
Embargo download and checksum calculation

### LINKED ISSUE #812: Allow download embargoed assets
Using the download link in the UI breaks for embargoed assets.

dchiquito:
Fixes #812

I missed download and checksum calculation.
Last embargo PR, I hope.


mvandenburgh (approved):
I think @waxlamp brought up a good point regarding the `if`/`else` statements that we need everywhere to differentiate between embargoed blobs and regular blobs, and I wonder if Django has something that we could use to make it cleaner. :shrug:

But that would be a separate discussion/PR, what you have here looks good to me

Issue #707, opened 2022-01-03
a few (missing) pieces in new DLP


satra:
- [ ] in this issue dandi/dandiarchive#876 we made sure that funding information was given additional preference, and if not present it showed as "No funding information available". this was to gently remind users that they weren't putting funding contributions in. the section that displays funding, it should be displayed with "No funding information available". https://github.com/dandi/dandi-archive/pull/817

- [x] also associated projects is now missing. that section should be at the bottom of the DLP.  https://github.com/dandi/dandi-archive/pull/818

- [x] assets summary doesn't collapse into row mode on a small display, unlike related resources and funding information. https://github.com/dandi/dandi-archive/pull/819

for the future, since it is hard to compare DLP completeness before and after on staging, we should craft a staging dandiset that has as much metadata/metadata variation as possible.

mvandenburgh:
This is still waiting on #817.

PR #818, opened 2022-01-18
Add Associated Projects card to DLP

mvandenburgh:
Fixes 2nd point in #707.

@satra what fields from associated projects should be displayed on the DLP? So far it displays just the `name` and `identifier`. 

To provide context since it's been some time since we've discussed DLP - this is only for the general "overview" display, in the future we will have separate tabs for the metadata, so we'll have a view like [this](https://www.figma.com/file/rBobiAvSLCsJmyYUYmU1OX/DLP?node-id=504%3A36884) for associated projects where we can display more detailed info with more/all of the fields.

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -177,11 +177,28 @@
         </div>
       </v-list>
     </v-card>
+
+    <MetadataCard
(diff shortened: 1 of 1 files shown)


satra:
just name and identifier is good for now.

satra (approved):
i tested a preview. looks good to me.

PR #821, opened 2022-01-18
Design doc: Add option to embargo draft states of dandisets

satra:


waxlamp (commented):
This seems like a reasonable feature, but I wonder if it is the same thing as embargo, rather than just hiding the draft version. That is, is this feature meant to be bound by the same rules/regulations as the NIH-recognized embargo period?

I am thinking forward to a time when we might be transferring large quantities of data back and forth between buckets, and what that would mean for user experience. I think we should revisit our two-bucket implementation strategy in case we want to make this "embargoed draft" idea a reality.

waxlamp:
@satra, what's the next action on this? Do we need a discussion to refresh and discuss the idea?

satra:
@waxlamp - i think the next steps are for your team perhaps to note what this will take. currently a embargoed dataset can have public bucket files. if the upload mechanism finds the checksum in the public bucket it won't add the bytes to the embargoed dataset.

doc/design/embargo-full.md
```diff
@@ -1,6 +1,15 @@
...
 This document takes the Embargo MVP as a predicate and describes the remaining features required for full Embargo support.
 
+## Switching drafts to embargoed on publish
+
+A user should be able to make the draft state of a published dataset embargoed. This has two possibilities.
+
+1. Starting with an embargoed dandiset: This involves creating an embargoed dataset, unembargoing and then publishing. At publish, the user can be given the option to embargo the draft version of the dataset.
+2. Starting with a published dandiset: This will embargo the draft version of a published dandiset. In this scenario any existing files will remain public.
+
+This would support use cases like adding new data to existing dandisets, but using all of the data for analyses, etc.,.
```
yarikoptic, line 11: Minor fix to commit, will do
```suggestion
This would support use cases like adding new data to existing dandisets, but using all of the data for analyses, etc.
```

yarikoptic:
As far as I see it we have `embargo_status` at the level of `Dandiset` not its `Version` (https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/api/models/dandiset.py#L46). Implementation of such a feature would mean shifting that property to a version thus allowing some versions (draft) to be embargoed while the others (and may be even some releases?) to be embargoed.  Overall -- might be not so trivial to implement all logic correctly.

But what would be the overall destiny for this PR -- should I expect us to finalize it soon or we would dwell on it? (if latter -- I would prefer to keep it in Draft mode)

waxlamp:
I would say move this back to draft. We have some big problems to solve with embargo, so I don't think this will be actionable at the moment, and the design may change if we solve some looming problems with embargo as it exists today.

yarikoptic:
AFAIK this embargo-full was never implemented (only embargo-mvp).  As such, this design document is rather "suggestive" so while thinking of the next design we could consult it, and avoid running around PRs and issue.  So I would vote to merge this PR - no point to keep it in draft.  If no objections voiced -- I will merge it early next week.

waxlamp:
> AFAIK this embargo-full was never implemented (only embargo-mvp). As such, this design document is rather "suggestive" so while thinking of the next design we could consult it, and avoid running around PRs and issue. So I would vote to merge this PR - no point to keep it in draft. If no objections voiced -- I will merge it early next week.

My objection is that design docs are not supposed to be merely "suggestive" but rather contain ideas that are largely vetted and consensed upon. In this particular case, we have already run into problems of abstraction, in that we don't embargo versions of dandisets (as you pointed up [upthread](https://github.com/dandi/dandi-archive/pull/821#issuecomment-1435069928)). On top of that, we are now developing the capability to hide individual assets in the bucket as "embargoed" (i.e., hidden with intent for inclusion among an embargoed dandiset). So at this point, the ability to "embargo a draft version" would simply be the ability to mark certain asset uploads as hidden, and then mark the dandiset as being embargoed (or something similar).

But that behavior would be the heart of this feature, which IMO requires a new round of requirements, etc., since it is offering specific refinements to what is now a sort of "bulk" behavior.

I would argue for closing this PR and readopting the relevant ideas in a new discussion and design doc. Let me know what you think, @yarikoptic.

yarikoptic:
@satra , this PR was out there for over two years and I do not think we ever felt or were imposed through some regulations the need for this feature.  With the new https://github.com/dandi/dandi-archive/blob/master/doc/design/embargo-redesign.md it probably would be relatively easy to implement on dandi-archive end.  But would add more of similar dance to backups2datalad (would need not only to migrate repo, but also to move URLs back under handling by datalad remote etc)... so - overall it would be an adventure.  And as not among our immediate "Initiatives" I will just close, but let's reopen / refile if comes of need.

PR #822, opened 2022-01-19
Embargoed Zarr models

### LINKED ISSUE #642: Add zarr embargo models
See https://github.com/dandi/dandi-api/issues/634

Normal asset embargo models have already been added, but the same needs to be done for `ZarrArchive` and `ZarrUploadFile`.

dchiquito:
For context, `ZarrArchive`s are roughly equivalent to `AssetBlob`s. `ZarrUploadFile`s are roughly equivalent to `Upload`s, although `ZarrUploadFiles` augment the `ZarrArchive`, while `Uploads` are converted into `AssetBlobs`.

Fixes #642


dandiapi/api/models/zarr.py
```diff
@@ -22,7 +23,9 @@
 
 
 class ZarrUploadFileManager(models.Manager):
-    def create_zarr_upload_file(self, zarr_archive: ZarrArchive, path: str, **kwargs):
+    def create_zarr_upload_file(
+        self, zarr_archive: ZarrArchive | EmbargoedZarrArchive, path: str, **kwargs
```
AlmightyYakob, line 27: TIL you could use the | operator for syntax :exploding_head:

dandiapi/api/models/zarr.py
```diff
@@ -143,9 +140,40 @@ def to_checksum(self) -> ZarrChecksum:
...
+    zarr_archive: 'ZarrArchive' = models.ForeignKey(
+        'ZarrArchive',
+        related_name='active_uploads',
+        on_delete=models.CASCADE,
+    )
+
+
+class EmbargoedZarrUploadFile(BaseZarrUploadFile):
+    class Meta:
+        get_latest_by = 'modified'
```
AlmightyYakob, line 159: Can this not be placed in `BaseZarrUploadFile`, since it's the same across this and `ZarrUploadFile`?

dandiapi/api/models/zarr.py
```diff
@@ -234,15 +260,42 @@ def cancel_upload(self):
...
+    upload_file_class = ZarrUploadFile
+
+    def s3_path(self, zarr_path: str | Path):
+        """Generate a full S3 object path from a path in this zarr_archive."""
+        return f'{settings.DANDI_DANDISETS_BUCKET_PREFIX}{settings.DANDI_ZARR_PREFIX_NAME}/{self.zarr_id}/{str(zarr_path)}'  # noqa: E501
+
+
+class EmbargoedZarrArchive(BaseZarrArchive):
+    class Meta:
+        get_latest_by = 'modified'
```
AlmightyYakob, line 290: Same question as with `EmbargoedZarrUploadFile`

AlmightyYakob (commented):
A couple questions but this all seems good to me, although I'm not the most familiar with Zarr files

dandiapi/api/models/zarr.py
mvandenburgh: Unfortunately, it seems like this syntax is only valid in Python 3.10 - https://www.python.org/dev/peps/pep-0604/ (dandi-api is on 3.8).

dchiquito: I think `from __future__ import annotations` will enable it, and that is conveniently already included in this file. Any objections to using it here?

dandiapi/api/models/zarr.py
dchiquito: You apparently can, neat.

AlmightyYakob:
Soft approval from me, but I don't feel fully qualified to stamp this. I'd wait for @mvandenburgh to chime in.

mvandenburgh (approved):
LGTM

PR #828, opened 2022-01-20
changes in ShareDialog: using inline style

### LINKED ISSUE #729: Fix design/placement for share icon
Right now the share icon is in two places: next to the title and in the right panel.

@satra - we were discussing this on Thursday, but wasn't sure if you preferred to have in both places or only next to the title?

djarecka:
adding inline style to the `ShareDialog` icon to be able to set it next to the title

fixes #729


web/src/views/DandisetLandingView/ShareDialog.vue
```diff
@@ -6,7 +6,10 @@
     max-width="500"
   >
     <template #activator="{ on }">
-      <div v-on="on">
+      <div
+        style="display: inline"
```
mvandenburgh, line 10: The `d-inline` class is a more idiosyncratic Vuetify way to do this.

```suggestion
        class="d-inline"
```

mvandenburgh (changes requested):

djarecka:
@mvandenburgh - i think I added your suggested, but this PR is still blocked. I've just merged current master,  could you please take a look if anything else is needed

web/src/views/DandisetLandingView/ShareDialog.vue
```diff
@@ -7,7 +7,10 @@
...
   >
     <template #activator="{ on }">
-      <div v-on="on">
+      <div
+        class="d-inline"
+        v-on="on"
+      >
         <v-icon
           color="primary"
         >
```
mvandenburgh, line 16: I think setting [`cursor` to `pointer`](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) here would make it more obvious to the user that the icon is clickable

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -8,8 +8,7 @@
       <v-row class="mx-2 my-2 mb-0">
         <v-col>
           <h1 :class="`font-weight-light ${$vuetify.breakpoint.xs ? 'text-h6' : ''}`">
-            {{ meta.name }}
-            <ShareDialog />
+            <ShareDialog />{{ meta.name }}
```
mvandenburgh, line 11:
```suggestion
            <ShareDialog />
            {{ meta.name }}
```
Minor nitpick, is there any issue with doing this? It's easier to read that way imo (unless of course, that messes up the formatting)

mvandenburgh:
@djarecka I merged latest `master` and left a couple small suggestions, PTAL. Other than that, this looks good to me.

web/src/views/DandisetLandingView/ShareDialog.vue
djarecka: not sure if I understand

waxlamp: If you roll your mouse over this share icon, you'll see that the mouse pointer stays an arrow; if you set the CSS `cursor` value to `pointer` for the `<v-icon>` element, instead it will display as the clicking finger that appears when you roll over a hyperlink.

You can set a class on the `<v-icon>` element and then add a rule to the `<style>` section, or you can (perhaps, not entire sure) inline the style on the `<v-icon>` itself. @mvandenburgh, is there a preferred option here (including a third way I didn't mention)?

mvandenburgh: Since vuetify makes heavy use of utility classes like `d-inline` above for styling, I usually prefer inline styles if applying the style to a single component so that you can see both the classes + styles in one glance (and generally that's how its done in the rest of the codebase). But AFAIK there's no best practice for this sort of thing.

satra:
hey folks  - what's the status of this?

djarecka:
it looks like there is still one more thing I should resolve

waxlamp:
@djarecka, let us know what you need to get this finished. @mvandenburgh may be available to meet with you if you want some direct collaboration.

web/src/views/DandisetLandingView/ShareDialog.vue
```diff
@@ -7,7 +7,10 @@
     max-width="500"
   >
     <template #activator="{ on }">
-      <div v-on="on">
+      <div
+        class="d-inline"
+        v-on="on"
+      >
         <v-icon
           color="primary"
```
mvandenburgh, line 15:
```suggestion
          color="primary"
          style="cursor: pointer;"
```

mvandenburgh: @djarecka this should resolve the last open thread - let me know if you have any questions. I think this PR is good to go after this is applied.

djarecka:
@mvandenburgh - sorry for delays and thank you for the help!

mvandenburgh (approved):

PR #830, opened 2022-01-20
Implement Unembargo

### LINKED ISSUE #679: Implement unembargoing
See [design doc](https://github.com/dandi/dandi-api/pull/619) ([old link](https://github.com/dandi/dandi-api/blob/96c85ca5b3c1932b450cb00d4c375f0a176e2261/doc/design/embargo-mvp.md?plain=1#L103-L121)):
> * New endpoint: `POST /api/dandisets/{dandiset_id}/unembargo`
>
>  Unembargo an embargoed dandiset.
>
>  Only permitted for owners and admins. If the `embargo_status` is `OPEN` or `UNEMBARGOING`, return 400.
>
>  Set the `embargo_status` to `UNEMBARGOING`, then dispatch the unembargo task.
>
>* Unembargo task
>
>  For every `Asset` with an `EmbargoedAssetBlob` in the dandiset, convert the `EmbargoedAssetBlob` into an `AssetBlob` by moving the data from the embargo bucket to the public bucket.
>  These could be >5GB, so the [multipart copy API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html) must be used.
>  The ETag and checksum must remain undisturbed; the only change should be where the data is stored.
>  Verify that the resulting unembargoed assets match one-for-one (in the database) with the embargoed assets that were copied.
>  Once finished, the `access` metadata field on the dandiset will be updated to `OpenAccess` and `embargo_status` is set to `OPEN`.
>
>  Before copying data, check if an existing `AssetBlob` with the same checksum has been uploaded already (this would have happened after uploading the embargoed data).
>  If so, use it instead of copying the `EmbargoedAssetBlob` data.

AlmightyYakob:
Closes #679

I'm not sure how accurately the tests replicate the behavior of S3 in production, @dchiquito maybe you have a better idea. If nothing else we might just want to keep an eye on it when it hits staging.


dandiapi/api/_copy.py
```diff
@@ -13,15 +18,21 @@
 PART_SIZE = 500 * 1024 * 1024  # 500 MB
```
AlmightyYakob, line 18: @dchiquito From what I can tell the maximum part size is `5GiB`. Should we change this to match that, so that we're not being intentionally inefficient?

dandiapi/api/tests/test_tasks.py
```diff
@@ -250,3 +252,108 @@ def test_validate_version_metadata_malformed_license(version: Version, asset: As
...
+    embargoed_asset: Asset = asset_factory(embargoed_blob=embargoed_asset_blob, blob=None)
+
+    # Assert properties before unembargo
+    assert embargoed_asset.embargoed_blob is not None
+    assert embargoed_asset.blob is None
+    assert embargoed_asset.embargoed_blob.etag != ''
+
+    # Run unembargo
+    draft_version = dandiset.draft_version
+    draft_version.assets.add(embargoed_asset)
```
dchiquito, line 289: I would put these in the setup blocks

dandiapi/api/tests/test_tasks.py
```diff
@@ -250,3 +252,108 @@ def test_validate_version_metadata_malformed_license(version: Version, asset: As
...
+
+    # Assert correct changes took place
+    assert dandiset.embargo_status == Dandiset.EmbargoStatus.OPEN
+    assert draft_version.metadata['access'] == 'dandi:OpenAccess'
+
+    asset: Asset = draft_version.assets.first()
+    assert asset.embargoed_blob is None
+    assert asset.blob is not None
+    assert asset.blob.etag == embargoed_asset_blob.etag
+
```
dchiquito, line 302: Can you add an assertion that the asset.blob.name or whatever is the expected value? It should be something like:
```python
assert asset.blob.name == f'test-prefix/blobs/{asset.asset_id[:3]}/{asset.asset_id[3:6]}/{asset.asset_id}'
```

dandiapi/api/tests/test_tasks.py
```diff
@@ -250,3 +252,108 @@ def test_validate_version_metadata_malformed_license(version: Version, asset: As
...
+    # Assert correct changes took place
+    assert dandiset.embargo_status == Dandiset.EmbargoStatus.OPEN
+    assert draft_version.metadata['access'] == 'dandi:OpenAccess'
+
+    asset: Asset = draft_version.assets.first()
+    assert asset.embargoed_blob is None
+    assert asset.blob is not None
+    assert asset.blob.etag == embargoed_asset_blob.etag
+    assert asset.blob == existing_asset_blob
```
dchiquito, line 359: Can you add a test for unembargoing a dandiset with a normal asset in it?

dandiapi/api/views/dandiset.py
```diff
@@ -262,6 +263,31 @@ def destroy(self, request, dandiset__pk):
...
+        },
+        operation_summary='Unembargo a dandiset.',
+        operation_description=(
+            'Unembargo an embargoed dandiset. Only permitted for owners and admins'
+            '. If the embargo status is OPEN or UNEMBARGOING, an HTTP 400 is returned.'
+        ),
+    )
+    @action(methods=['POST'], detail=True)
+    @method_decorator(permission_required_or_403('owner', (Dandiset, 'pk', 'dandiset__pk')))
+    def unembargo(self, request, dandiset__pk):
```
dchiquito, line 282: Can you add this view to the `test_embargo_visibility` parametrization?

dchiquito (changes requested):

dandiapi/api/tasks.py
```diff
@@ -164,3 +167,57 @@ def validate_version_metadata(version_id: int) -> None:
...
+    # Only the draft version is needed, since embargoed dandisets can't be published
+    draft_version: Version = dandiset.draft_version
+    embargoed_assets: List[Asset] = list(
+        Asset.objects.filter(versions__in=[draft_version], embargoed_blob__isnull=False)
+    )
+
+    # Copy all embargoed assets to public bucket
+    for asset in embargoed_assets:
+        etag = asset.embargoed_blob.etag
+        matching_blob = AssetBlob.objects.filter(etag=etag).first() if etag != '' else None
```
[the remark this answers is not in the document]
AlmightyYakob: I'm not sure, I was being safe :man_shrugging: If you don't think it's a serious concern I can remove it.

dandiapi/api/_copy.py
```diff
@@ -98,16 +105,26 @@ def _copy_object_s3(
...
 def _copy_object_minio(
     storage,
     source_bucket: str,
     source_key: str,
     dest_bucket: str,
     dest_key: str,
-):
+) -> CopyObjectResponse:
+    """Copy an object, returning the object key and etag."""
     # The Minio client will automatically use multipart upload if the file size is too big
```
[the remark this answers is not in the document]
AlmightyYakob: I wasn't aware that function existed in minio, could you point me to the relevant documentation?

dandiapi/api/tests/factories.py
```diff
@@ -141,10 +141,8 @@ def sha256(self):
 
     @factory.lazy_attribute
     def etag(self):
-        h = hashlib.md5()
-        h.update(self.blob.read())
-        self.blob.seek(0)
```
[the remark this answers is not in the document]
AlmightyYakob: My bad, I meant to keep that in. I'll re-add it.

dandiapi/api/tests/test_tasks.py
AlmightyYakob: I'm not sure what you mean by "setup blocks"?

dandiapi/api/views/dandiset.py
AlmightyYakob: 739f00c

dandiapi/api/tests/test_tasks.py
AlmightyYakob: 582272e

dandiapi/api/tests/test_tasks.py
AlmightyYakob: You mean a test where an embargoed dandiset contains only one asset, which is a normal (unembargoed) asset?

dandiapi/api/_copy.py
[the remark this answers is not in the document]
dchiquito: https://github.com/minio/minio-py/blob/b3ba3bf99fe6b9ff2b28855550d6ab5345c134e3/minio/api.py#L1711-L1744
Sadly it seems like it doesn't exist, `client.put_object` will always use a normal upload if the file is smaller than the part size :(
Our options are to continue using the fake ETags, or call `_create_multipart_upload`, `_upload_part`, and `_complete_multipart_upload` ourselves. I think having mismatched data is going to trip someone up at some point, so sadly I think we should probably do the legwork ourselves :(

There is also [compose_object](https://github.com/minio/minio-py/blob/b3ba3bf99fe6b9ff2b28855550d6ab5345c134e3/minio/api.py#L1369-L1376), but it looks like that also uses a normal upload for small files :'(

dandiapi/api/tasks.py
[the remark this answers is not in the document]
dchiquito: `BaseAssetBlob.etag = models.CharField(max_length=40, validators=[RegexValidator(f'^{ETAG_REGEX}$')])` if it were `""` it would be a SQL constraint violation, I don't think you need to check for it anywhere else.

dandiapi/api/tests/test_tasks.py
dchiquito: Sorry, I meant that these lines look like setting up test data rather than the code under test, so I would put them a few lines higher with the rest of the test setup code. I don't feel strongly, just sharing an opinion

dandiapi/api/tests/test_tasks.py
dchiquito: correct

dandiapi/api/tests/test_dandiset.py
```diff
@@ -694,6 +694,37 @@ def test_dandiset_rest_create_embargoed(api_client, user):
     }
 
 
+@pytest.mark.parametrize(
+    ('embargo_status', 'user_status', 'success'),
+    [
+        (Dandiset.EmbargoStatus.OPEN, 'owner', False),
+        (Dandiset.EmbargoStatus.OPEN, 'anonymous', False),
+        (Dandiset.EmbargoStatus.OPEN, 'not-owner', False),
+        (Dandiset.EmbargoStatus.EMBARGOED, 'owner', True),
```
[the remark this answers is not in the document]
AlmightyYakob: Personally I think this is fine, since this test is checking the combination of `embargo_status` and `user_status`. The fact that only one happens to be `True` doesn't really matter imo.

dandiapi/api/tests/test_dandiset.py
```diff
@@ -694,6 +694,37 @@ def test_dandiset_rest_create_embargoed(api_client, user):
...
+    if user_status == 'anonymous':
+        user = AnonymousUser
+    else:
+        user = user_factory()
+        api_client.force_authenticate(user=user)
+    if user_status == 'owner':
+        assign_perm('owner', user, dandiset)
+
+    response = api_client.post(f'/api/dandisets/{dandiset.identifier}/unembargo/')
+    assert (response.status_code == 200) == success
```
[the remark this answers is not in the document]
AlmightyYakob: I opted to parametrize the status code directly in 9a9e52a. Lmk if you think this is insufficient.

AlmightyYakob:
@dchiquito your concerns around duplicate part calculation and minio multipart copy have been addressed in 1da069d and 81a8eec, respectively.

dandiapi/api/models/upload.py
```diff
@@ -107,8 +107,8 @@ class Upload(BaseUpload):
     blob = models.FileField(blank=True, storage=get_storage, upload_to=get_storage_prefix)
     dandiset = models.ForeignKey(Dandiset, related_name='uploads', on_delete=models.CASCADE)
 
-    @classmethod
-    def object_key(cls, upload_id, dandiset: Dandiset):
+    @staticmethod
+    def object_key(upload_id, dandiset: Dandiset | None = None):
```
dchiquito, line 111: Can you update the method signature on `BaseUpload.object_key` as well?

dandiapi/api/tasks.py
```diff
@@ -164,3 +165,57 @@ def validate_version_metadata(version_id: int) -> None:
...
+@shared_task
+def unembargo_dandiset(dandiset_id: int):
+    """Unembargo a dandiset by copying all embargoed asset blobs to the public bucket."""
+    dandiset: Dandiset = Dandiset.objects.get(id=dandiset_id)
+
+    # Only the draft version is needed, since embargoed dandisets can't be published
+    draft_version: Version = dandiset.draft_version
+    embargoed_assets: List[Asset] = list(
+        Asset.objects.filter(versions__in=[draft_version], embargoed_blob__isnull=False)
+    )
```
dchiquito, line 179:
```suggestion
    embargoed_assets: List[Asset] = list(draft_version.assets.filter(embargoed_blob__isnull=False))
```

dandiapi/api/tasks.py
```diff
@@ -164,3 +165,57 @@ def validate_version_metadata(version_id: int) -> None:
...
+            assert resp.etag == asset.embargoed_blob.etag
+
+            # Assign blob
+            asset.blob = AssetBlob(
+                blob_id=asset.embargoed_blob.blob_id,
+                etag=resp.etag,
+                blob=resp.key,
+                size=asset.embargoed_blob.size,
+            )
+            asset.blob.save()
```
dchiquito, line 209: You might consider moving this logic to a method on `EmbargoedAssetBlob` to keep as much application logic on the model as possible, but not a big deal if you think here is better

dchiquito (changes requested):

dandiapi/api/copy.py
```diff
@@ -0,0 +1,130 @@
...
+class CopyPartGenerator(PartGenerator):
+    """Override the normal PartGenerator class to account for larger copy part sizes."""
+
+    @classmethod
+    def for_file_size(cls, file_size: int) -> CopyPartGenerator:
+        """Calculate sequential part sizes given a file size."""
+        if file_size == 0:
+            return cls(0, 0, 0)
+
+        part_size = gb(5)
```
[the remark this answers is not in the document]
AlmightyYakob: 5GB is the max part size for s3 multipart upload/copy. Since this is an S3 bucket to bucket copy, and nothing passes through the client, this seems the most efficient. 64mb is probably fine when uploading from a client to S3, but seems wildly inefficient at the speeds that S3 is capable of.

dandiapi/api/copy.py
```diff
@@ -0,0 +1,130 @@
...
+    return copy_object_multipart(
+        client,
+        source_bucket=source_bucket,
+        source_key=source_key,
+        dest_bucket=dest_bucket,
+        dest_key=dest_key,
+    )
+
+
+def copy_object_multipart(
```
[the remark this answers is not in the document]
AlmightyYakob: This is technically also valid for minio, which is why I named is as such. I suppose if we consider s3 against minio as s3 all the same, then your change makes sense.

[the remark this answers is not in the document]
dchiquito: I thought it make sense to label it S3 since it uses a boto3 client. If we had a `_copy_object_multipart_minio` which used a Minio client, we could obviously still use it with an S3 bucket, but I think it still makes sense to clarify with flavor of client is being used. Just my 2¢

dandiapi/api/copy.py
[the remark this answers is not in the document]
AlmightyYakob: For anybody else, after some offline discussion, we've decided to keep the current 64mb part size for now. If that causes serious bandwidth issues, we can re-evaluate.

dandiapi/api/models/upload.py
```diff
@@ -52,7 +52,7 @@ class Meta:
     size = models.PositiveBigIntegerField()
 
     @abstractclassmethod
-    def object_key(cls, upload_id, dandiset: Dandiset):  # noqa: N805
+    def object_key(cls, upload_id, dandiset: Dandiset | None = None):  # noqa: N805
```
dchiquito, line 55:
```
    @abstractstaticmethod
    def object_key(upload_id, dandiset: Dandiset | None = None):  # noqa: N805
```
Is that a thing?

dandiapi/api/tests/test_unembargo.py
```diff
@@ -0,0 +1,230 @@
...
+    part_size,
+    monkeypatch,
+):
+    # Since both fixtures are parametrized, only proceed when they use the same storage backend
+    if type(storage) != type(embargoed_storage):
+        pytest.skip('Skip tests with mismatched storages')
+
+    # Pretend like AssetBlob/EmbargoedAssetBlob were defined with the given storage
+    monkeypatch.setattr(AssetBlob.blob.field, 'storage', storage)
+    monkeypatch.setattr(EmbargoedAssetBlob.blob.field, 'storage', embargoed_storage)
```
dchiquito, line 73: this looks much more sophisticated than what I've been doing. Does it reset the values after the test completes?

dchiquito (approved):
Some departing pedantry, other than that looks good :+1:

dandiapi/api/models/upload.py
AlmightyYakob: Apparently yes https://docs.python.org/3/library/abc.html#abc.abstractstaticmethod

But you're really supposed to do this:

```python
    @staticmethod
    @abstractmethod
    def object_key(upload_id, dandiset: Dandiset | None = None):
```

I'll fix it.

dandiapi/api/tests/test_unembargo.py
AlmightyYakob: I believe so https://docs.pytest.org/en/6.2.x/monkeypatch.html

Namely

> All modifications will be undone after the requesting test function or fixture has finished. The raising parameter determines if a KeyError or AttributeError will be raised if the target of the set/deletion operation does not exist.

Which I really like :)

AlmightyYakob:
Looks like I'm getting some errors from `dandischema`, due to `localhost` not being a TLD. I'll look into fixing those.

dchiquito:
@AlmightyYakob IIRC [this](https://github.com/dandi/dandischema/blob/master/dandischema/models.py#L42-L46) is probably the thing that needs to be switched on

AlmightyYakob:
> @AlmightyYakob IIRC [this](https://github.com/dandi/dandischema/blob/master/dandischema/models.py?rgh-link-date=2022-02-01T23%3A13%3A54Z#L42-L46) is probably the thing that needs to be switched on

It seems this is already the case https://github.com/dandi/dandi-archive/blob/35c632c014a2d737db92e485d2b47d761970459e/dandiapi/settings.py#L111

and testing locally shows that it's available in the environment :thinking:

dandiapi/api/models/upload.py
waxlamp: > New in version 3.2.
> 
> Deprecated since version 3.3

Poor decorator though it was going to make it big in Pythonworld 😭

PR #831, opened 2022-01-20
Embargo permissions

### LINKED ISSUE #669: Return 401/403 for embargoed dandisets, versions, and assets
@yarikoptic wrote (https://github.com/dandi/dandi-api/pull/668#issuecomment-1006177484):
> FWIW, leaking knowledge about embargoed identifier existence is IMHO ok. And for that reason I even wonder if we should step away from GitHub inspired 404 into a proper 401 and 403 responses depending on either session is already authenticated or not.

When responding to a request for an embargoed resource, return 401 if unauthenticated, and 403 if authenticated but unauthorized.

dchiquito:
* Throw 401 whenever a client attempts to do anything to an embargoed version/asset without authenticating.
* Throw 403 whenever a client attempts to do anything to an embargoed version/asset without owning the dandiset.
* Add a single test for all embargo permissions. This should cover all negative testing for embargo permissions.

Fixes #669

### DIFF
--- a/dandiapi/api/tests/test_embargo.py
+++ b/dandiapi/api/tests/test_embargo.py
@@ -1,0 +1,77 @@
+import pytest
+
+from dandiapi.api.models import Dandiset
+
+
+@pytest.fixture(
+    params=[
+        Dandiset.EmbargoStatus.EMBARGOED,
+        Dandiset.EmbargoStatus.UNEMBARGOING,
+    ]
+)
+def embargo_status(request):
+    return request.param
+
+
+EMPTY_PAGINATION = {
+    'count': 0,
+    'next': None,
+    'previous': None,
+    'results': [],
+}
+
+
+@pytest.mark.parametrize(
+    ('method', 'url_format'),
+    [
+        ('get', '/api/dandisets/{dandiset.identifier}/versions/'),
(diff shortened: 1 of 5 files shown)


dandiapi/api/tests/test_embargo.py
```diff
@@ -0,0 +1,78 @@
...
+
+    url = url_format.format(dandiset=dandiset, asset=asset)
+    response = getattr(api_client, method)(url)
+    # The client is not authenticated, so all response codes should be 401
+    assert response.status_code == 401
+
+    api_client.force_authenticate(user=user)
+    response = getattr(api_client, method)(url)
+    # The client is now authenticated but not an owner, so all response codes should be 403
+    assert response.status_code == 403
```
AlmightyYakob, line 78: Is successful access tested elsewhere?

dchiquito: Searching `test.*embargo` gives the full catalog of embargo tests. Upload to embargoed dandisets is thoroughly tested. Other operations are tested more intermittently, but that should be sufficient since the `get_queryset()` functionality should extend to every operation in the ViewSet.

It would be nice to keep other embargo tests here, but I think it makes more sense to keep them next to their unembargoed counterparts, since a lot of logic is duplicated.

mvandenburgh (approved):
LGTM

PR #839, opened 2022-01-21
[wip] adding license field to the create dandiset page

### LINKED ISSUE #747: Force choice of license at dandiset creation time

djarecka:
Adding a new field - `license` to the create dandiset page as a mandatory field. 

closes #747

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -24,6 +24,16 @@
           outlined
           class="my-4"
         />
+        <v-combobox
+          v-model="license"
+          :items="['spdx:CC0-1.0', 'spdx:CC-BY-4.0']"
+          label="License*"
+          class="my-4"
+          multiple
+          outlined
+          dense
+        />
+
         <!-- TODO: remove v-if to enable embargoed dandiset creation -->
         <div v-if="false">
           <v-checkbox
@@ -118,6 +128,7 @@ export default defineComponent({
   setup(props, ctx) {
     const name = ref('');
     const description = ref('');
+    const license = ref('');
     const embargoed = ref(false);
     const awardNumber = ref('');
     const saveDisabled = computed(
@@ -135,7 +146,7 @@ export default defineComponent({
     }
 
     async function registerDandiset() {
-      const metadata = { name: name.value, description: description.value };
+      const metadata = { name: name.value, description: description.value, license: license.value };
 
       const { data } = embargoed.value
         ? await dandiRest.createEmbargoedDandiset(name.value, metadata, awardNumber.value)
@@ -147,6 +158,7 @@ export default defineComponent({
     return {
       name,
       description,
+      license,
       embargoed,
       awardNumber,
       saveDisabled,


djarecka:
@dchiquito @mvandenburgh - I know I'm not doing this properly, I hardcoded the licenses we are using. I found license types in schema.js file, but not sure what is the best way to extract the information from there..

satra:
we should discuss this a bit more in the context of not requiring license for embargoed dandisets. see: https://github.com/dandi/dandischema/issues/115

mvandenburgh:
> @dchiquito @mvandenburgh - I know I'm not doing this properly, I hardcoded the licenses we are using. I found license types in schema.js file, but not sure what is the best way to extract the information from there..

Types are only used for type-checking and are evaluated at compile time and then discarded, so unfortunately there's no way to access the license types in `schema.ts` at runtime. But, you should be able to extract them from the schema like this -

```
const licenses: ComputedRef<LicenseType[]> = computed(() => store.state.dandiset.schema.definitions.LicenseType.enum);
````

**Comment by djarecka** [comment]
@mvandenburgh - thank you, will try!

**Comment by waxlamp** [comment]
@djarecka, I converted this to a draft PR since you are still working on it.

File: web/src/views/CreateDandisetView/CreateDandisetView.vue [hunk 1]
```diff
@@ -75,6 +76,17 @@
             :rules="awardNumberRules"
           />
         </div>
+        <div v-if="embargoed===false">
+          <v-combobox
```
mvandenburgh, line 80: What are your thoughts on using a `v-select` instead of `v-combobox` here? IMO the extra autocomplete features `v-combobox` provides just confuse things, since we there's only two license types to choose from. Also, fwiw the meditor uses a `v-select` for licenses.

djarecka: sorry, I missed the comments.... I'm not sure if I had any specific thoughts, will try to use v-select

djarecka:
@mvandenburgh - I came back to this old PR, I merged the current master, but have some errors, that I don't fully understand. Could you please help me with it

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -117,6 +134,7 @@ export default defineComponent({
   setup(props, ctx) {
     const name = ref('');
     const description = ref('');
+    const license = ref('');
```
mvandenburgh, line 137:
```suggestion
    const license = ref<License>();
```

This is what is causing the type-check failures. `license` is implicitly defined as type `Ref<string>` here, but the schema dictates that it's an array of string literals, which is encapsulated in the `License` type. 

Making this change will make `license` instead have a type of `Ref<License | undefined>`, which works as long as you also update `saveDisabled` to return true if the dandiset is not embargoed and has an undefined license (which we want to do anyways), i.e. add `|| (!embargoed.value && !license.value)` to it (github code review unfortunately won't let me make that a formal suggestion b/c this PR doesn't modify that code yet).

mvandenburgh (changes requested):
@djarecka See below suggestions, they should fix the type-checking errors. Let me know if you have any questions.

web/src/views/CreateDandisetView/CreateDandisetView.vue
djarecka: did you have in mind `ref<LicenseType>()`?

mvandenburgh: No, `ref<License>()` (you would also have to add that to the import statement). `LicenseType` refers to a string, while `License` is an array of `LicenseTypes` that is populated by the `v-select` (see `schema.ts` for those type definitions)

djarecka:
@mvandenburgh - could you help me with the error, don't understand it

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -75,6 +75,17 @@
             :rules="awardNumberRules"
           />
         </div>
+        <div v-if="!embargoed">
+          <v-select
+            v-model="license"
+            :items="dandiLicenses"
+            :label="`License ${embargoed ? '' : '*'}`"
```
mvandenburgh, line 82:
```suggestion
        <div v-if="!embargoed">
          <v-select
            v-model="license"
            :items="dandiLicenses"
            label="License*"
```

Just realized, the v-if actually implies that this license dropdown won't appear for an embargoed dandiset, so my earlier suggestion doesn't make sense.

mvandenburgh (commented):
@djarecka I updated the tests as we discussed - this PR looks good to me after removing the redundant code I mentioned above.

PR #843, opened 2022-01-24
Only revalidate invalids in revalidate command

dchiquito:
Only revalidate invalid assets/versions when running the `revalidate` command.

Add `--revalidate-all` flag to disable that behavior.
Add `--dry-run` flag to only log the number of assets/versions being revalidated without revalidating.


waxlamp (approved):
LGTM (but I didn't test it out, of course)

PR #854, opened 2022-01-25
Update checksum calculation in zarr design doc

dchiquito:
The majority of the time taken when uploading zarr files is the API server updating the `.checksum` files. This design doc update describes a scheme for moving that update into a celery task.

@satra With this design, the upload to S3 would take 6 seconds, then the upload complete endpoint would take about another 6 seconds to verify that all the ETags are present and correct in S3. I estimate upload speeds of about 140 MB/min, although we won't know for sure until it's implemented.
The only way to further improve performance while still using the presigned URLs is to skip the ETag check, which would give 250-300 MB/min.

### DIFF
--- a/doc/design/zarr-support-3.md
+++ b/doc/design/zarr-support-3.md
@@ -56,7 +56,7 @@ The intended use of Zarr archives is primarily to support use cases where data a
 
   Completes an upload of a batch of zarr files.
   Fail if any of the checksums do not match.
-  Also, recursively update the tree hash for every parent node for each child.
+  Trigger the task which recursively updates the tree hash for every parent node for each child.
 
 * **DELETE /api/zarr/{zarr_id}/upload/**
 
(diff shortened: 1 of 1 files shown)


satra:
> the upload to S3 would take 6 seconds

is this per file?

dchiquito:
@satra no, it took 6 seconds for a batch of 255 files of ~150KB each.

doc/design/zarr-support-3.md
```diff
@@ -143,6 +143,54 @@ This bubbles up to the top of the zarr archive, where the final `.checksum` for
...
+The checksum update task takes a `zarr_id` as an argument.
+It will then run the following loop:
+* Select the `ZarrChecksumUpdate` with the greatest `depth` and status `PENDING`.
+  * If there is no such `ZarrChecksumUpdate`, terminate.
+* Select for update all `ZarrChecksumUpdate` with the same `parent_directory`.
+  These represent all the updates waiting to be applied to a `.checksum` file.
+  * If the select for update failed, another worker selected the same `parent_directory` in a race condition.
+    Catch the error and terminate silently.
+* Set the `status` of all selected `ZarrChecksumUpdates` to `IN_PROGRESS`.
+* Dispatch another checksum update task.
```
dchiquito, line 182: Only dispatch on the first loop to prevent exponential growth of the task queue.

dchiquito (commented):
some algorithmic complaints

dchiquito:
It has come to light that the checksum calculation isn't the bottleneck with zarr uploads. Any of the alternative upload schemes currently being considered will not benefit from this design change, so I am closing this.

PR #866, opened 2022-01-31
Add manifest link to DLP

### LINKED ISSUE #861: Add manifest location to DLP
This was present in the old DLP, but I didn't carry it over in the redesign.

Screenshot from 2022-01-28 16-04-16

mvandenburgh:
Fixes #861

Screenshot from 2022-01-31 11-14-10

### DIFF
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -231,7 +231,7 @@ const dandiRest = new Vue({
       return data;
     },
     assetManifestURI(identifier: string, version: string) {
-      return `${dandiApiRoot}dandisets/${identifier}/versions/${version}/assets`;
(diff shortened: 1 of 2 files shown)


mvandenburgh:
@satra `manifestLocation` is an array in the schema - does that mean there could be more than one manifest?

satra:
@mvandenburgh - it was an array to help support multiple locations that could deliver the array. for example for published dandisets it could be both a pointer to our api and the s3 bucket location. looks like we dropped our api location when publishing a dandiset. 

for the purpose of this PR, perhaps we can always point to the API endpoint for assets of a particular instance (draft/publishedversions on a specific server) instead of reading from metadata.

satra (approved):

PR #871, opened 2022-02-03
Fix inclusion of CopySourceRange param

AlmightyYakob:
When testing locally against minio, setting the `CopySourceRange` param to `''` was fine, but it seems like this isn't the case with S3 proper.

I've tested unembargo in staging with a file larger than 64mb, ensuring more than one part, and thus a non-empty `CopySourceRange` param. That succeeded, so I believe this fix will address the issue.

### DIFF
--- a/dandiapi/api/copy.py
+++ b/dandiapi/api/copy.py
@@ -52,12 +52,12 @@ class CopyObjectPart:
     include_range: bool = True
 
     @property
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
I've filed an issue for this https://github.com/minio/minio/issues/14246

dchiquito (approved):

PR #872, opened 2022-02-03
Sub hardcoded value with MINIO_STORAGE_ENDPOINT setting

AlmightyYakob:
Currently, these tests fails if running in docker, as the endpoint is `minio:9000` not `localhost:9000`. Plus, we probably shouldn't be hardcoding that anyway.

### DIFF
--- a/dandiapi/api/tests/test_zarr.py
+++ b/dandiapi/api/tests/test_zarr.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+from django.conf import settings
 import pytest
 
 from dandiapi.api.models import ZarrArchive, ZarrUploadFile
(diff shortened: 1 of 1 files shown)


dandiapi/api/tests/test_zarr.py
```diff
@@ -48,8 +49,8 @@ def test_zarr_rest_get(
 
 @pytest.mark.django_db
 def test_zarr_rest_get_very_big(authenticated_api_client, zarr_archive_factory):
-    ten_quadrillion = 10**16
-    ten_petabytes = 10**16
+    ten_quadrillion = 10 ** 16
+    ten_petabytes = 10 ** 16
```
satra, line 53: just a note that black just changed their styling to drop the spaces around this (i.e. what it looked like before the change). haven't followed the social world about the reaction.

AlmightyYakob: My bad, I must be running an old version of black.

dchiquito (approved):

PR #878, opened 2022-02-09
Remove github_username from admin search

### LINKED ISSUE #874: 500 error when searching for github id in user admin panel
go to the [user admin panel](https://api.dandiarchive.org/admin/auth/user/), enter a user id (e.g., waxlamp) and click on search.

dchiquito:
`github_username` is a display only property that we extract from the
user metadata, so it cannot be searched. Whenever searches tried to they
encountered an error.

fixes #874


waxlamp:
@satra, this fix will prevent further errors from arising. If you need github username search, that would constitute a separate chunk of work. Let us know.

mvandenburgh (approved):

satra:
this is fine for now.

PR #879, opened 2022-02-09
Add threads to delete_from_versioned_bucket

dchiquito:
This speeds up the versioned bucket delete script pretty substantially.

AWS doesn't have a good way to delete a lot of stuff from a versioned bucket, their [official recommendation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) is to just write your own script.

### DIFF
--- a/dev/delete_from_versioned_bucket.py
+++ b/dev/delete_from_versioned_bucket.py
@@ -11,6 +11,7 @@ This script was the best way I could find to bulk delete objects in an S3 direct
 You will need an AWS CLI profile set up with permission to delete objects from the desired bucket.
 See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html
 """
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import boto3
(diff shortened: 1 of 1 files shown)


dev/delete_from_versioned_bucket.py
```diff
@@ -33,7 +34,8 @@ def delete_version(bucket, thing):
 @click.option('--bucket', prompt='Bucket', help='The bucket to delete from')
 @click.option('--prefix', prompt='Path to delete', help='The prefix to delete all files under')
 @click.option('--force', is_flag=True, help='Skip confirmation prompts')
-def delete_objects(profile: str, bucket: str, prefix: str, force: bool):
+@click.option('--threads', prompt='Threads', help='How many threads to use')
```
AlmightyYakob, line 37: Would it be good to have some default here?

dchiquito: Set to 30

AlmightyYakob (approved):

PR #880, opened 2022-02-09
Add warning when an admin modifies owners of a dandiset they don't own

### LINKED ISSUE #858: Add a warning to ownership panel if non-owner admin is using it

mvandenburgh:
Closes #858.

Screenshot from 2022-02-09 13-56-24

### DIFF
--- a/web/src/components/DLP/DandisetOwnersDialog.vue
+++ b/web/src/components/DLP/DandisetOwnersDialog.vue
@@ -141,6 +141,41 @@
         </div>
       </v-col>
     </v-row>
+    <v-dialog
+      v-model="adminWarningDisplay"
+      max-width="60vw"
+      persistent
+    >
+      <v-card class="pb-3">
+        <v-card-title class="text-h5 font-weight-light">
+          WARNING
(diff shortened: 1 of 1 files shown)


web/src/components/DLP/DandisetOwnersDialog.vue
```diff
@@ -141,6 +141,41 @@
...
+        </v-card-text>
+        <v-card-text>
+          Would you like to proceed?
+        </v-card-text>
+
+        <v-card-actions>
+          <v-btn
+            color="error"
+            depressed
+            @click="save()"
```
AlmightyYakob, line 166: Usually you just pass the function itself
```suggestion
            @click="save"
```

AlmightyYakob (changes requested):
A couple of suggestions but LGTM

AlmightyYakob (approved):

waxlamp (approved):
This looks great. The comment on variable naming is not a blocker, but just a suggestion. My other suggestion would be to make "you are not an owner" bold within the alert text--but that's also not a blocker.

Issue #708, opened 2022-01-03
Finish typescript conversions


mvandenburgh:
This supersedes dandi/dandiarchive#855 by outlining specifically what is left in terms of converting the existing codebase to typescript.

- [x] Convert file browser component to typescript #882
- [x] Convert "utility" js files (`router.js`, `migrate.js`, etc) to typescript (#1009)
~~- [ ] Enable type checking in vue templates (#987)~~
~~This is blocked on converting the file browser component~~ moved to https://github.com/dandi/dandi-archive/issues/720

(I had a TODO written down to file this a long time ago but I can't find an existing issue for it, I guess I never got around to it)

mvandenburgh:
This isn't done, there's still 1 part left

mvandenburgh:
Turns out the last part of this is actually a dupe of https://github.com/dandi/dandi-archive/issues/720. I'll close this in favor of the more specific one.

PR #882, opened 2022-02-09
Convert `FileBrowser` to TypeScript/Composition API

mvandenburgh:
Thought it would make sense to do this before the file browser redesign.

Resolves in part #708

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -376,9 +376,6 @@ export default defineComponent({
       itemToDelete.value = null;
     }
 
-    // Refresh files/folders whenever URL changes
-    watch(location, getItems, { immediate: true });
-
     // Update URL if location changes
     watch(location, () => {
       const { location: existingLocation } = ctx.root.$route.query;
@@ -404,6 +401,7 @@ export default defineComponent({
     // go to the directory specified in the URL if it changes
     watch(() => ctx.root.$route, (route) => {
       location.value = route.query.location.toString() || rootDirectory;
+      getItems();
     }, { immediate: true });
 
     // Fetch dandiset if necessary
@@ -415,8 +413,6 @@ export default defineComponent({
         });
       }
     });
-
-    getItems();
 
     return {
       location,
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -84,9 +84,10 @@
               </span>
             </v-card-title>
             <v-progress-linear
-              v-if="$asyncComputed.items.updating"
+              v-if="updating"
               indeterminate
             />
+
             <v-divider v-else />
             <v-list>
               <v-list-item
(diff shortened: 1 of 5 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -259,167 +265,175 @@ export default {
...
-    fetchDandiset() {
-      store.dispatch.dandiset.fetchDandiset({
-        identifier: this.identifier,
-        version: this.version,
-      });
-    },
+    // Refresh files/folders whenever URL changes
+    watch(location, getItems);
+
+    watch(location, () => {
```
AlmightyYakob, line 382:
```suggestion

		// Update URL if location changes
    watch(location, () => {
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -259,167 +265,175 @@ export default {
...
+        query: { location: location.value },
+      } as RawLocation);
+    });
+
+    watch(items, () => {
+      if (items.value && !items.value.length) {
+        // If the API call returns no items, go back to the root (shouldn't normally happen)
+        location.value = rootDirectory;
+      }
+    });
```
AlmightyYakob, line 401:
```suggestion

    // If the API call returns no items, go back to the root (shouldn't normally happen)
    watch(items, () => {
      if (items.value && !items.value.length) {
        location.value = rootDirectory;
      }
    });
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -259,167 +265,175 @@ export default {
...
+    }
 
-    fetchDandiset() {
-      store.dispatch.dandiset.fetchDandiset({
-        identifier: this.identifier,
-        version: this.version,
-      });
-    },
+    // Refresh files/folders whenever URL changes
+    watch(location, getItems);
```
AlmightyYakob, line 380: This watcher can use `immediate` and the single `getItems` call below can be removed.

```suggestion
    watch(location, getItems, { immediate: true });
```

AlmightyYakob (changes requested):
A few style suggestions but otherwise looks good!

dchiquito (approved):
LGTM pending Jake's suggestions

web/src/views/FileBrowserView/FileBrowser.vue
mvandenburgh: @AlmightyYakob this introduces a bug - running it immediately will trigger another watcher which will result in another call to `getItems` with a blank `path_prefix` (i.e. the file browser will always change to the root folder shortly after the page loads)
Peek 2022-02-14 14-39

Notice the file browser initially displays the correct files, but then immediately changes to the files in the root directory. 

It seems this is happening because of a redundant watcher that is causing conflicts - I fixed it in https://github.com/dandi/dandi-archive/pull/882/commits/ad57e24f0cea8be0c011a1e2555c028874d61b3c. LMK if anything is unclear.

AlmightyYakob (approved):

PR #883, opened 2022-02-10
Turn on unembargo UI

mvandenburgh:
This removes the "feature flag" that was preventing the unembargo ui from rendering. I've also added additional logic to display an appropriate message when a dandiset is in the process of being unembargoed.

### DIFF
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -219,8 +219,8 @@ const dandiRest = new Vue({
       return data;
     },
     async unembargo(identifier: string) {
-      // TODO: implement this once the server endpoint is available
-      return identifier;
+      const { data } = await client.post(`dandisets/${identifier}/unembargo/`);
+      return data;
     },
     async info(): Promise<Info> {
       const { data } = await client.get('info/');
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -218,9 +218,8 @@ const dandiRest = new Vue({
       const { data } = await client.post(`dandisets/${identifier}/versions/draft/publish/`);
       return data;
     },
-    async unembargo(identifier: string) {
-      const { data } = await client.post(`dandisets/${identifier}/unembargo/`);
-      return data;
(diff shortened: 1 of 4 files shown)


web/src/rest.ts
```diff
@@ -219,8 +219,8 @@ const dandiRest = new Vue({
       return data;
     },
     async unembargo(identifier: string) {
-      // TODO: implement this once the server endpoint is available
-      return identifier;
+      const { data } = await client.post(`dandisets/${identifier}/unembargo/`);
+      return data;
```
AlmightyYakob, line 223: This isn't exactly wrong but that endpoint doesn't return any body, just status codes. I don't know if we need to change this but just wanted to point it out.

AlmightyYakob (dismissed):
Looks good, just a couple minor suggestions

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -382,13 +382,16 @@ export default defineComponent({
...
+      }
       return '';
     });
 
     const publishButtonDisabled: ComputedRef<boolean> = computed(
       () => !!(currentDandiset.value?.version_validation_errors.length
         || currentDandiset.value?.asset_validation_errors.length
-        || publishDisabledMessage.value),
+        || publishDisabledMessage.value) || currentDandiset.value?.dandiset.embargo_status !== 'OPEN',
     );
```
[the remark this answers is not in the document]
mvandenburgh: https://github.com/dandi/dandi-archive/pull/883/commits/07e9b8ac2908c896734d25be50503e47c49c6d51
Agreed, I needlessly complicated that for some reason

web/src/rest.ts
mvandenburgh: https://github.com/dandi/dandi-archive/pull/883/commits/653f249b7d1829cd8168b5aec7d19390323a85e5

AlmightyYakob (approved):

Issue #697, opened 2022-01-12
Add link to dandi jupyterhub onto main page or dandiset page


rly:
In our NWB outreach, we sometimes use or refer to the dandi jupyterhub, but as far as I can see, there is no link or easy way to get to the dandihub except if you happen to know the URL. Can this link be added to the main page and/or from a dandiset page? I vaguely remember this existing previously.

waxlamp:
@satra, should this go on the front page?

satra:
it should go on the header somewhere, so people can click on it from any page i think.

PR #891, opened 2022-02-11
Add dandihub link to navbar

waxlamp:
Finishes closing #697.

Note that this addition makes the navbar too crowded. However, once #890 is merged, this should be resolved.

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -130,7 +130,9 @@ import {
   insideIFrame as insideIFrameFunc,
   dandiRest,
 } from '@/rest';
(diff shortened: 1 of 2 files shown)


satra:
overall looks good, but gets crowded when i log in. perhaps there is some design to collapse some of the elements together, or replace with iconography.

waxlamp:
> overall looks good, but gets crowded when i log in. perhaps there is some design to collapse some of the elements together, or replace with iconography.

If you're ok with me merging this, please provide an approving review. Otherwise, we can wait for #890 to try to avert the crowding. Either way, let me know.

satra (approved):

PR #893, opened 2022-02-14
boost copyright to 2022 from 2019

yarikoptic:
or should we just make it correspond to current  year (programmatically) or the year of the last commit in the repo?

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
-          &copy; 2022 - present The DANDI Team<br>
+          &copy; 2019 - 2022 The DANDI Team<br>
           version
           <a
             class="version-link"
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
-          &copy; 2019 - present The DANDI Team<br>
+          &copy; 2022 - present The DANDI Team<br>
           version
           <a
             class="version-link"


waxlamp (commented):
We don't need this or any future changes to the copyright notice--it declares copyright from 2019 to "present".

yarikoptic:
IANAL , so -- doesn't copyright has some kind of expiration? (then keeping an old one would be detrimental although pragmatically irrelevant).  anyways, to me keeping old date in copyright notice just gives a "stale taste" to the website. Having 2019-2022 would show duration/maturity but didn't want to make it longer.

yarikoptic:
FWIW
```
$> curl --silent https://www.kitware.com/ | grep ©
                <ul id="menu-footer-lower-left-side" class=""><li id="menu-item-205" class="menu-item menu-item-type-custom menu-item-object-custom current-menu-item current_page_item menu-item-home menu-item-205"><a href="https://www.kitware.com" aria-current="page">©2021 Kitware, Inc</a></li>
```
I think the copyright year should correspond to the year of the last change since that is where possibly content which has copyright only at that time (not some years before since it might have not even existed) was introduced, and thus in fact reflect "freshness" not as a target of the manipulation, but rather as a "side-effect".

waxlamp:
IMO, we'd be much better off not overthinking how copyright law works, and just statically leaving this as "2019 - present" and never worrying about it again, but it's not a hill I want to die on or even really take a nap on 😄 so I would accept your idea for a programmatically set copyright notice.

waxlamp (approved):

web/src/components/DandiFooter.vue
```diff
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
-          &copy; 2019 - present The DANDI Team<br>
+          &copy; 2022 - present The DANDI Team<br>
```
yarikoptic, line 18: ok, I think it should be 

```suggestion
          &copy; 2019 - 2022 The DANDI Team<br>
```

as programmatically set by a bot @yarikoptic matching it to the year of the last change. The year when we forget to update it would mean we are no longer actively programmatically maintaining the project ;)

PR #897, opened 2022-02-15
Make ZarrArchives belong to a Dandiset

### LINKED ISSUE #884: Restrict ZarrArchives to belong to a Dandiset
* When creating a ZarrArchive, you must specify a Dandiset that the ZarrArchive will eventually belong to.
* You may not add an Asset containing a ZarrArchive to a dandiset that the ZarrArchive does not belong to. *This means no deduplication for zarrs!*
* ZarrArchives may only be modified (and therefore, uploaded to) by owners of the Dandiset.

dchiquito:
Fixes #884

attention @jwodder @yarikoptic this introduces a change in the zarr creation API endpoint:
image

`dandiset` now must be specified when creating a new ZarrArchive.


yarikoptic:
now that we have merged https://github.com/dandi/dandi-cli/pull/911 I have restarted dandi-cli integration test job, and it errorred out with 
```
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://localhost:8000/api/zarr/
```
see e.g. https://github.com/dandi/dandi-archive/runs/5205798389?check_suite_focus=true

dchiquito:
@yarikoptic The test is trying to save multiple zarrs with the same name in the same dandiset. This is resulting in a DB integrity error which is thrown as a 500, which is admittedly not great. I can change the error code, but the CLI tests will need to change to use different names for different zarrs in the same dandisets.

jwodder:
@dchiquito I can change the client to use the full asset path for the Zarr, but what will/should happen if a Zarr is uploaded to a Dandiset, then deleted, then another Zarr is uploaded at the same path?  (One way around this would be to make the Zarr name a UUID, but at that point, we might as well do away with the field entirely.)

dchiquito:
@jwodder Perhaps I'm not understanding your hypothetical, but I don't see any problems with that. When the zarr is deleted, all the data would be deleted from S3 and the row in the ZarrArchive table would be deleted, so there is nothing to conflict with the second zarr archive.

jwodder:
@dchiquito OK.  I thought the current lack of garbage collection might mean that the Zarr record stayed around after deletion.

jwodder:
PR created on client: https://github.com/dandi/dandi-cli/pull/912

yarikoptic:
merged in dandi-cli, reran job here

dchiquito:
@jwodder https://github.com/dandi/dandi-archive/runs/5219563693?check_suite_focus=true tests are still failing for the same reason. I assume two different zarrs with the same path are being created in the same dandiset.

jwodder:
@dchiquito Yes, the master branch does that.  Re-use/updating of Zarrs [has not been merged yet](https://github.com/dandi/dandi-cli/pull/907).

dandiapi/api/views/zarr.py
```diff
@@ -106,8 +111,14 @@ def create(self, request):
         serializer.is_valid(raise_exception=True)
 
         name = serializer.validated_data['name']
-        zarr_archive: ZarrArchive = ZarrArchive(name=name)
-        zarr_archive.save()
+        dandiset = get_object_or_404(
+            Dandiset.objects.visible_to(request.user), id=serializer.validated_data['dandiset']
+        )
+        zarr_archive: ZarrArchive = ZarrArchive(name=name, dandiset=dandiset)
```
AlmightyYakob, line 117: Would we need to have a permission check here to ensure the user creating the zarr is an owner on the specified dandiset? As is, it seems that any authenticated (but not necessarily authorized) user could create a zarr archive on any dandiset without one yet.

AlmightyYakob (changes requested):
Looks good, one small question

dandiapi/api/views/zarr.py
satra: yes.

dchiquito: Good catch, fixed

AlmightyYakob (approved):

PR #898, opened 2022-02-15
Add operation_id to zarr explore endpoint

### LINKED ISSUE #896: Fix duplicate zarr_read OpenAPI entry
There are two routes in our OpenAPI specification named `zarr_read`:
image
One of them should be explicitly renamed.

dchiquito:
Fixes #896

### DIFF
--- a/dandiapi/api/views/zarr.py
+++ b/dandiapi/api/views/zarr.py
@@ -204,6 +204,7 @@ class ZarrViewSet(ReadOnlyModelViewSet):
             pattern=r'.*',
         )
     ],
+    operation_id='zarr_file',
 )
 @api_view(['GET', 'HEAD'])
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/zarr.py
```diff
@@ -204,6 +204,7 @@ def delete_files(self, request, zarr_id):
             pattern=r'.*',
         )
     ],
+    operation_id='zarr_file',
```
waxlamp, line 207: Based on the difference in paths between the two clashing endpoints, I thought a name like `zarr_path_read` would be good, but I think something more descriptive than `zarr_file` is warranted.

dchiquito: Renamed to `zarr_content_read`

waxlamp (approved):

AlmightyYakob (approved):

PR #903, opened 2022-02-16
Use `validate` serializer method to check `blob_id` and `zarr_id`

AlmightyYakob:
I noticed this as a result of reviewing #897, but didn't want to derail that PR

This cleans up the logic around asset create/update.


dandiapi/api/views/asset.py
```diff
@@ -291,14 +300,12 @@ def create(self, request, versions__dandiset__pk, versions__version):
                 )
                 asset_blob = None
             zarr_archive = None
-        elif 'blob_id' not in serializer.validated_data and 'zarr_id' in serializer.validated_data:
+        else:
```
mvandenburgh, line 303: I actually think the explicit distinction between `blob_id` and `zarr_id` in these if/elif statements makes the intent here clearer. Also, if we ever have additional "special" assets like zarr, I fear that having the `else` block implicitly mean it's a zarr would confuse things

What do you think of keeping 
```
elif 'blob_id' not in serializer.validated_data and 'zarr_id' in serializer.validated_data`:
	.....
```
And then changing the `else` block on line 300 to be
```
else:
	raise NotImplementedError('.....')
```

(Obviously this would apply to line 395+ too)

AlmightyYakob: That's a good point, I'll change this

AlmightyYakob:
@mvandenburgh I included your suggestion, and was actually able to factor out more than I originally thought. Lmk what you think.

AlmightyYakob:
This will certainly conflict with #901, so I'll probably wait for that to merge, and then rebase off master.

dandiapi/api/views/asset.py
```diff
@@ -135,6 +135,29 @@ def _paginate_asset_paths(
     return {'results': paths.data, 'count': asset_count, 'next': next_url, 'previous': prev_url}
 
 
+def _maybe_validate_asset_metadata(asset: Asset):
```
dchiquito, line 138: what is conditional about this method?

dchiquito (approved):
Much better than what was there, thanks!

dandiapi/api/views/asset.py
AlmightyYakob: If this is a zarr asset, it doesn't validate metadata

mvandenburgh (approved):

PR #908, opened 2022-02-18
Update asset metadata when zarr is modified

### LINKED ISSUE #888: Update asset metadata when zarr is updated
Right now Assets are considered to be immutable, which does not work with the current model of having ZarrArchives change over time. Asset metadata should be updated whenever the ZarrArchive is updated.

dchiquito:
Fixes #888


AlmightyYakob (approved):
LGTM

PR #923, opened 2022-02-24
Add `dense` prop to meditor form elements

mvandenburgh:
This slightly improves the look of the meditor by condensing the sizes of the form inputs.

Before / After:

before
after


mvandenburgh:
@waxlamp - just pinging you to make you aware since you suggested it (and I do agree it's an improvement), no review needed

I'll merge this after CI passes, I don't think a code review is necessary for change as minor as this

PR #924, opened 2022-02-24
Fix meditor data loss issue

### LINKED ISSUE #728: Back button in meditor can lose user's work
Open the meditor, and make an edit. Then, instead of clicking the orange save icon, click on the browser back button. Note that the user's work is now lost.

I can think of two solutions here:
1. Use the standard browser mechanism of making a user confirm leaving a page when there is unsaved work therein.
2. Continuously autosave the metadata without the need for a save button.

(1) seems easier in the short term, but (2) seems like a more robust, modern solution to the problem on the whole.

mvandenburgh:
- Add `beforeunload` event listener that displays a browser prompt if user tries to navigate away from page (clicking refresh, ~~back button~~, closing window, etc) with unsaved meditor changes.
- Prevent meditor dialog window from being closed if there are unsaved meditor changes.

Closes #728 

Edit: this actually doesn't work for the back button. Due to how SPAs work, it looks like we'll need to use some `vue-router` hook to handle that case.

### DIFF
--- a/web/src/components/Meditor.vue
+++ b/web/src/components/Meditor.vue
@@ -187,7 +187,7 @@
 import type { JSONSchema7 } from 'json-schema';
 
 import {
-  defineComponent, ref, computed, ComputedRef,
+  defineComponent, ref, computed, ComputedRef, onMounted,
 } from '@vue/composition-api';
 
 import jsYaml from 'js-yaml';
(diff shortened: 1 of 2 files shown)


mvandenburgh:
Superseded by #947

PR #926, opened 2022-02-25
Add Dandiset Zarr Ingestion Script

AlmightyYakob:
This is a follow up to #914, which adds a method for dispatching ingestion for all zarrs within a given dandiset.

### DIFF
--- a/dandiapi/api/admin.py
+++ b/dandiapi/api/admin.py
@@ -26,7 +26,7 @@ from dandiapi.api.models import (
     ZarrArchive,
     ZarrUploadFile,
 )
-from dandiapi.api.tasks.zarr import ingest_zarr_archive
+from dandiapi.api.tasks.zarr import ingest_dandiset_zarrs, ingest_zarr_archive
 from dandiapi.api.views.users import social_account_to_dict
 
 admin.site.site_header = 'DANDI Admin'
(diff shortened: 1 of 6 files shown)


dandiapi/api/admin.py
```diff
@@ -97,6 +97,23 @@ class DandisetAdmin(GuardedModelAdmin):
...
+        for dandiset in queryset:
+            ingest_dandiset_zarrs(dandiset_id=dandiset.id)
+
+        # Return message
+        plural = 's' if queryset.count() > 1 else ''
+        self.message_user(
+            request,
+            f'Ingesting zarr archives for {queryset.count()} dandiset{plural}.',
+            messages.SUCCESS,
+        )
```
dchiquito, line 111: why not use ngettext here?

dchiquito (approved):
I'm curious what the use case of this task is. I think most dandisets will not have that many zarrs, so it seems simple enough to just select the zarrs in the dandiset and ingest those.

satra:
> most dandisets will not have that many zarrs

what do you consider that many? 108 will have about 4000 when completed.

dandiapi/api/admin.py
AlmightyYakob: It just seemed more messy than it was necessary in this case, Since the only change is `dandiset(s)`. However, if you think it's worthwhile to use `ngettext` here as well, I can change it.

PR #932, opened 2022-03-01
Allow running frontend/backend prod deploy actions from github UI

mvandenburgh:
This will let us bypass the release process and run the deploy actions directly from the GitHub UI, and will prevent us from having to cherry-pick out commits when releasing to production.


waxlamp (approved):

dchiquito (approved):

mvandenburgh:
We discussed this offline and decided to go with an approach based on intuit/auto instead.

Issue #934, opened 2022-03-01
Fix navbar responsive design styling


mvandenburgh:
The navbar behaves strangely when the window width is a certain size (i.e. buttons are misplaced/wrap incorrectly, the DANDI logo shrinks too much, etc). There's also some inconsistent behavior between Chromium-based browsers and Firefox. Overall, it seems like the navbar styling needs to be reworked significantly.

waxlamp:
@jtomeck, are you available to take a look at this issue and formulate a plan to fix it?

waxlamp:
Thanks @jtomeck! Let me know if you want to talk.

dandibot:
:rocket: Issue was released in [`v0.2.8`](https://github.com/dandi/dandi-archive/releases/tag/v0.2.8) :rocket:

PR #935, opened 2022-03-01
Make DANDI svg icon a fixed size

### LINKED ISSUE #933: DANDI logo isn't displayed in Firefox
The DANDI logo that links back to the home page doesn't show up if viewing the site with Firefox. It displays correctly with chromium-based browsers though (tested with brave and google chrome)

Screenshot from 2022-03-01 14-50-12

mvandenburgh:
Closes #933

This is just a stopgap measure to make sure the behavior of the DANDI logo is consistent across both chrome and firefox; there's an underlying issue with the styling that causes this issue and others that I have captured in #934.

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -47,8 +47,8 @@
         <v-img
           alt="DANDI logo"
           contain
-          max-height="48px"
-          max-width="120px"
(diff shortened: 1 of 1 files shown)


dchiquito (approved):
This works in Firefox and doesn't seem to behave any differently from before.

PR #941, opened 2022-03-03
Fix DRF UI for Asset endpoints

### LINKED ISSUE #917: DRF UI no longer shown for Asset endpoints
Endpoints that are part of the asset viewset no longer render the DRF HTML page and instead return raw JSON. Example https://api-staging.dandiarchive.org/api/dandisets/101233/versions/draft/assets/7e10c90b-3189-4608-91c2-5350273d4820/

(reported by @satra)

`git bisect` indicates [this commit](https://github.com/dandi/dandi-archive/commit/6be0088d079cf9dd7c766c17df19ed4b8ef096c2) broke it.

AlmightyYakob:
Closes #917

I'm not sure why this ever worked in the past but apparently using `JsonResponse` instead of just `Response` was the issue.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -159,17 +159,9 @@ class AssetViewSet(DetailSerializerMixin, GenericViewSet):
         self.raise_if_unauthorized()
         return super().get_queryset()
 
-    @swagger_auto_schema(
-        responses={
-            200: 'The asset metadata.',
-        },
-        manual_parameters=[ASSET_ID_PARAM],
-        operation_summary="Get an asset\'s metadata",
-        operation_description='',
-    )
     def retrieve(self, request, **kwargs):
         asset = self.get_object()
-        return JsonResponse(asset.metadata)
+        return Response(asset.metadata)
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/asset.py
```diff
@@ -159,17 +159,9 @@ def get_queryset(self):
...
         return super().get_queryset()
 
-    @swagger_auto_schema(
-        responses={
-            200: 'The asset metadata.',
-        },
-        manual_parameters=[ASSET_ID_PARAM],
-        operation_summary="Get an asset\'s metadata",
-        operation_description='',
-    )
```
mvandenburgh, line 169: Why remove this?

AlmightyYakob: The single manual asset ID parameter was unnecessary, as it's already covered by the `lookup_field` and `lookup_value_regex` class fields. However, I somehow overlooked the `responses` and `operation_summary` fields, so I'll add those back.

mvandenburgh (approved):
LGTM pending linting fix

PR #948, opened 2022-03-04
Default all viewsets to use IsApprovedOrReadOnly

### LINKED ISSUE #638: Make DRF permission classes opt-out instead of opt-in
We're currently using custom [DRF permission classes](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/permissions.py) to ensure that only approved users have access to write-enabled endpoints. These must be manually added to the endpoints' viewsets for them to take effect. 

There are some risks/potential security vulnerabilities that could come up with this:
1) There is now added overhead in that developers must remember to decorate new viewsets with these permission classes. 
2) If in the future we introduce any third party libraries that contain viewsets that write to our database, we can't add these permission classes to them. (thanks @danlamanna for catching this one!)

To fix this, I propose that we make these permission classes "opt-out" instead of "opt-in". In other words, they would be applied to every viewset automatically _unless_ a viewset explicitly says otherwise.

AlmightyYakob:
Closes #638

@dchiquito The test I added essentially mirrors the embargo visibility test https://github.com/dandi/dandi-archive/blob/23a5352683d05f9a8dbfef07880d7662886315dc/dandiapi/api/tests/test_embargo.py#L61-L86

Do you think it's worth unifying these two tests? The main difference is that my test includes a few more endpoints, and doesn't worry about embargo.


AlmightyYakob:
Actually after looking at this PR some more, I realized there are some issues with this. I'm going to put it back to draft and think about it. @dchiquito feel free to still leave your review if you've already started it.

dchiquito (approved):

dandiapi/api/tests/test_permission.py
```diff
@@ -0,0 +1,117 @@
...
+    url_format,
+    owner_required,
+):
+    dandiset = dandiset_factory()
+    version = draft_version_factory(dandiset=dandiset)
+    zarr = zarr_archive_factory(dandiset=dandiset)
+    asset = draft_asset_factory()
+    version.assets.add(asset)
+
+    url = url_format.format(dandiset=dandiset, asset=asset, zarr=zarr)
```
mvandenburgh, line 88: This is cool, for some reason it never occurred to me that you can do this to pass data from other fixtures into parameterized strings :open_mouth:

mvandenburgh (approved):

dandiapi/api/tests/test_permission.py
AlmightyYakob: Credits to @dchiquito for originally doing this in the embargo visibility tests.

PR #949, opened 2022-03-07
Fix DLP console errors

mvandenburgh:
#947 introduced some console errors on the DLP, causing tests to fail. This PR fixes them and should cause CI to pass again on that PR.


mvandenburgh:
Sorry, I just realized this is a bug that was actually introduced by #947, so... i 'll just fix it in that PR :sweat_smile:. Closing this

PR #951, opened 2022-03-07
Pin minio version

AlmightyYakob:
Some of the recent backend CI runs have been [failing](https://github.com/dandi/dandi-archive/runs/5454211381?check_suite_focus=true), and it seems to be tied to the latest `bitnami/minio` container image. I'm not sure the underlying cause, but pinning the version seems to fix the issue.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - ${DOCKER_RABBITMQ_CONSOLE_PORT-15672}:15672
 
   minio:
-    image: minio/minio:latest
+    image: minio/minio:2022.3.3
     # When run with a TTY, minio prints credentials on startup
     tty: true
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
       - ${DOCKER_RABBITMQ_CONSOLE_PORT-15672}:15672
 
   minio:
-    image: minio/minio:latest
+    image: minio/minio:2022.3.3
```
mvandenburgh, line 18: Is this a valid tag for the official non-bitnami minio docker image? I briefly glanced at https://hub.docker.com/r/minio/minio/tags and they seem to use a different versioning scheme.

AlmightyYakob: Good catch, that was lazy on my part. Seems like `minio/minio:RELEASE.2022-03-03T21-21-16Z` is the corresponding version. I'll update it.

mvandenburgh (approved):

PR #958, opened 2022-03-08
Sync with the girder 4 cookiecutter

mvandenburgh:
There's been some notable upstream changes in the cookiecutter that I think we should consider downstreaming to dandi. In particular,
- Update `django-composed-configuration`
	- This will get rid of the `django-girder-style` deprecation warnings we're seeing when running tests
	- This also has the side effect of upgrading to Django 4
- Update pytest `filterwarnings` to de-clutter test output
- Update to Python 3.9
	- Python 3.9 is now the "official" python version of the cookiecutter/girder 4.


waxlamp:
Thanks for doing this, @mvandenburgh.

We need to really make sure that the upgrade to Django 4 and Python 3.9 doesn't affect any front-facing functionality.

dchiquito (approved):
I see nothing concerning :+1:

AlmightyYakob (commented):
For some reason I never actually left a review despite looking at everything. LGTM

AlmightyYakob (approved):

PR #959, opened 2022-03-08
Don't crash web client when server is down, provide mechanism to display custom error message

### LINKED ISSUE #921: Frontend should fail gracefully if connection to API server fails
Currently, the frontend will crash and display a blank screen if it's unable to connect to the API server. Instead, this type of error should be caught and an error message should be displayed, perhaps with support for a custom error message via an environment variable.

mvandenburgh:
Closes #921 
Screenshot from 2022-03-10 15-09-12

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -31,3 +31,9 @@ For example, to migrate to schema version 0.5.1, run:
 ```bash
 yarn migrate 0.5.1
 ```
+
+
+### Environment Variables
+
+- VUE_APP_SERVER_DOWNTIME_MESSAGE
+  - A custom error message to be displayed when the backend server can't be reached
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -1,26 +1,65 @@
 <template>
-  <v-app class="dandi-app">
+  <v-app
+    v-if="verifiedServerConnection"
+    class="dandi-app"
+  >
     <AppBar />
-    <v-main>
+    <v-main v-if="connectedToServer">
       <UserStatusBanner />
       <router-view />
       <DandiFooter />
     </v-main>
+
+    <v-main
+      v-else
+      class="d-flex align-center text-center"
+    >
+      {{ SERVER_DOWNTIME_MESSAGE }}
+    </v-main>
   </v-app>
 </template>
 
-<script>
+<script lang="ts">
+import { defineComponent, onMounted, ref } from '@vue/composition-api';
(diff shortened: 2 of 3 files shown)


web/src/App.vue
```diff
@@ -1,26 +1,65 @@
...
+        store.dispatch.dandiset.fetchSchema(),
+        dandiRest.restoreLogin(),
+      ]).then(() => {
+        connectedToServer.value = true;
+      }).catch(() => {
+        connectedToServer.value = false;
+      }).finally(() => {
+        verifiedServerConnection.value = true;
+      });
+    });
```
AlmightyYakob, line 54: I don't see the need for two variables here. Couldn't you just use `connectedToServer`, defaulting to `true` and setting to `false` in the `catch` case?

web/README.md
```diff
@@ -31,3 +31,9 @@ For example, to migrate to schema version 0.5.1, run:
 ```bash
 yarn migrate 0.5.1
 ```
+
+
+### Environment Variables
+
+- VUE_APP_SERVER_DOWNTIME_MESSAGE
+  - A custom error message to be displayed when the backend server can't be reached
```
AlmightyYakob, line 39: I don't much see the benefit in providing this as an env var. Is there a specific need for this? I'd like to avoid adding extra environment variables unless they're necessary.

mvandenburgh: Yes, one of the requested features was to have a way to configure a custom message. During that Heroku outage a few weeks back, we wanted to display a message mentioning the specifics but had no way to quickly do that. Using an environment variable means we can go into the netlify gui/cli and just set it as opposed to needing to make a PR and cut a release.

web/src/App.vue
mvandenburgh: I believe that would introduce a race condition. If the server is down or `fetchSchema()`/`restoreLogin()` takes a long time, and just `connectedToServer` is used and defaulted to true, it's possible that `router-view` and it's children will render before `connectedToServer` is set properly. This would result in console errors since nothing in the store is initialized yet.

AlmightyYakob: I see. Although in it's current state, it appears you'd just see a blank page when connecting, if your connection was very slow. Even when testing with a deliberately slow connection this was pretty brief though, so I think this is fine.

AlmightyYakob (approved):

PR #963, opened 2022-03-11
Type check vue template code

mvandenburgh:
Adds type checking for vue templates and fixes any type errors that were uncovered.


web/src/views/DandisetLandingView/ShareDialog.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <v-dialog
+    v-if="meta && meta.name"
```
AlmightyYakob, line 3: Just curious, are you able to use the optional chaining operator in the template?

```suggestion
    v-if="meta?.name"
```
If so, I'd recommend we use this in the template wherever possible.

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -267,7 +270,7 @@ export default defineComponent({
   },
   setup(props, ctx) {
     const location = ref(rootDirectory);
-    const itemToDelete = ref(null);
+    const itemToDelete = ref();
```
AlmightyYakob, line 273: I don't see a need to type this as `any`

```suggestion
    const itemToDelete = ref<AssetFile | null>(null);
```

AlmightyYakob (changes requested):
Overall looks good, a few comments.

One other thing is that if possible I'd suggest creating separate PRs for separate issues, it seems like this PR addresses both template type checking and composition API conversion. Although I image the type checker might have been giving you reasons to do this all at once.

web/src/views/DandisetLandingView/ShareDialog.vue
mvandenburgh: I wish, unfortunately not :(

https://github.com/vuejs/vue/issues/11088

mvandenburgh:
> One other thing is that if possible I'd suggest creating separate PRs for separate issues, it seems like this PR addresses both template type checking and composition API conversion. Although I image the type checker might have been giving you reasons to do this all at once.

Yeah, there were type errors in those components that were only detected once I ran the new type checker on them, and they were small enough that it was easier to just rewrite them using composition API. But yes I will try to be more careful with that going forward, it probably should've been a separate PR set to merge into this one.

web/src/views/FileBrowserView/FileBrowser.vue
mvandenburgh: You're right, that's the correct type; I also had to make some additional changes to satisfy the type checker due to the limited support for typescript in Vue templates.

On another note unrelated to this PR, just curious, I noticed you used this syntax for declaring the ref type
`const itemToDelete = ref<AssetFile | null>(null);`
instead of
`const itemToDelete: Ref<AssetFile | null> = ref(null);`.

Is there any specific reasoning behind that? I've been using the latter because feels cleaner to me and is similar to the syntax for Python type hints + other languages I've worked with that are actually statically typed, but I've seen both syntaxes used and I'm not sure what the advantages of using one over the other are.

AlmightyYakob:
Looks good, the only thing I'm curious about is why 2d29534 is included in this PR? This commits seems to be taken from https://github.com/dandi/dandi-archive/pull/959, even though that's in master?

mvandenburgh:
> Looks good, the only thing I'm curious about is why [2d29534](https://github.com/dandi/dandi-archive/commit/2d29534aa29858b098707e6749084b3bedc7b1b1) is included in this PR? This commits seems to be taken from #959, even though that's in master?

Strange, I must have messed something up when I merged master. I rebased and the commits appear to be correct now.

PR #964, opened 2022-03-11
Prevent responsive breakpoints from closing meditor

### LINKED ISSUE #915: Prevent meditor from closing upon reaching responsive breakpoints
To reproduce: go to any DLP, open the meditor, then resize the window smaller in the  horizontal direction until the meditor disappears. Alternatively, resize the window to be narrow first, then open the meditor, and enlarge the window until the meditor disappears.

In both cases, the meditor closing will lose any unsaved changes; the meditor should stay open instead.

mvandenburgh:
Fixes #915 

This basically just moves the `Meditor` to a higher level in the component tree. Before, the tree looked like this

```
<App>
  <router-view>
    <DandisetLandingView>
      <DandisetSidebar>
        <DandisetActions>
          <Meditor />
        </DandisetActions>
      </DandisetSidebar>
    </DandisetLandingView>
  </router-view>
</App>
```

Now, it looks like this
```
<App>
  <router-view>
    <DandisetLandingView>
      <Meditor />
    </DandisetLandingView>
  </router-view>
</App>
```

### DIFF
--- a/web/src/components/Meditor/Meditor.vue
+++ b/web/src/components/Meditor/Meditor.vue
@@ -1,222 +1,227 @@
 <template>
-  <v-card
-    v-if="schema && model"
-    v-page-title="model.name"
-    class="overflow-hidden"
+  <v-dialog
+    v-model="open"
+    max-width="85vw"
   >
-    <v-dialog
-      v-model="loadFromLocalStoragePrompt"
-      persistent
-      max-width="60vh"
+    <v-card
+      v-if="schema && model && editorInterface"
+      v-page-title="model.name"
+      class="overflow-hidden"
     >
-      <v-card class="pb-3">
-        <v-card-title class="text-h5 font-weight-light">
-          Attention
-        </v-card-title>
-        <v-divider class="my-3" />
-        <v-card-text>
-          You have previously edited this dandiset and have unsaved data.
-        </v-card-text>
-        <v-card-text>
-          Would you like to load that data?
-        </v-card-text>
+      <v-dialog
+        v-model="loadFromLocalStoragePrompt"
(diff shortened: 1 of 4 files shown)


web/src/components/Meditor/Meditor.vue
```diff
@@ -1,222 +1,227 @@
 <template>
-  <v-card
-    v-if="schema && model"
-    v-page-title="model.name"
-    class="overflow-hidden"
+  <v-dialog
```
mvandenburgh, line 2: FYI the github diff view makes this look like a big change, but the only substantial change in this template is the entire thing is now wrapped in the `v-dialog` (which was moved from `DandisetActions.vue`).

waxlamp: Random thought to enhance communication about this kind of change: have one commit that just wraps the component in the `<v-dialog>`, but doesn't do the correct indentation, then the next commit with the indentation changes. Then you can link to each commit independently to demonstrate that this is actually a small change.

AlmightyYakob: I actually disagree with this, since that would purposely leave the first commit in an incorrect state, and could cause confusion. IMO this is just an unfortunate reality of these sorts of changes.

AlmightyYakob (approved):
LGTM, functions properly :+1:

web/src/components/Meditor/Meditor.vue
waxlamp: Yeah, I agree with that assessment as well. It depends on if you want each commit to be a logical expression of some bit of change, or a complete and correct state all on its own.

PR #973, opened 2022-03-17
Use `intuit/auto` to manage releases

### LINKED ISSUE #968: Use intuit/auto to manage releases

mvandenburgh:
Closes #968 

Summary of what this gets us -

- `auto` handles creating releases for us. To cut a release, simply tag a PR with the `release` label. 
- I've included the `conventional-commits` auto plugin, which lets us use the [conventional commits](https://intuit.github.io/auto/docs/generated/conventional-commits) format in commits to tell `auto` whether this is a MAJOR, MINOR, or PATCH release. Then, on the next release `auto` will bump the version number accordingly. 
	- `auto` defaults to a PATCH release if no "conventional commits" are in the PR.
- For every release, `auto` will attach the `released` label to any issues that were closed by a PR in that release. 
	- This functionality provided by the [`released`](https://intuit.github.io/auto/docs/generated/released) plugin.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # fetch history for all branches and tags
+
+      - name: Download latest auto
+        run: |
+          auto_download_url="$(curl -fsSL https://api.github.com/repos/intuit/auto/releases/latest | jq -r '.assets[] | select(.name == "auto-linux.gz") | .browser_download_url')"
+          wget -O- "$auto_download_url" | gunzip > ~/auto
+          chmod a+x ~/auto
+
+      - name: Create Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ~/auto shipit
--- a/.autorc
+++ b/.autorc
@@ -1,0 +1,14 @@
+{
+  "plugins": [
+    "git-tag",
+    "all-contributors",
+    "conventional-commits",
+    "first-time-contributor",
+    "released"
+  ],
+  "owner": "dandi",
+  "repo": "dandi-archive",
+  "name": "dandibot",
+  "email": "dandibot@mit.edu",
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
```
waxlamp, line 10: Is this recommended by the auto people? It would be better to have just one special code phrase to skip this workflow. But then again, I don't really understand what this is for. Maybe to prevent some kind of horrific recursive infinite release loop?

waxlamp (commented):
Neat! How do we test this? Just stick it in and try a PR with the release tag?

.github/workflows/release.yml
mvandenburgh: > Maybe to prevent some kind of horrific recursive infinite release loop?

Yup - https://intuit.github.io/auto/docs/build-platforms/github-actions

waxlamp: Weird that they have two trigger phrases 🤷🏾

mvandenburgh:
> Neat! How do we test this? Just stick it in and try a PR with the release tag?

Yeah. Or we could even attach a release tag to this PR, since the github action will immediately take effect when this is merged.

waxlamp (approved):
I think go ahead and test this by adding the appropriate tag, and pushing an empty commit with a conventional-commits message that bumps the minor version number. (Minor because the patch level gets updated by default, and perhaps this change could herald our shift from 0.1.50 to 0.2.0; other opinions on this plan welcome.)

dchiquito:
@mvandenburgh dandi-cli and dandischema have something set up where you can label a PR with `patch`, `minor`, or `major` and it will increment the version appropriately before releasing. Is that enabled in this setup, or do we need to use conventional commits? I personally like the labels more, since you don't need to worry about version incrementing while committing your code.

AlmightyYakob:
@mvandenburgh Does this still allow for manual release? I.e. if we needed to release for some reason independent of a PR. Would that interfere with this setup?

mvandenburgh:
> @mvandenburgh Does this still allow for manual release? I.e. if we needed to release for some reason independent of a PR. Would that interfere with this setup?

Manual releases should continue to work, as long as the release tag follows semantic version format.

AlmightyYakob (approved):

mvandenburgh:
:rocket: PR was released in [`v0.2.0`](https://github.com/dandi/dandi-archive/releases/tag/v0.2.0) :rocket:

PR #990, opened 2022-03-29
Set up pyppeteer e2e tests

### LINKED ISSUE #738: E2E test overhaul to enable testing publish, file browser, etc
At the moment we are unable to test some of the core features of our system in our E2E tests since dandi-cli is a prerequisite to uploading assets. There should be some way to put the database into a predetermined "testing" state that contains sample dandisets with assets that can be used in testing. See discussion in dandi/dandiarchive#861

mvandenburgh:
This PR 

- sets up `girder-pytest-pyppeteer` to be run locally through tox or in CI with GitHub Actions
- converts two of the simpler puppeteer e2e tests to pyppeteer
	- I was thinking the remaining conversions could be done incrementally. This would ease review and also let us tackle these as time/cycles permit. In the meantime we can continue running the puppeteer-based tests in CI alongside the pyppeteer ones (which is exactly what I've set up in this PR). I've filed #995 for this.
- Separate E2E tests into their own CI job
	-  As part of this I've changed things so the E2E tests run for all changes, not just frontend changes. It's completely plausible for a backend-only change to break the frontend, which I must not have thought about when I set it up that way.

Closes #738.


dandiapi/api/tests/e2e/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    await page.reload()
+    await wait_for_navigation(page)
+
+
+async def contains_text(page: Page, text: str):
+    """Determine if the given page contains the given text."""
+    from pyppeteer.errors import TimeoutError
+
+    try:
+        await page.waitForXPath(f'//*[contains(text(), "{text}")]', {'timeout': 500})
```
dchiquito, line 38:
```suggestion
        await page.waitForXPath(f'//*[contains(., "{text}")]', {'timeout': 500})
```
Say you are searching for the string `Name: Foo Bar` when the DOM looks like this:
```
<div>
  Name:
  <span>Foo Bar</span>
</div>
```
`.` will work, while `text()` will not. See https://stackoverflow.com/a/41862373 for a fuller discussion.

dchiquito (approved):

PR #991, opened 2022-03-30
Bump dandischema to 0.6.0

### LINKED ISSUE #925: zarr_content_read: include sizes for the "files" (and possibly "directories")
motivation: https://github.com/dandi/dandi-cli/pull/923#issuecomment-1050296836  -- in dandi-cli, to possibly avoid relatively expensive checksumming of the content to decide either a (changed) file within zarr needs to be reuploaded, we first check the size(s) of the file.  If that one changed -- we know that we need to reupload even without computing its checksum.

It would be helpful if that API endpoint also returned sizes of the files (if it can do it without extra runtime penalty)

dchiquito:
https://github.com/dandi/dandischema/pull/120 made some changes to the zarr API:
* Add `size` of files/directories to the serialization
* Only record the name of the file/directory in the .checksum file instead of the entire path relative to the zarr root.
* Add suffix `-{file_count}--{file_size}` to directory digests
* ZarrChecksum.md5 -> ZarrChecksum.digest, since directories are no longer technically just an MD5
* ZarrChecksum.path -> ZarrChecksum.name, since only the name is stored rather than the whole path.

A depressingly large number of changes are required to enact this change.

Fixes #925

### DIFF
--- a/dandiapi/api/zarr_checksums.py
+++ b/dandiapi/api/zarr_checksums.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+from collections.abc import Mapping
 from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Mapping, Optional
+from typing import TYPE_CHECKING, Optional
 
 from django.conf import settings
 from django.core.files.base import ContentFile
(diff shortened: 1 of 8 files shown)


AlmightyYakob (approved):
This all looks good to me, albeit a lot to take it. Just one suggestion really

dandiapi/api/zarr_checksums.py
```diff
@@ -3,7 +3,7 @@
 from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Mapping, Optional
```
mvandenburgh, line 6:
```suggestion
from collections.abc import Mapping
from typing import TYPE_CHECKING, Optional
```

`typing.Mapping` is deprecated now https://peps.python.org/pep-0585/#implementation

dchiquito: TIL, cool

PR #997, opened 2022-04-05
Use history mode instead of hash mode (remove hash from GUI URL)

### LINKED ISSUE #785: Web UI (vue): switch from "hash" to "history" mode?
For the problem statement/goals, see dandi/dandiarchive#427.

I am no vue anything (not even a novice) so may be this is a wrong direction:  Quick search lead me to https://stackoverflow.com/questions/34623833/how-to-remove-hashbang-from-url which in turn to https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations which suggests that there is a "history" mode which would make it possible to make URLs look normal (without placing all args into a fragment).

with switching away from using `hash` (fragment) mode in web ui we would be able to place `dandiarchive.org` behind reverse proxy (#427) to provide any necessary consistent mapping from permalinks to web ui and/or API end points without redirects.

AlmightyYakob:
Closes #785

### DIFF
--- a/web/netlify.toml
+++ b/web/netlify.toml
@@ -9,3 +9,8 @@ command = "yarn run build"
   VUE_APP_DANDI_API_ROOT = "https://api-staging.dandiarchive.org/api/"
   VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
   VUE_APP_SENTRY_ENVIRONMENT = "staging"
+
+[[redirects]]
+from = "/*"
+to = "/index.html"
+status = 200
--- a/web/src/router.ts
+++ b/web/src/router.ts
@@ -51,4 +51,4 @@ const routes: RouteConfig[] = [
   },
 ];
 
-export default new Router({ routes });
+export default new Router({ mode: 'history', routes });
--- a/web/src/views/HomeView/HomeView.vue
+++ b/web/src/views/HomeView/HomeView.vue
@@ -68,6 +68,11 @@ export default defineComponent({
         router.replace(trimmed);
       }
     });
+
+    return {
+      // Hack to satisfy template type checker
+      $vuetify: (ctx.root as any).$vuetify,
+    };
   },
 });
 </script>
--- a/web/src/views/HomeView/HomeView.vue
+++ b/web/src/views/HomeView/HomeView.vue
@@ -50,11 +50,25 @@
 <script lang="ts">
 import StatsBar from '@/views/HomeView/StatsBar.vue';
 import DandisetSearchField from '@/components/DandisetSearchField.vue';
-import { defineComponent } from '@vue/composition-api';
+import { computed, defineComponent, watchEffect } from '@vue/composition-api';
 
 export default defineComponent({
   name: 'HomeView',
   components: { StatsBar, DandisetSearchField },
+  setup(props, ctx) {
+    /**
+    * Redirect old hash URLS to the correct one. This is only done on
+    * the home page, since any URL with a hash will default to here.
+    */
+    const currentRoute = computed(() => ctx.root.$route);
+    watchEffect(() => {
+      const router = ctx.root.$router;
+      if (currentRoute.value.hash) {
+        const trimmed = router.currentRoute.hash.replace('#', '');
+        router.replace(trimmed);
+      }
+    });
+  },
 });
 </script>
 
--- a/web/src/views/HomeView/StatsBar.vue
+++ b/web/src/views/HomeView/StatsBar.vue
@@ -51,7 +51,7 @@ export default defineComponent({
         name: 'dandisets',
         value: dandisets.value,
         description: 'A DANDI dataset including files and dataset-level metadata',
-        href: '/#/dandiset',
+        href: '/dandiset',
       },
       { name: 'users', value: users.value },
       { name: 'total data size', value: filesize(size.value, { round: 0, base: 10, standard: 'iec' }) },


AlmightyYakob:
> @AlmightyYakob - the current redirector does a few things that only a server can do. e.g. [https://github.com/dandi/redirector/blob/master/serve.py#L132](https://github.com/dandi/redirector/blob/master/serve.py?rgh-link-date=2022-04-05T15%3A12%3A47Z#L132) (search for other HEAD requests and also the sitemap generation). if this is something that netlify could do, we can simply make netlfiy be the canonical source for `dandiarchive.org` and `staging.dandiarchive.org`.
> 
> would you be able to look into whether netlify can do those things?
> 
> if netlify cannot do that we can copy over this change to the redirector and serve the compiled js from the redirector. netlify can continue to do deployment previews, but it won't serve gui.dandiarchive.org.
> 
> we will want to integrate the redirector deployment into terraform and perhaps rewrite that server with fastapi instead of sanic.

I don't think Netlify would be able to do this, since to my knowledge it can't do programmatic redirection, just that defined in the `Netlify.toml`. However, I'd urge **strongly** against trying to replace Netlify with the redirector for serving the GUI, as Netlify provides us a lot of benefits, deploy previews just being one of them.

satra:
i wouldn't replace netlify with the redirector for anything other than serving the instance project. if netlify pushes the compiled js somewhere, we would simply use that. or even use netlify as a proxy. we want to continue having it do deploy previews.

the main goal here is to remove gui.dandiarchive.org as a target. so only that bit would be removed from netlify.

let's go through a few options:

1. keep netlify and let it serve from dandiarchive.org as the base url: 

this would require understanding that any server side requests being done by the redirector has to come from the api server (which is ok). we would need to ensure that we turn on all the pieces of server side rendering in netlify such that our pages get indexed, etc.,. i believe there was an issue to this effect. 

what we lose in this mode is to respond to anything at dandiarchiv.org, since it becomes a client side app. we need to ensure that the CLI understands this and only makes requests to the api. i.e. anyone developing tools has to understand that the web app is not a server, and needs to know what the server is.

2. keep redirector and simply serve the netlify compiled code.

allows us to provide responses to requests and we simply serve the js code for the instance.

i'm ok with implementing 1, but conveying that tools should be built against api will be important.

AlmightyYakob:
> we would need to ensure that we turn on all the pieces of server side rendering in netlify such that our pages get indexed, etc.,. i believe there was an issue to this effect.

As of now I'd say this is out of scope, since currently our GUI is a Singe Page App (SPA), and thus doesn't really play nice with search engines (since mainly JS is returned from netlify, which is then rendered and injected into the html). The topic of SEO and search engine indexing is a separate discussion IMO, as it would be a _much_ larger change.

satra:
> As of now I'd say this is out of scope, since currently our GUI is a Singe Page App (SPA), and thus doesn't really play nice with search engines (since mainly JS is returned from netlify, which is then rendered and injected into the html). 

there is a separate issue open about this. netlify has settings that automatically help with search that can be enabled. but the current role of the sitemap generator is related to search and is done by the redirector service at present.

yarikoptic:
> i.e. anyone developing tools has to understand that the web app is not a server, and needs to know what the server is.

agree, and let's keep https://dandiarchive.org/server-info providing that information so in principle client could "ask" a server about what API server to talk to.  BUT also `/api` could be proxied (or redirected?) to `https://api.dandiarchive.org/api` .

Overall -- in addition to https://github.com/dandi/dandi-archive/pull/997#issuecomment-1089031540 I still see a possible `3.` in making "redirector" into a reverse proxy to "serve" at least the web UI from its respective underlying netflify, and otherwise  not  seen/visited directly by the users (don't know yet if possible to enforce redirection for direct access to https://gui.dandiarchive.org to the reverse proxy - should probably be possible).

dchiquito (approved):
Outside of the larger deployment discussion, I think removing the hash from the URL is a UX improvement and should be done now regardless.

satra:
> I think removing the hash from the URL is a UX improvement and should be done now regardless.

how will existing urls in papers that have the `#` be redirected?

AlmightyYakob:
> how will existing urls in papers that have the `#` be redirected?

Wouldn't this just be a simple update to the redirector, to remove the `#` from the url to redirect to?

satra:
> Wouldn't this just be a simple update to the redirector, to remove the # from the url to redirect to?

the redirector serves the domain `dandiarchive.org` not `gui.dandiarchive.org`. the `#` is only in the gui urls that people have copied and pasted into their papers.

AlmightyYakob:
> the redirector serves the domain `dandiarchive.org` not `gui.dandiarchive.org`. the `#` is only in the gui urls that people have copied and pasted into their papers.

Oh, I didn't realize people were using the GUI url with the `#` included. I thought the whole point of the redirector was so people could use stable dandiset URLs and not need to use GUI URLs directly.. It might be possible to check for that on the client.

AlmightyYakob:
> that was indeed the hope, the reality is people prefer copying from browser url bar rather than the share icon.

Did some quick experimenting with the UI, and I think I could include a change to this PR that would automatically redirect those old URLs to the correct ones.

satra:
> I think I could include a change to this PR that would automatically redirect those old URLs to the correct ones.

that would be great.

AlmightyYakob:
@satra [07dc0f0](https://github.com/dandi/dandi-archive/pull/997/commits/07dc0f04554da7fb8cbd828aa847556821f7506d) contains the change I mentioned. I'd encourage you to play around with the deploy preview to make sure it redirects things as you expect. It's pretty simple but from what I can tell it seems to perform exactly as expected.

satra:
> I can tell it seems to perform exactly as expected.

yup. seems to do the right thing.

mvandenburgh (approved):
LGTM

dchiquito (approved):

Issue #996, opened 2022-04-05
[urgent] if user registration is incomplete, user cannot take further steps


satra:
@mvandenburgh - i think this one is for you.

there are a few users in incomplete state. i know some of them and didn't realize their membership did not go through. i don't know what causes this state, but the worrying part is that in this state, we don't get an email and they cannot reregister.

(some of the incompletes are also possibly spam).

dandibot:
:rocket: Issue was released in [`v0.2.2`](https://github.com/dandi/dandi-archive/releases/tag/v0.2.2) :rocket:

PR #998, opened 2022-04-05
Pin django-allauth to 0.46.0

mvandenburgh:
The latest version of `django-allauth` has a "github confirmation" screen that allows users to bypass github and create a username/password account. It seems that the update to django 4 triggered a fresh `pip install` on Heroku which resulted in this dependency being bumped.

This is a short-term fix for #996 that will prevent users from creating invalid account; in the meantime I will figure out how to disable this screen so we can just use the latest version.


mvandenburgh:
@dchiquito @AlmightyYakob @waxlamp just pinging to make you aware of this - I'll merge and release this once CI passes.

mvandenburgh:
Not sure why pip install is failing in CI. Regardless I have solved this problem using the latest allauth version in #999 - closing this

PR #1006, opened 2022-04-06
Prevent double-publishing

### LINKED ISSUE #955: Display feedback during publish process
Currently, the DLP gives no feedback to the user after clicking "publish". The publish occurs quickly (within the span of a single request; i.e., maximum of 30 seconds) but the page does not indicate that anything is happening. Instead:
- display a throbber after the publish process begins
- respond to the 2xx success code with a success message and/or reloading of the DLP
- respond to error codes with a toast or other error indicator

### LINKED ISSUE #956: Prevent double publishing of dandisets
Initiating a publish for a dandiset leaves the dandiset in a publishable state, meaning that a dedicated agent could rapidly fire off more than one publish request, both of which could succeed, resulting in two different versions of a dandiset being created from the same draft version. Our domain model disallows this, so we need to add a "publishing" state to the system. This "publishing" state would have some advantages:
- it would prevent exactly this double publishing situation
- it would enable the DLP to properly display the status of a dandiset (using, e.g., a disabled publish button with a throbber any time a publishing dandiset's DLP is loaded)

mvandenburgh:
Closes #956
Closes #955

Currently, there is a race condition where two or more simultaneous calls to the `/publish` endpoint may result in multiple publishes occurring. This PR fixes this in two parts:

1) On the backend, the publishing process is now a celery task.
2) On the frontend, when the user clicks the publish button it checks if a publish is ongoing, and if one is then it polls the server continuously until its finished and then displays a toast message linking to the new version.


dandiapi/api/views/version.py
```diff
@@ -102,10 +103,27 @@ def update(self, request, **kwargs):
...
+                # `nowait` is used here so that this request aborts and returns a 400 error as soon
+                # as it sees that the target Version is locked. The default behavior of
+                # `select_for_update` is to block until the lock can be acquired, but in this case
+                # we don't need to do that and incur the performance hit that comes with it.
+                queryset.select_for_update(nowait=True),
+                **kwargs,
+            )
+        except DatabaseError:
+            raise ValidationError('Dandiset is already in the process of being published')
+        self.check_object_permissions(self.request, old_version)
```
mvandenburgh, line 125: Unfortunately we lost a bit of the convenience/safety that `get_object` provides with automatic permission checking, so it must be called explicitly here. It seems that there's been some upstream discussions about supporting `select_for_update` in `get_object` on both the Django and DRF sides, but nothing has come of it yet. ([Django issue](https://code.djangoproject.com/ticket/27477), [DRF issue](https://github.com/encode/django-rest-framework/issues/4675))

waxlamp: The check probably deserves a comment.

dandiapi/api/views/version.py
```diff
@@ -102,10 +103,27 @@ def update(self, request, **kwargs):
...
+            old_version: Version = get_object_or_404(
+                # `nowait` is used here so that this request aborts and returns a 400 error as soon
+                # as it sees that the target Version is locked. The default behavior of
+                # `select_for_update` is to block until the lock can be acquired, but in this case
+                # we don't need to do that and incur the performance hit that comes with it.
+                queryset.select_for_update(nowait=True),
+                **kwargs,
+            )
+        except DatabaseError:
+            raise ValidationError('Dandiset is already in the process of being published')
```
waxlamp, line 124: Should this really be a `ValidationError`?

Also, why is this needed, if `select_for_update()` with `nowait` results in a 404?

Final question: if a 404 arises, it should be converted into some other 400-series error specific to "already being published" right?

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -457,7 +458,8 @@ export default defineComponent({
       }
     }
 
-    async function publish() {
+    // Throttle this function so spam-clicking the button doesn't send multiple requests
+    const publish = throttle(async () => {
```
waxlamp, line 462: Why not just disable the button instead, then poll to find out when the publish is finish to re-enable?

I guess the juicier question is what to expect if a bunch of people happen to all click publish on the same Dandiset at the same time. It seems like one of them would "win" the race, and the rest would receive an error message that needs to be handled?

dandiapi/api/views/version.py
mvandenburgh: > Also, why is this needed, if `select_for_update()` with `nowait` results in a 404?

`select_for_update()` with `nowait` doesn't result in a 404, it throws a `DatabaseError` if the row is locked. Then, from there I catch the `DatabaseError` and raise a `ValidationError` which does raise a 404. The original comment didn't explain this well, so I rewrote it.

> Should this really be a ValidationError?

DRF provides `ValidationError` and `ParseError` exceptions as a conveniences as opposed to manually returning a `Response` with status code. I see what you mean though, I changed it to explicitly return a `Response` with a 423 error instead.

> Final question: if a 404 arises, it should be converted into some other 400-series error specific to "already being published" right?

Yes, I changed it to a 423 (Resource is locked) error.

web/src/views/DandisetLandingView/DandisetPublish.vue
mvandenburgh: > Why not just disable the button instead, then poll to find out when the publish is finish to re-enable?

Throttling seemed simpler, but I think you're right that it's not enough. In addition to the race scenario you described, I hadn't considered that even in the case of a single person clicking publish, if the dandiset has a lot of assets it may take more than a second.

> It seems like one of them would "win" the race, and the rest would receive an error message that needs to be handled?

I agree, this seems like the best solution here - I'll implement this

dandiapi/api/views/version.py
mvandenburgh: Added https://github.com/dandi/dandi-archive/pull/1006/commits/f2c41c3fa54edd43db34b043b28ed03d48d62380

dandiapi/api/views/version.py
waxlamp: Perfect!

mvandenburgh:
> Why not just disable the button instead, then poll to find out when the publish is finish to re-enable?
>
> I guess the juicier question is what to expect if a bunch of people happen to all click publish on the same Dandiset at the same time. It seems like one of them would "win" the race, and the rest would receive an error message that needs to be handled?

@waxlamp I implemented this. This is what happens when two people try to publish at the same time (I added some `sleep()` calls to the backend to make the publish take a long time, I don't think it would every take this long in a real publish)

dandipublish

dchiquito (approved):

djarecka:
@yarikoptic - you mean adding tests to the changes in vue code?

mvandenburgh:
After coming back to this with fresh eyes after 2 months, I think my approach here is fundamentally flawed. The implementation here does technically work, but it relies on the assumption that _a locked `Version` row means it is being published_. This may be a safe assumption for now, but if we need to lock `Versions` for other purposes in the future this will no longer work. 

I believe the ideal solution would be to introduce a `PUBLISHING` status to the `Version.status` field + add a check to the start of the publish endpoint that returns a 429 if `version.status == PUBLISHING`. However, this introduces a potential race condition if another request occurs _after_ checking `version.status` but _before_ setting it to PUBLISHING. So, we need to lock the `Version` using `select_for_update`. This in turn creates another problem, which is if a version takes a long time to be published, the second request thread will be stuck waiting to acquire the lock.

Taking a step back, I think publishing would be more appropriate as a celery task. iirc, this issue was motivated by the fact that dandisets with tens of thousands of assets can take a few seconds (or more) to publish. That reason alone is enough evidence that this should be a celery task, I think. Plus, the concurrency logic becomes a lot simpler than what I have currently-

```python
@transaction.atomic
@action(detail=True, methods=['POST'])
def publish(self, request, **kwargs):
  version: QuerySet[Version] = self.get_queryset().select_for_update().filter(**kwargs).first()
  if version.status == "PUBLISHING":
    return 429
  elif version.version != 'draft':
    return 400
  # other checks.....
  .....
  publish_task.delay(version.pk) # version.status would be set to PUBLISHED at the end of this task
  return 200
```

waxlamp:
I think this is a good design overall. Some thoughts:
1. I think `409 Conflict` might be a better error code than `429 Too Many Requests`. The latter is meant as a rate limiting condition, whereas this is straight up a disallowed action based on the state of the resource.
2. One final edge case to consider: after the `Version` has moved into the `PUBLISHING` state, what do we do if the publishing task crashes or otherwise doesn't complete? One idea is to give the user a way to cancel the publish (thus manually returning the state to `PUBLISHED` and letting them try again).
3. I think you've solved this problem through design, but if "the second request thread [is] stuck waiting to acquire the lock", I think you could do a non-blocking check instead of a blocking one. Essentially, after a short timeout of a few seconds, the second thread could receive a `408 Request Timeout` indicating that it should try again later.

> After coming back to this with fresh eyes after 2 months, I think my approach here is fundamentally flawed. The implementation here does technically work, but it relies on the assumption that _a locked `Version` row means it is being published_. This may be a safe assumption for now, but if we need to lock `Versions` for other purposes in the future this will no longer work.
> 
> I believe the ideal solution would be to introduce a `PUBLISHING` status to the `Version.status` field + add a check to the start of the publish endpoint that returns a 429 if `version.status == PUBLISHING`. However, this introduces a potential race condition if another request occurs _after_ checking `version.status` but _before_ setting it to PUBLISHING. So, we need to lock the `Version` using `select_for_update`. This in turn creates another problem, which is if a version takes a long time to be published, the second request thread will be stuck waiting to acquire the lock.
> 
> Taking a step back, I think publishing would be more appropriate as a celery task. iirc, this issue was motivated by the fact that dandisets with tens of thousands of assets can take a few seconds (or more) to publish. That reason alone is enough evidence that this should be a celery task, I think. Plus, the concurrency logic becomes a lot simpler than what I have currently-
> 
> ```python
> @transaction.atomic
> @action(detail=True, methods=['POST'])
> def publish(self, request, **kwargs):
>   version: Version = self.get_queryset().select_for_update().filter(**kwargs).first()
>   if version.status == "PUBLISHING":
>     return 429
>   elif version.version != 'draft':
>     return 400
>   # other checks.....
>   .....
>   version.status = 'PUBLISHING'
>   version.save()
>   publish_task.delay(version.pk) # version.status would be set to PUBLISHED at the end of this task
>   return 200
> ```

waxlamp:
> > I think this is a good design overall. Some thoughts:
> > 
> > 1. I think `409 Conflict` might be a better error code than `429 Too Many Requests`. The latter is meant as a rate limiting condition, whereas this is straight up a disallowed action based on the state of the resource.
> 
> Whoops, I meant to put `423 Locked` instead of `429 Too Many Requests`. "Locked" seems more descriptive for this scenario to me, what do you think @waxlamp ?
> 

I think that code would be fine. It's probably more suitable than 409, but I don't think it matters much which 4xx code we pick for this.

mvandenburgh:
> One idea is to give the user a way to cancel the publish (thus manually returning the state to `PUBLISHED` and letting them try again).

UI-wise, I think using the cancel button introduced in #1122 would be perfect here.

waxlamp:
@mvandenburgh, status update on this PR? AFAIK, the design discussions we've had here are still valid. Confirm?

mvandenburgh:
> @mvandenburgh, status update on this PR? AFAIK, the design discussions we've had here are still valid. Confirm?

Yes, everything here is still valid, I just have to implement the design. This PR sort of fell off my radar in favor of higher priority bugs, but those seem to be mostly resolved so I'll rebase this and start working on it again

dandiapi/api/tasks/__init__.py
```diff
@@ -184,3 +185,45 @@ def unembargo_dandiset(dandiset_id: int):
...
+
+    # Publish any draft assets
+    draft_assets = old_version.assets.filter(published=False).all()
+    for draft_asset in draft_assets:
+        draft_asset.publish()
+    Asset.objects.bulk_update(draft_assets, ['metadata', 'published'])
+
+    AssetVersions.objects.bulk_create(
+        [AssetVersions(asset_id=asset.id, version_id=new_version.id) for asset in draft_assets]
+    )
```
mvandenburgh, line 219: All of these loops that iterate over every asset in a dandiset are probably going to lead to memory issues when publishing dandisets with large numbers of assets. These are copied 1-1 from the original publishing endpoint code, so I'm not going to attempt to fix them here; I'll do some memory profiling and make a follow up PR once I come up with a solution to this

dandiapi/api/views/version.py
```diff
@@ -105,72 +104,55 @@ def update(self, request, **kwargs):
...
+                        status.HTTP_400_BAD_REQUEST,
+                    )
+            return Response(resp_text, status=resp_status)
+
+        version.status = Version.Status.PUBLISHING
+        version.save(update_fields=['status'])
+
+        transaction.on_commit(lambda: publish_task.delay(version.id))
+
+        return Response(None, status=status.HTTP_200_OK)
```
danlamanna, line 155: This could be 202 Accepted.

dandiapi/api/tasks/__init__.py
```diff
@@ -184,3 +185,45 @@ def unembargo_dandiset(dandiset_id: int):
...
     dandiset.save(update_fields=['embargo_status'])
+
+
+@shared_task
+@atomic
+def publish_task(version_id: int):
+    old_version: Version = Version.objects.get(id=version_id)
+    new_version: Version = old_version.publish_version
+
+    new_version.doi = doi.create_doi(new_version)
```
danlamanna, line 196: This is unrelated, but could you make a follow on PR moving this into on_commit? This is a bug waiting to happen.

danlamanna (approved):

dandiapi/api/tasks/__init__.py
waxlamp: Could you explain further about the bug risk?

mvandenburgh: @waxlamp as it is now, if something goes wrong and the transaction gets rolled back, the DOI is still created.

waxlamp: Ah makes sense, thanks!

mvandenburgh:
@jwodder @yarikoptic the CLI tests are failing here. It appears to be happening because of the changes to the semantics of the publishing process in this PR, so we'll need some updates on the CLI side to unblock this. 

Currently, the publishing process is like this -
- User sends a POST request to the `/publish` endpoint
- The server receives the request, performs the publish operation, sets the `status` of the draft version to `PUBLISHED`, and responds with a 200 OK and info about the newly published version.

This PR updates it to do the following -
- User sends a POST request to the `/publish` endpoint
- The server receives the request, sets the `status` of the draft version to `PUBLISHING`, queues up an async worker task to publish the dandiset, and responds with a 202 ACCEPTED.
- Asynchronously, a worker picks up the publishing task off the queue, performs the publish, and sets the status of the draft version to `PUBLISHED`.

To tell when a version has finished publishing, you can poll the `/api/dandisets/<identifier>/versions/draft/info/` endpoint, and when the `status` field in that response is `"PUBLISHED"`, you can call `/api/dandisets/<identifier>/versions` to get the newly published version (they are sorted by creation time so the first version with a status of `"PUBLISHED"` in that response is guaranteed to be the most recently published version). 

Can you make the necessary changes to the CLI to account for this? Please let me know if anything is unclear or if you need any more information.

Issue #1004, opened 2022-04-06
api/zarr/ -- staging 500s out


yarikoptic:
I was thinking to try for https://github.com/dandi/dandi-archive/issues/925#issuecomment-1090483484 but:

```shell
$> curl -X GET "https://api-staging.dandiarchive.org/api/zarr/" -H  "accept: application/json"

<!doctype html>
<html lang="en">
<head>
  <title>Server Error (500)</title>
</head>
<body>
  <h1>Server Error (500)</h1><p></p>
</body>
</html>
```

dandibot:
:rocket: Issue was released in [`v0.2.4`](https://github.com/dandi/dandi-archive/releases/tag/v0.2.4) :rocket:

PR #1010, opened 2022-04-06
Add force argument in `ingest_zarr_archive`

AlmightyYakob:
Addresses underlying cause of #1004.

@dchiquito Right now, the management command defaults to `force=False`. I'd think most of the time we're using the management command, we'd want to use `force=True`. Do you think we should default `force=True` in the management command? I'm leaning towards yes.

I'm also wondering how we can improve the process of checksum updates. That is, if there is ever a need again to update the checksum format, it'd be good to have some automated process around this, rather than someone needing to manually run a management command. I get the feeling that any automated solution would be overly costly in the average case of not needing to update. @dchiquito any thoughts?


dchiquito (approved):
I think the least destructive option should always be the default, so I vote for `force=False`. The first thing I always do after digging up a management command is checking the `--help` to see what options it has, and possibly reviewing the source code so I know exactly what it's going to do.

I also don't think an automated process is a good idea here. Having a silently watching daemon thingy that will swoop in and reingest everything when it feels like something has changed sounds creepy to me, not reassuring. If we make a change to the data format, we should expect to have to do some data migration, which I would always expect to be a manual process. I would not trust some automated script written in the distant past to handle that migration for me.

PR #1017, opened 2022-04-08
Create design doc for replacing Redirector with Netlify

### LINKED ISSUE #727: Write a design doc for hosting the archive at dandiarchive.org

AlmightyYakob:
Closes #727

This should be very straightforward to implement.

### DIFF
--- a/doc/design/apex-domain-netlify.md
+++ b/doc/design/apex-domain-netlify.md
@@ -1,0 +1,38 @@
+# Serving the GUI at dandiarchive.org with Netlify
+Currently, the GUI is served at gui.dandiarchive.org, while the apex domain dandiarchive.org is controlled by the redirector. The redirector performs two main tasks
+
+1. Redirect requests from dandiarchive.org to gui.dandiarchive.org
+2. Serve JSON information about deployed services (e.g. urls for the API, web ui, and jupyterhub) at dandiarchive.org/server-info.
+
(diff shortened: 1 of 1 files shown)


satra (approved):
thanks @AlmightyYakob - this looks good to me.

doc/design/apex-domain-netlify.md
```diff
@@ -0,0 +1,38 @@
...
+
+[[redirects]]
+from = "/server-info"
+to = "/server-info.json"
+status = 200
+```
+
+## Obtaining the static JSON file
+
+Since Netlify is merely serving content for us, we need a way to include this JSON file as a local file, either as a part of, or a prerequisite to the build process. Rather than hardcoding this file in the source code, where it must be regularly updated, the API `/api/info` endpoint could be augmented to include this information as well. The Netlify build process would then simply be updated to download this file and serve it as shown above. The fetching and downloading of this file can be done through Nelify [Build Plugins](https://docs.netlify.com/configure-builds/build-plugins/).
```
waxlamp, line 31: If we're already serving this information at `/api/info`, is there a specific reason we need to also serve this file from the frontend? (attn: @satra, @yarikoptic)

AlmightyYakob: Obviously Satra and Yarik can answer, but I believe it's to have an index into the api (as well as other related services) attached to the main site itself. For other instances it might not be as simple as just `api.<instance>`.

satra: yup. we wanted a single point of entry into the services and that should be from the domain endpoint.

PR #1018, opened 2022-04-08
Add netlify plugin and redirection

AlmightyYakob:
Implementation of #1017

~~Blocked by https://github.com/dandi/dandi-infrastructure/pull/119~~

This essentially adds the info from the existing https://dandiarchive.org/server-info endpoint to our `api/info` endpoint, and setups up netlify to serve that information directly. 

After this is merged, we'll need to coordinate with @satra and @yarikoptic to transfer the GUI to be hosted directly at https://dandiarchive.org, replacing the redirector.


AlmightyYakob:
The netlify portion of this is working as intended, and visible at https://deploy-preview-1018--gui-staging-dandiarchive-org.netlify.app/server-info.

Now all that's left is to update the `api/info` endpoint to include the info that is currently being served at https://gui-staging.dandiarchive.org/server-info.

web/netlify.toml
```diff
@@ -10,7 +10,17 @@ command = "yarn run build"
   VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
   VUE_APP_SENTRY_ENVIRONMENT = "staging"
 
+
+[[redirects]]
+from = "/server-info"
+to = "/server-info.json"
```
dchiquito, line 16:
```suggestion
to = "https://api.dandiarchive.org/server-info.json"
```
Rather than bake the `server-info.json` file into netlify, why not just use the `status = 200` to proxy for the API server directly?

dchiquito (changes requested):
I see some linting errors that would probably be resolved with a `tox -e format`. Looks like the Github actions configuration also needs to be updated to include `DJANGO_DANDI_JUPYTERHUB_URL`.

dandiapi/api/views/info.py
```diff
@@ -25,12 +43,18 @@ class ApiInfoSerializer(serializers.Serializer):
 )
 @api_view()
 def info_view(self):
-
     serializer = ApiInfoSerializer(
         data={
             'schema_version': settings.DANDI_SCHEMA_VERSION,
             'schema_url': schema_url,
             'version': versioneer.get_version(),
+            "cli_minimal_version": "0.14.2",
```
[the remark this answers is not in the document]
AlmightyYakob: I agree. This PR is mainly to maintain parity with what exists already at https://dandiarchive.org/server-info

web/netlify.toml
AlmightyYakob: It was a request from @yarikoptic and @satra to not redirect, but to directly serve this information, to account for programmatic access and redirects not necessarily being followed.

AlmightyYakob: Oh, I think I see what you mean, e.g. https://docs.netlify.com/routing/redirects/rewrites-proxies/. This would be worth giving a shot.

AlmightyYakob:
It turns out our current deploy infrastructure actually won't support this, so we'll have to pause on this until it's addressed.

web/netlify.toml
AlmightyYakob: For posterity: This is possible but we can't change this value based on deploy context, so [fd4576d](https://github.com/dandi/dandi-archive/pull/1018/commits/fd4576d8b17e67dd9254f3ae77fb4d9b83c129dc) updates the plugin to proxy the request to the `api/info` endpoint. This should be more robust anyway.

AlmightyYakob:
@dchiquito I know I'm flip flopping on this but do you see any real reason to actually include the Juypterhub url as an env var? I don't see it ever changing in a way that warrants an env var. If you agree, I'll revert the changes that include that env var and just hardcode that url.

yarikoptic:
I think it should stay env bar since deployment specific. We might get staging hub, and anyone's deployment could use their hub

AlmightyYakob:
> I think it should stay env bar since deployment specific. We might get staging hub, and anyone's deployment could use their hub

Okay, this will require a CLI change to fix integration tests so I'll make a PR there.

AlmightyYakob:
Blocked on https://github.com/dandi/dandi-cli/pull/959 for passing cli integraiton tests.

dchiquito (approved):

PR #1022, opened 2022-04-12
Add support for glob/regex filtering on `NestedAsset` list endpoint

### LINKED ISSUE #1002: find files matching a regex
how do i best search for files matching a path regex? given the sizes of dandiset 26 and 108, i would prefer not having to get all the assets to filter it.  with the hackathon coming up, this would be a good thing to be able to support filters at some endpoint.

```
sub-I46/*/*.ome.tiff
*.ngff
```

related #230

mvandenburgh:
Closes #1002.

Note that Postgres doesn't support querying with globs, but it does support querying with a regex. So, the approach I've taken to support globbing here is to convert the glob pattern to a regex pattern and then use that in the query. Python has a built-in function (`fnmatch.translate`) that converts a glob expression to a regex, but the regex syntax it uses is incompatible with Postgres's regex syntax. I found [this 81-line library](https://github.com/Salamek/glob2regex) that does work with Postgres, so I've used that here.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -37,9 +37,7 @@ from dandiapi.api.models import Asset, AssetBlob, Dandiset, Version, ZarrArchive
 from dandiapi.api.models.asset import BaseAssetBlob, EmbargoedAssetBlob
 from dandiapi.api.tasks import validate_asset_metadata
 from dandiapi.api.views.common import (
-    ASSET_GLOB_PARAM,
     ASSET_ID_PARAM,
-    ASSET_REGEX_PARAM,
     PATH_PREFIX_PARAM,
     VERSIONS_DANDISET_PK_PARAM,
     VERSIONS_VERSION_PARAM,
@@ -47,6 +45,7 @@ from dandiapi.api.views.common import (
 )
 from dandiapi.api.views.serializers import (
     AssetDetailSerializer,
+    AssetListSerializer,
     AssetPathsQueryParameterSerializer,
     AssetPathsResponseSerializer,
     AssetSerializer,
@@ -511,12 +510,14 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
 
-    @swagger_auto_schema(manual_parameters=[ASSET_GLOB_PARAM, ASSET_REGEX_PARAM])
+    @swagger_auto_schema(query_serializer=AssetListSerializer)
     def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
+        serializer = AssetListSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
 
-        glob_pattern: str | None = self.request.query_params.get('glob')
-        regex_pattern: str | None = self.request.query_params.get('regex')
+        queryset = self.filter_queryset(self.get_queryset())
+        glob_pattern: str | None = serializer.validated_data.get('glob')
+        regex_pattern: str | None = serializer.validated_data.get('regex')
 
         if regex_pattern is not None:
             try:
@@ -531,7 +532,7 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
 
         if glob_pattern is not None:
             queryset = queryset.filter(
-                path__iregex=glob_pattern.replace("*", ".*").replace(".", "\\.")
+                path__iregex=glob_pattern.replace('*', '.*').replace('.', '\\.')
             )
 
         page = self.paginate_queryset(queryset)
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 try:
     from storages.backends.s3boto3 import S3Boto3Storage
 except ImportError:
@@ -33,6 +35,7 @@ from dandiapi.api.models import Asset, AssetBlob, Dandiset, Version, ZarrArchive
 from dandiapi.api.models.asset import BaseAssetBlob, EmbargoedAssetBlob
 from dandiapi.api.tasks import validate_asset_metadata
 from dandiapi.api.views.common import (
+    ASSET_GLOB_PARAM,
     ASSET_ID_PARAM,
     PATH_PREFIX_PARAM,
     VERSIONS_DANDISET_PK_PARAM,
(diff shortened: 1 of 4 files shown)


AlmightyYakob:
> Python has a built-in function (`fnmatch.translate`) that converts a glob expression to a regex, but the regex syntax it uses is incompatible with Postgres's regex syntax.

What exactly from the translation is incompatible with Postgres? I'd hope that any differences are minor and could be addressed directly. I'd rather not pull in an obscure 3rd party library if it's not necessary.

dandiapi/api/views/asset.py
```diff
@@ -505,6 +512,35 @@ def destroy(self, request, versions__dandiset__pk, versions__version, **kwargs):
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
 
+    @swagger_auto_schema(manual_parameters=[ASSET_GLOB_PARAM, ASSET_REGEX_PARAM])
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+
+        glob_pattern: str | None = self.request.query_params.get('glob')
+        regex_pattern: str | None = self.request.query_params.get('regex')
```
AlmightyYakob, line 520: Unless there's a good reason against it, I'd suggest defining a serializer for these params

```python
class AssetListSerializer(serializers.Serializer):
    glob = serialzers.CharField(required=False)
    regex = serialzers.CharField(required=False)
```

and then just using like so

```python
    @swagger_auto_schema(query_serializer=AssetListSerializer)
    def list(self, request, *args, **kwargs):
        serializer = AssetListSerializer(data=request.query_params)

        queryset = self.filter_queryset(self.get_queryset())
        glob_pattern: str | None = serializer.validated_data['glob']
        regex_pattern: str | None = serializer.validated_data['regex']
```

IMO `manual_parameters` should be used as a last resort, for when the normal mode of data serialization can't be used.

mvandenburgh:
> > Python has a built-in function (`fnmatch.translate`) that converts a glob expression to a regex, but the regex syntax it uses is incompatible with Postgres's regex syntax.
> 
> What exactly from the translation is incompatible with Postgres? I'd hope that any differences are minor and could be addressed directly. I'd rather not pull in an obscure 3rd party library if it's not necessary.

`fnmatch.translate('*.json')` returns `(?s:.*\.json)\Z` while Postgres expects something like `^.*\.json$`. I'm unsure of the differences between the two, although I've never seen the `?s` `\Z` syntax from the first one before. I do know that support for regex varies across different languages. For example, apparently Python and JavaScript have slight differences in their regex syntaxes that make them incompatible. There's a surprisingly lack of information online about how to translate between the `fnmatch` format and postgres/Django format :(. I'll take another look

AlmightyYakob:
> `fnmatch.translate('*.json')` returns `(?s:.*\.json)\Z` while Postgres expects something like `^.*\.json$`. I'm unsure of the differences between the two, although I've never seen the `?s` `\Z` syntax from the first one before. I do know that support for regex varies across different languages. For example, apparently Python and JavaScript have slight differences in their regex syntaxes that make them incompatible. There's a surprisingly lack of information online about how to translate between the `fnmatch` format and postgres/Django format :(. I'll take another look

It seems to me a lot of this has to do with regex modes and differences with chars used. For example, the `?:s` is the python regex mode for `include newlines in .`, which doesn't really apply to us. Toying around with it for a bit, I got this regex replacement which should solve most of our problems

```python
re.sub(r'\(\?s?m?:([^\)]+)\)\\Z', r'^\1$', glob)
```

This essentially removes the `(?s:)` and `\\Z` part from the string. It also catches the `m` (multiline) modifier).

dandiapi/api/views/common.py
```diff
@@ -33,6 +33,20 @@ class DandiPagination(PageNumberPagination):
...
+    description='Glob pattern to filter asset paths with.',
+)
+
+ASSET_REGEX_PARAM = openapi.Parameter(
+    'regex',
+    openapi.IN_QUERY,
+    type=openapi.TYPE_STRING,
+    description='Regex pattern to filter asset paths with.',
+)
+
```
AlmightyYakob, line 49: Related to my other comment

waxlamp:
> > `fnmatch.translate('*.json')` returns `(?s:.*\.json)\Z` while Postgres expects something like `^.*\.json$`. I'm unsure of the differences between the two, although I've never seen the `?s` `\Z` syntax from the first one before. I do know that support for regex varies across different languages. For example, apparently Python and JavaScript have slight differences in their regex syntaxes that make them incompatible. There's a surprisingly lack of information online about how to translate between the `fnmatch` format and postgres/Django format :(. I'll take another look
> 
> It seems to me a lot of this has to do with regex modes and differences with chars used. For example, the `?:s` is the python regex mode for `include newlines in .`, which doesn't really apply to us. Toying around with it for a bit, I got this regex replacement which should solve most of our problems
> 
> ```python
> re.sub(r'\(\?s?m?:([^\)]+)\)\\Z', r'^\1$', glob)
> ```
> 
> This essentially removes the `(?s:)` and `\\Z` part from the string. It also catches the `m` (multiline) modifier).

I don't love either including that library (even if it seems to be harmless) or this homebrew regex (even if we could write tests for it). The docs for fnmatch.translate imply that the globbing star is just converted to `.*`, while regular dots are converted to `\.`. (Not sure what happens to `**`, but we can defer that functionality for now.) Could we just do that?

AlmightyYakob:
> I don't love either including that library (even if it seems to be harmless) or this homebrew regex (even if we could write tests for it). The docs for fnmatch.translate imply that the globbing star is just converted to `.*`, while regular dots are converted to `\.`. (Not sure what happens to `**`, but we can defer that functionality for now.) Could we just do that?

I think it's worth testing out at least. For the record, `**` isn't handled by `fnmatch`, that's only handled by the [`glob`](https://docs.python.org/3/library/glob.html#module-glob) module.

dandiapi/api/views/asset.py
mvandenburgh: Oh nice, I didn't realize this lets you replace `manual_parameters` - there's other places in the code where `manual_parameters` is used, I wonder if similar refactors could be done there

mvandenburgh: Filed #1026 so we don't forget about this

mvandenburgh:
Agreed that @waxlamp's solution is cleaner. I tested it and it seems to work perfectly, I'll update this to do it that way :+1:

dandiapi/api/views/serializers.py
```diff
@@ -197,3 +197,8 @@ class AssetPathsQueryParameterSerializer(serializers.Serializer):
     path_prefix = serializers.CharField(default='')
     page = serializers.IntegerField(default=1)
     page_size = serializers.IntegerField(default=DandiPagination.page_size)
+
+
+class AssetListSerializer(serializers.Serializer):
+    glob = serializers.CharField(required=False)
+    regex = serializers.CharField(required=False)
```
AlmightyYakob, line 204: As is, both glob and regex can be specified. We should probably catch this and error in the serializer.

```suggestion
class AssetListSerializer(serializers.Serializer):
    glob = serializers.CharField(required=False)
    regex = serializers.CharField(required=False)
    
    def validate(self, data):
        if 'glob' in data and 'regex' in data:
            raise serializers.ValidationError(
                {'glob': 'Cannot specify both glob and regex'},
            )
        return data
```

AlmightyYakob (changes requested):
Just one request otherwise looks good

dandiapi/api/views/serializers.py
mvandenburgh: Why not allow both to be specified?

AlmightyYakob: They could conflict and so leaves it ambiguous what should occur. If both are applied, order would matter, and that's fairly opaque to the user. Unless there's a convincing reason to require both be used at the same time I'd suggest we only allow one at a time.

mvandenburgh: Good point, I didn't consider that order would matter - I'll add that validation and another test for this scenario

AlmightyYakob (approved):

AlmightyYakob:
@mvandenburgh it seems the `test_zarr_rest_list_filter` test failed? Have you seen that for yourself or is that just a fluke?

mvandenburgh:
> @mvandenburgh it seems the `test_zarr_rest_list_filter` test failed? Have you seen that for yourself or is that just a fluke?

I assumed it was a fluke because I'm not seeing it when I run tests locally, but it happened again when I reran the job :confused:

PR #1036, opened 2022-04-18
Fix disabled Heroku Github integration for staging and production deployment

AlmightyYakob:
This is a temporary fix to address a recent [Heroku Incident](https://status.heroku.com/incidents/2413), which resulted in Heroku disabled Heroku Github integration for all Heroku applications. This requires us to temporarily add back a Heroku CLI based deployment mechanism, until Github integration is restored.


mvandenburgh (approved):
lgtm

PR #1046, opened 2022-04-20
Fix incorrect fields in info endpoint

### LINKED ISSUE #1044: Key names in `/server-info` response changed
`https://dandiarchive.org/server-info` currently returns a JSON object with the keys `"cli_minimal_version"` and `"cli_bad_versions"`.  However, previously, these keys were named with hyphens instead of underscores, and the client depends on the old names, and so the client is now broken.

AlmightyYakob:
Closes #1044

The test is pretty barebones but will ensure tests fail if the endpoint structure is modified.

This also fixes a typo in the `DJANGO_DANDI_JUPYTERHUB_URL` env var in `dev/.env.docker-compose` and `dev/.env.docker-compose-native`.

### DIFF
--- a/dandiapi/api/tests/test_info.py
+++ b/dandiapi/api/tests/test_info.py
@@ -1,0 +1,21 @@
+from django.conf import settings
+import versioneer
+
+from dandiapi.api.views.info import schema_url
+
+
+def test_rest_info(api_client):
+    resp = api_client.get('/api/info/')
+    assert resp.status_code == 200
+    assert resp.json() == {
+        'schema_version': settings.DANDI_SCHEMA_VERSION,
+        'schema_url': schema_url,
+        'version': versioneer.get_version(),
+        'cli-minimal-version': '0.14.2',
+        'cli-bad-versions': [],
+        'services': {
+            'api': {'url': settings.DANDI_API_URL},
+            'webui': {'url': settings.DANDI_WEB_APP_URL},
+            'jupyterhub': {'url': settings.DANDI_JUPYTERHUB_URL},
+        },
+    }
--- a/dandiapi/api/tests/test_info.py
+++ b/dandiapi/api/tests/test_info.py
@@ -1,7 +1,7 @@
 from django.conf import settings
-import versioneer
 
 from dandiapi.api.views.info import schema_url
+import versioneer
 
 
 def test_rest_info(api_client):
--- a/dandiapi/api/views/info.py
+++ b/dandiapi/api/views/info.py
@@ -23,14 +23,23 @@ class ApiServicesSerializer(serializers.Serializer):
 
 
 class ApiInfoSerializer(serializers.Serializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Need to add fields here to allow for hyphens in name
+        self.fields.update(
+            {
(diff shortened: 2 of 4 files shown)


dandiapi/api/views/info.py
```diff
@@ -23,14 +23,23 @@ class ApiServicesSerializer(serializers.Serializer):
...
 
 class ApiInfoSerializer(serializers.Serializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Need to add fields here to allow for hyphens in name
+        self.fields.update(
+            {
+                'cli-minimal-version': serializers.CharField(),
+                'cli-bad-versions': serializers.ListField(child=serializers.CharField()),
```
mvandenburgh, line 33: In addition to the test, I think we should have a comment somewhere here saying that the CLI expects these exact keywords. Someone who doesn't know that could come along and change it + update the test without being aware of it.

dandiapi/api/tests/test_info.py
```diff
@@ -0,0 +1,21 @@
...
+        'schema_url': schema_url,
+        'version': versioneer.get_version(),
+        'cli-minimal-version': '0.14.2',
+        'cli-bad-versions': [],
+        'services': {
+            'api': {'url': settings.DANDI_API_URL},
+            'webui': {'url': settings.DANDI_WEB_APP_URL},
+            'jupyterhub': {'url': settings.DANDI_JUPYTERHUB_URL},
+        },
+    }
```
mvandenburgh, line 21: Since this endpoint is something that the CLI depends on, I think having a similar test as part of the CLI integration tests would be valuable (thoughts @jwodder @yarikoptic?)

mvandenburgh (approved):
LGTM - just had one small suggestion

dandiapi/api/tests/test_info.py
jwodder: There effectively is such a test already: https://github.com/dandi/dandi-cli/blob/00ee4ddf8785a91ff5bd893497b018ef5fce6437/dandi/tests/test_utils.py#L303

AlmightyYakob: That's a really good point. I think adding a CLI Integration test is important, since that's what would break if we were to change this. We probably want to merge this as soon as possible, but that test could be added after the fact.

AlmightyYakob: @jwodder I see a couple of problems with that test

1. We're not specifying the `DANDI_REDIRECTOR_BASE` environment variable currently, so this test is either skipped or tested against the currently deployed prod (not totally sure of the behavior from looking at that file). In either case, the current change isn't tested.
2. Even if we did specify that env variable, this redirection doesn't happen locally, it only happens in Netlify, so the test would fail in that way.

We should probably at least add an almost identical test aimed at the locally spun up API instance, instead of prod.

AlmightyYakob:
CLI Integration tests are failing, but they seem to be due to a prior change, not this one (evidenced by this same failure occurring prior to this PR [here](https://github.com/dandi/dandi-archive/runs/6098618164?check_suite_focus=true)). It seems the integration tests need some additional work on top of this, per our conversation on this PR.

I'm going to merge this, since this is currently causing issue in prod, and these changes can be applied after the fact.

PR #1050, opened 2022-04-20
Misc. Django/Python improvements

mvandenburgh:
I noticed a bunch of code that did things in a slightly incorrect and/or outdated way, this PR fixes them -

1) Updates any use of `.filter(~Q(...))` to `.exclude(...)`. `exclude` behaves the same and is the "correct" way to do this sort of query in Django
2) Update DRF viewsets to use `rest_framework.generics.get_object_or_404` instead of `django.shortcuts.get_object_or_404`.
	- DRF's version of the function wraps the vanilla Django one and adds some DRF specific validation/error handling https://stackoverflow.com/questions/68001954/difference-between-django-shortcuts-and-rest-framework-generics
3) Replace instances of `typing.Union` and `typing.Optional` with new union type syntax

Technically, 3) isn't necessarily incorrect (from what I understand, `typing.Optional` and `typing.Union` are not deprecated, at least not yet), but I find the `x | y` union syntax to be much more readable. I don't feel strongly on this though :man_shrugging:

### DIFF
--- a/dandiapi/api/dashboard.py
+++ b/dandiapi/api/dashboard.py
@@ -2,7 +2,7 @@ from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied
-from django.db.models import Exists, OuterRef, Q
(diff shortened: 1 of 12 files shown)


AlmightyYakob (approved):
LGTM

PR #1054, opened 2022-04-20
Fix version in `/api/info/`

### LINKED ISSUE #222: /info: automated `version`ing  by versioneer needs fixing
#221 added versioning by versioneer but seems deployment is not entirely kosher with it since getting
```
{
  "schema_version": "0.2.0",
  "schema_url": "https://raw.githubusercontent.com/dandi/schema/master/releases/0.2.0/dandiset.json",
  "version": "0+unknown"
}
```
I guess it depends on how it gets deployed (may be fetch is happening without fetching the git tags or smth like that)

### LINKED ISSUE #1045: `version` in `/server-info` is "0+unknown"
`https://dandiarchive.org/server-info` currently returns a JSON object in which the `"version"` field has a value of `"0+unknown"`.  This causes a test failure in the dandi-cli code, as it has a test that depends on that field being greater than "1.2.0".

AlmightyYakob:
Closes #1045
Closes #222

I don't believe there's anything to do with the infrastructure, since this would change on each deploy, but I could be wrong.

Requesting more reviews than usual since this modifies our staging/prod deploy a bit.


dev/.env.docker-compose
```diff
@@ -11,3 +11,4 @@ DJANGO_DANDI_DANDISETS_EMBARGO_BUCKET_NAME=dandi-embargo-dandisets
 DJANGO_DANDI_WEB_APP_URL=http://localhost:8085
 DJANGO_DANDI_API_URL=http://localhost:8000
 DJANGO_DANDI_JUPYTERHUB_URL=https://hub.dandiarchive.org/
+DJANGO_DANDI_API_VERSION=v0.0.0
```
yarikoptic, line 14: so local containerized instances used by dandi-cli would not have it set, or these compose recipes aren't used for dandi-cli? (attn @jwodder)

jwodder: `.env.docker-compose` is only used when running Docker Compose from within the dandi-archive repository.  The environment variables used by dandi-cli's Docker Compose setup are set in its `docker-compose.yml` file.

AlmightyYakob: This variable also defaults to `v0.0.0` if not supplied.

.github/workflows/update-release-branch.yml
```diff
@@ -14,14 +14,30 @@ jobs:
...
           git reset --hard ${{ github.ref }}
           git push --force origin release
 
+  production-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # fetch history for all branches and tags
+          token: ${{ secrets.GH_TOKEN }} # use PAT with permissions to push to master
```
mvandenburgh, line 28: Why is this token needed here? I think we should stick to the default `GITHUB_TOKEN` with limited permissions when possible

mvandenburgh (changes requested):
I think the new `checkout` action can be changed to just use the default token provided by github actions - see my comment below.

.github/workflows/update-release-branch.yml
AlmightyYakob: Good catch, I agree. If using the default do I need to specify a token here? Or can the entire `token` field be left out, and it will default to `GITHUB_TOKEN`?

mvandenburgh: Yeah, `actions/checkout` defaults to `GITHUB_TOKEN` if `token` isn't provided.

.github/workflows/backend-staging-deploy.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Get latest git tag
         run: |
           git fetch --prune --unshallow
-          echo "API_VERSION=$(git describe --abbrev=0)" >> $GITHUB_ENV
+          echo "API_VERSION=$(git describe)" >> $GITHUB_ENV
```
mvandenburgh, line 18: We use `git describe --tags` to get the version string for the frontend https://github.com/dandi/dandi-archive/blob/master/web/vue.config.js#L10, would it make sense to do that here too?

dandiapi/api/views/info.py
```diff
@@ -6,8 +6,6 @@
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
 
-import versioneer
```
dchiquito, line 9: Can you also remove versioneer?

dandiapi/settings.py
```diff
@@ -77,6 +77,9 @@ def mutate_configuration(configuration: Type[ComposedConfiguration]):
     # considered a temporary fix.
     DANDI_SCHEMA_VERSION = values.Value(default='0.6.2', environ=True)
 
+    # The version of the API
+    DANDI_API_VERSION = values.Value(default='v0.0.0', environ=True)
```
dchiquito, line 81: What's the purpose of having a nonsensical API version?

.github/workflows/backend-staging-deploy.yml
AlmightyYakob: What's the benefit of adding `--tags`? As far as I can tell, it just allows the inclusion of "lightweight" tags. From the `git-tag` man docs,

```
Annotated tags are meant for release while lightweight tags are meant for private or temporary
object labels. For this reason, some git commands for naming objects (like git describe) will
ignore lightweight tags by default.
```

It seems that we always use annotated tags, so I'm inclined to just leave it as `git describe`, but I don't know if it'll make much of a difference.

mvandenburgh: Ah, I didn't know it was for lightweight tags - I think we're fine with just `git describe` then. t looks like `auto` always uses annotated tags https://github.com/intuit/auto/blob/main/plugins/git-tag/src/index.ts#L65-L70

dandiapi/settings.py
AlmightyYakob: Aside from production, this value really isn't important, so I set a default just to reduce any unnecessary friction. If you think it should be required I can remove the default too, I don't feel too strongly.

dandiapi/api/views/info.py
AlmightyYakob: If we do that, what should we use for the version in `setup.py`? Should we just pull it out of the `DJANGO_DANDI_API_VERSION` env var?

AlmightyYakob:
Based on some out of band conversations, it looks like we're going to take a different approach to how we handle versioning. For the time being I'm going to mark this PR as draft until that's fleshed out.

AlmightyYakob:
This PR has been updated to use `setuptools_scm`. This allows for the version to be computed determined by git as we intended. This still works in heroku as git info is still present when heroku pip installs our application.

dchiquito (approved):

Issue #1065, opened 2022-04-26
Fix failing tests


AlmightyYakob:
In master (and all branches off of the current master), `check-migrations` is failing, as well as the `test_create_dev_dandiset` test (evidenced by [this](https://github.com/dandi/dandi-archive/runs/6180990890?check_suite_focus=true) CI run). It appears dandischema validation is now failing.

dandibot:
:rocket: Issue was released in [`v0.2.18`](https://github.com/dandi/dandi-archive/releases/tag/v0.2.18) :rocket:

PR #1066, opened 2022-04-26
Remove erroneous select_for_update and update tests to use transactions

### LINKED ISSUE #1064: TransactionManagementError: select_for_update cannot be used outside of a transaction.
This is being caused by an erroneous `select_for_update` call outside of a transaction. We should fix this and update our tests to ensure that this kind of problem is caught in the future.

Sentry Issue: [DANDI-API-7Y](https://sentry.io/organizations/dandiarchive/issues/3224809704/?referrer=github_integration)

https://github.com/dandi/dandi-archive/blob/fe7b79cbe4ff7a8af75dbb1f7edb0a724877cf1e/dandiapi/api/views/zarr.py#L217

AlmightyYakob:
Closes #1064

To verify that the updated tests do catch this kind of error, you can roll back 98c1158b56743400905637afd362a28311eec64c and see that the bug is now caught by tests.


AlmightyYakob:
Tests will fail due to https://github.com/dandi/dandi-archive/issues/1065

AlmightyYakob:
> LGTM It's unfortunate that you have to manually specify `transaction=True` if a transaction is used in the code being tested.... I could see this being a problem again in the future.

I agree, but I also don't think it's a good idea to try to default to using transactions, since they'd slow down our already not that fast tests. We'll probably have to keep our eye on this.

Issue #1063, opened 2022-04-26
zarr ingest


satra:
i triggered zarr ingest on a small set of 9 files. it ingested 4 and stopped (i think). didn't see any other files ingested.

https://dandiarchive.org/dandiset/000108/draft/files?location=sub-MITU01%2Fses-20210521h17m17s06%2Fmicr%2F

it would be good to know how to determine what succeeded and failed.

on a related note, the zarr ingest end point requires a zarr id in both the api url and as a payload for the post. shouldn't one be sufficient?

satra:
perhaps another related issue. this is on a previously ingested file.

```
$ dandi upload sub-MITU01_ses-20210521h17m17s06_sample-178_stain-LEC_run-1_chunk-2_SPIM.ngff
2022-04-26 14:27:45,953 [    INFO] Found 1 files to consider
PATH                                                                                      SIZE    ERRORS UPLOAD STATUS                          MESSAGE                                                                
...s06/micr/sub-MITU01_ses-20210521h17m17s06_sample-178_stain-LEC_run-1_chunk-2_SPIM.ngff 52.5 GB   0           ERROR                           Error 400 while sending POST request to https://api.dandiarchive.org...Summary:                                                                                  52.5 GB               1 ERROR                         1 Error 400 while sending POST request to https://api.dandiarchive.o...
2022-04-26 14:50:35,571 [    INFO] Logs saved in /home/satra/.cache/dandi-cli/log/20220426182704Z-662907.log
```

and 

```
2022-04-26T14:50:35-0400 [DEBUG   ] urllib3.connectionpool 662907:140131214083840 https://api.dandiarchive.org:443 "POST /api/zarr/2fe1afcb-bc07-44e7-8a9e-cd2337320163/ingest/ HTTP/1.1" 400 31
2022-04-26T14:50:35-0400 [DEBUG   ] dandi 662907:140131214083840 Response: 400
2022-04-26T14:50:35-0400 [ERROR   ] dandi 662907:140131214083840 Error 400 while sending POST request to https://api.dandiarchive.org/api/zarr/2fe1afcb-bc07-44e7-8a9e-cd2337320163/ingest/: "Zarr archive already ingested"
[...]

PR #1067, opened 2022-04-26
Add external viewer for NGFF files

### LINKED ISSUE #906: add context viewer for ngff files
for the api version this is the url

> 'https://neuroglancer-demo.appspot.com/#!{"dimensions":{"d0":[1,""],"d1":[1,""],"d2":[1,""],"d3":[1,""],"d4":[1,""]},"displayDimensions":["d2","d3","d4"],"position":[0.5,0.5,1060.5,1299.5,2234.5],"crossSectionScale":1.4804251528842283,"projectionOrientation":[-0.1748984158039093,-0.3958210051059723,-0.21921014785766602,0.8744616508483887],"projectionScale":128,"layers":[{"type":"image","source":"zarr://https://api-staging.dandiarchive.org/api/zarr/b35681cd-936a-477c-ac51-06661612d4df.zarr/0?dimension_separator=/","tab":"source","shaderControls":{"normalized":{"range":[0,5279]}},"crossSectionRenderScale":25.62082955327362,"name":"4?dimension_separator="}],"selectedLayer":{"visible":true,"layer":"4?dimension_separator="},"layout":"4panel"}'

for the direct s3 version here is the url

> 'https://neuroglancer-demo.appspot.com/#!{"dimensions":{"d0":[1,""],"d1":[1,""],"d2":[1,""],"d3":[1,""],"d4":[1,""]},"displayDimensions":["d2","d3","d4"],"position":[0.5,0.5,1060.5,1299.5,2234.5],"crossSectionScale":1.4804251528842283,"projectionOrientation":[-0.1748984158039093,-0.3958210051059723,-0.21921014785766602,0.8744616508483887],"projectionScale":128,"layers":[{"type":"image","source":"zarr://https://dandi-api-staging-dandisets.s3.amazonaws.com/zarr/b35681cd-936a-477c-ac51-06661612d4df/0?dimension_separator=/","tab":"source","shaderControls":{"normalized":{"range":[0,5279]}},"crossSectionRenderScale":25.62082955327362,"name":"4?dimension_separator="}],"selectedLayer":{"visible":true,"layer":"4?dimension_separator="},"layout":"4panel"}'

mvandenburgh:
Adds https://kitware.github.io/itk-vtk-viewer/app/ as a "External service" for .ngff files in the file browser gui.

closes #906

### DIFF
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -254,9 +254,24 @@ class AssetFolderSerializer(serializers.Serializer):
     modified = serializers.DateTimeField()
 
 
+class AssetFileSerializer(AssetSerializer):
+    class Meta(AssetSerializer.Meta):
+        fields = AssetSerializer.Meta.fields + ['url']
+
+    url = serializers.SerializerMethodField(method_name='get_url')
+
+    def get_url(self, asset: Asset):
+        if asset.is_blob:
+            return asset.blob.s3_url
+        elif asset.is_embargoed_blob:
+            return asset.embargoed_blob.s3_url
+        else:
+            return asset.zarr.s3_url
+
+
 class AssetPathsResponseSerializer(serializers.Serializer):
     folders = serializers.DictField(child=AssetFolderSerializer())
-    files = serializers.DictField(child=AssetSerializer())
+    files = serializers.DictField(child=AssetFileSerializer())
 
 
 class AssetPathsQueryParameterSerializer(serializers.Serializer):
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -258,15 +258,7 @@ class AssetFileSerializer(AssetSerializer):
     class Meta(AssetSerializer.Meta):
         fields = AssetSerializer.Meta.fields + ['url']
 
-    url = serializers.SerializerMethodField(method_name='get_url')
-
-    def get_url(self, asset: Asset):
-        if asset.is_blob:
-            return asset.blob.s3_url
-        elif asset.is_embargoed_blob:
-            return asset.embargoed_blob.s3_url
-        else:
-            return asset.zarr.s3_url
+    url = serializers.URLField(source='s3_url')
 
 
 class AssetPathsResponseSerializer(serializers.Serializer):
--- a/web/src/types/index.ts
+++ b/web/src/types/index.ts
@@ -94,4 +94,5 @@ export interface AssetFile extends AssetStats {
   asset_id: string,
   path: string,
   services: AssetServices[],
+  url: URL;
 }
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -186,6 +186,8 @@
                         v-for="el in item.services"
                         :key="el.name"
                         :href="el.url"
+                        target="_blank"
+                        rel="noopener"
                       >
                         <v-list-item-title class="font-weight-light">
                           {{ el.name }}
@@ -255,6 +257,13 @@ const EXTERNAL_SERVICES = [
     maxsize: 1e9,
     endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
   },
+
+  {
+    name: 'VTK/ITK Viewer',
+    regex: /\.ome\.zarr$/,
+    maxsize: Infinity,
+    endpoint: 'https://kitware.github.io/itk-vtk-viewer/app/?gradientOpacity=0.3&image=',
+  },
 ];
 
 export default defineComponent({
@@ -286,14 +295,13 @@ export default defineComponent({
       dandiRest.user
       && owners.value?.includes(dandiRest.user?.username)
     ));
-
-    function getExternalServices(asset_id: string, name: string, size: number) {
-      const { identifier, version } = props;
+    function getExternalServices(name: string, asset: AssetFile) {
       return EXTERNAL_SERVICES
-        .filter((service) => new RegExp(service.regex).test(name) && size <= service.maxsize)
+        .filter((service) => new RegExp(service.regex)
+          .test(name) && (asset.size || 0) <= service.maxsize)
         .map((service) => ({
           name: service.name,
-          url: `${service.endpoint}${dandiRest.assetDownloadURI(identifier, version, asset_id)}`,
+          url: `${service.endpoint}${asset.url}`,
         }));
     }
 
@@ -352,8 +360,7 @@ export default defineComponent({
       // Add items
       newItems.push(...Object.keys(files).map(
         (key) => {
-          const { asset_id, size } = files[key];
-          const services = getExternalServices(asset_id, key, size || 0);
+          const services = getExternalServices(key, files[key]);
           return {
             ...files[key],
             name: key,
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -180,6 +180,15 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
         else:
             return self.zarr.digest
 
+    @property
+    def s3_url(self) -> str:
+        if self.is_blob:
+            return self.blob.s3_url
+        elif self.is_embargoed_blob:
+            return self.embargoed_blob.s3_url
+        else:
+            return self.zarr.s3_url
+
     def _populate_metadata(self):
         download_url = settings.DANDI_API_URL + reverse(
             'asset-download',
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -12,7 +12,7 @@ from rest_framework.test import APIClient
 
 from dandiapi.api.models import Asset, AssetBlob, EmbargoedAssetBlob, Version
 from dandiapi.api.models.dandiset import Dandiset
-from dandiapi.api.views.serializers import AssetFolderSerializer, AssetSerializer
+from dandiapi.api.views.serializers import AssetFileSerializer, AssetFolderSerializer
 
 from .fuzzy import HTTP_URL_RE, TIMESTAMP_RE, URN_RE, UTC_ISO_TIMESTAMP_RE, UUID_RE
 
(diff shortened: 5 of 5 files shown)


web/src/rest.ts
```diff
@@ -233,8 +233,13 @@ const dandiRest = new Vue({
     assetManifestURI(identifier: string, version: string) {
       return `${dandiApiRoot}dandisets/${identifier}/versions/${version}/assets/`;
     },
-    assetDownloadURI(identifier: string, version: string, uuid: string) {
-      return `${dandiApiRoot}assets/${uuid}/download/`;
+    assetDownloadURI(asset: Asset|AssetFile) {
+      // @ts-ignore next-line
+      const { asset_id, zarr } = asset;
+      if (zarr) {
+        return `https://dandiarchive.s3.amazonaws.com/zarr/${zarr}`;
```
mvandenburgh, line 240: @dchiquito is there a way to retrieve this s3 url from the API instead? I *think* the `/zarr/{zarr_id}.zarr/{path}` endpoint will do it, but I get a 404 when I try to call it like this -
https://api.dandiarchive.org/api/zarr/21e4e2e9-9474-4568-8dee-ae9154316ddd.zarr/

satra: >  but I get a 404 when I try to call it like this -
https://api.dandiarchive.org/api/zarr/21e4e2e9-9474-4568-8dee-ae9154316ddd.zarr/

that's because ingest hasn't been completed for all of them. (see #1063 issue)

mvandenburgh: Update: I was able to get this endpoint working in staging, it seems it doesn't actually provide what we need here https://api-staging.dandiarchive.org/api/zarr/94fdcde1-afa8-437f-8d7b-c75967295843.zarr/

mvandenburgh: > i would suggest using the s3 http endpoint if possible, since it gets the api requests to a much more horizontally scalable infrastructure.

We actually already have an `fsspec`-compliant zarr API that makes this very easy (thanks @dchiquito for pointing me to that), so unless there are scalability concerns that I'm not aware of, I think it'd be better to use that. Otherwise, we would need to augment the asset path endpoint to also return s3 urls.

satra: the asset metadata has s3 urls in it. and perhaps that could be retrieved on click rather than when listing the asset. since a click will be a much more infrequent event, there is no need to add the return to the asset path endpoint.

i think we are already seeing ngff api limits from the data backup script.

web/src/rest.ts
```diff
@@ -233,8 +233,13 @@ const dandiRest = new Vue({
...
     },
-    assetDownloadURI(identifier: string, version: string, uuid: string) {
-      return `${dandiApiRoot}assets/${uuid}/download/`;
+    assetDownloadURI(asset: Asset | AssetStats) {
+      // @ts-ignore next-line
+      const { asset_id, zarr } = asset;
+      if (zarr) {
+        return `${dandiApiRoot}zarr/${zarr}.zarr`;
+      }
+      return `${dandiApiRoot}assets/${asset_id}/download/`;
```
satra, line 242: unless dandiset is embargoed this should use the s3 locations.

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -255,6 +259,13 @@ const EXTERNAL_SERVICES = [
     maxsize: 1e9,
     endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
   },
+
+  {
+    name: 'VTK/ITK Viewer',
+    regex: '.ngff$',
```
satra, line 265:
```suggestion
    regex: '.ome.zarr$',
```

yarikoptic: Let's allow for both?
```suggestion
    regex: '.(ngff|ome.zarr)$',
```

satra: didn't we decide that ngff is not a valid extension?

yarikoptic: according to BIDS -- yeah.  But what if we do have such a folder/file which might have... anyways -- good either way to me

AlmightyYakob (dismissed):
I thought I had already given this

mvandenburgh:
@satra re: your concerns about performance using the zarr API - after testing a few zarr files from dandiset 108 (for example, https://kitware.github.io/itk-vtk-viewer/app/?gradientOpacity=0.3&image=https://api.dandiarchive.org/api/zarr/2fe1afcb-bc07-44e7-8a9e-cd2337320163.zarr), I'm not seeing any abnormally high response times. Now that we have Sentry Performance tier I think we should merge this as-is and try the simpler solution first. And then if sentry reports any performance problems, we can re-evaluate the approach and move to s3 urls.

web/src/views/FileBrowserView/FileBrowser.vue
mvandenburgh: @satra @yarikoptic I went with `.ome.zarr` only here - if you agree, please resolve this conversation

satra:
@mvandenburgh - it's not about performance as an individual but what if 500 people in a class access it. or 1000 people at a conference. going directly to s3 would be horizontally scalable, as the dandi api only needs to serve 500 or 1000 requests for metadata, but not the real time zarr browsing. so for public dandisets, i would still suggest directly using the s3 location. since this is visualization, it's more about such use cases.

waxlamp:
> @mvandenburgh - it's not about performance as an individual but what if 500 people in a class access it. or 1000 people at a conference. going directly to s3 would be horizontally scalable, as the dandi api only needs to serve 500 or 1000 requests for metadata, but not the real time zarr browsing. so for public dandisets, i would still suggest directly using the s3 location. since this is visualization, it's more about such use cases.

I'm actually quite curious what the performance would be in such a situation. I'd suggest merging this to staging, then setting up a load test under these conditions to see what happens.

satra:
you would need some real scale data on there to test. i think that using the api endpoint for zarr should be a separate PR, and using direct s3 download should be what is included and merged at present.

satra:
i should also say, for this dataset, this viewer is the most trivial form of an external viewer, the ideal form would be to view all the stains and chunks in a session. currently, matt's viewer can't support that, but neuroglancer can, which is still an open related issue. 

to check out you can get this file: https://www.dropbox.com/s/de5esf6yam8pyuk/000108-dashboard_err.html?dl=0 and then open it in a browser. in the samples coverage 63 - 124, click on any of the colored square on the column of sample 98. this is the kind of thing we will want to enable on that context viewer next. 

that view is retrieving information from 36 different zarr files simultaneously directly from their dandi s3 paths. and if you really want to get fancy you could construct that neuroglancer url in two forms, one using the api and one directly going to s3.

web/src/views/FileBrowserView/FileBrowser.vue
yarikoptic: Fine with me: Although dandi-cli supports both, so we theoretically can end up with .ngff/ folders. But inability to view them would only encourage to make them properly named.

satra: i think the cli should stop supporting `.ngff` it's something lee and i used when we got started with writing them, but now that ome folks chimed in on it as sticking with `.ome.zarr` i think we can stick with that.

satra:
@waxlamp - i just realized that the dashboard i create can easily be altered to use either the api zarr endpoint or the s3 location. i will try this tomorrow. 

i updated the dashboard here to clean up the view a bit. this now shows each stain separately and then blended together. if you want to see something that loads 5 different zarr files, click on 97.

https://dl.dropbox.com/s/gzu58xq7etwtbu1/000108-dashboard.html

waxlamp:
> @waxlamp - i just realized that the dashboard i create can easily be altered to use either the api zarr endpoint or the s3 location. i will try this tomorrow.
> 
> i updated the dashboard here to clean up the view a bit. this now shows each stain separately and then blended together. if you want to see something that loads 5 different zarr files, click on 97.
> 
> https://dl.dropbox.com/s/gzu58xq7etwtbu1/000108-dashboard.html

This is cool! I'm not quite sure what I'm looking at/clicking on here--I wonder if you could demonstrate this for us at Monday's meeting, and then perhaps we could use the two alternative versions to build a bit of a load test to try to collect some data.

mvandenburgh:
As we discussed, I updated this to use s3 urls - @AlmightyYakob can you give this another review?

dandiapi/api/views/serializers.py
```diff
@@ -254,9 +254,24 @@ class AssetFolderSerializer(serializers.Serializer):
...
+
+    url = serializers.SerializerMethodField(method_name='get_url')
+
+    def get_url(self, asset: Asset):
+        if asset.is_blob:
+            return asset.blob.s3_url
+        elif asset.is_embargoed_blob:
+            return asset.embargoed_blob.s3_url
+        else:
+            return asset.zarr.s3_url
```
AlmightyYakob, line 269: IMO this abstraction should probably be done on the Asset model directly

```python
    @property
    def s3_url(self):
        if self.is_blob:
            return self.blob.s3_url
        elif self.is_embargoed_blob:
            return self.embargoed_blob.s3_url
        else:
            return self.zarr.s3_url
```

Then here, we could just do
```suggestion
    url = serializers.URLField(source='s3_url')
```

web/src/types/index.ts
```diff
@@ -52,6 +52,7 @@ export interface Asset {
   created: string,
   modified: string,
   version: Version,
+  zarr?: string,
```
AlmightyYakob, line 55: Shouldn't we also have the `blob` field here?
```suggestion
  zarr?: string,
  blob?: string,
```

AlmightyYakob (changes requested):

web/src/types/index.ts
mvandenburgh: Yes. However, there seems to be a lot of issues with the Asset TS types, so I figured I'd make a follow up PR to this one that cleans them up and ensures that they match up with the DRF serializers. In fact, that `zarr` field isn't even needed for this PR since we're using s3 urls now, so I'll drop that from this PR to keep things consistent. 

I might look into something like https://github.com/adenh93/django-typomatic to help us keep our TypeScript interfaces in sync with our DRF serializers.

AlmightyYakob (approved):

Issue #1012, opened 2022-04-08
dandischema alignment with releases


satra:
add check to github workflow for tests to see if the repo is using the latest dandischema in setup. for example, `0.7.0` was released and i thought that was critical to fixing the publishing issue on staging, yet it is not in current master.

yarikoptic:
What about adding a quick workflow to CI of dandi-archive which checks for most recent dandischema release and if there is a newer one than dandi-archive uses -- fail?!  May be the check should be of the `=~0.MINOR.PATCH` kind, i.e. allow for PATCH releases of dandischema .

WDYT @satra and @waxlamp ?

waxlamp:
This is a good idea. I'd say we could add a GH action that "fails" (without preventing merge) if the two are not aligned. (I'm staying general here in case we ever want to update dandischema without pulling the archive along--if that is a rare or otherwise atypical situation, we can just have a failure there prevent merge as usual.)

A separate matter: right now, the dandischema version in use is 0.7.1--I want to make sure that's correct. @satra, @yarikoptic?

PR #1074, opened 2022-04-28
Boost dandischema and DANDI_SCHEMA_VERSION

yarikoptic:
@satra have mentioned the need to upgrade dandischema library switching dandischema library from 0.6. to 0.7. series. Since there just happened 0.7.1 bugfix release -- made it the target version

- is this enough and safe to do?
- why don't we use `'dandischema=~0.7.1'` i.e. to allow for further versions within 0.7 series?
- whenever upgrade propagates, we would need to merge/release https://github.com/dandi/dandi-cli/pull/987 to do similar upgrade on client side
- related ref: #1012 
- why do we hardcode DANDI_SCHEMA_VERSION (diff here to switch to 0.6.3) and not just import it from dandischema? (`dandischema/consts.py:DANDI_SCHEMA_VERSION = "0.6.3"`)

this is pretty much a "motivational PR". Dear @waxlamp , I am assigning it to you to reassign/change/finalize/etc.


AlmightyYakob:
> @satra have mentioned the need to upgrade dandischema library switching dandischema library from 0.6. to 0.7. series. Since there just happened 0.7.1 bugfix release -- made it the target version
> 
> * is this enough and safe to do?

If the tests pass, I don't see why not.

> * why don't we use `'dandischema=~0.7.1'` i.e. to allow for further versions within 0.7 series?

If dandi-schema verisons comply with semantic versioning then pinning to `~=0.7.1` would be ideal.

> * why do we hardcode DANDI_SCHEMA_VERSION (diff here to switch to 0.6.3) and not just import it from dandischema? (`dandischema/consts.py:DANDI_SCHEMA_VERSION = "0.6.3"`)

I think we should keep the setting just in case it ever needs to be overridden, but I think the default value should be pulled from dandischema.

I can push a change that fixes CI and updates this behavior.

yarikoptic:
`dandi-cli` tests are failing because we still install `dandischema-0.6.0 ` in those tests since dandi-cli is awaiting for dandi-archive to get this upgrade.  So, let's proceed and possibly adjust dependencies etc later per discussion above. Filed a dedicated https://github.com/dandi/dandi-archive/issues/1078 for that

PR #1077, opened 2022-05-02
Add *ZarrArchive sizes into total_size

### LINKED ISSUE #1076: total size of archive doesn't reflect size of individual dandisets
at present (4pm ET apr 30) 108 is going through ingest updates and reports a total size of 77TB. however, the total size of the archive is reported on the dashboard at 66 TB.  i just deleted all the older hdf5 assets and microscopy folders.

yarikoptic:
If is completed, Closes #1076 

- first commit is a basic RF, worth adopting if the rest not
- follow up commit implements tentative solution
- last commit provides an attempt at establishing a test but I am not sure how to augment draft asset factory so it would allow for assetblob **or** zarrfile

TODOs
- [x] fixup/extend testing

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -298,11 +298,15 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
     @classmethod
     def total_size(cls):
         # delayed import to avoid present imports circularity
-        from .zarr import ZarrArchive, EmbargoedZarrArchive
+        from .zarr import ZarrArchive
+
         return sum(
             cls.objects.filter(assets__versions__isnull=False)
             .distinct()
             .aggregate(size=models.Sum('size'))['size']
             or 0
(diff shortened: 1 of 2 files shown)


yarikoptic:
@dandi/dandiarchive  folks -- guidance on "how to cook up the test" would be appreciated. May be @satra or @jwodder knows?

satra:
@yarikoptic - after ingest a zarr blob should have a size element associated with it. perhaps there is an ingest related test and that hooks into the size related test.

yarikoptic:
Dear @jwodder please give it a shot to figure out how to deal with the test factory for an asset to be able to point to the zarr archive (not just blob)

yarikoptic:
@satra -- if you like it -- merge it.  I have not tested locally with upload of zarr but unittest "works" now and I verified that it no longer crashes locally.  Might test later with upload

satra (approved):

yarikoptic:
FWIW -- confirmed on local instance with upload of zarr that works as desired!

yarikoptic:
will not release right away since we better fix up other things and have it tested in staging

PR #1082, opened 2022-05-05
Empty commit to cut a release

yarikoptic:
Added `patch` label as was done in https://github.com/dandi/dandi-archive/pull/1042 although it seems more of an `internal` thing ;)

The primary motivation for the release -- bring up in sync with staging on using the DANDI_SCHEMA 0.6.3 since can't even release dandi-cli with that support yet because of that.


yarikoptic:
ok, let's have a shot at it!

PR #1084, opened 2022-05-05
Extend docs with information about devel env

yarikoptic:
From my experience while trying to make it running locally
on my Debian OS laptop

Minimal changes. If anyone wants to expand -- please suggest the diffs.


yarikoptic:
ok, since no objections or further recommendations, so we do not conflict with possible improvements by @djarecka -- merging

Issue #1085, opened 2022-05-05
Should guard better for user metadata being malformed/missing?


yarikoptic:
Trying to get a local instance of API server + web UI for development.
I have DJANGO logged in in `/swagger` under my email/password.  Tried to authorize and get a token but

![image](https://user-images.githubusercontent.com/39889/166959931-bf3d82b9-3af6-4f4c-ab08-6a359c46a276.png)

I went to `/admin` interface and added first and last name for my record in Users.

and then now when I go to web UI at http://localhost:8085/ I get "Connection to server failed." and following traceback in console

```
$> [15:34:50] INFO     "GET /api/info/ HTTP/1.1" 200 344                                                                                   basehttp.py:187
           ERROR    Internal Server Error: /api/users/me/                                                                                    log.py:241
                    Traceback (most recent call last):                                                                                                 
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/django/core/handlers/exception.py",              
                    line 55, in inner                                                                                                                  
                        response = get_response(request)                                                                                               
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/django/core/handlers/base.py", line              
                    197, in _get_response                                                                                                              
                        response = wrapped_callback(request, *callback_args, **callback_kwargs)                                                        
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/django/views/decorators/csrf.py", line           
                    54, in wrapped_view                                                                                                                
                        return view_func(*args, **kwargs)                                                                                              
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/django/views/generic/base.py", line              
                    84, in view                                                                                                                        
                        return self.dispatch(request, *args, **kwargs)                                                                                 
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/rest_framework/views.py", line 509, in           
                    dispatch                                                                                                                           
                        response = self.handle_exception(exc)                                                                                          
                      File "/home/yoh/proj/dandi/dandi-archive/venvs/dev3/lib/python3.9/site-packages/rest_framework/views.py", line 469, in           
                    handle_exception                                                                                                                   
                        self.raise_uncaught_exception(exc)                                                                                             
[...]

PR #1086, opened 2022-05-05
BF(workaround): to avoid crash for user lacking metadata - return INCOMPLETE

yarikoptic:
Underlying issue is somewhere in the logic/instructions so that locally created
user lacks .metadata somehow, although it seems to be available within /admin
interface among Users table for that user.  So must be some glue missing etc.
With this workaround it should still work ok, but would avoid crashing
and due to INCOMPLETE I would expect that user lacking any super powers
otherwise granted

Note: pycharm checker highlights `.metadata` since it is not defined in `User` class at all.  Is it indeed just ducktyped into it?

Workaround for https://github.com/dandi/dandi-archive/issues/1085

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
+
 from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.models import User
 from django.db.models import Q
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/users.py
```diff
@@ -15,6 +17,19 @@
...
+
+def _get_user_status(user: User):
+    # Workaround for https://github.com/dandi/dandi-archive/issues/1085
+    # TODO: remove/robustify some other way whenever underlying reason is
+    # identified
+    metadata = getattr(user, 'metadata', None)
+    if metadata:
+        return metadata.status
+    logger.error("User %s lacks .metadata. Returning user's status as INCOMPLETE", user)
+    return UserMetadata.Status.INCOMPLETE
```
yarikoptic, line 31: @djarecka make it here

```suggestion
    return UserMetadata.Status.APPROVED
```

if you want to allow that user to be able to click 'Create dataset' etc

yarikoptic:
since I keep getting annoyed to need this patch to do anything with local instance, I will merge this PR although a proper solution for #1085 is yet to be developed to populate `.metadata` properly for a local user.

PR #1088, opened 2022-05-06
Fix opening Meditor when clicking on "Fix Issues" in the list of dandiset metadata issues

### LINKED ISSUE #950: "Fix Issues" button on DLP navigates to empty page
This seems to have broken at some point. I believe the intended behavior would be to simply open the meditor. An additional benefit would be to navigate to the first tab in which errors occur, but I suppose it isn't necessary, just a thought.

djarecka:
opening Meditor when clinging "FIX ISSUES".
fixes #950

@yarikoptic - see if this is what you wanted. I used `openMeditor` instead of `meditorLink`. I'm pretty sure there was soe idea behind it, so don't know if I should remove it completely

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -421,18 +421,6 @@ export default defineComponent({
       () => !!(!publishButtonDisabled.value && user.value?.admin && !isOwner.value),
     );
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you @djarecka !!!  CI is green, works great for me, I think we know now how/why it got broken, so let's proceed!

PR #1092, opened 2022-05-09
checking the validation status

### LINKED ISSUE #756: Show indicator of running validation job in DLP
per a brief discussion during stand-up and inspired by the question of @bendichter on slack last week on the status of the dandiset after a sizeable upload, I think we might improve UX if besides purely "Red" state, we add a "Yellow" ("in progress") state.

ATM we have only Red as 

image

which would be red if there are issues with metadata (such as missing mandatory fields) but also if some fields (such as sha256) are missing because jobs are still in the queue/running (to compute sha256 etc), and no user-actionable errors detected so far. 

I am not sure how feasible/easy it would be to implement in current state of the workflow ATM since most likely there is no association between "missing sha256" and "job X is yet to run/running to compute that".

djarecka:
fixes #756 

checking if the validation status is `Pending` or `Invalid` before providing the error

@yarikoptic - please see if this is what you expect. I didn't want to copy too much so I put multiple `if/else` statement in the existing block, but perhaps I should create entire new object (or `<v-row>`)

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -146,14 +146,14 @@
                 </v-row>
               </v-card>
             </template>
-            <span v-if="currentDandiset.status === 'Pending'">
+            <span v-if="currentDandiset && currentDandiset.status === 'Pending'">
               Wait for validation or reload the page
             </span>
             <span v-else>Fix issues with metadata</span>
           </v-tooltip>
         </template>
         <v-card
-          v-if="currentDandiset.status === 'Invalid'"
+          v-if="currentDandiset && currentDandiset.status === 'Invalid'"
           class="pa-1"
         >
           <v-list
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -136,7 +136,8 @@
                     </div>
 
                     <div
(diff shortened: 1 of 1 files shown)


web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -129,7 +129,14 @@
...
                     <div
-                      v-if="currentDandiset"
+                      v-if="currentDandiset && currentDandiset.status === 'Pending'"
+                      class="text-caption"
+                    >
+                      Validation is pending.
+                    </div>
+
+                    <div
+                      v-else-if="currentDandiset && currentDandiset.status === 'Invalid'"
```
yarikoptic, line 139:
```suggestion
                      v-else-if="currentDandiset && currentDandiset.status === 'Invalid'" && currentDandiset.version_validation_errors.length
```
since I think it would be marked invalid also if there are asset errors. So we end up with smth like:

image

yarikoptic (changes requested):
indeed we might be getting too much into if/else partitioning which would have some holes/incorrect behavior but IMHO it would already be better/more adequate.  Left just one suggestion having observed an odd display

yarikoptic:
eh -- now I think I understand what you meant by needing a v-row:

image
 
I have not spotted that you were just adjusting the text instead of an entire element which is what seems needed here or we would have some odd empty "Fix Issues" etc.

djarecka:
@yarikoptic - perhaps I copy entire block of code and add `if` "higher", might be cleaner...

djarecka:
opened #1097 instead

PR #1097, opened 2022-05-11
checking the validation status, new attempt 

### LINKED ISSUE #756: Show indicator of running validation job in DLP

djarecka:
fixes #756 

instead of #1092

checking if the validation status is `Pending` before proving the metadata error

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -93,7 +93,57 @@
     </v-row>
 
     <v-row
-      v-if="currentDandiset.version_validation_errors.length "
+      v-if="currentDandiset.status === 'Pending'"
+      class="mb-4 px-1"
+      no-gutters
+    >
+      <v-menu
+        :nudge-width="200"
+      >
+        <template #activator="{ on: menu, attrs }">
+          <v-tooltip bottom>
+            <template #activator="{ on: tooltip }">
+              <v-card
+                class="amber lighten-5 no-text-transform"
+                outlined
+                v-bind="attrs"
+                v-on="{ ...tooltip, ...menu }"
+              >
+                <v-row class="align-center px-4">
+                  <v-col
+                    cols="1"
+                    class="justify-center py-0"
+                  >
+                    <v-icon
+                      color="warning"
+                      class="mr-1"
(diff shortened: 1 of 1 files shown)


web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -93,7 +93,57 @@
...
+                      v-if="currentDandiset"
+                      class="text-caption"
+                    >
+                      Validation of the dandiset is pending.
+                    </div>
+                  </v-col>
+                </v-row>
+              </v-card>
+            </template>
+            <span>Wait for validation or reload the page</span>
```
yarikoptic, line 139:
```suggestion
            <span>Reload the page to see if validation is over.</span>
```

it seems that our app is not that "React"ife to `v-if="currentDandiset.status === 'Pending'"` !  I had waited enough, nothing changed so "Wait for validation" is IMHO misleading.  Thus I am proposing to adjust the message, and filed a dedicated https://github.com/dandi/dandi-archive/issues/1098 for the reactive updates

yarikoptic: otherwise -- seems good to me!!!  I dislike such verbosity needed for adding a single row and can't assess the correctness -- but seems to work well!

djarecka: yes, I noticed that it needs reload, but wasn't sure if doing something wrong or the changes are needed somewhere else... thanks for opening the issue, can think about it later

I'm sure that the code could be shorter, but wanted to be as similar as possible to the other card

yarikoptic (approved):
thank you @djarecka !

PR #1103, opened 2022-05-13
BF: populate repository field for asset + strip for asset and version

### LINKED ISSUE #1081: Complains: repository: URL host invalid, top level domain required
I was testing if https://github.com/dandi/dandi-archive/issues/961 is still an issue after dandischema upgrade.  Created a new dandiset (since old ones still have old error message; separate issue/question: #1080): https://gui-staging.dandiarchive.org/dandiset/200589 but it gives

image

whenever 

```shell
$> grep repository /tmp/dandiset.json 
  "repository": "https://gui-staging.dandiarchive.org",
```

@satra -- do you know where this error message comes from and what does it really want from me? e.g. does it want it to be `https://dandiarchive.org` ? (which would be inappropriate for staging)

yarikoptic:
Since that field is "computed" (or populated) lets make it explicitly
populated and stripped.  dandimodel does default atm to one
which should correspond but it seems might fail.

Hopefully this fixes #1081 . First I will try to reproduce on current staging, and then see if this addresses it...

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -191,12 +191,12 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
             s3_url = self.embargoed_blob.s3_url
         else:
             s3_url = self.zarr.s3_url
-
         metadata = {
             **self.metadata,
             'id': f'dandiasset:{self.asset_id}',
             'path': self.path,
             'identifier': str(self.asset_id),
+            'repository': settings.DANDI_WEB_APP_URL,
             'contentUrl': [download_url, s3_url],
             'contentSize': self.size,
             'digest': self.digest,
@@ -289,6 +289,7 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
             'digest',
             'datePublished',
             'publishedBy',
+            'repository',
         ]
         return {key: metadata[key] for key in metadata if key not in computed_fields}
 
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -193,13 +193,13 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
             s3_url = self.zarr.s3_url
         metadata = {
             **self.metadata,
-            'id': f'dandiasset:{self.asset_id}',
-            'path': self.path,
-            'identifier': str(self.asset_id),
-            'repository': settings.DANDI_WEB_APP_URL,
-            'contentUrl': [download_url, s3_url],
             'contentSize': self.size,
+            'contentUrl': [download_url, s3_url],
             'digest': self.digest,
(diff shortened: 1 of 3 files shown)


yarikoptic:
observations: 
- the original issue #1081 is already resolved somehow on staging (we had some fixes merged/released recently) in that `repository` is simply not populated for an asset on it. Even tests here test that there is no 'repository' so they failed now that I have added it.
- but that makes it odd on how it got that field at all... Eureka: may be I was running dandi-cli locally in the env where I did set that `DJANGO_DANDI_API_URL` env var (well -- just did `source ./dev/export-env.sh` in dandi-archive repo I guess), `dandischema` (while used by dandi-cli) picked it up and populated it in the metadata record uploaded for the asset and server did not strip it off.  @djarecka -- may be that is how you reproduced it, in that it was the same environment in which you ran dandi-archive locally ?

So, it seems that stripping/re-populating anyways is the best way to proceed to guarantee consistency.
I will also push code RF to have those fields lists sorted in the code I touched since it gets hard to locate one and there is no reason to not have them sorted.

dandiapi/api/models/asset.py
```diff
@@ -191,15 +191,15 @@ def _populate_metadata(self):
...
-            'path': self.path,
-            'identifier': str(self.asset_id),
-            'contentUrl': [download_url, s3_url],
             'contentSize': self.size,
+            'contentUrl': [download_url, s3_url],
             'digest': self.digest,
+            'id': f'dandiasset:{self.asset_id}',
+            'identifier': str(self.asset_id),
+            'path': self.path,
+            'repository': settings.DANDI_WEB_APP_URL,
```
yarikoptic, line 202: note: `repository` is the field I added here

dandiapi/api/models/asset.py
```diff
@@ -281,14 +281,15 @@ def save(self, *args, **kwargs):
...
             'contentSize',
-            'digest',
+            'contentUrl',
             'datePublished',
+            'digest',
+            'id',
+            'identifier',
+            'path',
             'publishedBy',
+            'repository',
```
yarikoptic, line 292: note: `repository` is the field I added here

dandiapi/api/models/version.py
```diff
@@ -188,18 +188,19 @@ def citation(cls, metadata):
...
             'dateCreated',
             'datePublished',
-            'publishedBy',
+            'doi',
+            'id',
+            'identifier',
             'manifestLocation',
+            'name',
+            'publishedBy',
+            'repository',
```
yarikoptic, line 201: note: `repository` is the field I added here as well for consistency -- I do not see why we should not strip it from version as well

yarikoptic:
old, didn't gain review, original problem depicted in https://github.com/dandi/dandi-archive/issues/1081 seems was resolved through some other means.

PR #1107, opened 2022-05-25
ENH: increase page size from 25 to 100

yarikoptic:
Should result in x4 fold reduction of logs from such requests,
speed up client interactions etc.

Motivation: https://github.com/dandi/dandi-cli/issues/1018

edit 1: possible "visible" side effect is 100 assets to be displayed per page in file navigator per dandiset. Might be considered pros or cons

### DIFF
--- a/dandiapi/api/views/common.py
+++ b/dandiapi/api/views/common.py
@@ -5,7 +5,7 @@ max_page_size = 1000
 
 
 class DandiPagination(PageNumberPagination):
-    page_size = 25
+    page_size = 100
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
The web client [specifies a page_size for the file browser](https://github.com/dandi/dandi-archive/blob/master/web/src/views/FileBrowserView/FileBrowser.vue#L232), so that won't be an issue.

yarikoptic:
coolio, let's give it a shot then

PR #1116, opened 2022-06-10
Replace `girder-pytest-pyppeteer` usage with `pytest-pyppeteer`

mvandenburgh:


mvandenburgh:
This was an experiment I did to see what the capabilities of `pytest-pyppeteer` are, and whether or not it could replace `girder-pytest-pyppeteer`. It lacks some features that `girder-pytest-pyppeteer` has, and e2e testing isn't the highest priority right now, so for the moment I think we're better staying with what we have.

PR #1120, opened 2022-06-15
Add ngff rename script as a management command

mvandenburgh:
Adds the script I wrote to rename .ngff files to .ome.zarr as a management command so we have it for future reference

### DIFF
--- a/dandiapi/api/management/commands/rename_ngff.py
+++ b/dandiapi/api/management/commands/rename_ngff.py
@@ -34,7 +34,7 @@ def rename_ngff(dandiset_id: int | None):
                 zarr.name = new_name
                 zarr.save(update_fields=['name'])
 
-                for asset in zarr.assets.all():
+                for asset in zarr.assets.iterator():
                     if asset.path.endswith('.ngff'):
                         new_path = f'{asset.path[:-5]}.ome.zarr'
                         click.echo(
--- a/dandiapi/api/management/commands/rename_ngff.py
+++ b/dandiapi/api/management/commands/rename_ngff.py
@@ -28,7 +28,7 @@ def rename_ngff(dandiset_id: int | None):
         new_name = f'{zarr.name[:-5]}.ome.zarr'
 
         # only rename the ngff if a ome.zarr hasn't been uploaded already
-        if not ZarrArchive.objects.filter(name=new_name).exists():
+        if not ZarrArchive.objects.filter(name=new_name, dandiset=zarr.dandiset).exists():
             click.echo(f'Renaming ZarrArchive ({zarr.id}) from {zarr.name} to {new_name}.')
             with transaction.atomic():
                 zarr.name = new_name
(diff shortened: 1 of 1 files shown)


AlmightyYakob (dismissed):
Looks good overall, just one suggestion

dandiapi/api/management/commands/rename_ngff.py
```diff
@@ -0,0 +1,44 @@
...
+        click.echo(f'Renaming .ngff files to .ome.zarr for dandiset {dandiset_id}...')
+        qs = qs.filter(dandiset_id=dandiset_id)
+    else:
+        click.echo('Renaming .ngff files to .ome.zarr for all dandisets...')
+
+    for zarr in qs.prefetch_related('assets'):
+        new_name = f'{zarr.name[:-5]}.ome.zarr'
+
+        # only rename the ngff if a ome.zarr hasn't been uploaded already
+        if not ZarrArchive.objects.filter(name=new_name).exists():
```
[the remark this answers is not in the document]
mvandenburgh: Good catch, I restricted it to the current zarr's dandiset https://github.com/dandi/dandi-archive/pull/1120/commits/38c7dd1e1b0aab4b046491c3237aee087f1f3c75. `ZarrArchives` are associated with `Dandisets` and not `Versions`, so I believe that is sufficient

mvandenburgh:
@AlmightyYakob can you review the changes here?

dandiapi/api/management/commands/rename_ngff.py
```diff
@@ -0,0 +1,44 @@
...
+        new_name = f'{zarr.name[:-5]}.ome.zarr'
+
+        # only rename the ngff if a ome.zarr hasn't been uploaded already
+        if not ZarrArchive.objects.filter(name=new_name, dandiset=zarr.dandiset).exists():
+            click.echo(f'Renaming ZarrArchive ({zarr.id}) from {zarr.name} to {new_name}.')
+            with transaction.atomic():
+                zarr.name = new_name
+                zarr.save(update_fields=['name'])
+
+                for asset in zarr.assets.all():
```
AlmightyYakob, line 37: Should we use `.iterator` here?

mvandenburgh: I suppose so, since a `ZarrArchive` could have a lot of associated assets, and the primary use case for this script would be running it in a `heroku run` environment (where we're constrained by the dyno's memory quota).

yarikoptic:
FWIW dandi move command could be used. See https://github.com/dandi/dandi-cli/issues/1081#issuecomment-1191798525 for an example with regex.

mvandenburgh:
> FWIW dandi move command could be used. See [dandi/dandi-cli#1081 (comment)](https://github.com/dandi/dandi-cli/issues/1081#issuecomment-1191798525) for an example with regex.

Oh that might be a better option for this kind of operation than a management command. @satra requested that this script be committed to the repo, so I'll leave it up to him whether or not this should still be merged.

AlmightyYakob (approved):
LGTM if we end up wanting to merge.

yarikoptic:
> Jun 15

was the date for this PR birth... `dandi move` https://github.com/dandi/dandi-cli/pull/1024 was initiated May 31st but merged into master of dandi-cli Jun 27 .  I do feel that this PR is just no longer needed, but @satra might have different thoughts.

satra:
i think this is a more generic operation at the level of the database if necessary. i would keep it in the folder as an example of how to operate/rename on many assets, but most likely we will never use this specific script (there should be no ngff files moving forward).

PR #1121, opened 2022-06-15
Fix broken search bar

### LINKED ISSUE #1068: Search bar broken in frontend
To reproduce -

- Go to https://dandiarchive.org/dandiset, type something in the search bar, and hit enter. This will work correctly
- Try to do another search from the search results page - this will not work and will just clear the search bar

mvandenburgh:
Fixes #1068 

The search bar now properly works when used from the `/search` page. 

There were two issues -
1) The `submit` event emitted by the search bar on form submission wasn't being caught. So, the default behavior of the browser (refreshing the page) kicked in.
2) A stale, non-reactive copy of the `vue-router` `route` object was being used, which prevented any changes to the search field from being recognized.

### DIFF
--- a/dandiapi/api/tests/e2e/test_search.py
+++ b/dandiapi/api/tests/e2e/test_search.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pyppeteer.page import Page
+    from pyppeteer.element_handle import ElementHandle
+
+import pytest
+from utils import contains_text, wait_for_navigation
+
+from dandiapi.api.models import Version
+
+
+@pytest.mark.pyppeteer
+@pytest.mark.django_db
+async def test_search(
+    page: Page,
+    webpack_server: str,
+    published_version_factory,
+    asset_factory,
+):
+    # Create some dandisets with assets
+    dandisets: list[Version] = [published_version_factory() for _ in range(20)]
+    for dandiset in dandisets:
+        asset = asset_factory()
+        asset.versions.set([dandiset])
+
+    # Search for each of them using the search bar and ensure they are in the results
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
LGTM if it passes tests and works as expected.

AlmightyYakob (approved):

yarikoptic:
Could a rudimentary unit test be added? What tests did you have in mind  @waxlamp ?

waxlamp:
> Could a rudimentary unit test be added? What tests did you have in mind @waxlamp ?

I just meant the test suite as is. This doesn't seem like something that requires a unit or integration test, though we can add one if you insist (the feature was broken for a long time, after all). But, now that it's fixed, I am confident that it works properly, and adding a test might be more cost than it's worth.

Let us know what you think.

yarikoptic:
I personally just tend to code-in test even before introducing the fix, so whenever I add a fix, I know that it is fixed *now* and in any future refactoring/fix/etc where I might forget to manually check for that particular behavior.  As the team is "fluid" and we do not have a dedicated "testing/QA" team, encoding assumptions and discovered incorrect-and-then-fixed behavior doesn't come back.  I am curious now -- how difficult is to code in the test for this particular behavior?

mvandenburgh:
> I am curious now -- how difficult is to code in the test for this particular behavior?

Since this bug was purely UI-only, it would need to be a browser-based e2e test similar to https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/tests/e2e/test_cookies.py. I guess for the test we could populate the DB with a couple different dandisets, and then search for each using the search bar and assert that they appear in the search results.

mvandenburgh:
> how difficult is to code in the test for this particular behavior?

@yarikoptic not too difficult, I added a test for it.

PR #1122, opened 2022-06-15
Add publish "checklist"

### LINKED ISSUE #310: Publish checklist
Everything that needs to happen for publish

## Code to write

- [x] Finish validation #302
- [x] Delete computed/injected properties from incoming metadata before it is saved in DB
- [x] Use Dorota's DOI publish function
- [x] Migrate all metadata to schema version 0.4.1
  - [x] @satra mentioned he would write a function that takes 0.3.1 compliant metadata and outputs 0.3.2(?)
  - [x] Update computed asset metadata https://github.com/dandi/dandi-api/pull/316#discussion_r647639084
  - [x] We would just take this function and write a django script to apply it to all draft versions
  - [x] Bump the `SCHEMA_VERSION` in https://github.com/dandi/dandi-infrastructure/blob/master/terraform/api.tf#L27
- [x] Merge https://github.com/dandi/dandiarchive/pull/730 into staging ( @mvandenburgh )
- [x] Figure out what's up with dandi-cli testing errors
- [x] Thoroughly test publish in staging
  - [x] The Publish button should only be enabled when version+asset metadata is valid, and should have error messages otherwise
  - [x] Manifest files are written and have the correct content
  - [x] DOIs are generated (and published?) correctly on the test server.
    - [x] @yarikoptic Clarify if publishing on the test server is cool, since it's so very very permanent
  - [x] dandi-cli can pull correctly from published dandisets
  - [x] Everything in the UI still looks acceptable for published dandisets.
- [x] [Do not allow deletes on dandisets with published versions](https://github.com/dandi/dandi-api/issues/288)

## Deploying
- [x] Merge `staging` to `master`
- [x] Run migrate_version_metadata script
- [x] Update schemaVersion in Terraform
- [x] Migrate asset metadata (@yarikoptic, @satra  et al )
- [x] Update DOI server/user to production in https://github.com/dandi/dandi-infrastructure/blob/master/terraform/api.tf#L28-L29
  - [x] Update the DOI password in the Terraform cloud secret variables
- [x] Are we publishing everyone's dandisets for them, or are we letting users deal with that? (We are letting users publish their own dandisets.)
- [x] Investigate Yarik's publishing failure
- [x] Allow users to publish (currently only admins have permission)

### LINKED ISSUE #1090: add checklist to publish button.
this would be a nice checklist to add when someone hits publish and before the publish actually happens. 

```
To maximize utility, have you checked the list of authors and keywords and have you added:
1. a descriptive title (e.g., Data related to foraging behavior in bees vs Smith et al 2022)
2. a description that clearly describes the dataset for another neuroscientist
3. a reference to a protocol in protocols.io detailing how you collected the data? this can also be another publication added as a resource.
4. a reference to an ethics protocol number if you have an institutional review board review the study.
5. funding information: dandi treats funding agencies as contributors. you can add multiple contributing institutions. unless you are the funder, you can add a new contributor of type organization, uncheck include contributor in citation, include dcite:Funder role and add the relevant award info.
6. reference pointers to code in github, related publications, etc.,.
```

one can also ask for a description/commit message related to the publication

mvandenburgh:
Adds a dialog with the checklist from #1090  that pops up when the publish button is clicked.

Screenshot from 2022-06-15 16-15-52

Closes #1090

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -425,7 +425,7 @@ import { RawLocation } from 'vue-router';
 import { open as openMeditor } from '@/components/Meditor/state';
 
 const PUBLISH_CHECKLIST = [
-  'A descriptive title (e.g., "Data related to foraging behavior in bees" rather than "Smith et al 2022")',
+  'A descriptive title (e.g., <span class="font-italic">Data related to foraging behavior in bees</span> rather than <span class="font-italic">Smith et al 2022</span>)',
   'A clear, informative description that is helpful for other neuroscientists',
   'A reference to a protocol in protocols.io detailing how you collected the data (this can also be another publication added as a resource)',
   'A reference to an ethics protocol number if you are working with an Institutional Review Board on this study',
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -104,8 +104,8 @@
                   <v-card-text>
                     <v-list>
                       <span class="text-body-1 font-weight-bold">
-                        To maximize utility, have you checked the list of
-                        authors and keywords and have you added:
+                        For best results, please check the following
+                        items before you publish:
                       </span>
 
                       <v-list-item
@@ -423,12 +423,12 @@ import { RawLocation } from 'vue-router';
 import { open as openMeditor } from '@/components/Meditor/state';
 
 const PUBLISH_CHECKLIST = [
-  'a descriptive title (e.g., Data related to foraging behavior in bees vs Smith et al 2022)',
-  'a description that clearly describes the dataset for another neuroscientist',
-  'a reference to a protocol in protocols.io detailing how you collected the data? this can also be another publication added as a resource.',
-  'a reference to an ethics protocol number if you have an institutional review board review the study.',
-  'funding information: dandi treats funding agencies as contributors. you can add multiple contributing institutions. unless you are the funder, you can add a new contributor of type organization, uncheck include contributor in citation, include dcite:Funder role and add the relevant award info.',
-  'reference pointers to code in github, related publications, etc.,.',
+  'A descriptive title (e.g., "Data related to foraging behavior in bees" rather than "Smith et al 2022")',
+  'A clear, informative description that is helpful for other neuroscientists',
+  'A reference to a protocol in protocols.io detailing how you collected the data (this can also be another publication added as a resource)',
+  'A reference to an ethics protocol number if you are working with an Institutional Review Board on this study',
+  'Funding information (DANDI treats funding agencies as contributors, so you can add multiple contributing institutions as needed. If you are the funder, you can add a new contributor of type "organization", uncheck "include contributor in citation", set the role as "dcite:Funder", and include the relevant award information)',
+  'References to code in GitHub, related publications, etc.',
 ];
 
 function getValidationErrorIcon(errorField: string): string {
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -107,13 +107,15 @@
                         For best results, please check the following
                         items before you publish:
                       </span>
-
+                      <!-- Note: this is safe as we aren't rendering any user-generated input -->
+                      <!-- eslint-disable vue/no-v-html -->
                       <v-list-item
                         v-for="(item, i) in PUBLISH_CHECKLIST"
                         :key="`checklist_item_${i}`"
                         class="text-body-2 my-1"
-                        v-text="`${i+1}. ${item}`"
+                        v-html="`<span>${i+1}. ${item}</span>`"
                       />
+                      <!-- eslint-enable vue/no-v-html -->
                     </v-list>
                   </v-card-text>
 
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -75,17 +75,69 @@
                 </span>
               </v-col>
             </v-row>
-            <v-btn
-              block
-              :color="showPublishWarning ? 'error' : 'success'"
-              depressed
-              :disabled="publishButtonDisabled"
-              @click="publish"
+            <div
+              class="text-center"
             >
-              Publish
-              <v-spacer />
-              <v-icon>mdi-upload</v-icon>
-            </v-btn>
+              <v-dialog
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):
LGTM

waxlamp:
Don't merge yet--I want to copy-edit the dialog text.

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -75,17 +75,69 @@
...
+                        authors and keywords and have you added:
+                      </span>
+
+                      <v-list-item
+                        v-for="(item, i) in PUBLISH_CHECKLIST"
+                        :key="`checklist_item_${i}`"
+                        class="text-body-2 my-1"
+                        v-text="`${i+1}. ${item}`"
+                      />
+                    </v-list>
```
waxlamp, line 117: Is there a way for these to contain HTML content, rather than just text? (This will benefit from being able to, e.g., italicize certain parts of the items.)

It seems like the HTML-native way to do this (if we really want a numbered list) is to just use an `<ol>` and let the renderer fill in the numbers. What's the vuetify way to do that?

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -370,6 +422,15 @@ import { draftVersion, VALIDATION_ICONS } from '@/utils/constants';
 import { RawLocation } from 'vue-router';
 import { open as openMeditor } from '@/components/Meditor/state';
 
+const PUBLISH_CHECKLIST = [
+  'a descriptive title (e.g., Data related to foraging behavior in bees vs Smith et al 2022)',
+  'a description that clearly describes the dataset for another neuroscientist',
+  'a reference to a protocol in protocols.io detailing how you collected the data? this can also be another publication added as a resource.',
+  'a reference to an ethics protocol number if you have an institutional review board review the study.',
+  'funding information: dandi treats funding agencies as contributors. you can add multiple contributing institutions. unless you are the funder, you can add a new contributor of type organization, uncheck include contributor in citation, include dcite:Funder role and add the relevant award info.',
+  'reference pointers to code in github, related publications, etc.,.',
```
waxlamp, line 431: Some wholesale copy editing:

```suggestion
  'A descriptive title (e.g., Data related to foraging behavior in bees vs Smith et al 2022)',
  'A clear, informative description that is helpful for other neuroscientists',
  'A reference to a protocol in protocols.io detailing how you collected the data (this can also be another publication added as a resource)',
  'A reference to an ethics protocol number if you are working with an Institutional Review Board on this study',
  'Funding information (DANDI treats funding agencies as contributors, so you can add multiple contributing institutions as needed. If you are the funder, you can add a new contributor of type "organization", uncheck "include contributor in citation", set the role as "dcite:Funder", and include the relevant award information)',
  'References to code in GitHub, related publications, etc.',
```

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -370,6 +422,15 @@ import { draftVersion, VALIDATION_ICONS } from '@/utils/constants';
 import { RawLocation } from 'vue-router';
 import { open as openMeditor } from '@/components/Meditor/state';
 
+const PUBLISH_CHECKLIST = [
+  'a descriptive title (e.g., Data related to foraging behavior in bees vs Smith et al 2022)',
```
waxlamp, line 426: I don't think "Data related to foraging behavior in bees vs Smith et al 2022" is a good example of a dandiset title (specifically because of the "vs Smith et al 2022" part).

Ah--the "vs" is introducing an example of a bad title? (@satra, do I have that right?) If so, then we should change this to:

```suggestion
  'a descriptive title (e.g., "Data related to foraging behavior in bees" rather than "Smith et al 2022")',
```

(This is also a case where I would prefer to use italics to highlight the text, which relates to my question above about HTML content inside of a `<v-list-item>`.)

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -370,6 +422,15 @@ import { draftVersion, VALIDATION_ICONS } from '@/utils/constants';
 import { RawLocation } from 'vue-router';
 import { open as openMeditor } from '@/components/Meditor/state';
 
+const PUBLISH_CHECKLIST = [
```
waxlamp, line 425: This doesn't matter much, but what is the general JavaScript style policy on defining constants using all-majuscule snake case? My gut tells me this should just be `publishChecklist`.

waxlamp (changes requested):
I suggested some cosmetic and content changes to the checklist items. @satra, can you give those a sanity check to make sure I didn't twist the meaning of any of them?

@mvandenburgh and @AlmightyYakob, of course I'd like your feedback on those changes as well.

waxlamp:
@mvandenburgh, @satra, please take a look at my copy edits and let me know what you think.

web/src/views/DandisetLandingView/DandisetPublish.vue
satra: i think this one conflicts with the previous commit (the first bullet). otherwise the changes LGTM.

satra (approved):
approving the text changes

web/src/views/DandisetLandingView/DandisetPublish.vue
mvandenburgh: MDN docs say all capitals https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage.

Personally I think using all-capital snake case makes it clearer that the array itself is constant, and signals that you shouldn't be doing `PUBLISH_CHECKLIST.push(...)` anywhere.

web/src/views/DandisetLandingView/DandisetPublish.vue
mvandenburgh: There doesn't appear to be a way to use `<ol>` in vuetify. You can leverage the `tag` prop to make the `v-list` and `v-list-item` elements use `<ol>` and `<li>` respectively, but it still doesn't render any numbers
Screenshot from 2022-07-19 13-09-39

web/src/views/DandisetLandingView/DandisetPublish.vue
mvandenburgh: > (This is also a case where I would prefer to use italics to highlight the text, which relates to my question above about HTML content inside of a `<v-list-item>`.)

https://github.com/dandi/dandi-archive/pull/1122/commits/444804d5e5155b623938f45eb9d3b20a6f04c34f

mvandenburgh:
@waxlamp I implemented your suggestions and modified the list items to render as `innerHTML` instead of `innerText`.

waxlamp (approved):

PR #1125, opened 2022-06-20
Present dialog if attempting to remove self from dandiset

### LINKED ISSUE #754: Don't allow self-removal from dandiset owners without a warning
Currently it is possible to use your ownership permissions to remove yourself as an owner. This becomes an irreversible action, requiring contacting another owner or the admins to be re-added. Since it's such an easy action to take (at least, in the DLP redesign), we should put up a guard to stop it from happening by mistake.

AlmightyYakob:
Closes #754

### DIFF
--- a/web/src/components/DLP/DandisetOwnersDialog.vue
+++ b/web/src/components/DLP/DandisetOwnersDialog.vue
@@ -284,7 +284,11 @@ export default defineComponent({
     }
 
     const selfRemovalWarningDialog = ref(false);
-    function removeOwner(owner: User) {
+    function removeOwner(owner: User | null) {
+      if (owner === null) {
+        throw new Error('Cannot remove null owner from dandiset!');
+      }
+
       // If current user, open dialog and wait for second call to this function
       if (ownerIsCurrentUser(owner) && selfRemovalWarningDialog.value === false) {
         selfRemovalWarningDialog.value = true;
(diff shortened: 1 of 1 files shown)


mvandenburgh:
I think the intent isn't to completely remove the ability for one to remove themselves as an owner, but that there should be some sort of confirmation prompt warning them before they go through with it.

web/src/components/DLP/DandisetOwnersDialog.vue
```diff
@@ -100,6 +100,35 @@
...
+                Remove yourself from this Dandiset?
+              </v-card-title>
+              <v-card-subtitle>
+                To regain ownership of this dandiset, you will
+                need another owner or an admin to add you.
+              </v-card-subtitle>
+              <v-card-actions>
+                <v-spacer />
+                <v-btn
+                  color="secondary"
```
mvandenburgh, line 118: The rest of the `v-btn`'s on the DLP have their shadows removed, I think this would apply here for UI style consistency
```suggestion
                  color="secondary"
                  elevation="0"
```

web/src/components/DLP/DandisetOwnersDialog.vue
```diff
@@ -100,6 +100,35 @@
...
+                  @click="selfRemovalWarningDialog = false"
+                >
+                  Cancel
+                </v-btn>
+                <v-btn
+                  color="error"
```
mvandenburgh, line 124: See above comment
```suggestion
                  color="error"
                  elevation="0"
```

mvandenburgh (changes requested):
Looks good, just some suggested style changes to keep the UI consistent with the existing DLP styling

mvandenburgh (approved):
LGTM

PR #1126, opened 2022-06-21
Use QuerySet.iterator in asset paths endpoint

mvandenburgh:
When iterating over large querysets, using `[QuerySet.iterator](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#iterator)` is more memory-efficient. This may mitigate or solve the OOM errors with dandiset 26.


AlmightyYakob (approved):
This is cool! It might be worth investigating what else in our codebase can be optimized this way

mvandenburgh:
> This is cool! It might be worth investigating what else in our codebase can be optimized this way

Agreed! The asset paths endpoint immediately came to mind when I found out about `iterator`, but it's definitely worth taking a close look at where else this can be used i think - I'll file an issue

PR #1127, opened 2022-06-21
Add timeout to gunicorn in heroku

mvandenburgh:
Downstreaming https://github.com/girder/cookiecutter-girder-4/pull/190 from the cookiecutter.

This will help distinguish Heroku request timeouts from other errors that cause a `SystemExit` in sentry.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 release: ./manage.py migrate
-web: gunicorn --bind 0.0.0.0:$PORT dandiapi.wsgi
+web: gunicorn --bind 0.0.0.0:$PORT dandiapi.wsgi --timeout 25
 # celery-beat: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery beat --loglevel INFO
 # Rather than using a dedicated worker for Celery Beat, we simply use the -B option on the priority task worker.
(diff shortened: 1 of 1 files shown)


waxlamp:
What's a good way to test this? Maybe an endpoint with an artificial delay that we can introduce and then remove later?

mvandenburgh:
> What's a good way to test this? Maybe an endpoint with an artificial delay that we can introduce and then remove later?

Yeah I think that's the only way we could reliably test this. We could add a `time.sleep` call to an endpoint and test in staging.

waxlamp:
> > What's a good way to test this? Maybe an endpoint with an artificial delay that we can introduce and then remove later?
> 
> Yeah I think that's the only way we could reliably test this. We could add a `time.sleep` call to an endpoint and test in staging.

Can you file an issue and assign it to yourself?

AlmightyYakob:
FWIW I think making a temporary endpoint and merging it to master to test staging is the wrong approach. You can create this endpoint and run gunicorn locally, which imo is sufficient.

mvandenburgh:
Oh yeah i think you're right. For some reason I thought we had to test this in Heroku because we're testing the Heroku timeout, but in reality we are replacing the Heroku 30-second timeout with a 25-second timeout built in to gunicorn 

I'll test it locally with gunicorn, @waxlamp does that sound good?

PR #1128, opened 2022-06-21
Add verbose logging to `ingest_zarr_archive` task

mvandenburgh:
This should help identify where the ingest_zarr_archive task is getting stuck.


AlmightyYakob (approved):
LGTM. Although this will have a perforation, I think it's okay in the short term to help us diagnose issues. We should definitely remove these logs after we've addressed these concerns. I'll file an issue

PR #1131, opened 2022-06-23
Add support for cloning assets to `import_dandisets` script

mvandenburgh:
This will help us test #1126 in staging.

### DIFF
--- a/dandiapi/api/management/commands/import_dandisets.py
+++ b/dandiapi/api/management/commands/import_dandisets.py
@@ -1,6 +1,7 @@
 from urllib.parse import urljoin
 from uuid import uuid4
 
+from django.conf import settings
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.db import transaction
 import djclick as click
@@ -22,14 +23,11 @@ def import_assets_from_response(api_url: str, asset_api_response: dict, version:
         zarr = result.get('zarr')
 
         if blob:
-            asset_metadata = requests.get(
-                urljoin(
-                    api_url,
-                    f'/api/dandisets/{version.dandiset.identifier}/versions/'
-                    f'{version.version}/assets/{result["asset_id"]}/',
-                ),
-                {'page_size': 1000},
-            ).json()
+            asset_metadata = {
+                'schemaVersion': settings.DANDI_SCHEMA_VERSION,
+                'encodingFormat': 'text/plain',
+                'schemaKey': 'Asset',
+            }
             asset = Asset.objects.create(
                 blob=asset_blob, metadata=asset_metadata, path=result['path']
             )
--- a/dandiapi/api/management/commands/import_dandisets.py
+++ b/dandiapi/api/management/commands/import_dandisets.py
@@ -1,14 +1,65 @@
 from urllib.parse import urljoin
+from uuid import uuid4
 
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.db import transaction
 import djclick as click
 import requests
 
-from dandiapi.api.models import Dandiset, Version
+from dandiapi.api.models import Asset, AssetBlob, Dandiset, Version, ZarrArchive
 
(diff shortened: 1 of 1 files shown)


dandiapi/api/management/commands/import_dandisets.py
```diff
@@ -1,14 +1,65 @@
...
+                status=zarr_info['status'],
+            )
+            asset = Asset.objects.create(zarr=zarr_archive, path=zarr_info['name'])
+            click.echo(f'    Importing ZarrArchive "{asset.path}"')
+            version.assets.add(asset)
+
+    # Handle API pagination
+    if asset_api_response.get('next'):
+        next_assets = requests.get(asset_api_response.get('next')).json()
+        import_assets_from_response(api_url, next_assets, version)
```
AlmightyYakob, line 56: If a version has lots of assets, this could recurse considerably, potentially leading to higher than desired memory usage. I'd think doing something like the following would be sufficient:

```python
results = asset_api_response['results']
while results:
    for result in results:
        ...

    # Pagination
    next_link = results.get('next')
    if not next_link:
        break

    results = requests.get(next_link).json()['results']
```

dandiapi/api/management/commands/import_dandisets.py
```diff
@@ -69,23 +133,33 @@ def import_dandiset_from_response(
...
+            include_assets=include_assets,
         )
 
     # Handle API pagination
     if dandiset_api_response.get('next'):
         new_dandisets = requests.get(dandiset_api_response.get('next')).json()
-        import_dandisets_from_response(api_url, new_dandisets, identifier_offset, should_replace)
+        import_dandisets_from_response(
+            api_url, new_dandisets, identifier_offset, should_replace, include_assets
+        )
```
AlmightyYakob, line 162: Same comment here regarding recursion (for the record happy to have a discussion about this, I just don't see a huge reason to use recursion here).

dandiapi/api/management/commands/import_dandisets.py
```diff
@@ -1,14 +1,65 @@
...
+
+        if blob:
+            asset_metadata = requests.get(
+                urljoin(
+                    api_url,
+                    f'/api/dandisets/{version.dandiset.identifier}/versions/'
+                    f'{version.version}/assets/{result["asset_id"]}/',
+                ),
+                {'page_size': 1000},
+            ).json()
```
AlmightyYakob, line 32: Is this level of fidelity required for these closed assets? I'd think the default populated metadata would be sufficient, since the underlying data doesn't match anyway. This would half the amount of requests made for non-zarr assets.

AlmightyYakob (changes requested):

dandiapi/api/management/commands/import_dandisets.py
mvandenburgh: For some reason, recursion felt more natural to me for this. But I think you're right that iteration is more practical here, especially consider we'll be running this on Heroku dynos with limited memory. I'll update it to use iteration

dandiapi/api/management/commands/import_dandisets.py
mvandenburgh: Good point, none of the metadata would be valid anyway :stuck_out_tongue:, I updated it to use the same minimal asset metadata blob for every asset.

mvandenburgh:
@AlmightyYakob I think it would be cleaner and keep this PR's scope smaller if I open a second PR on top of this one that converts all of the recursion to iteration instead of doing it here. Does that sound good to you?

AlmightyYakob:
> @AlmightyYakob I think it would be cleaner and keep this PR's scope smaller if I open a second PR on top of this one that converts all of the recursion to iteration instead of doing it here. Does that sound good to you?
Good point, sounds good to me.

AlmightyYakob (approved):

PR #1134, opened 2022-06-28
Optimize dandiset list endpoint

### LINKED ISSUE #1132: Optimize dandiset list endpoint
The main dandiset listing endpoint is taking an disproportional amount of time to complete (>3s on the main instance https://api.dandiarchive.org/api/dandisets/). We should look into this, I suspect we're making lots of redundant / unnecessary DB calls.

AlmightyYakob:
Closes #1132

Previously, this endpoint was making heavy use of model properties in the `Dandiset` and `Version` models, in order to populate various computed / annotated fields. This caused many repeated queries (~900 SQL queries for ~350 dandisets, so around 2.5 queries per dandiset).

I've consolidated this aggregation into both a larger SQL query, and some manual aggregation in python. After this change, there are only 3 SQL queries being made by this endpoint (out of the 5 or 6 total queries you might see in the Django Debug Toolbar).

Looking back, much of this was caused by rampant use of Django model properties. I think in the future we should discourage this pattern, as it leads to lots of repeated queries (for not all that much abstraction).

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -161,7 +161,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
         responses={200: DandisetListSerializer(many=True)},
     )
     def list(self, request, *args, **kwargs):
-        qs = self.get_queryset().annotate()
+        qs = self.get_queryset()
         dandisets = self.paginate_queryset(self.filter_queryset(qs))
 
         # Query versions, to supply num/count of assets.
@@ -187,7 +187,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
         # see all drafts first, then any published. Once we have seen both one draft
         # and one published, we're done for that dandiset.
         dandisets_to_versions = {}
-        for version in versions:
+        for version in versions.iterator():
             version: Version
 
             if version.dandiset_id not in dandisets_to_versions:
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -5,7 +5,7 @@ from dandischema.models import Dandiset as PydanticDandiset
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Count, OuterRef, Subquery, Sum
+from django.db.models import Count, OuterRef, Subquery, Sum, F
 from django.db.models.functions import Coalesce
 from django.db.utils import IntegrityError
 from django.http import Http404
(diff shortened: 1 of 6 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -154,6 +156,54 @@ def get_object(self):
...
+            .order_by('-version', '-modified')
+        )
+
+        # Determine draft and most recently published for each dandiset.
+        #
+        # Since ordering is version descending, then modified descending, we will
+        # see all drafts first, then any published. Once we have seen both one draft
+        # and one published, we're done for that dandiset.
+        dandisets_to_versions = {}
+        for version in versions:
```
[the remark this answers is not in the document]
AlmightyYakob: Both great suggestions, I'll implement them. 



> so we're looking at at least 1000 `Version` objects here, possibly many more since there is no limit to how many versions a dandiset can have.

Yeah this is an unfortunate limitation of the query. I'd realistically like to limit the amount of returned versions to one draft version and one published version (the most recent one) per dandiset. I originally wrote a query that did that, using `distinct` on the dandiset IDs, but annotation/aggregations don't work properly with that. I might add another SQL query to this, since one extra query would probably be worth not iterating over all versions associated with any of the dandisets here.

dandiapi/api/views/dandiset.py
```diff
@@ -154,6 +156,65 @@ def get_object(self):
...
+        )
+        published = (
+            base_query.exclude(version='draft')
+            .alias(latest=Subquery(latest_dandiset_version))
+            .filter(version=F('latest'))
+        )
+        drafts = base_query.filter(version='draft')
+
+        # Union with drafts
+        versions = published.union(drafts).order_by('dandiset_id', '-version')
```
AlmightyYakob, line 196: @mvandenburgh I think I figured out how to slim down the list of versions to just one draft version and one published version (if it exists) per dandiset. I wish I could make this useful to more places in the codebase than just here. If you can think of anywhere else this might apply please lmk.

dandiapi/api/views/serializers.py
```diff
@@ -86,6 +86,63 @@ class Meta:
...
+    class Meta:
+        model = Version
+        fields = [
+            'version',
+            'name',
+            'asset_count',
+            'size',
+            'status',
+            'created',
+            'modified',
```
mvandenburgh, line 101:
```suggestion
            'modified',
            'dandiset',
```

The current endpoint also returns the associated `Dandiset` nested within the `Version`.

mvandenburgh (changes requested):

dandiapi/api/views/serializers.py
AlmightyYakob: Including that in this endpoint is redundant, since the dandiset is at the top level anyway. I'd prefer to solve this in the CLI, since I don't see any reason to keep that in the listing endpoint.

AlmightyYakob: Oh my bad, those are e2e tests. In any case, is there a good reason to keep that redundant nesting?

mvandenburgh: Probably not. I think the "RESTful" way to do it would be to call the `dandiset` endpoint to fetch that data instead of packing it into the version endpoint

AlmightyYakob: It's not even a restful thing, my point is that, prior to this change, this was the structure of the response:

```
"results": [
    {
        "identifier": "000001",
		...(all dandiset info),
        "draft_version": {
            "version": "draft",
            "name": "test",
            ...,
            dandiset: {
                "identifier": "000001",
                ...(all info repeated again)
            }
        }
    }
]
```

So the dandiset information is needlessly included twice. However, I totally forgot to fix the use of it in the web app, so it's a good thing your pyppeteer CI caught it. If you're okay with removing the nested `dandiset` parameter, I'll make a fix to the front-end. We should also inform the collaborators, in case the CLI makes use of it.

mvandenburgh: Oh. Wow I didn't realize that info was literally repeated, yeah i agree that should be removed.

AlmightyYakob:
According to the CLI folks this won't cause any issues

mvandenburgh (approved):
LGTM

PR #1143, opened 2022-06-29
Add memray dev deps to setup.py

AlmightyYakob:
Memray seems to be a pretty useful tool for memory profiling, so this PR adds the relevant packages to the `dev` and `test` extra requires.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -103,6 +103,11 @@ Useful sub-commands include:
 To automatically reformat all code to comply with
 some (but not all) of the style checks, run `tox -e format`.
 
+### Profiling with Memray
+To include a memory profile with your tests, add `--memray` at the end of your test command invocation. For example, to run a memory profile with all tests, you would run `tox -e test -- --memray`. This can be used in conjuction with other pytest CLI flags (like `-k`) as well. See the `pytest-memray` [docs](https://github.com/bloomberg/pytest-memray) for more invocation details.
+
+#### NOTE: If you have an existing dandi-archive installation in which you have previously run tox, you may need to recreate the tox environment (by adding `-r` to your tox invocation) the first time you attempt to use memray. If your attempt to use the `--memray` flag fails with `pytest: error: unrecognized arguments: --memray`, this is likely why.
+
 ## dandiarchive.org WEB Interface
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -82,6 +83,7 @@
             'pytest-cov',
             'pytest-django',
             'pytest-factoryboy',
+            'pytest-memray',
```
mvandenburgh, line 86: If we're adding this, we could also consider adding `--memray` to the pytest invocation in `tox.ini`. WDYT @AlmightyYakob? If not, I think it's at least worth mentioning in the README that you can do `tox -e test -- --memray` since this plugin is installed.

AlmightyYakob:
> Tests are failing with an API endpoint returning an incorrect response... it looks like it is the user search endpoint, could it be related to the changes in #1124?

Yes I've noticed that, it seems to only happen in CI which is weird, not locally. I'll keep taking a look...

setup.py
AlmightyYakob: I considered that, I think the latter approach (documenting the option in the README) is the right approach. I'll add that.

mvandenburgh:
> > Tests are failing with an API endpoint returning an incorrect response... it looks like it is the user search endpoint, could it be related to the changes in #1124?
> 
> Yes I've noticed that, it seems to only happen in CI which is weird, not locally. I'll keep taking a look...

For consistency, I'll mention that this was fixed in #1146

mvandenburgh (approved):

DEVELOPMENT.md
```diff
@@ -103,6 +103,11 @@ Useful sub-commands include:
 To automatically reformat all code to comply with
 some (but not all) of the style checks, run `tox -e format`.
 
+### Profiling with Memray
+To include a memory profile with your tests, add `--memray` at the end of your test command invocation. For example, to run a memory profile with all tests, you would run `tox -e test -- --memray`. This can be used in conjuction with other pytest CLI flags (like `-k`) as well. See the `pytest-memray` [docs](https://github.com/bloomberg/pytest-memray) for more invocation details.
```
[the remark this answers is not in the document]
AlmightyYakob: Good catch, thanks.

PR #1146, opened 2022-06-30
Sort user search results

mvandenburgh:
#1124 removed the explicit ordering of user search results, which I believe is causing the flaky CI failures.

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -118,6 +118,7 @@ def users_search_view(request: Request) -> HttpResponseBase:
             # `extra_data` isn't an actual JSON field so we need to search this way
             | Q(socialaccount__extra_data__iregex=rf'"login": "[^"]*{re.escape(username)}[^"]*"')
         )
+        .order_by('date_joined')
     )[:10]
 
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
Good find! Before we merge this though, I want to understand:

1. Why this wasn't caught in CI when the PR was merged
2. Why this error doesn't show up when run locally

If we can codify the reason for this in a test that'd be ideal.

waxlamp:
Instead of removing the ordering from the search results, why can't we apply the same ordering to the ground truth in testing?

Keeping determinism in our search results seems like an important quality to maintain.

AlmightyYakob (approved):

waxlamp:
> > Instead of removing the ordering from the search results, why can't we apply the same ordering to the ground truth in testing?
> > Keeping determinism in our search results seems like an important quality to maintain.
> 
> The problem is the PR that caused this removed the ordering from the results, so this PR adds that ordering back in. Looking at my two questions now, it seems obvious that both are caused by the non-deterministic ordering so I don't think any more tests are needed.

Thanks, not sure how I missed that. The plus sign and the green background in your one-line diff should have been enough to clue me in 😝

PR #1152, opened 2022-06-30
Use `logging.info` for user errors instead of `logging.error`

### LINKED ISSUE #1151: Error while validating asset 35379
Sentry Issue: [DANDI-API-96](https://sentry.io/organizations/dandiarchive/issues/3389632183/?referrer=github_integration)

```
Error while validating asset 35379
```

This is happening due to the usage of `logging.error` to log the error message. However, I don't think this is an appropriate use of the ERROR log verbosity level, since it's not an application error, but rather a user error.

mvandenburgh:
Fixes #1151 

There's a couple places where logs indicating "errors" on the part of the client are logged as ERROR level, which are then reported as a sentry error. I think ERROR-level logging should be reserved for actual programming errors or abnormal system failures, not user "errors" like validation failures that are a normal part of the system. Any thoughts @AlmightyYakob @waxlamp?

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -227,7 +227,7 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
             except Exception:
                 # The assets summary aggregation may fail if any asset metadata is invalid.
                 # If so, just use the placeholder summary.
-                logger.error('Error calculating assetsSummary', exc_info=1)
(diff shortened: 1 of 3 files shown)


waxlamp:
Agreed.

waxlamp (approved):

AlmightyYakob (approved):
LGTM

PR #1158, opened 2022-07-05
Fix missing package error when first using docker-compose setup

AlmightyYakob:
~~Due to the following line~~

https://github.com/dandi/dandi-archive/blob/5d721709ad06041debbee93ca592ad470dd719ab/dandiapi/__init__.py#L8

~~The existing dev docker setup doesn't work as it's described in the `README.md`, as the `dandiapi` folder isn't copied into the container at the time of pip install, causing the `dandiapi` package to not be found. This wasn't an issue before, since we didn't rely on pip actually having knowledge of our `dandiapi` package to run django. However, due to our current versioning setup, any time our app is run, it runs the above line, to determine the current version. At that point, it finds that the `dandiapi` package isn't installed, causing an error.~~

~~We can't just copy the `dandiapi` directory into the container before pip install is run, since when we overwrite the current directory with our `docker-compose` setup, that folder will be gone, leaving us with the same problem. There's then two solutions:~~

~~1. Copy the `dandiapi` folder into the container at build time, before pip install. Then instead of mounting the entire directory with `docker-compose`, mount only the files/folders we need, so that the `.egg-info` folder isn't overwritten.~~
~~2. Add a step in the README to run pip install, which will then place a `.egg-info` folder in the cloned repo _outside_ of the container. After that, it will always get mounted into the container when docker-compose is run, and behave normally.~~

~~This PR takes the latter approach, since imo maintaining an exhaustive list of all files needed in the container could lead to all kinds of issues, which would likely not be caught since the dev setup isn't tested.~~

---

My initial report was a red herring, this PR has been updated to address the real cause, outlined in https://github.com/dandi/dandi-archive/pull/1158#issuecomment-1194641157.

### DIFF
--- a/dandiapi/__init__.py
+++ b/dandiapi/__init__.py
@@ -1,8 +1,8 @@
+from importlib.metadata import version
+
 # This project module is imported for us when Django starts. To ensure that Celery app is always
 # defined prior to any shared_task definitions (so those tasks will bind to the app), import
 # the Celery module here for side effects.
-from pkg_resources import get_distribution
-
 from .celery import app as _celery_app  # noqa: F401
 
-__version__ = get_distribution('dandiapi').version
+__version__ = version('dandiapi')
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -19,7 +19,10 @@ COPY ./setup.py /opt/django-project/setup.py
 # Copy git folder for setuptools_scm
 COPY ./.git/ /opt/django-project/.git/
 
-RUN pip install --editable /opt/django-project[dev]
+# Don't install as editable, so that when the directory is mounted over with docker-compose, the
+# installation still exists (otherwise the dandiapi.egg-info/ directory would be overwritten, and
+# the installation would no longer exist)
+RUN pip install /opt/django-project[dev]
 
 # Use a directory name which will never be an import name, as isort considers this as first-party.
 WORKDIR /opt/django-project


yarikoptic:
@jwodder please have a look if aligns with your understanding of the moving pieces here

AlmightyYakob:
> @AlmightyYakob Instead of using `pkg_resources.get_distribution()`, wouldn't you solve the problem by using `importlib.metadata.version()` (or `importlib-metadata` on pre-Python 3.8)? Aside from the fact that `pkg_resources` is deprecated, `importlib.metadata` operates on distribution packages rather than import packages, so it shouldn't need the code to be present at compile time.

Interesting, I wasn't aware of this. I can try that out, and if that solves our issues more elegantly, I'll close this PR in favor of that one.

AlmightyYakob:
> @AlmightyYakob Instead of using `pkg_resources.get_distribution()`, wouldn't you solve the problem by using `importlib.metadata.version()` (or `importlib-metadata` on pre-Python 3.8)? Aside from the fact that `pkg_resources` is deprecated, `importlib.metadata` operates on distribution packages rather than import packages, so it shouldn't need the code to be present at compile time.

What's your source for this behavior? In any test that I run, I need to install `dandiapi` prior, like with `get_distribution`. Otherwise, I'll get

```
importlib.metadata.PackageNotFoundError: dandiapi
```

jwodder:
@AlmightyYakob Yes, dandiapi needs to be installed no matter what, but the code shouldn't need to be there at install time.  I interpreted this line from the OP:

> as the `dandiapi` folder isn't copied into the container at the time of pip install, causing the `dandiapi` package to not be found

as referring to [this folder](https://github.com/dandi/dandi-archive/tree/master/dandiapi) and the `dandiapi` [import package](https://packaging.python.org/en/latest/glossary/#term-Import-Package), not the [distribution package](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package).  Is that note the case?  My reading of [`django.Dockerfile`](https://github.com/dandi/dandi-archive/blob/master/dev/django.Dockerfile) indicates that the distribution package but not the import package is present when `pip install` runs in the Dockerfile, so the distribution package metadata should have access to the version number from version control.  Could you clarify exactly what's happening with Docker and packaging and what the problem you're trying to solve is?

AlmightyYakob:
> My reading of [`django.Dockerfile`](https://github.com/dandi/dandi-archive/blob/master/dev/django.Dockerfile?rgh-link-date=2022-07-25T20%3A04%3A15Z) indicates that the distribution package but not the import package is present when `pip install` runs in the Dockerfile, so the distribution package metadata should have access to the version number from version control. Could you clarify exactly what's happening with Docker and packaging and what the problem you're trying to solve is?

It seems I was a bit confused initially on this. You're correct in that when we run `pip install` in the container, the distribution package is available, and the version can be read in the final container (I've just checked that). The reason for the initial error is that since we do an editable install of `dandiapi`, when we mount the entire local directory with `docker-compose`, the local `dandiapi.egg-info/` folder is overwritten (as is anything else in the directory). So at that point, there is no `dandiapi` distribution package installed, since the only record of that was kept in the `dandiapi.egg-info` folder.

I've tested just installing the package locally without `--editable`, and it seems to solve this issue, while still allowing for hot reloading. So I'll update this PR to reflect those changes. Your point about importlib is still valid though, so I'll probably incorporate that change as well.

mvandenburgh (approved):
The only potential problem I see here is anyone using docker to run things would lose the ability to easily make code changes to installed libraries (mainly thinking of our use of `dandischema` here), since we're doing non-editable mode pip install. I suppose we could just say, if you want the ability to do that you must run the python code natively, outside of docker. Either way, it's not something I've had to do often enough to be concerned about.

AlmightyYakob:
> The only potential problem I see here is anyone using docker to run things would lose the ability to easily make code changes to installed libraries (mainly thinking of our use of `dandischema` here), since we're doing non-editable mode pip install. I suppose we could just say, if you want the ability to do that you must run the python code natively, outside of docker. Either way, it's not something I've had to do often enough to be concerned about.

This was already the case though, wasn't it? Any third party libraries aren't present on the local machine itself, they're installed in the docker container's site packages. How does this change modify that setup?

mvandenburgh:
> > The only potential problem I see here is anyone using docker to run things would lose the ability to easily make code changes to installed libraries (mainly thinking of our use of `dandischema` here), since we're doing non-editable mode pip install. I suppose we could just say, if you want the ability to do that you must run the python code natively, outside of docker. Either way, it's not something I've had to do often enough to be concerned about.
> 
> This was already the case though, wasn't it? Any third party libraries aren't present on the local machine itself, they're installed in the docker container's site packages. How does this change modify that setup?

Looking back, I'm not sure what I was getting at here - I was referring to the case where you might do an editable install of your local `dandischema` repo, so that you could make changes to the code live, but this change doesn't prevent one from doing that. I think this change is fine :+1:

PR #1159, opened 2022-07-05
Fix excessive memory usage in asset summary calculation

### LINKED ISSUE #1149: Fix excessive memory usage during asset summary calculation
This line https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py#L224-L226 loads metadata of every Asset into memory as a dict using .values, and then coerces it into a list. This means that we're essentially storing two copies of the asset metadata in memory for **every Asset** in the Version being validated. So for dandisets like https://dandiarchive.org/dandiset/000026, it appears that we're loading 80,000 JSON blobs into memory at once.

I'll run asset summary calculation for a large dandiset like that locally with memray and update this issue with the results.

mvandenburgh:
This line [https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py#L224-L226](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py?rgh-link-date=2022-06-30T20%3A20%3A35Z#L224-L226) loads metadata of every Asset into memory as a dict using .values, and then coerces it into a list. This means that we're essentially storing two copies of the asset metadata in memory for every Asset in the Version being validated. So for dandisets like [dandiarchive.org/dandiset/000026](https://dandiarchive.org/dandiset/000026), it appears that we're loading 80,000 JSON blobs into memory at once.

This PR changes this line to 

1) Use `QuerySet.values_list(..., flat=True)` to convert the queryset directly into a list of asset metadata JSON blobs, instead using .values and then recreating the list from scratch.
2) Use `QuerySet.iterator()` to produce a generator from the previous `values_list` queryset to avoid loading every asset metadata into memory at once. Note that the dandi-schema `aggregate_assets_summary` function takes any `Iterable` as an argument, so it's okay to pass a generator instead of a list here.

Fixes #1149

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -203,7 +203,7 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         ]
         return {key: metadata[key] for key in metadata if key not in computed_fields}
 
-    def _populate_metadata(self, version_with_assets: Version = None):
+    def _populate_metadata(self, version_with_assets: Version | None = None):
 
         # When validating a draft version, we create a published version without saving it,
         # calculate it's metadata, and validate that metadata. However, assetsSummary is computed
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime
 import logging
+from typing import TYPE_CHECKING
 
 from dandischema.metadata import aggregate_assets_summary
 from django.conf import settings
@@ -13,6 +14,9 @@ from django_extensions.db.models import TimeStampedModel
 from dandiapi.api.models.metadata import PublishableMetadataMixin
 
 from .dandiset import Dandiset
+
+if TYPE_CHECKING:
+    from .asset import Asset
 
 logger = logging.getLogger(__name__)
 
@@ -221,8 +225,12 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         }
         if version_with_assets.id:
             try:
+                assets: models.QuerySet[Asset] = version_with_assets.assets
                 summary = aggregate_assets_summary(
-                    [asset['metadata'] for asset in version_with_assets.assets.values('metadata')]
+                    # There is no limit to how many assets a dandiset can have, so use
+                    # `values_list` and `iterator` here to keep the memory footprint
+                    # of this list low.
+                    assets.values_list('metadata', flat=True).iterator()
                 )
             except Exception:
                 # The assets summary aggregation may fail if any asset metadata is invalid.


mvandenburgh:
I used memray to profile this function call on a dandiset with 37,000 assets - without this PR, it took 276 MB of memory; with it, it took 16 MB of memory.

dandiapi/api/models/version.py
```diff
@@ -203,7 +207,7 @@ def strip_metadata(cls, metadata):
         ]
         return {key: metadata[key] for key in metadata if key not in computed_fields}
 
-    def _populate_metadata(self, version_with_assets: Version = None):
+    def _populate_metadata(self, version_with_assets: Version | None = None):
```
yarikoptic, line 210: Note: ha, I was expecting `Optional[Version]` but apparently it isn't the convention in dandi-archive:

```shell
(git)lena:~/proj/dandi/dandi-archive[master]git
$> git grep 'Optional\['
(dev3) 1 34993 ->1.....................................:Tue 05 Jul 2022 03:51:59 PM EDT:.
(git)lena:~/proj/dandi/dandi-archive[master]git
$> git grep '| None'    
dandiapi/api/copy.py:    def copy_range(self) -> str | None:
dandiapi/api/mail.py:def build_message(subject: str, message: str, to: list[str], html_message: str | None = None):
dandiapi/api/models/dandiset.py:    def identifier(self) -> str | None:
dandiapi/api/models/upload.py:    def object_key(upload_id, dandiset: Dandiset | None = None):  # noqa: N805
dandiapi/api/models/upload.py:    def object_key(upload_id, dandiset: Dandiset | None = None):
dandiapi/api/models/zarr.py:    def checksum(self) -> str | None:
dandiapi/api/tasks/zarr.py:def yield_files(bucket: str, prefix: str | None = None):
dandiapi/api/tasks/zarr.py:    def read_checksum_file(self) -> ZarrChecksumListing | None:
dandiapi/api/views/asset.py:        glob_pattern: str | None = serializer.validated_data.get('glob')
dandiapi/api/zarr_checksums.py:    def read_checksum_file(self) -> ZarrChecksumListing | None:
```
although according to  https://docs.python.org/3/library/typing.html#typing.Optional  it is only a recent "addition"

```
Changed in version 3.10: Optional can now be written as X | None. 
```

dandiapi/api/models/version.py
```diff
@@ -221,8 +225,12 @@ def _populate_metadata(self, version_with_assets: Version = None):
         }
         if version_with_assets.id:
             try:
+                assets: models.QuerySet[Asset] = version_with_assets.assets
                 summary = aggregate_assets_summary(
-                    [asset['metadata'] for asset in version_with_assets.assets.values('metadata')]
+                    # There is no limit to how many assets a dandiset can have, so use
+                    # `values_list` and `iterator` here to keep the memory footprint
+                    # of this list low.
+                    assets.values_list('metadata', flat=True).iterator()
```
yarikoptic, line 233: and having Python generator instead of `list`, i.e.

```suggestion
                    (asset['metadata'] for asset in version_with_assets.assets.values('metadata'))
```

would not be *as good*? (although I already like seeing a single `'metadata'` just want to make sure I understand what is happening)

dandiapi/api/models/version.py
AlmightyYakob: You're right, since we only run 3.9, this wouldn't usually work, but we have the following line at the top of the file

```
```

which allows that syntax.

AlmightyYakob (approved):
LGTM

dandiapi/api/models/version.py
mvandenburgh: @yarikoptic Using `version_with_assets.assets.values('metadata')` caused a `list` of the following format to be loaded all at once into memory:
```
[
	{
		'metadata': { ..... }
	},
	{
		'metadata': { ..... }
	},
	{
		'metadata': { ..... }
	},
 ....
]
```
So, the list comprehension was needed to flatten the list into the correct format for the dandi-schema `aggregate_assets_summary` function. I've changed this instead to use `assets.values_list('metadata', flat=True)`, which accomplishes flattening the list via SQL without needing the additional list comprehension.

PR #1184, opened 2022-07-12
Improve manifest file writing performance

### LINKED ISSUE #985: Improve memory performance of write_manifest_files task
https://github.com/dandi/dandi-archive/pull/977 fixed the initial bugs with the `write_manifest_files` task consuming too much memory when running on a dandiset with more than 40,000 assets by giving the task a dedicated worker with a lower concurrency. Some additional steps that can be taken to lower the memory usage:

* Use https://docs.djangoproject.com/en/4.0/ref/models/querysets/#iterator to limit django's model caching.
* Test if `YAMLRenderer` will accept an iterator instead of a list to avoid allocating the entire list at once.
* Subclass `YAMLRenderer` or drop down to the python `yaml` module to stream the data directly to S3, or stream to disk and then stream to S3. This avoids ever needing to have the entire file in memory at once.

danlamanna:
This significantly speeds up and reduces the memory usage of the write_manifest_files task that takes place on publish.

The manifest files are now written in a streaming fashion, and the YAML output now uses the C implementation. Worst case time/memory usage went from ~2m20s and ~1gb ram to ~20s and ~25mb (constant time) memory usage.

As a result, I'm removing the manifest-worker since this task doesn't need to be treated differently than any others on the main worker.

Fixes #985

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -9,5 +9,3 @@ web: gunicorn --bind 0.0.0.0:$PORT dandiapi.wsgi --timeout 25
 worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q celery -B
 # The checksum-worker calculates blob checksums and updates zarr checksum files
 checksum-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q calculate_sha256,ingest_zarr_archive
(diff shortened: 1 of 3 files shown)


mvandenburgh:
@danlamanna you'll also need to remove the resource from Terraform (https://github.com/dandi/dandi-infrastructure/blob/master/terraform/api.tf#L56-L61) after this PR is released.

PR #1187, opened 2022-07-12
Add time limits to celery tasks

danlamanna:
This adds (relaxed) time limits to the tasks where it makes sense. This will help to prevent "runaway tasks" from potentially blocking the queue.


mvandenburgh (approved):
LGTM

PR #1190, opened 2022-07-13
Add "/server-info" frontend endpoint in development builds

### LINKED ISSUE #1108: Q: provide /server-info "natively" (or redirected to) by web service?
ATM it is handled via redirect on netlify level

https://github.com/dandi/dandi-archive/blob/12e1c5e146bf046b68e933c4a8e628afc4a3e0a0/web/netlify/plugins/server-info-build/index.js#L10

so it is not available in docker-compose'd instance of the archive, thus no other but the netlify instances would have it.  But dandi-cli uses `/server-info` to identify the actual endpoints for e.g. API service etc, so that is what it accesses whenever given some URL.  So it would be nice to have that endpoint also for docker-compose instances.  We have been relying on redirector before, so was tested and now we need to adjust the test https://github.com/dandi/dandi-cli/issues/977 accordingly but there is no such endpoint in docker-compose setup.

waxlamp:
This PR adds an "endpoint" to the frontend that proxies to the backend server info endpoint, emulating the mechanism in place on the Netlify-controlled deploys.

This works but only if you ask for the path complete with trailing slash, and I'm not sure why. Reviewers, perhaps you can recommend a better way to make this work so that the trailing slash is not required.

Closes #1108.

### DIFF
--- a/web/vue.config.js
+++ b/web/vue.config.js
@@ -34,7 +34,19 @@ module.exports = {
   devServer: {
     // The default port 8080 conflicts with Girder
     port: 8085,
+
+    // This proxy emulates the Netlify redirect to the backend (on the staging
+    // and production deploys).
+    proxy: {
+      '^/server-info/$': {
+        target: process.env.VUE_APP_DANDI_API_ROOT,
+        pathRewrite: {
+          '/server-info': '/info',
+        },
+      },
+    },
   },
+
   chainWebpack: (config) => {
     config
       .plugin('moment-locales')


waxlamp:
> I think a better solution would be to proxy `/api`, making it a standard that a `$gui_url/api` is the location for retrieving information.

I agree, and I'll file an issue to make this the standard way the API is proxied through to the front end, but if we can merge this PR, it will provide the CLI team immediate value, and then we can get to the fuller solution afterwards.

yarikoptic:
confirming that it works as intended:

```
$> curl http://localhost:8085/server-info/
{"schema_version":"0.6.3","schema_url":"https://raw.githubusercontent.com/dandi/schema/master/releases/0.6.3/dandiset.json","version":"0.2.39.post19+g9b92512f","services":{"api":{"url":"http://localhost:8000/api"},"webui":{"url":"http://localhost:8085"},"jupyterhub":{"url":"https://hub.dandiarchive.org/"}},"cli-minimal-version":"0.14.2","cli-bad-versions":[]}
```
(whenever on master it gives html page about JS needed etc).

yarikoptic (approved):

AlmightyYakob (approved):

AlmightyYakob (approved):

PR #1191, opened 2022-07-13
Escape . in regex for EXTERNAL_SERVICES

yarikoptic:
but may be JS has somewhat different syntax for regexes? dunno how we would have missed it in original review, so may be I am just wrong.

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -244,14 +244,14 @@ const sortByName = (a: AssetStats, b: AssetStats) => {
 const EXTERNAL_SERVICES = [
   {
     name: 'Bioimagesuite/Viewer',
-    regex: '.nii(.gz)?$',
+    regex: '\.nii(\.gz)?$',
     maxsize: 1e9,
     endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=',
   },
 
   {
     name: 'MetaCell/NWBExplorer',
-    regex: '.nwb$',
+    regex: '\.nwb$',
     maxsize: 1e9,
     endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
   },
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -244,14 +244,14 @@ const sortByName = (a: AssetStats, b: AssetStats) => {
 const EXTERNAL_SERVICES = [
   {
     name: 'Bioimagesuite/Viewer',
-    regex: '\.nii(\.gz)?$',
+    regex: /\.nii(\.gz)?$/,
     maxsize: 1e9,
     endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=',
   },
 
   {
     name: 'MetaCell/NWBExplorer',
-    regex: '\.nwb$',
+    regex: /\.nwb$/,
     maxsize: 1e9,
     endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
   },


waxlamp:
@yarikoptic is this PR fixing a specific bug you ran into?

mvandenburgh:
Good catch! I did some more testing to figure out the exact behavior here -

```node
> r = new RegExp('.nwb$').test('test.nwb')
true
> r = new RegExp('.nwb$').test('testnwb')
true
> r = new RegExp('.nwb$').test('testnmwb')
false
```
It seems that it is indeed interpreting the `.` as the special regex character and not a literal. 

Testing out your proposed fix -
```
> r = new RegExp('\.nwb$').test('test.nwb')
true
> r = new RegExp('\.nwb$').test('testnwb')
true
```

It doesn't work. Since the regex is a regular JS string, the backslash itself must be escaped so that the `RegExp` can parse it:
```
> r = new RegExp('\\.nwb$').test('testnwb')
false
> r = new RegExp('\\.nwb$').test('test.nwb')
true
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -244,14 +244,14 @@ const sortByName = (a: AssetStats, b: AssetStats) => {
 const EXTERNAL_SERVICES = [
   {
     name: 'Bioimagesuite/Viewer',
-    regex: '.nii(.gz)?$',
+    regex: '\.nii(\.gz)?$',
```
mvandenburgh, line 247:
```suggestion
    regex: '\\.nii(\\.gz)?$',
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -244,14 +244,14 @@ const sortByName = (a: AssetStats, b: AssetStats) => {
...
     maxsize: 1e9,
     endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=',
   },
 
   {
     name: 'MetaCell/NWBExplorer',
-    regex: '.nwb$',
+    regex: '\.nwb$',
```
mvandenburgh, line 254:
```suggestion
    regex: '\\.nwb$',
```

mvandenburgh (changes requested):

web/src/views/FileBrowserView/FileBrowser.vue
mvandenburgh, line 247:
```suggestion
    regex: /\.nii(\.gz)?$/,
```

web/src/views/FileBrowserView/FileBrowser.vue
mvandenburgh, line 254:
```suggestion
    regex: /\.nwb$/,
```

mvandenburgh (changes requested):
@yarikoptic after our discussion earlier today I looked back into this - `new RegExp(...)` is only needed if defining regex patterns with strings. If you use a JS _pattern_ like I did in the suggestions below, you don't need the extra escape. I think that syntax is cleaner, let me know what you think.

web/src/views/FileBrowserView/FileBrowser.vue
yarikoptic: I have no clue in JS, so if you are saying this specification is somehow better - I would just trust you. But how or why is `/.../` form better than `'...'` -- does it compile it into regex or smth like that?

mvandenburgh: `/.../` only works for literal values, whereas `new RegExp(...)` allows for dynamic values since it accepts regular strings. `/.../` is more performant, since it doesn't have the overhead of "converting" the string into a regex. There's more info here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#creating_a_regular_expression

mvandenburgh (approved):
Looks like this works now. The incorrect behavior can be observed on staging https://gui-staging.dandiarchive.org/dandiset/202400/draft/files?location=, while the correct behavior is now happening for that dandiset on the deploy preview for this branch https://deploy-preview-1191--gui-staging-dandiarchive-org.netlify.app/dandiset/202400/draft/files?location=

yarikoptic:
Thanks @mvandenburgh  for the confirmation!  Since I was the author I guess it is my duty to have it merged! Clicking the green button!

Issue #1117, opened 2022-06-11
getting timeouts, 500s, and good number of errors logged -- need help with analysis/triage


yarikoptic:
While backing up dandisets on drogon, 

<details>
<summary>we keep running into various 500s, timeouts etc</summary> 

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ PATH=/home/dandi/git-annexes/10.20220525+git57-ge796080f3-1~ndall+1/usr/lib/git-annex.linux:$PATH  python -m tools.backups2datalad --pdb -l WARNING -J 5 --target /mnt/backup/dandi/dandisets update-from-backup --zarr-target /mnt/backup/dandi/dandizarrs --backup-remote dandi-dandisets-dropbox --zarr-backup-remote dandi-dandizarrs-dropbox --gh-org dandisets --zarr-gh-org dandizarrs 000108
A newer version (0.40.1) of dandi/dandi-cli is available. You are using 0.40.0
2022-06-10T10:46:04-0400 [WARNING ] backups2datalad Retrying GET request to https://api.dandiarchive.org/api/assets/6164186b-16c2-4bc0-9036-6804f5934019/ in 1.000000 seconds as it raised ReadTimeout:
2022-06-10T10:46:36-0400 [WARNING ] backups2datalad Retrying GET request to https://api.dandiarchive.org/api/assets/6164186b-16c2-4bc0-9036-6804f5934019/ in 1.800000 seconds as it raised ConnectTimeout:
2022-06-10T10:47:04-0400 [WARNING ] backups2datalad Retrying GET request to https://api.dandiarchive.org/api/assets/6164186b-16c2-4bc0-9036-6804f5934019/ in 3.240000 seconds as it raised ConnectTimeout:
2022-06-10T10:47:36-0400 [WARNING ] backups2datalad Retrying GET request to https://api.dandiarchive.org/api/assets/6164186b-16c2-4bc0-9036-6804f5934019/ in 5.832000 seconds as it raised ConnectTimeout:
2022-06-10T10:48:27-0400 [WARNING ] backups2datalad Retrying GET request to https://api.dandiarchive.org/api/assets/6164186b-16c2-4bc0-9036-6804f5934019/ in 10.497600 seconds as it raised ConnectTimeout:
[...]

PR #1199, opened 2022-07-18
Require metadata/asset schemaVersion

### LINKED ISSUE #1193: Provided metadata has no schema version
for now I think it is ok to assume that `ValueError` raised while calling a validation function is due to validation failure.  I think it is needed to investigate how `ValueError: Provided metadata has no schema version` could come about (check the logs/traceback there to see how it got there) since AFAIK dandi-cli shouldn't provide such records, so might be somewhere on web frontend.

_Originally posted by @yarikoptic in https://github.com/dandi/dandi-archive/issues/1117#issuecomment-1183554239_

danlamanna:
Fixes #1193 and #1117. 

This makes schemaVersion required with a database constraint and patches up the primary place we weren't defaulting it to `settings.DANDI_SCHEMA_VERSION`. I also merged two tests that were testing nearly identical things but had already started to drift apart. I'd like to review this with @mvandenburgh before merging.


mvandenburgh:
@danlamanna #1212 fixed some issues with the CLI integration tests. They seem to be different than the ones you're encountering here, but it might be worth rebasing on `master` to see if anything changes

dandiapi/api/migrations/0031_asset_asset_metadata_has_schema_version_and_more.py
```diff
@@ -0,0 +1,38 @@
...
+from django.db import migrations, models, transaction
+
+
+def populate_empty_asset_schema_versions(apps, _):
+    Asset = apps.get_model('api', 'Asset')
+
+    with transaction.atomic():
+        for asset in Asset.objects.filter(metadata__schemaVersion__isnull=True).iterator():
+            asset.metadata.setdefault('schemaVersion', settings.DANDI_SCHEMA_VERSION)
+            asset.save(update_fields=['metadata'])
```
AlmightyYakob, line 13: I don't see this being done with version metadata anywhere, shouldn't that also be added?

mvandenburgh: The API always populates the schemaVersion when a dandiset is created (https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/views/dandiset.py#L246), so it's less likely to be missing a schemaVersion (the only scenario I believe, is if someone did a PUT to an existing version's metadata with it missing).

I double checked, and `Version.objects.filter(metadata__schemaVersion__isnull=True).exists()` returns false on both production and staging.

mvandenburgh (commented):
LGTM - although we'll need to figure out why the CLI tests are failing before merging.

mvandenburgh (approved):

dandiapi/api/management/commands/migrate_published_version_metadata.py
```diff
@@ -21,10 +21,6 @@ def migrate_published_version_metadata(dandiset: str, published_version: str, to
     )
     metadata = version.metadata
 
-    # If there is no schemaVersion, assume to_version
-    if 'schemaVersion' not in metadata:
-        metadata['schemaVersion'] = to_version
-
```
satra, line 27: is this being removed for a specific reason? or because we no longer have such a possibility?

AlmightyYakob: The latter, since attempting to exclude `schemaVersion` from the metadata would fail at the database level.

Issue #1111, opened 2022-06-01
Archive allows asset paths that are empty, non-normalized, or absolute


jwodder:
The archive does not prohibit creating assets with the following types of paths:

- empty paths
- non-normalized paths (i.e., containing `.` or `..` as a path component or containing multiple consecutive slashes)
- absolute paths (i.e., paths that start with a slash)

Additionally, trying to create an asset with a path that contains a NUL character leads to a 500 error.

All of these types of asset paths are problematic and should be forbidden.

MCVE:

```python
import os
from pathlib import Path
from tempfile import TemporaryDirectory

from dandi.consts import dandiset_metadata_file
from dandi.dandiapi import DandiAPIClient
from dandi.files import dandi_file

with DandiAPIClient.for_dandi_instance(
    "dandi-staging", token=os.environ["DANDI_API_KEY"]
) as client:

    d = client.create_dandiset(
        "Dandiset with non-normalized file paths",
        {
            "schemaKey": "Dandiset",
            "name": "Test Dandiset",
            "description": "A test Dandiset",
            "contributor": [
                {
                    "schemaKey": "Person",
                    "name": "Wodder, John",
                    "roleName": ["dcite:Author", "dcite:ContactPerson"],
                }
            ],
            "license": ["spdx:CC0-1.0"],
        },
    )

    dandiset_id = d.identifier
    print("DANDISET ID:", dandiset_id)
    try:
        with TemporaryDirectory() as tmpdir:
            dspath = Path(tmpdir)
            (dspath / dandiset_metadata_file).write_text(f"identifier: '{dandiset_id}'\n")
            p = dspath / "file.txt"
            p.write_text("This is test text.\n")
            df = dandi_file(p, dandiset_path=dspath)
            for path in [
                "",
                "./foo",
                "../foo",
                "foo/.",
                "foo/..",
                "foo/./bar",
                "foo/../bar",
                "/foo",
                "/",
                "foo//bar",
                # This one causes an internal server error:
                #"foo\0bar",
            ]:
                print("Uploading", path, "...")
                df.upload(d, {"path": path, "description": "A file"})

            print("Assets in Dandiset:")
            for asset in d.get_assets():
                print(f" - {asset.path!r}")
    finally:
        print("Deleting Dandiset ...")
        d.delete()
```

waxlamp:
I will take it as given that we all agree with John that these are "problematic and should be forbidden".

[...]

PR #1201, opened 2022-07-19
Handle forbidden characters in asset path

DeepikaGhodki:
Fixes [1111](https://github.com/dandi/dandi-archive/issues/1111)

**Current path validations for an asset are as follows:**

1. Must not be empty or null
2. Must not be an absolute path (start with /)
3. Must not be a directory path (./ or ../) 
4. Must not contain only `.` in sub paths
5. Must not contain `\0`
6. Must not contain repeating `/`

This PR contains validations for checking the above using regular expressions for character matching. 

**TODO** 

- [x] Fix affected test cases using absolute asset path


dandiapi/api/tests/test_asset.py
```diff
@@ -538,6 +538,51 @@ def test_asset_create(api_client, user, draft_version, asset_blob):
...
+        f'/api/dandisets/{draft_version.dandiset.identifier}'
+        f'/versions/{draft_version.version}/assets/',
+        {'metadata': metadata, 'blob_id': asset_blob.blob_id},
+        format='json',
+    )
+
+    if valid:
+        assert resp.status_code == 200, resp.data
+    else:
+        assert resp.status_code != 200, resp.data
```
mvandenburgh, line 583: I think making the status code itself a parameter instead of `valid`, and then just turning this line into
```python
assert resp.status_code == expected_status_code, resp.data
```
would be better, because 
1) it strengthens the test by testing for an explicit status code. The endpoint could return a 500 error, and this test would still pass as it is now.
2) It decreases the amount of code paths through the test. Tests should be simple and ideally not have any complex logic that changes their behavior depending on a parameter. If you find yourself writing conditionals like this in tests, I find that it usually means a) the test needs to be parametrized differently (which is the case here), or b) the test needs to be broken up into multiple smaller-scoped tests.

DeepikaGhodki: Thank you for pointing this out, this makes a lot of sense.

dandiapi/api/views/asset.py
```diff
@@ -409,6 +434,7 @@ def create(self, request, versions__dandiset__pk, versions__version):
         _maybe_validate_asset_metadata(asset)
 
         serializer = AssetDetailSerializer(instance=asset)
+        print(serializer.data)
```
danlamanna, line 437:
```suggestion
```

danlamanna (changes requested):

dandiapi/api/views/asset.py
```diff
@@ -251,9 +252,33 @@ def validate(self, data):
                 {'blob_id': 'Exactly one of blob_id or zarr_id must be specified.'}
             )
 
-        if 'path' not in data['metadata']:
+        if 'path' not in data['metadata'] or not data['metadata']['path']:
```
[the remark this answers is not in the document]
DeepikaGhodki: This changes the error response structure and breaks an existing test. So, leaving it as is for now.

dandiapi/api/views/asset.py
```diff
@@ -251,9 +252,33 @@ def validate(self, data):
...
             raise serializers.ValidationError({'metadata': 'No path specified in metadata.'})
 
+        path = data['metadata']['path']
```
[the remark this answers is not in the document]
DeepikaGhodki: Need the path string for other checks, moving `PurePosixPath` to `is_absolute()` directly

danlamanna (approved):

PR #1208, opened 2022-07-22
De-Duplicate API calls in DLP

### LINKED ISSUE #1166: Remove duplicate DLP API calls
Here's a screen grab of the network panel when loading the Dandiset Landing Page. Notice all of the duplicate calls

image

These should be investigated and removed.

DeepikaGhodki:
Resolves #1166

- Remove watchEffect
- ~~Set version count while fetching all versions instead of getting it from server~~
- ~~A new state called `versionCount` is added to facilitate this~~
- Fetch the latest version (instead of fetching by count)


web/src/rest.ts
```diff
@@ -161,8 +162,7 @@ const dandiRest = new Vue({
       }
     },
     async mostRecentVersion(identifier: string) {
-      // TODO: find a way to do this with fewer requests
-      const count = (await this.versions(identifier))?.count;
+      const count = store.state.dandiset.versionCount; // todo - this.$store.state is undefined?
```
DeepikaGhodki, line 165: Unable to resolve cyclic dependency between store.js and rest.js. Alternate approach tried:
```
computed: {
    versionCount(): Number {
      return this.$store?.direct.state.dandiset.versionCount;
    },
}
```

```
async mostRecentVersion(identifier: string) {
      const count = this.versionCount;
```

AlmightyYakob: I think we should just avoid using the store here in the first place. We can just pass `versionCount` as a parameter to this function.

web/src/store/dandiset.ts
```diff
@@ -84,24 +89,27 @@ const dandisetModule = defineModule({
...
       const { dispatch } = dandisetActionContext(context);
       await dispatch.uninitializeDandisets();
 
-      // this can be done concurrently, don't await
-      dispatch.fetchDandisetVersions({ identifier });
+      // first fetch all versions, set total count
+      // use count to fetch latest dandiset, must be sequential
+      await dispatch.fetchDandisetVersions({ identifier });
       await dispatch.fetchDandiset({ identifier, version });
       await dispatch.fetchOwners(identifier);
```
AlmightyYakob, line 102: I realize this would involve a backend change, but I think the way to properly handle this would be to add a sort parameter to the version list endpoint, which would allow us to sort versions by most recent, and just choose the first one from there (removing the need to call `mostRecentVersion` from within `fetchDandiset`. We could then store that in a variable in the store as well (named something like `mostRecentVersion`.)

I think it would be within scope of this PR to add that, but if you would rather just get this merged we can punt on that and file an issue.

AlmightyYakob (changes requested):

web/src/store/dandiset.ts
DeepikaGhodki: I agree, sorting would be a more straightforward way of getting the most recent version. I don't mind continuing the work here itself. Meanwhile, I will move this to draft.

web/src/rest.ts
DeepikaGhodki: What is the benefit of passing a parameter over using a store? In other words, does using a store create an issue?

AlmightyYakob: I don't think we should be using the store in this file at all. Currently there's a separation of the store and the API wrapper, and I think to reduce complexity we should keep it that way. Everything that's currently being done (tracking `versionCount` in the store) can still be done, but we can simply just pass that value as an argument to `mostRecentVersion`, instead of importing the store and then reading it.

If this PR is updated to add sorting by most recent, then I suspect the use of `mostRecentVersion` will become unnecessary anyway, since we can just grab it from the response and save it directly in the store.

DeepikaGhodki: Got it!

AlmightyYakob (approved):

PR #1216, opened 2022-07-26
Add metadata flag for asset listing

### LINKED ISSUE #397: /assets and /paths -- add 'include=["metadata", "validation"]' option
Generalization of the #100 , inspired by https://github.com/dandi/dandi-cli/issues/699 , and somewhat of an "extension" to https://github.com/dandi/dandi-api/pull/312 . 
If we (user, web-ui, dandi-cli) could optionally request extra information from per-asset end-points, this would 
- avoid necessity to bombard API server with per-asset requests (e.g. for metadata and/or `/validation`).  We can easily implement that in client (as we already do for `metadata`) but IMHO it should be done within API to actually reduce load on the server and make it a better UX since would not require thousands of requests to be sent to the API server
- make it possible for web UI to display that information without incurring significant cost/delay as what #312 is trying to avoid for one particular aspect -- download URL.
- both `/assets` and `/paths` are very similar in this respect, only that `/paths` would provide an "aggregate" across underlying a "directory" assets.  It would be IMHO a really nice UX to be able to see that there are problems with assets under a given "directory" and what they are.

### LINKED ISSUE #1217: Optionally include asset metadata in paginated asset listing

AlmightyYakob:
Closes #1217.

Addresses in part #397.

This specifically addresses the metadata flag for the asset list endpoint. If we want to address other aspects of that issue, they should be filed as their own respective issues.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -533,13 +533,14 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
             glob_pattern = re.escape(glob_pattern)
             queryset = queryset.filter(path__iregex=glob_pattern.replace('\\*', '.*'))
 
-        # Paginate if necessary
+        # Paginate and return
         page = self.paginate_queryset(queryset)
         if page is not None:
-            queryset = page
+            serializer = self.get_serializer(page, many=True, metadata=include_metadata)
+            return self.get_paginated_response(serializer.data)
 
         serializer = self.get_serializer(queryset, many=True, metadata=include_metadata)
-        return self.get_paginated_response(serializer.data)
+        return Response(serializer.data)
 
     @swagger_auto_schema(
         manual_parameters=[PATH_PREFIX_PARAM],
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -517,7 +517,9 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
         serializer.is_valid(raise_exception=True)
 
         # Don't include metadata field if not asked for
-        queryset: QuerySet[Asset] = self.filter_queryset(self.get_queryset().select_related('blob'))
+        queryset: QuerySet[Asset] = self.filter_queryset(
+            self.get_queryset().select_related('blob', 'embargoed_blob', 'zarr')
+        )
         include_metadata = serializer.validated_data['metadata']
         if not include_metadata:
             queryset = queryset.defer('metadata')
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -511,14 +511,18 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
 
-    @swagger_auto_schema(query_serializer=AssetListSerializer)
+    @swagger_auto_schema(query_serializer=AssetListSerializer, responses={200: AssetSerializer()})
     def list(self, request, *args, **kwargs):
         serializer = AssetListSerializer(data=request.query_params)
         serializer.is_valid(raise_exception=True)
 
-        queryset = self.filter_queryset(self.get_queryset())
-        glob_pattern: str | None = serializer.validated_data.get('glob')
+        # Don't include metadata field if not asked for
+        queryset: QuerySet[Asset] = self.filter_queryset(self.get_queryset().select_related('blob'))
+        include_metadata = serializer.validated_data['metadata']
+        if not include_metadata:
+            queryset = queryset.defer('metadata')
 
+        glob_pattern: str | None = serializer.validated_data.get('glob')
         if glob_pattern is not None:
             # Escape special characters in the glob pattern. This is a security precaution taken
             # since we are using postgres' regex search. A malicious user who knows this could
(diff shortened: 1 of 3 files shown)


dandiapi/api/views/asset.py
```diff
@@ -511,14 +511,18 @@ def destroy(self, request, versions__dandiset__pk, versions__version, **kwargs):
...
-    @swagger_auto_schema(query_serializer=AssetListSerializer)
+    @swagger_auto_schema(query_serializer=AssetListSerializer, responses={200: AssetSerializer()})
     def list(self, request, *args, **kwargs):
         serializer = AssetListSerializer(data=request.query_params)
         serializer.is_valid(raise_exception=True)
 
-        queryset = self.filter_queryset(self.get_queryset())
-        glob_pattern: str | None = serializer.validated_data.get('glob')
+        # Don't include metadata field if not asked for
+        queryset: QuerySet[Asset] = self.filter_queryset(self.get_queryset().select_related('blob'))
```
yarikoptic, line 520: just to make sure -- that `.select_related('blob')` wouldn't just select assets with blobs thus leaving behind zarr assets, would it?  not clear to me why that change is done for this PR

and shouldn't comment above that line be moved below this line?

dandiapi/api/views/asset.py
```diff
@@ -527,13 +531,13 @@ def list(self, request, *args, **kwargs):
             glob_pattern = re.escape(glob_pattern)
             queryset = queryset.filter(path__iregex=glob_pattern.replace('\\*', '.*'))
 
+        # Paginate if necessary
         page = self.paginate_queryset(queryset)
         if page is not None:
-            serializer = self.get_serializer(page, many=True)
-            return self.get_paginated_response(serializer.data)
+            queryset = page
```
yarikoptic, line 537: I know nothing about django/querysets and serializers, but 
- when page could be `None`?
- is it expected to reuse original `queryset` from above if page is None?

logically to me page would be `None` if there is no assets, but then why do we need any serializers and original querysets at all then?

yarikoptic (commented):
just a few questions ;)

dandiapi/api/views/asset.py
AlmightyYakob: `select_related('blob')` joins the `blob` foreign key into the single query, so that we don't need to fetch `blob` from the database individually for every asset when it's serialized into the response (causing n queries where n is the number of assets). It's a performance optimization.

dandiapi/api/views/asset.py
AlmightyYakob: I'm not sure under what circumstances `page` is None, but it's seemingly possible, since this code is taken almost verbatim from the source code for the Django Rest Framework [`ListModelMixin`](https://github.com/encode/django-rest-framework/blob/a1b35bb44b7c9251c8b7bc995aa6598044f1d3ef/rest_framework/mixins.py#L33-L46) class.

dandiapi/api/views/asset.py
yarikoptic: Is that change necessary to the purpose of the PR (adding metadata) or just a general optimization?

dandiapi/api/views/asset.py
yarikoptic: Referenced code is
```python

        page = self.paginate_queryset(queryset)
        if page is not None:
            serializer = self.get_serializer(page, many=True)
            return self.get_paginated_response(serializer.data)

        serializer = self.get_serializer(queryset, many=True)
        return Response(serializer.data)
```
and it
- doesn't overload queryset (readability)
- returns response in case of page being None 

I guess should be matched here?

dandiapi/api/views/asset.py
AlmightyYakob: Just a general optimization, the codebase is littered with unoptimized endpoints.

yarikoptic: then ideally should have been done at least in a separate commit with an explanation alike given [above](https://github.com/dandi/dandi-archive/pull/1216#discussion_r930467945) for others to be able to understand why it is done in a commit message.
Otherwise poor souls like yours truly would spend numerous hours while trying to understand why it was done in the context of the "add metadata subrecord" -- e.g. "was blob somehow needed for metadata?".

tip of trade: I use emacs only for magit mode which allows for convenient staging of hunts and rewrite of commits. Otherwise, could be done "manually" via `git add -i` and choosing one hunk at a time.

re "litter" -- filed a dedicated https://github.com/dandi/dandi-archive/issues/1221 . Let's try to keep PRs and commits atomic regarding their purpose.

PS with this I consider this conversation indeed resolved.

mvandenburgh (approved):

dandiapi/api/views/common.py
```diff
@@ -58,3 +60,20 @@ class DandiPagination(PageNumberPagination):
...
     pattern=r'(0\.\d{6}\.\d{4})|draft',
 )
+
+
+def paginated_response(cls, queryset: QuerySet, *args, **kwargs):
+    """
+    Return a paginated response from the provided queryset, using the provided cls.
+
+    Note:
+        * cls should inherit rest_framework.generics.GenericAPIView
```
yarikoptic, line 70: FWIW `should` can be easily expressed in code via 

```python
assert issubclass(cls, GenericAPIView)
```

thus alerting a developer who needs to RTFM without making him reveal to his peers that he didn't bother to RTFM.
I bet that check is fast enough to not worry about impact on performance.

yarikoptic (approved):
Since I trust nobody, especially an assumption that developers would read all the docstrings thoroughly, I have left a suggestion and will leave another little one. But otherwise, in general, this PR looks ok'ish to me now

dandiapi/api/views/asset.py
```diff
@@ -511,14 +512,20 @@ def destroy(self, request, versions__dandiset__pk, versions__version, **kwargs):
...
     def list(self, request, *args, **kwargs):
         serializer = AssetListSerializer(data=request.query_params)
         serializer.is_valid(raise_exception=True)
 
-        queryset = self.filter_queryset(self.get_queryset())
-        glob_pattern: str | None = serializer.validated_data.get('glob')
+        # Don't include metadata field if not asked for
+        queryset: QuerySet[Asset] = self.filter_queryset(
+            self.get_queryset().select_related('blob', 'embargoed_blob', 'zarr')
+        )
```
yarikoptic, line 523:
```suggestion
        queryset: QuerySet[Asset] = self.filter_queryset(
            self.get_queryset().select_related('blob', 'embargoed_blob', 'zarr')
        )
        # Don't include metadata field if not asked for
```
since `queryset` construction IMHO has nothing to do with that comment.

yarikoptic:
And thank you @AlmightyYakob for cooking it up so quickly!

Issue #1221, opened 2022-07-27
Do a sweep over end points to optimize them


yarikoptic:
Echoing @AlmightyYakob from https://github.com/dandi/dandi-archive/pull/1216#discussion_r931302779 : 

> ... the codebase is littered with unoptimized endpoints.

the code base should be sweeped for majority of such optimization to be done in one concerted effort instead of including them into unrelated modifications within codebase, and without appropriate "documentation" in commit messages.

waxlamp:
Now that we are tracking performance problems in Sentry, I don't think we need to make a concerted effort to solve all the problems at once.

Since we are set up to detect and fix performance issues as they arise, and because in fact each performance issue will be unrelated to the others (aside from all being "performance issues"), I'm going to close this issue in favor of identifying specific issues as they arise.

As I do this, I want to remind you that solving these issues has been a major theme of our technical debt paydown in recent days/weeks, so this is still very much on our minds, and we've made some good strides already, with more to come.

PR #1222, opened 2022-07-27
Fix asset list performance issue

danlamanna:
Fix n+1 problem in the asset listing endpoint.

Partially addresses #1221.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -137,7 +137,7 @@ class AssetFilter(filters.FilterSet):
 
 
 class AssetViewSet(DetailSerializerMixin, GenericViewSet):
-    queryset = Asset.objects.all().order_by('created')
+    queryset = Asset.objects.all().select_related('zarr').order_by('created')
 
     serializer_class = AssetSerializer
     serializer_detail_class = AssetDetailSerializer


dandiapi/api/views/asset.py
```diff
@@ -136,7 +136,7 @@ class Meta:
 
 
 class AssetViewSet(DetailSerializerMixin, GenericViewSet):
-    queryset = Asset.objects.all().order_by('created')
+    queryset = Asset.objects.all().select_related('zarr').order_by('created')
```
yarikoptic, line 139: Ignorant in django me wonders: 
- Since it seems to be a generic (not some zarr-specific) viewset, why only `zarr` to be select_related whenever e.g. in https://github.com/dandi/dandi-archive/pull/1216/files#diff-00bb976899730dcae9a407d74d487ec5ba609713007bb4bad93ca68da650cc1cR522 it is

```python
self.get_queryset().select_related('blob', 'embargoed_blob', 'zarr')
```
? e.g. I do see use of `.embargoed_blob` and `.blob` in the code below.
- do we really want *always* (since I guess it defines default behavior at this level) to do that (i.e. always add some minor overhead) or should it may be tuned per method depending on what it needs?

yarikoptic, line 139: how does this change relate to https://github.com/dandi/dandi-archive/pull/1216/files#diff-00bb976899730dcae9a407d74d487ec5ba609713007bb4bad93ca68da650cc1cR522 mentioned above -- isn't it also for the "listing of assets"?

mvandenburgh (approved):
Since this fixes the immediate performance problems we're seeing, I think this is good to go and we can iterate on a followup PR to determine if more `select_related` calls are needed.

dandiapi/api/views/asset.py
yarikoptic: While chatting with @waxlamp the answers are
- only `zarr` because some profiling on sentry showed that merging on that particular one was beneficial. Other ones might also be useful to be added, but nobody checked etc
- the answer is "may be" ;)

yarikoptic: different view. but indeed shows an alternative approach of "take all 3 since we know they would come useful"

PR #1223, opened 2022-07-29
Remove flake8 linting rule

danlamanna:
IME this rule rarely catches problems and regularly frustrates in the cases where uppercase variable names are often necessary e.g. migrations.


mvandenburgh (approved):
Seems fine to me.

PR #1229, opened 2022-08-03
Use tuples for Admin model actions and inlines

AlmightyYakob:
It seems the release of Django 4.1 caused a breaking change with our admin models. As far as I can tell from the [release notes](https://docs.djangoproject.com/en/4.1/releases/4.1/), nothing else in our codebase was affected.

This PR also pins our django version to `4.1.*`, to avoid this in the future.


mvandenburgh (approved):

waxlamp:
Did this close #1230?

yarikoptic:
most likely since CI above is green whenever it was AFAIK failing before .

yarikoptic:
FWIW re-triggered https://github.com/dandi/dandi-cli/pull/1003 showed them green so should be good now.

PR #1231, opened 2022-08-05
Fix incorrect type hints

mvandenburgh:
I noticed some incorrect type hints while working on the API code. We don't do static type checking for our python code, so this went unnoticed.


waxlamp:
Can you also file an issue to add type checking to our CI process?

mvandenburgh:
> Can you also file an issue to add type checking to our CI process?

Filed #1232.

AlmightyYakob (approved):

PR #1239, opened 2022-08-10
RF: sort long listings of keys in metadata records

yarikoptic:
Originally suggested as a part of the
https://github.com/dandi/dandi-archive/pull/1103

Consistent sorted order improves DX by not needing to do linear search
for a specific field and simplifying matching of listings (e.g. to see if
specific field is lacking from needing to be considered autogenerated etc).

### DIFF
--- a/dandiapi/api/manifests.py
+++ b/dandiapi/api/manifests.py
@@ -127,9 +127,9 @@ def write_collection_jsonld(version: Version):
             .render(
                 {
                     '@context': version.metadata['@context'],
-                    'id': version.metadata['id'],
                     '@type': 'prov:Collection',
                     'hasMember': list(version.assets.values_list('metadata__id', flat=True)),
+                    'id': version.metadata['id'],
                 },
             )
             .decode()
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -262,12 +262,12 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
 
         metadata = {
             **self.metadata,
-            'id': f'dandiasset:{self.asset_id}',
-            'path': self.path,
-            'identifier': str(self.asset_id),
-            'contentUrl': [download_url, s3_url],
             'contentSize': self.size,
+            'contentUrl': [download_url, s3_url],
             'digest': self.digest,
+            'id': f'dandiasset:{self.asset_id}',
+            'identifier': str(self.asset_id),
+            'path': self.path,
         }
         schema_version = metadata['schemaVersion']
         metadata['@context'] = (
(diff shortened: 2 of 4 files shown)


waxlamp:
The field ordering as written in the code has a semantic meaning that's useful to the team. Since text editors have strong features for text searching, I'm going to close this without merging.

yarikoptic:
> The field ordering as written in the code has a semantic meaning that's useful to the team.

Please then describe the "semantic meaning" in an alternative PR on top of each such record to supersede this one so others would know e.g. in what order to add a new field to satisfy such semantic.  Insofar I do not catch any semantic here, besides a possible historical aspect of order in adding them.

If there is no really semantic meaning, I would strongly prefer to have fields in some deterministic order, e.g. sorted.  It is not only about being able to "search" for a field but to be able to more easily read the code.  As more eyes look at the code through time, we better strive to have code more readable, and thus I would reincarnate this PR.

waxlamp:
> > The field ordering as written in the code has a semantic meaning that's useful to the team.
> 
> Please then describe the "semantic meaning" in an alternative PR on top of each such record to supersede this one so others would know e.g. in what order to add a new field to satisfy such semantic. Insofar I do not catch any semantic here, besides a possible historical aspect of order in adding them.
> 
> If there is no really semantic meaning, I would strongly prefer to have fields in some deterministic order, e.g. sorted. It is not only about being able to "search" for a field but to be able to more easily read the code. As more eyes look at the code through time, we better strive to have code more readable, and thus I would reincarnate this PR.

This is basically an instance of [Chesteron's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence).  There is probably not a very detailed specification of the semantic ordering of these fields, but that doesn't mean that alphabetizing them is the right thing to do. For instance, keeping the `id` field at the top of the list as a general principle is useful, as is placing other principal information (per model) near the top.

I am running off of the idea that when a team puts these kinds of decisions into the code, there is often an implicit understanding of why things are the way they are (this [paper](https://pages.cs.wisc.edu/~remzi/Naur.pdf) and this [article](https://hiringengineersbook.com/post/autonomy/) go into this idea in depth).  As for knowing where to put a new field, I would say that code review would expose if and how to change the placement, and through that process that implicit shared model that exists in the minds of the team can be extended to the mind of the new contributor.

I realize this may seem like an overblown argument, but it is actually important in a subtle way: I need for the team that put this code together to retain autonomy over such decisions as changing to an alphabetized field listing, even when not doing so doesn't have a clear cut justification. Again, this is because of the (unavoidably) implicit modeling that is done between the engineering team. And in turn, that leads to Chesterton's fence as the principle by which to avoid what may seem like obvious improvements to the codebase.

yarikoptic:
I will not fight over this, but I do remain of an opinion that current order is nearly arbitrary (since can't be codified) and will just let to less efficient coding etc.

PR #1243, opened 2022-08-15
Configure mypy

### LINKED ISSUE #1232: Perform static type-checking on Python code
We should add static type-checking to our CI process for the backend python code. Our tox config already includes a [`type` target that runs `mypy`](https://github.com/dandi/dandi-archive/blob/master/tox.ini#L21-L29), ~~so maybe it's just a matter of enabling it there~~. (edit: it's not, this is going to require some more investigation as the current `tox -e type` command results in a bunch of celery errors)

mvandenburgh:
Closes #1232.

This makes the necessary changes to `tox.ini` to enable `mypy` static type-checking locally and in CI. Type-checking can be run via `tox -e type`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,3 +11,25 @@ line_length = 100
 force_sort_within_sections = true
 # Combines "as" imports on the same line
 combine_as_imports = true
+
+[tool.mypy]
+ignore_missing_imports = true
+show_error_codes = true
+disable_error_code = ["attr-defined", "var-annotated"]
+follow_imports = "skip" # Don't follow imports into other files. This should be removed once all type errors have been resolved.
+exclude = [
+    "^dandiapi/api/admin.py",
+    "^dandiapi/api/mail.py",
+    "^dandiapi/api/tests/",
+    "^dandiapi/api/management/",
+    "^dandiapi/api/migrations/",
+    "^dandiapi/api/models/",
+    "^dandiapi/api/tasks/",
+    "^dandiapi/api/views/",
+]
+
+# Re-enable these when https://github.com/typeddjango/django-stubs/issues/417 is fixed.
+# plugins = ["mypy_django_plugin.main", "mypy_drf_plugin.main"]
+
+# [tool.django-stubs]
+# django_settings_module = "dandiapi.settings"
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 [tox]
 envlist =
     lint,
+    type,
     test,
     check-migrations,
 
@@ -19,14 +20,12 @@ commands =
     flake8 --config=tox.ini {posargs:.}
 
 [testenv:type]
-skipsdist = true
-skip_install = true
 deps =
     mypy
-    django-stubs
-    djangorestframework-stubs
+extras =
+    dev
 commands =
-    mypy {posargs:.}
+    mypy {posargs:dandiapi/}
 
 [testenv:format]
 skipsdist = true
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,9 +1,0 @@
-[mypy]
-plugins =
-    mypy_django_plugin.main
-
-[mypy.plugins.django-stubs]
-django_settings_module = "dandiapi.settings"
-
-[mypy-celery.*]
-ignore_missing_imports = True
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,9 @@ setup(
             'ipython',
             'tox',
             'boto3-stubs[s3]',
+            'django-stubs',
+            'djangorestframework-stubs',
+            'types-setuptools',
             'memray',
         ],
         'test': [


pyproject.toml
```diff
@@ -11,3 +11,25 @@ line_length = 100
...
+exclude = [
+    "^dandiapi/api/admin.py",
+    "^dandiapi/api/mail.py",
+    "^dandiapi/api/tests/",
+    "^dandiapi/api/management/",
+    "^dandiapi/api/migrations/",
+    "^dandiapi/api/models/",
+    "^dandiapi/api/tasks/",
+    "^dandiapi/api/views/",
+]
```
mvandenburgh, line 29: For now, I've excluded all files and directories that have type errors. I'll make follow up PRs where I fix the type errors in each of these and remove the corresponding entry from `exclude`.

pyproject.toml
```diff
@@ -11,3 +11,25 @@ line_length = 100
 force_sort_within_sections = true
 # Combines "as" imports on the same line
 combine_as_imports = true
+
+[tool.mypy]
+ignore_missing_imports = true
+show_error_codes = true
+disable_error_code = ["attr-defined", "var-annotated"]
```
mvandenburgh, line 18: These two error cases produced a lot of type errors, many of which were incorrect because of django typing stubs not working. I've disabled them for now, and if django typing stubs ever get integrated with `django-configurations` properly, we can re-enable them.

pyproject.toml
```diff
@@ -11,3 +11,25 @@ line_length = 100
...
+    "^dandiapi/api/models/",
+    "^dandiapi/api/tasks/",
+    "^dandiapi/api/views/",
+]
+
+# Re-enable these when https://github.com/typeddjango/django-stubs/issues/417 is fixed.
+# plugins = ["mypy_django_plugin.main", "mypy_drf_plugin.main"]
+
+# [tool.django-stubs]
+# django_settings_module = "dandiapi.settings"
```
mvandenburgh, line 35: Since we are using `django-configurations` (well, technically `composed-configuration` does, not dandi), there's some additional considerations for running these django plugins. There's an open issue to support it https://github.com/typeddjango/django-stubs/issues/417, but it hasn't been implemented yet

brianhelba: DANDI _does_ use `django-configuration` directly. That's why things like this: https://github.com/dandi/dandi-archive/blob/master/manage.py#L14 are necessary.

I agree that due to the lack of proper support, it's safer to disable these plugins for now.

tox.ini
```diff
@@ -20,13 +21,31 @@ commands =
 
 [testenv:type]
 skipsdist = true
-skip_install = true
+setenv =
+    DJANGO_CONFIGURATION = DevelopmentConfiguration
+passenv =
```
brianhelba, line 26: If we were using the Mypy plugin, then the settings file would be getting parsed and these should be necessary. However, this shouldn't be necessary just to import all the files. Do you know what is requiring these?

brianhelba, line 26: Also, Tox has a better syntax to copy values from another environment: https://tox.wiki/en/latest/config.html#substitution-for-values-from-other-sections

Use that, instead of repeating yourself.

tox.ini
```diff
@@ -3,6 +3,7 @@ envlist =
     lint,
     test,
     check-migrations,
+    type,
```
brianhelba, line 6: Maybe run this before the `test` step, since it should be a bit faster and produce more obvious errors.

.github/workflows/backend-ci.yml
```diff
@@ -51,6 +51,9 @@ jobs:
           DJANGO_MINIO_STORAGE_ENDPOINT: localhost:9000
           DJANGO_MINIO_STORAGE_ACCESS_KEY: minioAccessKey
           DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
+          DJANGO_STORAGE_BUCKET_NAME: dandi-bucket
```
brianhelba, line 54: You shouldn't need to set this if you're using `TestingConfiguration`.

tox.ini
```diff
@@ -20,13 +21,31 @@ commands =
 
 [testenv:type]
 skipsdist = true
-skip_install = true
+setenv =
+    DJANGO_CONFIGURATION = DevelopmentConfiguration
```
brianhelba: If this is necessary, how about `TestingConfiguration`?

However, I still don't think it ought to be needed at all.

tox.ini
mvandenburgh: Yeah that's exactly why - I was using `mypy_django_plugin` previously, before I realized it didn't work with `django-configuration` - updated

tox.ini
mvandenburgh: :+1:

tox.ini
mvandenburgh: Nope, not needed.

tox.ini
mvandenburgh: Seems like everything is working properly now without this. But thanks, that's good to know :+1:

brianhelba: So can this be removed now?

brianhelba (changes requested):
Open question about `passenv` in `tox.ini`.

tox.ini
mvandenburgh: My bad, I thought you were referring to the `setenv` section here, not `passenv`. But yes, `passenv` can also be removed now, I've pushed that change.

tox.ini
```diff
@@ -20,13 +21,15 @@ commands =
 
 [testenv:type]
 skipsdist = true
```
brianhelba, line 23: I'd remove `skipsdist` too. The distinction between it and `skip_install` isn't super clear (as it needs to make an sdist to do the install, but I practically think of them as coupled Tox settings.

tox.ini
```diff
@@ -20,13 +21,15 @@ commands =
...
-skip_install = true
 deps =
     mypy
     django-stubs
     djangorestframework-stubs
+    types-setuptools
```
brianhelba, line 28:
```suggestion
```

We shouldn't need this. We're not checking `setup.py`, we're looking directly inside `dandiapi/`.

tox.ini
brianhelba, line 23:
```suggestion
```

I'd remove `skipsdist` too. The distinction between it and `skip_install` isn't super clear (as it needs to make an sdist to do the install, but I practically think of them as coupled Tox settings.

tox.ini
```diff
@@ -19,14 +20,14 @@ commands =
     flake8 --config=tox.ini {posargs:.}
 
 [testenv:type]
-skipsdist = true
-skip_install = true
 deps =
     mypy
     django-stubs
     djangorestframework-stubs
+extras =
```
brianhelba, line 27: I see that `boto3-stubs[s3]` is in the `[dev]` extra. All the type stubs ought to be in the same place. Either:
* Put all stubs in the `[dev]` extra (or another new extra name, if you want), so they can be easily installed for the benefit of developers' IDEs for auto-complete. This Tox env can then install that extra.
* Put all stubs under `[testenv:type]`

tox.ini
mvandenburgh: We're using `setuptools_scm` for versioning in this file https://github.com/dandi/dandi-archive/blob/master/dandiapi/__init__.py. We could ignore it, but I don't think it hurts that much either to add it as a dependency. :shrug: I added it back as a `dev` extra dependency, but I'm fine with just excluding this one file too if installing `types-setuptools` is too overkill - any thoughts @brianhelba @AlmightyYakob ?

tox.ini
mvandenburgh: Since we're going to be enforcing type checking in CI, I think it's reasonable to include this as a `[dev]` extra so developers get auto-complete in their IDEs.

tox.ini
brianhelba: I didn't realize you had that import of `pkg_resources` in your actual project module. I think it's correct to include `types-setuptools` for that. Sorry for my confusion.

@AlmightyYakob This is totally out of scope of this PR, but note there's a way to get the package version from Python 3.8+ standard libraries (rather than setuptools): https://github.com/pypa/setuptools_scm/#retrieving-package-version-at-runtime

brianhelba (approved):
LGTM. :shipit: 

@mvandenburgh Thanks for your perseverance in working out some of these low-level usage details. I'd like to upstream some of these change to the Girder 4 cookiecutter and I've learned a lot myself from this experience.

mvandenburgh:
I'll ping @waxlamp @danlamanna @AlmightyYakob @DeepikaGhodki here just so you all are aware of this change. In particular, note that this will now run in CI alongside the other tox checks, after linting and before the pytest tests.

PR #1246, opened 2022-08-15
Fix type errors in `models`

mvandenburgh:
Removes `dandiapi/api/models` from the excluded files list in mypy, and fixes all type errors reported by it.

### DIFF
--- a/dandiapi/api/models/upload.py
+++ b/dandiapi/api/models/upload.py
@@ -41,13 +41,13 @@ class BaseUpload(TimeStampedModel):
 
     @staticmethod
     @abstractmethod
-    def object_key(upload_id, dandiset: Dandiset | None = None):  # noqa: N805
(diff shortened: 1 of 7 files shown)


dandiapi/api/models/upload.py
```diff
@@ -41,13 +41,13 @@ class Meta:
 
     @staticmethod
     @abstractmethod
-    def object_key(upload_id, dandiset: Dandiset | None = None):  # noqa: N805
```
mvandenburgh, line 44: Note this code change slightly changes the way that `BaseUpload.object_key()` is called. Since `EmbargoedUpload` doesn't require `dandiset`, but `Upload` does, I changed `dandiset` to be a mandatory keyword argument.

AlmightyYakob (approved):

PR #1247, opened 2022-08-18
Require Python 3.10

mvandenburgh:
Updates our development environment and Heroku deployment to require the latest Python 3.10.

This will let us use new typing syntax like the `X | Y` union operator without having to "cheat" with `from __future__ import annotations`, in addition to all the other new features Python 3.10 introduces.

I've also run [pyupgrade](https://github.com/asottile/pyupgrade) on the codebase to update to the latest syntax and remove deprecated code.


.github/workflows/cli-integration.yml
```diff
@@ -61,10 +61,6 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
-      - name: Install hdf5
-        if: matrix.python == '3.9'
-        run: sudo apt-get update && sudo apt-get install -y libhdf5-dev
```
mvandenburgh, line 66: @jwodder is this step safe to remove now that we're not testing against Python 3.9?

jwodder: If the tests pass without it, yes.

AlmightyYakob (approved):
LGTM, I'd only ask to make sure that the updated dev containers works locally as well, since that isn't tested by CI (although maybe that's covered by the CLI tests, I'm not sure).

mvandenburgh:
> LGTM, I'd only ask to make sure that the updated dev containers works locally as well, since that isn't tested by CI (although maybe that's covered by the CLI tests, I'm not sure).

Good catch. I recreated the dev containers locally and they work fine.

PR #1250, opened 2022-08-23
Fix type errors in `mail.py`

mvandenburgh:


dandiapi/api/mail.py
```diff
@@ -21,10 +21,10 @@
 
 
 def build_message(subject: str, message: str, to: list[str], html_message: str | None = None):
-    message = mail.EmailMultiAlternatives(subject=subject, body=message, to=to)
```
mvandenburgh, line 24: `message` is already defined as an argument of type `str`, so I just renamed this variable

AlmightyYakob (approved):

PR #1255, opened 2022-08-26
Rearrange file structure

mvandenburgh:
Celery beat tasks and the dashboard view are not in the `tasks/` and `views/` directories, which makes them difficult to find, IMO


mvandenburgh:
We also might want to consider moving all of the DRF `ViewSets` into a `rest/` directory, and leave the `views/` directory only for actual Django `Views`.

AlmightyYakob (approved):
LGTM. I agree somewhat with the point of separating DRF viewsets and regular django views, although I don't think it's too critical.

PR #1256, opened 2022-08-30
Create django app for zarr functionality

danlamanna:
This creates a django app specifically for zarr. This segments off most of the zarr functionality (views, models, commands, and tests) into its own module for readability. The diff numbers are inflated due to autogenerated migrations - Django has a hard time moving models between apps so some trickery is necessary to get it to noop. To be clear, the database schema and functionality remains the same, all that's happening in this PR is code moving between files and imports being renamed.


AlmightyYakob (approved):
All makes sense, nothing really jumps out at me. Tested locally and everything worked without issue.

PR #1265, opened 2022-09-06
Migrate Vuex store to Pinia

### LINKED ISSUE #1261: Upgrade Vuex to Pinia
[Pinia](https://pinia.vuejs.org/introduction.html) has replaced Vuex as the official state management solution for Vue. Now that we're on Vue 2.7 as of #1258, we should upgrade our frontend to it. 

In particular, this will let us abandon the [`direct-vuex` wrapper](https://github.com/paroi-tech/direct-vuex) and get full TypeScript support natively in our store.

mvandenburgh:
Closes #1261 

For reference, I mostly followed [this guide](https://pinia.vuejs.org/cookbook/migration-vuex.html). 

One thing to note is that Pinia has a slightly different paradigm than Vuex when it comes to "stores" and "modules". Vuex had one central _store_ that could be divided into several _modules_. On the other hand, Pinia has no concept of a first-class _module_, instead, you have multiple _stores_. So, in this PR I have converted our Vuex store with a `dandiset` module, to a single Pinia `DandisetStore`.


web/src/rest.ts
```diff
@@ -37,11 +40,7 @@ const dandiRest = new Vue({
     };
   },
   computed: {
-    schemaVersion(): string {
-      // Use injected $store instead of importing to
-      // avoid dependency cycle
-      return this.$store?.direct.getters.dandiset.schemaVersion;
-    },
+    ...mapState(useDandisetStore, ['schemaVersion']),
```
AlmightyYakob, line 43: We should probably migrate this component to use composition API as well, but that's out of scope for this PR. Do you mind filing an issue?

AlmightyYakob (changes requested):
LGTM just a couple suggestions

web/src/rest.ts
mvandenburgh: https://github.com/dandi/dandi-archive/issues/1267

AlmightyYakob (approved):

web/src/stores/dandiset.ts
```diff
@@ -0,0 +1,112 @@
...
+    async uninitializeDandisets() {
+      this.dandiset = null;
+      this.versions = null;
+      this.owners = null;
+      this.loading = false;
+    },
+    async initializeDandisets({ identifier, version }: Record<string, string>) {
+      this.uninitializeDandisets();
+
+      // this can be done concurrently, don't await
```
waxlamp, line 55: Is this a TODO?

web/src/stores/dandiset.ts
```diff
@@ -0,0 +1,112 @@
...
+export const useDandisetStore = defineStore('dandiset', {
+  state: (): State => ({
+    dandiset: null,
+    versions: null,
+    loading: false, // No mutation, as we don't want this mutated by the user
+    owners: null,
+    schema: null,
+  }),
+  getters: {
+    version: (state) => (state.dandiset ? state.dandiset.version : draftVersion),
```
waxlamp, line 28: Can we apply TS types to these functions? It would make the list of getters that much more legible.

web/src/stores/dandiset.ts
```diff
@@ -0,0 +1,112 @@
...
+      if (!sanitizedVersion) {
+        this.dandiset = null;
+        this.loading = false;
+        return;
+      }
+
+      try {
+        const data = await dandiRest.specificVersion(identifier, sanitizedVersion);
+        this.dandiset = data;
+      } catch (err) {
```
waxlamp, line 83: Is there a need to examine `err` in case something needs to be raised to Sentry?

web/src/App.vue
```diff
@@ -23,19 +23,21 @@
 import { onMounted, ref } from 'vue';
 
 import { dandiRest } from '@/rest';
-import store from '@/store';
+import { useDandisetStore } from '@/stores/dandiset';
```
waxlamp, line 26: Is is the right usage pattern to name this function with a `use` prefix? In my mind, that signals that it's a [Vue composable](https://vuejs.org/guide/reusability/composables.html), whereas this just straight up returns a store object. If the usage convention is to prefix this application with `use` too, that's fine. Otherwise, I think it would be good to stick to the conventions.

waxlamp (commented):
This is very cool. Thanks @mvandenburgh!

web/src/stores/dandiset.ts
mvandenburgh: Good catch, yes we should only catch expected 400 errors here, and raise `err` otherwise. I'll make a follow up PR for that.

web/src/App.vue
mvandenburgh: I had the same thought initially, but yes that is the convention https://pinia.vuejs.org/cookbook/migration-vuex.html#converting-a-single-module

web/src/stores/dandiset.ts
mvandenburgh: Unlike Vuex with `direct-vuex`, Pinia is fully typed, so the types of every function parameter/return value here are inferred automatically without needing to manually specify them.

mvandenburgh: Oh, maybe I'm misunderstanding what you're saying. Are you just asking for explicit types to be added? Like
```suggestion
    version: (state): string => (state.dandiset ? state.dandiset.version : draftVersion),
```

web/src/stores/dandiset.ts
mvandenburgh: No, I believe it's just there to clarify why there's no `await` in front of the following method so someone doesn't see it and think it's a bug. It was there in the old vuex code, I just copied it over.

web/src/stores/dandiset.ts
waxlamp: Yeah, I meant the latter. It's not necessary to fill them all in now, but I think one of the major benefits of TypeScript is that the types are checked by the machine, but also legible to the human.

web/src/stores/dandiset.ts
waxlamp: The confusion I had was between what you just said (explaining why the first statement has no `await`) and suggesting that someone should remove the following two `await`s and make all three operations concurrent.

PR #1266, opened 2022-09-07
Fix display of validation errors when there's no associated field

### LINKED ISSUE #750: Validation error message for bad schema version is malformed
The error reads like this: `: Metadata version 0.4.3 is not allowed. Allowed are: 0.5.1.` (on, e.g., [staging dandiset 000001](https://deploy-preview-857--gui-dandiarchive-org.netlify.app/#/dandiset/000001).

danlamanna:
Fixes #750.

The `field` for validation errors can be empty if it's referring to a general metadata error. This makes the GUI account for those scenarios.


mvandenburgh (approved):
LGTM

PR #1270, opened 2022-09-08
Enable sentry performance for frontend

mvandenburgh:


mvandenburgh:
> LGTM, have you tested this to ensure it comes through sentry correctly?

Yes, I tested locally with the staging sentry DSN and it works

PR #1273, opened 2022-09-09
Use direct foreign keys with django-guardian

danlamanna:
Using direct foreign keys allows us to guarantee referential integrity for permissions. Doing this migration already revealed several orphaned permission relations. More on direct foreign keys is covered in the documentation here https://django-guardian.readthedocs.io/en/stable/userguide/performance.html#direct-foreign-keys.


mvandenburgh (approved):
LGTM

PR #1280, opened 2022-09-19
Don't revalidate version metadata on PUT unless it has changed

### LINKED ISSUE #541: Do not trigger version validation when no changes were made
Currently, opening the meditor and clicking "Save" without changing anything triggers revalidation and rewriting the draft manifests. We should check to make sure content was actually modified before setting `status = Version.Status.PENDING`.

mvandenburgh:
Closes #541

### DIFF
--- a/dandiapi/api/tests/test_version.py
+++ b/dandiapi/api/tests/test_version.py
@@ -1,9 +1,16 @@
+from __future__ import annotations
+
 from datetime import datetime
+from time import sleep
+from typing import TYPE_CHECKING
 
 from django.conf import settings
 from django.contrib.auth.models import User
 from guardian.shortcuts import assign_perm
 import pytest
+
+if TYPE_CHECKING:
+    from rest_framework.test import APIClient
 
 from dandiapi.api import tasks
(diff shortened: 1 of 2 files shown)


dandiapi/api/tests/test_version.py
```diff
@@ -878,3 +885,26 @@ def test_version_rest_delete_draft_admin(api_client, admin_user, draft_version):
     assert response.status_code == 403
     assert response.data == 'Cannot delete draft versions'
     assert draft_version in Version.objects.all()
+
+
+@pytest.mark.django_db
+def test_version_no_metadata_updates(api_client: APIClient, admin_user, draft_version: Version):
```
AlmightyYakob, line 891: I think this should be renamed to something like
```
test_version_rest_update_no_changed_metadata
```

to keep in line with the other metadata update tests, as well as moved up in the file to live with them.

AlmightyYakob (changes requested):
One small suggestion otherwise LGTM

AlmightyYakob (approved):

PR #1281, opened 2022-09-19
Only suppress 4xx errors in pinia action

mvandenburgh:
Follow-up from [this conversation](https://github.com/dandi/dandi-archive/pull/1265#discussion_r973302209).


mvandenburgh:
@waxlamp I don't think a review is necessary for this, but just pinging so you're aware

PR #1282, opened 2022-09-19
Fix asset unembargo method

### LINKED ISSUE #1264: Asset metadata is not updated during unembargo
Follow up from https://github.com/dandi/dandisets/issues/257

The asset `unembargo` method only calls save on the blob and embargoed_blob fields https://github.com/dandi/dandi-archive/blob/6134e15122e7156c718ef50b7f3bd8c9f98b6d6e/dandiapi/api/models/asset.py#L276

Because of this, the asset metadata is never repopulated, and so the old embargoed URL is still present in the metadata. Retrieving the current s3 url for any unembargoed assets that have since not been saved again, returns a path within the public bucket, not the embargoed bucket.

We should either include `metadata` in the list of fields to update, or just save the entire model.

### LINKED ISSUE #1278: assets can still be missing sha256, may be due to "unembargo"
originally observed/reported in https://github.com/dandi/dandisets/issues/257#issuecomment-1240994331 but likely better reside in a dedicated report here since might require some fixing of code.  Assets in 000235 are lacking sha256 despite them being out there for quite awhile.  I believe the same problem is overall in 00023[5678] which we had to exclude from backup for now since we do need sha256.  Since those "appeared" only recently but are "low" in their id (we have 000337 ATM) I think they are unembargoed and iirc @AlmightyYakob suggested that there could be remaining issues in unemabrgoe'ing process.  So - the code should be fixed and assets in these dandisets have their sha256 computed.

related question: Do we have some way to figure out that a dandiset was unemabrgoed by looking at metadata?

AlmightyYakob:
Closes #1264
Closes #1278 

There were a couple of things wrong with the asset unembargo method
1. The metadata wasn't saved, so the metadata remained out of date and incorrect
2. The sha256 value wasn't copied to the unembargoed asset blob, leaving that field null.

@yarikoptic I believe the latter is the cause of #1278.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -261,18 +261,19 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
             # Assert files are equal
             assert resp.etag == self.embargoed_blob.etag
 
-            # Assign blob
+            # Assign blob (changing only blob)
             self.blob = AssetBlob(
-                blob_id=self.embargoed_blob.blob_id,
-                etag=resp.etag,
                 blob=resp.key,
+                blob_id=self.embargoed_blob.blob_id,
+                sha256=self.embargoed_blob.sha256,
+                etag=self.embargoed_blob.etag,
                 size=self.embargoed_blob.size,
             )
             self.blob.save()
 
         # Save updated blob field
         self.embargoed_blob = None
-        self.save(update_fields=['blob', 'embargoed_blob'])
+        self.save()
 
     def publish(self):
         """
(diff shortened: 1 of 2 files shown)


yarikoptic:
coolio! Thank you @AlmightyYakob . Note that "CI for backend / test" seems to be errorring out legitimately since in `test_asset_unembargo[published]`. `botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found` so needs a look I guess.

mvandenburgh (approved):
LGTM

PR #1285, opened 2022-09-20
Use `exact` prop consistently

### LINKED ISSUE #792: Consistently use exact in router links
By default, include `exact` alongside any usage of `:to`.

https://router.vuejs.org/api/#exact

waxlamp:
Closes #792

Reviewers, please especially look at f4790be63307f95296e467cb0b0968e61923a6f0 and tell me if that `exact` is actually supposed to be there in some other capacity (it has no corresponding `:to`, and doesn't look like it's supposed to).

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -24,7 +24,6 @@
             >
               <v-list-item-content
                 v-if="!navItem.external"
-                exact
                 text
                 class="text-md"
               >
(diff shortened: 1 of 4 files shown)


mvandenburgh (approved):
LGTM

> Reviewers, please especially look at f4790be and tell me if that exact is actually supposed to be there in some other capacity (it has no corresponding :to, and doesn't look like it's supposed to).

I looks like it was placed there erroneously. The `v-list-item-content` docs make no mention of an `exact` prop.
https://vuetifyjs.com/en/api/v-list-item-content/

brianhelba (commented):
I haven't looked at this comprehensively to be sure that all the appropriate places were covered, but the changes here seem reasonable.

There is a legitimate use case for `to` without `exact`, where nested pages should cause a top-level link to me marked as active. I don't know if that applies to any of these changes.

PR #1290, opened 2022-09-22
Allow for using django core exceptions with DRF 

danlamanna:
This creates a custom exception handler for DRF which will properly handle core django exceptions (like `ValidationError`and `PermissionDenied`) and raise them as the DRF versions. I think we should still explicitly the context specific exceptions when we can (e.g. in DRF views) - but this will allow internal APIs that aren't context aware to raise the builtin django exceptions.

### DIFF
--- a/dandiapi/drf_utils.py
+++ b/dandiapi/drf_utils.py
@@ -1,0 +1,30 @@
+from django.core.exceptions import (
+    PermissionDenied as DjangoPermissionDenied,
+    ValidationError as DjangoValidationError,
+)
+from django.http import Http404
+from rest_framework import exceptions as drf_exceptions
+from rest_framework.response import Response
+from rest_framework.serializers import as_serializer_error
+from rest_framework.views import exception_handler
+
+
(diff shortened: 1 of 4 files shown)


dandiapi/drf_utils.py
```diff
@@ -0,0 +1,30 @@
...
+    and be used from DRF endpoints.
+    """
+    if isinstance(exc, DjangoValidationError):
+        exc = drf_exceptions.ValidationError(as_serializer_error(exc))
+
+    if isinstance(exc, Http404):
+        exc = drf_exceptions.NotFound()
+
+    if isinstance(exc, DjangoPermissionDenied):
+        exc = drf_exceptions.PermissionDenied()
```
AlmightyYakob, line 26: So this seems to only re-wrap 3 specific django errors. Is there a reason you chose just these three? Are these 3 the "legitimate" ones that should be handle gracefully, while other would result in 500 errors?

danlamanna: These are the 3 that I could think of, yeah. I think it's fine to add more if we discover them.

AlmightyYakob (approved):

PR #1300, opened 2022-10-03
Convert dandiset search test to jest/puppeteer, remove pyppeteer config from project

### LINKED ISSUE #995: Port remaining jest/puppeteer tests to pyppeteer
This is a follow up to #738. Now that we have pyppeteer configured with some of the puppeteer tests converted, the rest will need to be converted.

- [ ] Convert puppeteer tests to pyppeteer
- [ ] Remove puppeteer test CI job

mvandenburgh:
This is the final test that needed to be converted to jest/puppeteer from pyppeteer. I also removed all configuration from CI and tox for pyppeteer.

Closes #995


web/test/src/tests/dandisetsPage.test.js
```diff
@@ -26,4 +27,60 @@ describe('dandisets page', () => {
...
+
+    const dandisetIdentifierNameMapping = new Map();
+
+    // Create 10 dandisets
+    for (let i = 0; i < 10; i += 1) {
+      // Generate random dandiset name and description
+      const name = faker.lorem.words();
+      const description = faker.lorem.sentences();
+
+      /* eslint-disable no-await-in-loop */
```
danlamanna, line 42: It seems like this sort of rule should be disabled in the entire tests folder, right?

web/test/src/tests/dandisetsPage.test.js
```diff
@@ -26,4 +27,60 @@ describe('dandisets page', () => {
     await expect(page).toMatch(`Contact ${lastName}, ${firstName}`);
     await expect(page).toMatch(`Updated on ${moment(new Date()).format('LL')}`);
   });
+
+  it('search for dandisets', async () => {
+    await registerNewUser();
+
+    const dandisetIdentifierNameMapping = new Map();
+
+    // Create 10 dandisets
```
danlamanna, line 36: Did you give a lot of thought to scaffolding these tests in python vs js? I don't think it makes much of a difference in the short term, but if we find ourselves writing a lot of these tests we might want to revisit it.

danlamanna (approved):
This is awesome! Less code and more reliable testing ftw.

web/test/src/tests/dandisetsPage.test.js
mvandenburgh: This is the only place currently where we use `await` inside a loop, but yeah I guess it would be good to include as a global setting for the tests directory since this is one of the exceptions where awaiting inside a loop makes sense.

web/test/src/tests/dandisetsPage.test.js
mvandenburgh: No, I tried to keep it as simple as possible and didn't try coming up with a way to scaffold test data. If we do end up writing a lot of these tests, we'll definitely need to consider that, especially in the context of things like publishing and zarr which would **require** some sort of test data setup beforehand, since in normal circumstances those require asset upload via the CLI. I think we still need to decide what the future of tests like these looks like though, so I did the bare minimum for now.

This is one major disadvantage of using js instead of python, where we had direct access to `pytest-django`.

PR #1303, opened 2022-10-03
Convert `Meditor.vue` to `<script setup>`

mvandenburgh:


web/src/components/Meditor/Meditor.vue
```diff
@@ -57,18 +58,18 @@
                 <template #activator="{ on }">
                   <v-icon
                     left
-                    :color="allModelsValid ? 'success' : 'error'"
+                    :color="modelValid ? 'success' : 'error'"
```
mvandenburgh, line 61: Note, this was aliased as `allModelsValid` in the returned object of the original `setup()` function. Instead of replicating that behavior in the new <script> section, I think it's clearer to just use the `modelValid` variable directly.

mvandenburgh: (ref: https://github.com/dandi/dandi-archive/pull/1303/files#diff-5dff88078a7fb94bc8e18f93f81fb9cc63cbe2af517e0e488f0612ea91b4be6cL426)

AlmightyYakob (approved):

PR #1311, opened 2022-10-04
Improved pagination design in file browser

### LINKED ISSUE #725: reduce pagination thumbs
when a folder has a large number of files, it would be useful to reduce the number of pages displayed at the bottom:

image

mvandenburgh:
Fixes #725.

I implemented Jared's design from #725, and also added an error message when an invalid path is entered by the user. 
Previously, a user would only run into this error if they manually edited the query params in the browser URL bar, so our very basic solution was to just kick the user back to the root directory of the dandiset if that happened. But now that there's an input to specify the page number, it's much easier for a user to enter something invalid. So, I've changed this behavior to display an error message instead of kicking them back to the root. 
I still think the UX here could use some significant improvement, but for the purposes of this PR I think it's sufficient.

### DIFF
--- a/web/src/components/FileBrowser/FileBrowserPagination.vue
+++ b/web/src/components/FileBrowser/FileBrowserPagination.vue
@@ -77,5 +77,10 @@ const emit = defineEmits(['changePage']);
 const pageInput = ref(props.page.toString());
 
 watch(() => props.page, (newPage) => { pageInput.value = newPage.toString(); });
-watch(pageInput, (newPage) => emit('changePage', Number(newPage)));
+watch(pageInput, (newPage) => {
+  const page = Number(newPage);
+  if (page > 0 && page <= props.pageCount) {
+    emit('changePage', page);
+  }
+});
 </script>
--- a/web/src/components/FileBrowser/FileBrowserPagination.vue
+++ b/web/src/components/FileBrowser/FileBrowserPagination.vue
@@ -32,6 +32,7 @@
       :maxlength="pageCount"
       :rules="[pageIsValid]"
     />
+    <span>of {{ pageCount }}</span>
     <v-btn
       class="mx-2"
       color="white"
--- a/web/src/components/FileBrowser/FileBrowserPagination.vue
+++ b/web/src/components/FileBrowser/FileBrowserPagination.vue
@@ -1,0 +1,80 @@
+<template>
+  <div
+    class="d-flex justify-center align-center text-center"
+  >
+    <v-btn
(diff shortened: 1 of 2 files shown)


mvandenburgh:
This seems to be a good staging dandiset to try this out on https://deploy-preview-1311--gui-staging-dandiarchive-org.netlify.app/dandiset/201029/draft/files?location=

satra:
it would be nice to know how many pages there are.

waxlamp:
> it would be nice to know how many pages there are.

@mvandenburgh, is it known on each render how many total files/pages there are? If yes, it should be straightforward to add an `of <N>` after the combobox.

mvandenburgh:
> > it would be nice to know how many pages there are.
> 
> @mvandenburgh, is it known on each render how many total files/pages there are? If yes, it should be straightforward to add an `of <N>` after the combobox.

Yes, I can add that. Sorry, I thought I replied to @satra's comment but I guess I didn't submit it.

AlmightyYakob:
Maybe when the user inputs an invalid page, it could error but without making the failing call to the API (so just staying on the last valid page)? Currently, if they input something invalid, the page immediately shrinks, and it's more awkward to use.

waxlamp (approved):
Changes look good. I think there's a very subtle issue with saying "no items found at the specified path" for a non-existent folder (it should be a "harsher" error message to distinguish between a folder with no items and a folder that doesn't exist--even though the latter doesn't actually happen in DANDI) but for now I think this is fine.

waxlamp:
> Changes look good. I think there's a very subtle issue with saying "no items found at the specified path" for a non-existent folder (it should be a "harsher" error message to distinguish between a folder with no items and a folder that doesn't exist--even though the latter doesn't actually happen in DANDI) but for now I think this is fine.

I forgot to mention: I think [Jake's comment](https://github.com/dandi/dandi-archive/pull/1311#issuecomment-1289485990) is a good idea; you could implement that now or defer to later--I'll leave that between you and Jake.

mvandenburgh:
> Maybe when the user inputs an invalid page, it could error but without making the failing call to the API (so just staying on the last valid page)? Currently, if they input something invalid, the page immediately shrinks, and it's more awkward to use.

Implemented this https://github.com/dandi/dandi-archive/pull/1311/commits/7626033fb0e9fb2ca14f55c963155d2a002d4093

mvandenburgh:
Coming back to this, I noticed a bug with the "error" page - marking this as draft until I fix that

Issue #1109, opened 2022-05-31
Archive allows the same asset path to be both a directory and a file


jwodder:
When uploading assets to a Dandiset (as well as when changing an asset's `path`), the archive does not prohibit having one asset with path `foo` and another with path `foo/bar`, regardless of what order the assets are created in.  This is a problem, as it means that "`foo`" is both a directory and a file.

MVCE for the uploading case:

```python
import os
from pathlib import Path
from shutil import rmtree
from tempfile import TemporaryDirectory

from dandi.consts import dandiset_metadata_file
from dandi.dandiapi import DandiAPIClient
from dandi.files import dandi_file

with DandiAPIClient.for_dandi_instance(
    "dandi-staging", token=os.environ["DANDI_API_KEY"]
) as client:
    d = client.create_dandiset(
        "Dandiset with Mismatched File Types",
        {
            "schemaKey": "Dandiset",
            "name": "Test Dandiset",
            "description": "A test Dandiset",
            "contributor": [
                {
                    "schemaKey": "Person",
                    "name": "Wodder, John",
                    "roleName": ["dcite:Author", "dcite:ContactPerson"],
                }
            ],
            "license": ["spdx:CC0-1.0"],
        },
    )
    dandiset_id = d.identifier
    print("DANDISET ID:", dandiset_id)
    with TemporaryDirectory() as tmpdir:
        dspath = Path(tmpdir)
        (dspath / dandiset_metadata_file).write_text(f"identifier: '{dandiset_id}'\n")
        (dspath / "subdir").mkdir()

        p = dspath / "subdir" / "file.txt"
        p.write_text("This is test text.\n")
        df = dandi_file(p, dandiset_path=dspath)
        print("Uploading subdir/file.txt ...")
        df.upload(d, {"description": "A file"})

        p.unlink()
        p = dspath / "subdir" / "file.txt" / "subfile.dat"
        p.parent.mkdir()
        p.write_text("This is inner text.\n")
        df = dandi_file(p, dandiset_path=dspath)
        print("Uploading subdir/file.txt/subfile.dat ...")
        df.upload(d, {"description": "A file"})

        rmtree(dspath / "subdir")
        p = dspath / "subdir"
        p.write_text("This is outer text.\n")
        df = dandi_file(p, dandiset_path=dspath)
        print("Uploading subdir ...")
        df.upload(d, {"description": "A file"})

        print("Assets in Dandiset:")
        for asset in d.get_assets():
[...]

PR #1312, opened 2022-10-04
Optimize Asset Paths

### LINKED ISSUE #652: directory depth error
image

dandiset 108 is at least 9 directories deep. When I click on the 9th "0", the website teleports me back to level 1. I am not sure if this is a level issue or maybe there are too many elements to display or some other reason.

image

AlmightyYakob:
Closes #652

This is a performance optimization to how we handle asset paths. This implementation greatly improves read times for the asset paths endpoint, as simplifies the UI FileBrowser component.

Some notes:
* This PR is messier than I'd like due to us not having a proper asset service layer yet. I expect that will come soon, and that when it does, the asset paths service layer will become entirely internal to that.
* Since this PR adds restrictions to the Asset `path` field, any offending assets in production will need to be dealt with (likely manually) _before_ this is merged/deployed.
* After deploy, the `ingest_asset_paths` management command will need to be run, which means there will be a small period of time where the file browser is not functional.


dandiapi/api/models/asset_paths.py
```diff
@@ -0,0 +1,47 @@
...
+
+from django.db import models
+from django.db.models import Q
+from django_extensions.db.models import TimeStampedModel
+
+from dandiapi.api.models.asset import Asset
+from dandiapi.api.models.version import Version
+
+
+class AssetPath(TimeStampedModel):
```
mvandenburgh, line 11: What is the reasoning for using `TimeStampedModel` on these models? If we want to consider the modification of an `AssetPath` or `AssetPathRelation` as a modification of the `Asset` itself, I wonder if we should just directly update `AssetPath.asset.modified`. In general, I worry about introducing more unnecessary timestamps into the system considering how messy that aspect of the system is currently.

dandiapi/api/models/asset_paths.py
```diff
@@ -0,0 +1,47 @@
...
+    path = models.CharField(max_length=512)
+
+    # TODO: Determine what should be done on delete
```
mvandenburgh, line 14: I would think `models.CASCADE` would make sense here, right? If an `Asset` is deleted, there's no reason to keep any of its `AssetPath`s around any longer.

dandiapi/api/views/asset.py
```diff
@@ -571,48 +507,29 @@ def paths(self, *args, **kwargs):
...
-
-        items = folders
-        items.update(files)
-        resp = self._paginate_asset_paths(items)
-        return Response(resp)
+        # Fetch child paths
+        path: str = query_serializer.validated_data['path_prefix']
+        children_paths = search_asset_paths(path, version)
+        if children_paths is None:
+            raise serializers.ValidationError('Specified path not found.')
```
mvandenburgh, line 524: [`NotFound`](https://www.django-rest-framework.org/api-guide/exceptions/#notfound) might be a more appropriate exception to raise instead of a serializer `ValidationError`; the values provided to the serializer might be completely "valid" in that they pass validation, but still point to path that doesn't exist. Plus, it raises a 404 instead of a 400.
```suggestion
            raise NotFound('Specified path not found.')
```

mvandenburgh (commented):
This is awesome! 

How will we migrate the existing data to this new model? Would someone run the `ingest_asset_paths` management command manually?

dandiapi/api/models/asset_paths.py
AlmightyYakob: I agree, this predated our discussion regarding `TimeStampedModel`. I'll change it.

AlmightyYakob:
> How will we migrate the existing data to this new model? Would someone run the `ingest_asset_paths` management command manually?

Yes that's correct.

dandiapi/api/models/asset.py
```diff
@@ -105,8 +109,16 @@ class Status(models.TextChoices):
         VALID = 'Valid'
         INVALID = 'Invalid'
 
+    def validate_path(path: str):
+        if path.startswith('/'):
+            raise ValidationError('Path must not begin with /')
+        if not re.match(ASSET_PATH_REGEX, path):
+            raise ValidationError(f'Path violates regex: {ASSET_PATH_REGEX}')
```
danlamanna, line 116: I think we should avoid putting regular expressions into our error messages. Something like "Path is improperly formatted" is probably fine.

dandiapi/api/models/asset_paths.py
```diff
@@ -0,0 +1,47 @@
...
+    asset = models.ForeignKey(
+        Asset, null=True, blank=True, related_name='leaf_paths', on_delete=models.CASCADE
```
danlamanna, line 16: I don't have any better suggestions, but it is counter intuitive that asset paths might not have an asset. #namingthingsishard

dandiapi/api/models/asset.py
AlmightyYakob: My concern was that someone might not know what constitutes a valid asset path, and so wouldn't know how to fix their error. Any suggestions regarding that?

dandiapi/api/services/asset/paths/__init__.py
```diff
@@ -0,0 +1,110 @@
+from __future__ import annotations
```
danlamanna, line 1: I wouldn't put any of the code in this PR underneath `services`. With the exception of `search_asset_paths`, none of these are really user facing business logic. Given we're still not putting read operations into the services module either, I think all of these should be put in something like `dandiapi.api.asset_paths`.

dandiapi/api/models/asset.py
danlamanna: I agree it's a problem, but given the level of dependency on hard coded error messages right now - I don't want to create an error message with a very-likely-to-change-in-the-future regex. I think the solution is general documentation, since end users are often not going to able to decipher a regex-laden error message anyway. Either way I don't think there's any blocker to this PR right now, I would just make the error message more generic.

dandiapi/api/models/asset_paths.py
AlmightyYakob: It's tricky, since if we did that it would allow for unilaterally deleting an asset, taking the asset paths with them, but without updating the asset paths tree. E.g., if we deleted an asset with that was `5kb` in size, all parent paths would have their `aggregate_size` field +5kb larger than it should be, as well as the `aggregate_files` 1 file larger than it should be.

I'm actually inclined to to `models.PROTECT` here, since it would enforce that you need to call `delete_asset_paths` before deleting the asset.

mvandenburgh:
Now that #1006 is merged, the migrations here will need to be recreated.

dandiapi/api/models/asset.py
waxlamp: For error messages such as this (where the fix is in scope for a CS person, but perhaps not for someone in a different field) we can also put a link in the error message encouraging them to get in touch via the helpdesk.

dandiapi/api/services/asset/paths/__init__.py
AlmightyYakob: https://github.com/dandi/dandi-archive/pull/1312/commits/a2c9444961e5ef8aa3ba4cb89120beafd0544876

dandiapi/api/models/asset_paths.py
AlmightyYakob: https://github.com/dandi/dandi-archive/pull/1312/commits/9ee70ad4ffc9af64008bcae5f57073e4477cafed

dandiapi/api/models/asset_paths.py
mvandenburgh: Ah, right. Yeah I agree with using `models.PROTECT` to enforce that `delete_asset_paths` is called before deleting an asset.

AlmightyYakob: https://github.com/dandi/dandi-archive/pull/1312/commits/76eb5ee1915d6fa1d68776415ec17cfbe3682e8a

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -230,20 +244,40 @@ import { trimEnd } from 'lodash';
...
 
-const parentDirectory = '..';
 const rootDirectory = '';
-
 const FILES_PER_PAGE = 15;
 
-const sortByName = (a: AssetStats, b: AssetStats) => {
-  if (a.name > b.name) {
+// AssetService is slightly different from Service
+export interface AssetService {
```
mvandenburgh, line 253:
```suggestion
interface AssetService {
```

Does this need to be exported? I don't see it used anywhere outside of this component.

dandiapi/api/views/asset.py
AlmightyYakob: [`2bbd583` (#1312)](https://github.com/dandi/dandi-archive/pull/1312/commits/2bbd583b504ab4adadf5632526fe36f6bc79faee)

dandiapi/api/models/asset.py
AlmightyYakob: [`312d74d` (#1312)](https://github.com/dandi/dandi-archive/pull/1312/commits/312d74d45372619678d0a84ffdac211fc2b50bb2)

dandiapi/api/asset_paths.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@transaction.atomic()
+def update_asset_paths(old_asset: Asset, new_asset: Version, version: Version):
+    delete_asset_paths(old_asset, version)
+    add_asset_paths(new_asset, version)
+
+
+@transaction.atomic()
+def publish_version(draft_version: Version, published_version: Version):
```
danlamanna, line 143: Is this not called anywhere?

AlmightyYakob: No it's not, good catch. This should be called in the version publish task, I can fix that.

danlamanna:
`dandiapi/api/services/asset/__init__.py` can be deleted as well. (Can't leave a comment on it because it's empty)

dandiapi/api/asset_paths.py
AlmightyYakob: [`c171288` (#1312)](https://github.com/dandi/dandi-archive/pull/1312/commits/c171288d5f89001c7adf9eb07379aafa0c7cde9c)

dandiapi/api/tests/test_asset.py
```diff
@@ -35,130 +35,44 @@ def test_asset_blob_and_zarr(draft_asset, zarr_archive):
...
-def test_asset_rest_path(
-    api_client,
-    draft_version_factory,
-    asset_factory,
-    asset_blob_factory,
-    path,
-    asset_paths,
-    expected,
-):
+def test_asset_rest_path(api_client, draft_version_factory, asset_factory):
```
yarikoptic, line 38: I thought that the idea was to have no effects on API, so am I right to assume that this test should have still worked with all parametrizations? (I would have preferred to keep it even if you think it would not be "as useful as before")

yarikoptic:
FWIW, as I was one among those who "questioned" the efficiency of a "flat list of paths" representation when it was introduced in original DANDI design, and performance improvements are always welcome but I think the cons of such approach should also be considered. It
- raises the implementation complexity. More models: `AssetPath`, `AssetPathRelation`; necessity to make sure to call into the AssetPath datastructure operations for anything which might be changing `path` of any asset.
- adds information duplication (having "path" field and then `AssetPath*` structures to represent any particular path in the tree)
  - would need adding of some `fsck` across the models to ensure consistency. **Is this PR comes with some kind of such check function/end point which could trigger such checking? How long would it take to `fsck` the entire archive ATM**
- adds performance penalty for some operations, e.g. renaming/moving paths into another folder might require `O(depth)` operations (adjust all parents in the tree for size, [ref])https://github.com/dandi/dandi-archive/pull/1312/files#diff-8b69af35cb401375428f41693ecb95e9f791080123167e6faef32023e0928b4eR52)) and thus possibly started to ask for locking to avoid race conditions etc. **Do we have guestimate of the impact from such RF on `dandi move` in some typical case?**

An alternative, and generic, could be - caching of views based on the last-modified timestamp of the dandiset version (I thought I filed an issue but I failed to find it), on which we rely anyways already in backups and have check introduced which would alert us if there are changes without timestamp boost.

dandiapi/api/tests/test_asset.py
AlmightyYakob: > I thought that the idea was to have no effects on API, so am I right to assume that this test should have still worked with all parametrizations? (I would have preferred to keep it even if you think it would not be "as useful as before")

The parametrization was removed here as it was no longer relevant to test this from the asset tests, and similar tests exist in the asset paths tests.

waxlamp:
> * raises the implementation complexity. More models: `AssetPath`, `AssetPathRelation`; necessity to make sure to call into the AssetPath datastructure operations for anything which might be changing `path` of any asset.

It does indeed cost some complexity to gain better correctness and performance for the file browsing functionality. However, this is a case where we actually *failed* to model properly this corner of the system. In that sense, the new models are required to capture the domain logic here.

I'm not sure what you mean exactly by the need to call into AssetPath, but we can take steps as needed to make this part a bit easier. For example, if we identify which things have a need to "change `path` of any asset" (i.e., `dandi move`?) we can expose high-level API to enable those things without having to recompose complex operations in each place it's needed.

> * adds information duplication (having "path" field and then `AssetPath*` structures to represent any particular path in the tree)
>   
>   * would need adding of some `fsck` across the models to ensure consistency. **Is this PR comes with some kind of such check function/end point which could trigger such checking? How long would it take to `fsck` the entire archive ATM**

We are indeed denormalizing the path data in the name of maintaining performance. However, we are not particularly worried about loss of consistency, since the system doesn't modify the existing path data (assets must be copied first in order to change their paths, or else they need to be deleted and re-added, etc.). If we add an "fsck" type of script, it would be a maintenance script we could run to help us debug in case we see something going wrong, rather than something that would be exposed via the API.

> * adds performance penalty for some operations, e.g. renaming/moving paths into another folder might require `O(depth)` operations (adjust all parents in the tree for size, [ref])https://github.com/dandi/dandi-archive/pull/1312/files#diff-8b69af35cb401375428f41693ecb95e9f791080123167e6faef32023e0928b4eR52)) and thus possibly started to ask for locking to avoid race conditions etc. **Do we have guestimate of the impact from such RF on `dandi move` in some typical case?**
> 

The design is intended to accelerate the common case of reading path information, at a slight expense for deletion and addition operations. But the complexity is not `O(depth)`, but rather `O(1)`. The necessary updates to the database happen in about two or three bulk queries. I expect the impact on something like `dandi move` should be negligible.

Furthermore, we should talk about exactly what kind of thing `dandi move` enables. If there is a safer, faster way to support that operation, we should look into it.

> An alternative, and generic, could be - caching of views based on the last-modified timestamp of the dandiset version (I thought I filed an issue but I failed to find it), on which we rely anyways already in backups and have check introduced which would alert us if there are changes without timestamp boost.

This approach carries a lot of risk (as you know, cache invalidation is one of the [two hard things](https://martinfowler.com/bliki/TwoHardThings.html) in computer science). It is hard to guess at the performance boost this approach would give us, measured against the risk of incorrect updates, etc. On top of that, we are able to solve other problems with our current approach with an actual model of file paths (efficient updates of file sizes and file counts; pagination).

Jake's PR brings much more good than harm, and I think we should (after final reviews) deploy it to staging, play around with it, and gain some more confidence that it's working.

dandiapi/api/asset_paths.py
```diff
@@ -0,0 +1,151 @@
...
+    leaf.refresh_from_db()
+    assert leaf.aggregate_size == 0
+    assert leaf.aggregate_files == 0
+
+    # Delete leaf node and any other paths with no contained files
+    AssetPath.objects.filter(aggregate_files=0).delete()
+
+
+@transaction.atomic()
+def update_asset_paths(old_asset: Asset, new_asset: Version, version: Version):
```
danlamanna, line 142: The type of `new_asset` is incorrect, fyi @mvandenburgh this error is visible in my editor but seems to be passing tests.

danlamanna (approved):

mvandenburgh (approved):

jwodder:
@AlmightyYakob Does this resolve #1109?

AlmightyYakob:
> @AlmightyYakob Does this resolve #1109?

No, but it would be trivial to address with these new models. Since there's been so much churn on this PR, I'd like to avoid fixing it here and do a follow up afterwards.

dandiapi/api/asset_paths.py
```diff
@@ -0,0 +1,151 @@
...
+        'asset__blob',
+        'asset__embargoed_blob',
+        'asset__zarr',
+    )
+    path_ids = (
+        AssetPathRelation.objects.filter(parent=path, depth=1)
+        .values_list('child', flat=True)
+        .distinct()
+    )
+    return qs.filter(id__in=path_ids)
```
danlamanna, line 50: One last nitpick before merging. This queryset is unordered which raises a testing warning:
```
UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'dandiapi.api.models.asset_paths.AssetPath'> QuerySet.
```

PR #1315, opened 2022-10-07
Restore DLP loading bar, add skeleton loader

### LINKED ISSUE #1013: Utilize "loading states" when pages take a bit to load
When using the app, I've noticed there are times when all that shows up is the header and footer and it takes a few seconds for the content area to show up. I think using skeleton loaders would be very beneficial here.

https://vuetifyjs.com/en/components/skeleton-loaders/

mvandenburgh:
I noticed that we have a visual loading indicator on the DLP that wasn't appearing [due to a bug](https://github.com/dandi/dandi-archive/blob/master/web/src/views/DandisetLandingView/DandisetLandingView.vue#L12-L20). This PR fixes the conditional so that that loading bar appears properly. I also went ahead and added a `v-skeleton-loader` to the page as the linked issue suggests.

Closes #1013.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -10,7 +10,6 @@
       />
     </v-toolbar>
     <v-container
-      v-if="currentDandiset"
       fluid
       class="grey lighten-4 pa-0"
     >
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
Looks great! Leaving a technical review to Jake as requested.

AlmightyYakob (approved):
Looks great!

PR #1320, opened 2022-10-11
Render error message in UI when uncaught exception happens

### LINKED ISSUE #768: Show error message when interaction with dandi-api results in error
"Inspired" by https://github.com/dandi/dandi-api/issues/208
ATM there is no visible feedback to the user and it just leaves user puzzled either he/she has done something incorrectly or it is an internal issue.  I think web UI should have some popup message (which can disappear after a while) which would inform that error has happened, may be with at least the end-point and type (error code and time stamp?) to assist in locating that error in the logs later.

Also, see: https://github.com/dandi/dandiarchive/pull/544#discussion_r559791480

mvandenburgh:
Adds a [global error handler](https://v2.vuejs.org/v2/api/#errorHandler) that renders a snackbar error message whenever an uncaught exception occurs.

Closes #768.

dandi-error-handler

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -16,11 +16,33 @@
     >
       {{ SERVER_DOWNTIME_MESSAGE }}
     </v-main>
+    <v-snackbar
+      :value="showError"
+      :timeout="-1"
+      top
+      right
+      color="error"
+    >
+      <span>
+        Sorry, something went wrong on our side (the developers have been notified).
+        Please try that operation again later.
+      </span>
+      <template #action="{ attrs }">
+        <v-btn
+          color="white"
+          text
+          v-bind="attrs"
+          @click="showError = false"
+        >
+          Close
+        </v-btn>
+      </template>
+    </v-snackbar>
   </v-app>
 </template>
 
 <script setup lang="ts">
-import { onMounted, ref } from 'vue';
+import Vue, { onMounted, ref } from 'vue';
 
 import { dandiRest } from '@/rest';
 import { useDandisetStore } from '@/stores/dandiset';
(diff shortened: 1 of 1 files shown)


web/src/App.vue
```diff
@@ -16,11 +16,21 @@
...
     </v-main>
+    <v-snackbar
+      v-model="showError"
+      :timeout="6000"
+      top
+      right
+      color="error"
+    >
+      Sorry, something went wrong on our side.
+      The developers have been notified, please try that operation again later.
```
waxlamp, line 27: Just a small suggestion to avoid a run-on sentence.

```suggestion
      Sorry, something went wrong on our side
      (the developers have been notified). Please try that operation again later.
```

waxlamp (approved):
This is an elegant solution. Nice work!

The only other idea I have is to set the timeout to `-1` (so that the snackbar doesn't auto-dismiss), and include a "close" button (like the example in the v-snackbar documentation). The rationale is to alert a user to an error when they're not paying attention (e.g., they flip to a different tab and come back 10 seconds later, or get up from their computer, etc.)

I don't want that to hold up this PR, though it should be an easy fix if you decide to include it.

yarikoptic:
looks good to me, thank you @mvandenburgh ! 
Is there a dandiset on staging so I could go to https://deploy-preview-1320--gui-staging-dandiarchive-org.netlify.app/dandiset and enjoy it ? webshot demonstration seems to be on 000362 but I can't find it there

mvandenburgh:
> looks good to me, thank you @mvandenburgh ! Is there a dandiset on staging so I could go to [deploy-preview-1320--gui-staging-dandiarchive-org.netlify.app/dandiset](https://deploy-preview-1320--gui-staging-dandiarchive-org.netlify.app/dandiset) and enjoy it ? webshot demonstration seems to be on 000362 but I can't find it there

That was taken on my local development environment, where I intentionally introduced an error to trigger the alert. But, you can go to https://deploy-preview-1320--gui-staging-dandiarchive-org.netlify.app/dandiset/search?search=test&page=7&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false to see it in action (that links to some search results with an invalid page number, which currently results in an exception and is another bug on my list to fix)

PR #1338, opened 2022-10-20
Fix asset update/deletion when associated with zarr

### LINKED ISSUE #1336: New AssetPaths lead to dandi-cli integration testing failing
Merging #1312 seems has lead to failing testing against dandi-cli in master

- Given that it was green within PR, please check if that "integration" testing is actually using the code from the PR
- Tests trigger 500 from the API (this is just a sample, didn't check other, total it was 10 of them)
```
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://localhost:8000/api/dandisets/000224/versions/draft/assets/05ccca9a-7381-4666-b043-c4d2408eb334/
```

AlmightyYakob:
Closes #1336 

The reported issue was caused by the following:

* When a zarr is created and associated to an asset (but not ingested) it initially has a size of zero
* Ingest can then be performed, giving it a non-zero size
* If the containing asset is then updated / deleted, all parent paths are updated, decrementing the size of the asset. However, now that size is non-zero, and was never updated, so a possibly zero value is being decremented by a non-zero value, causing the value to be negative, and throwing an integrity error.

This PR updates the `ingest_zarr_archive` task to delete all asset paths that are associated with the zarr, and then re-add them after the ingestion is finished. This is essentially the same process as updating an asset path (path deletion and immediate addition), but with zarr ingestion in the middle.

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -160,14 +160,14 @@ def add_version_asset_paths(version: Version):
 def add_zarr_paths(zarr: ZarrArchive):
     """Add all asset paths that are associated with a zarr."""
     # Only act on draft assets/versions
-    for asset in zarr.assets.filter(published=False):
-        for version in asset.versions.filter(version='draft'):
+    for asset in zarr.assets.filter(published=False).iterator():
+        for version in asset.versions.filter(version='draft').iterator():
             add_asset_paths(asset, version)
 
 
 def delete_zarr_paths(zarr: ZarrArchive):
     """Remove all asset paths that are associated with a zarr."""
     # Only act on draft assets/versions
-    for asset in zarr.assets.filter(published=False):
-        for version in asset.versions.filter(version='draft'):
+    for asset in zarr.assets.filter(published=False).iterator():
+        for version in asset.versions.filter(version='draft').iterator():
             delete_asset_paths(asset, version)
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -6,6 +6,7 @@ from django.db import transaction
 from django.db.models import Count, F, QuerySet
 
 from dandiapi.api.models import Asset, AssetPath, AssetPathRelation, Version
+from dandiapi.zarr.models import ZarrArchive
 
 ####################################################################
 # Dandiset and version deletion will cascade to asset path deletion.
@@ -154,3 +155,19 @@ def add_version_asset_paths(version: Version):
     """Add every asset from a version."""
     for asset in version.assets.iterator():
         add_asset_paths(asset, version)
+
+
+def add_zarr_paths(zarr: ZarrArchive):
+    """Add all asset paths that are associated with a zarr."""
+    # Only act on draft assets/versions
+    for asset in zarr.assets.filter(published=False):
+        for version in asset.versions.filter(version='draft'):
+            add_asset_paths(asset, version)
+
+
+def delete_zarr_paths(zarr: ZarrArchive):
+    """Remove all asset paths that are associated with a zarr."""
+    # Only act on draft assets/versions
+    for asset in zarr.assets.filter(published=False):
+        for version in asset.versions.filter(version='draft'):
+            delete_asset_paths(asset, version)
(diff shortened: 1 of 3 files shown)


AlmightyYakob:
It seems this is still failing for some reason. Investigating...

AlmightyYakob:
@jwodder I can no longer reproduce this error locally, using your [MVCE](https://github.com/dandi/dandi-archive/issues/1336#issuecomment-1285636599). I was originally able to, but it seems that inexplicably (no local code changes), those now succeed. I've also reset my local DB to no avail. Yet, they still fail in CI.

Would #1339 fix this issue? Even the master tests fail, so it would seem the answer is no, but I'm not totally sure on the inner-workings of those tests.

Are you still able to reproduce this locally (with the changes from this branch)?

jwodder:
@AlmightyYakob Testing locally:

* The MVCEs I posted fail with a 500 error when run against staging
* The dandi-cli tests fail when run against a locally-built image of dandi-archive's master branch and pass when using this branch

Note that the CLI integration tests here currently use the master branch of dandi-archive regardless of what PR is being tested.  Once #1339 and https://github.com/dandi/dandi-cli/pull/1141 are merged, the tests will go back to using the current PR.

AlmightyYakob:
> @AlmightyYakob Testing locally:
> 
> * The MVCEs I posted fail with a 500 error when run against staging
> * The dandi-cli tests fail when run against a locally-built image of dandi-archive's master branch and pass when using this branch
> 
> Note that the CLI integration tests here currently use the master branch of dandi-archive regardless of what PR is being tested. Once #1339 and [dandi/dandi-cli#1141](https://github.com/dandi/dandi-cli/pull/1141) are merged, the tests will go back to using the current PR.

I see, that makes sense now. Thanks!

dandiapi/api/asset_paths.py
```diff
@@ -154,3 +155,19 @@ def add_version_asset_paths(version: Version):
...
+        for version in asset.versions.filter(version='draft'):
+            add_asset_paths(asset, version)
+
+
+def delete_zarr_paths(zarr: ZarrArchive):
+    """Remove all asset paths that are associated with a zarr."""
+    # Only act on draft assets/versions
+    for asset in zarr.assets.filter(published=False):
+        for version in asset.versions.filter(version='draft'):
+            delete_asset_paths(asset, version)
```
mvandenburgh, line 173: Would it be worth using `.iterator()` for all the querysets here? As long as these functions are only called inside tasks, I don't see any downsides to doing so. In normal operations, I don't think we'd ever have a `ZarrArchive` with enough `Assets` (or one of those `Assets` with enough `Versions`) to make these querysets large enough to blow up memory, but theoretically a malicious actor could create one like that and crash the worker instance.

@AlmightyYakob  @danlamanna any thoughts on this? Maybe this is just super-paranoia on my part from all the worker memory issues we had in the past due to iterating over large querysets w/out using `.iterator()`.

mvandenburgh (approved):
One question about using `.iterator()`, but this LGTM. I do think that it's not ideal to have to delete and recreate the asset paths on zarr ingest, but this seems more like a flaw in the overall design of Zarr/how we kind of consider zarrs `Assets`, but not really in some places.

dandiapi/api/asset_paths.py
AlmightyYakob: I'm okay with it. I don't see any real downsides in doing so.

yarikoptic:
> This PR updates the `ingest_zarr_archive` task to delete all asset paths that are associated with the zarr, and then re-add them after the ingestion is finished. 

are we talking about paths **within** zarr, ie thousands of them?  do you have an estimate of runtime necessary to delete and add e.g. 200k paths?

AlmightyYakob:
> are we talking about paths **within** zarr, ie thousands of them? do you have an estimate of runtime necessary to delete and add e.g. 200k paths?

No, we don't store the paths within zarr files. This is referring to all of the assets that point to zarrs. Zarrs have to be handled differently here, since zarr ingestion can update the asset size, yet subverts the normal asset update process.

PR #1339, opened 2022-10-21
Disable pulling of all Docker images from Docker Hub

jwodder:
See https://github.com/dandi/dandi-archive/issues/1336#issuecomment-1286926124 etc.; depends on https://github.com/dandi/dandi-cli/pull/1141.

### DIFF
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -47,6 +47,7 @@ jobs:
           - master
     env:
       DANDI_ALLOW_LOCALHOST_URLS: 1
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
So just to clarify, prior to this change, the docker image was being built locally, but was then not being used, as an image from dockerhub was pulled and used instead?

jwodder:
@AlmightyYakob Correct.  And this started when https://github.com/dandi/dandi-cli/pull/1121 was merged on the 3rd.

AlmightyYakob (approved):

PR #1342, opened 2022-10-24
Add file upload instructions to file browser UI

### LINKED ISSUE #731: Finish DLP redesign implementation
- [x] dandi/dandiarchive#951
- [x] dandi/dandiarchive#863
- [x] dandi/dandi-api#706  
- [x] #732

### LINKED ISSUE #732: Implement "Add files" DLP design
[Figma](https://www.figma.com/file/rBobiAvSLCsJmyYUYmU1OX/DLP?node-id=504%3A33148)

mvandenburgh:
Closes #732 
Closes #731 
Screenshot from 2022-10-24 14-01-28

### DIFF
--- a/web/src/components/FileBrowser/FileUploadInstructions.vue
+++ b/web/src/components/FileBrowser/FileUploadInstructions.vue
@@ -34,20 +34,13 @@
         <span class="text-subtitle-1">Don't have DANDI CLI?</span>
         <div>
           <span class="text-body-2 grey--text text--darken-1">
-            Install the Python client (DANDI CLI) in a Python 3.7+ environment using command:
+            <span class="text-body-2 grey--text text--darken-1">
+              Follow the installation instructions in the
+              <a href="https://www.dandiarchive.org/handbook/10_using_dandi/#dandi-python-client">
+                DANDI handbook
+              </a> .
+            </span>
           </span>
-        </div>
-        <div
-          class="d-flex justify-center"
-          style="font-family: monospace;"
-        >
-          <v-sheet
-            color="black"
-            width="60%"
-            class="white--text pl-2 py-1 text-left"
-          >
-            <span>> pip install "dandi>=0.13.0"</span>
-          </v-sheet>
         </div>
       </div>
     </v-row>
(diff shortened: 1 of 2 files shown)


web/src/components/FileBrowser/FileUploadInstructions.vue
```diff
@@ -0,0 +1,62 @@
...
+    </v-row>
+  </v-sheet>
+</template>
+
+<script setup lang="ts">
+import { computed } from 'vue';
+import { useDandisetStore } from '@/stores/dandiset';
+
+const store = useDandisetStore();
+const dandisetIdentifier = computed(() => store.dandiset?.dandiset.identifier);
```
waxlamp, line 61: There's a small typing problem here. `dandisetIdentifier` has type `string | undefined` but you're assuming it's defined at the point of use in the template.

waxlamp (approved):
If you can fix the typing issue, that would be good to include here. Otherwise, because this is a PR to non-master branch, I'm fine with you merging it pending a bigger review of that branch when it goes to master.

danlamanna:
To avoid the installation instruction versions becoming stale, should we link to https://www.dandiarchive.org/handbook/10_using_dandi/#dandi-python-client instead?

web/src/components/FileBrowser/FileUploadInstructions.vue
mvandenburgh: Fixed https://github.com/dandi/dandi-archive/pull/1342/commits/5c98f19e2097faa605b16ebf80de5a3a21c86989

mvandenburgh:
> To avoid the installation instruction versions becoming stale, should we link to [dandiarchive.org/handbook/10_using_dandi/#dandi-python-client](https://www.dandiarchive.org/handbook/10_using_dandi/#dandi-python-client) instead?

Good idea, I think that makes more sense. I added that link and a link to the file upload instructions.

mvandenburgh:
Updated to only change the text about installing the CLI.

waxlamp:
> @waxlamp and @mvandenburgh - given this is a web app, how about we create the text that it reads in some location separate from the web app, and it can pull the info, which we can update async from development of the app.
> 

If this text were going to be in rapid flux (or if it turns out to be like that) I think we could consider this approach. For a first implementation, I prefer the simplicity of what we have in this PR.

> btw, i love this feature.

🥳

danlamanna (approved):

PR #1343, opened 2022-10-24
Optimize version/zarr asset path ingestion

AlmightyYakob:
This unblocks asset path ingestion in staging/prod. Previously, transactions were being used in such an inefficient way that ingestion would essentially halt.

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+import contextlib
 import os
 
-from django.db import transaction
-from django.db.models import Count, F, QuerySet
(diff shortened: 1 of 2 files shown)


waxlamp (approved):

danlamanna:
I'd like to take a look at this more closely this afternoon. If it's urgent feel free to merge it now and I can revisit it later.

danlamanna (commented):
I think managing transaction control by passing a boolean value around is a bit surprising. It's also wrapping the entire function body in a transaction, rather than isolating some specific portion of the logic, so it doesn't add much value to a caller. I think we should either drop the flag altogether, or determine which behavior ought to be conditional (bulk updating of AssetPaths or something like that).

PR #1348, opened 2022-10-27
Disable APPROVE button if user is already approved

### LINKED ISSUE #1341: disable user buttons on review page
very low priority, but should be very little effort as well. 

on the review user admin page, it would be nice to disable accept button for accepted users.

waxlamp:
This conditionally disables the approve button when a user's status is `APPROVED`.

The `REJECT` button remains enabled in case we need to revoke approval from a user for some reason (attn: @satra, is this correct?)

Closes #1341.

### DIFF
--- a/dandiapi/api/templates/dashboard/user_approval.html
+++ b/dandiapi/api/templates/dashboard/user_approval.html
@@ -80,7 +80,11 @@
         <form method="POST">
           {% csrf_token %}
           <input type="hidden" name="status" value="REJECTED">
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
Should we also disable the REJECT button when the user is already rejected?

waxlamp:
> ~Should we also disable the REJECT button when the user is already rejected?~ oh i see you've already brought that up in #1341

I was actually asking what to do with the REJECT button for approved users (before realizing it should stay enabled), but I hadn't thought of disabling it for rejected users... added in 2d76f2d.

Issue #1347, opened 2022-10-27
incorrect zarr digest for "Complete" fc3f8c83-f4a9-48e4-a67c-aa24b530f82a


yarikoptic:
We have both API reporting
```
(dandisets) dandi@drogon:/mnt/backup/dandi/tmp/fc3f8c83-f4a9-48e4-a67c-aa24b530f82a$ curl --silent -X 'GET'   'https://api.dandiarchive.org/api/zarr/fc3f8c83-f4a9-48e4-a67c-aa24b530f82a/' | jq . 
{
  "name": "sub-MITU01/ses-20220321h14m52s06/micr/sub-MITU01_ses-20220321h14m52s06_sample-23_stain-LEC_run-1_chunk-4_SPIM.ome.zarr",
  "dandiset": "000108",
  "zarr_id": "fc3f8c83-f4a9-48e4-a67c-aa24b530f82a",
  "status": "Complete",
  "checksum": "edfb9e937142aceca60acef5fe5ba98a-4845--7049299697",
  "upload_in_progress": true,
  "file_count": 4845,
  "size": 7049299697
}
```
and 
```
(dandisets) dandi@drogon:/mnt/backup/dandi/tmp/fc3f8c83-f4a9-48e4-a67c-aa24b530f82a$ s3cmd -q -c ~/.s3cfg-dandi-backup get s3://dandiarchive/zarr-checksums/fc3f8c83-f4a9-48e4-a67c-aa24b530f82a/.checksum -  2>/dev/null
[...]

PR #1349, opened 2022-10-28
ENH: expose force for zarr_ingest API call

yarikoptic:
Reference for the need:  https://github.com/dandi/dandi-archive/issues/1347#issuecomment-1294933546

TODOs if decide to proceed
- [ ] check that help_text is the right way to go to document it. (if it is -- why other similar fields are not documented anywhere?)
- [ ] add test
- [ ] ensure correct response/reaction upon use by non-admin user (ATM taken the same behavior as for another admin-only operation of creating dandiset with specified id)


waxlamp:
It turns out we already have appropriate management commands in the codebase to initiate forced reingestion. Rather than update the API to do something we can already do as admins, I'm going to close this PR.

yarikoptic:
how can we do this as admins @waxlamp?

yarikoptic:
FTR, I have tried in admin interface but that didn't work as I outlined in https://github.com/dandi/dandi-archive/issues/1347#issuecomment-1296010873 .

waxlamp:
There is a management command that does the force reingestion. The admin console only supports *non*-force reingestion (i.e., for Zarrs that have not yet been ingested at all). As such, it is not a very useful feature, and in the name of simplification we may even think about removing it (probably after we improve the general process for Zarrs, such that we won't need a feature like this as often).

In terms of yesterday's discussion: the management command is the technician's secret button to initiate force reingestion (which means "jumping wires on the circuit board with a needle" isn't actually necessary), while the admin console has something akin to the "close door" button in an elevator.

PR #1354, opened 2022-11-04
Fix oauth client error

### LINKED ISSUE #1353: User can't login from DLP
looks as horrible as  just

```
Error: invalid_request
Mismatching redirect URI.
```
on the white screen.     To reproduce -- logout and go to https://dandiarchive.org/dandiset/000115/draft or https://dandiarchive.org/dandiset/000115/0.210914.1732 and click on "Login with Github" button on top right.   Then I was not able to login even from https://dandiarchive.org/dandiset/000115  or https://dandiarchive.org/ until I force-refreshed that https://dandiarchive.org/ page.

mvandenburgh:
[The `girder-oauth-client` library defaults to `window.location` for the OAuth redirect URL](https://github.com/girder/girder-oauth-client/blob/master/src/oauth-client.ts#L22). That means when trying to log in from, for example, https://dandiarchive.org/dandiset/000108, the oauth client was sending `https://dandiarchive.org/dandiset/000108` as the redirect URL. 

Note that this approached worked fine before #997 since we were using "fake" hash-based URLs, but now that we're using history mode, `https://dandiarchive.org/dandiset/000108` != `https://dandiarchive.org/` for redirect URLs.

This PR updates our oauth client to always use the current URL's `origin` instead of the entire URL. Note, explicitly specifying the redirect URL was introduced in girder-oauth-client 0.8.0, so this PR also updates us to that version.

Fixes #1353

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.7",
-    "@girder/oauth-client": "^0.7.7",
+    "@girder/oauth-client": "^0.8.0",
     "@koumoul/vjsf": "^2.5.2",
     "@sentry/integrations": "^7.13.0",
     "@sentry/tracing": "^7.13.0",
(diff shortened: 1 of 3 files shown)


waxlamp (approved):
Code looks great and it looks like it works!

As a followup we should look at bouncing a freshly logged-in user to the page they were on when they logged in.

PR #1360, opened 2022-11-08
Separate backend CI into separate jobs

mvandenburgh:
IMO it's annoying to have to open the job logs and sift through the entire log output from flake8, mypy, pytest, and the migrations check just to figure out why one of them failed. Any objections to running these in separate jobs so it's easy to distinguish? This will make it easier to tell at-a-glance which one failed, and also declutter the log outputs.


danlamanna (approved):

waxlamp:
Good idea, no objections.

AlmightyYakob (approved):

PR #1363, opened 2022-11-10
Internal publishing service

mvandenburgh:
Refactors publishing into a service layer module.

### DIFF
--- a/dandiapi/api/services/publish/__init__.py
+++ b/dandiapi/api/services/publish/__init__.py
@@ -1,0 +1,138 @@
+from django.contrib.auth.models import User
+from django.db import transaction
+
+from dandiapi.api import doi
+from dandiapi.api.asset_paths import add_version_asset_paths
+from dandiapi.api.models import Asset, Dandiset, Version
+from dandiapi.api.services.exceptions import NotAllowed
+from dandiapi.api.services.publish.exceptions import (
+    DandisetAlreadyPublished,
+    DandisetAlreadyPublishing,
+    DandisetBeingValidated,
+    DandisetInvalidMetadata,
+    DandisetNotLocked,
+    DandisetValidationPending,
+)
+from dandiapi.api.tasks import write_manifest_files
+
+
+def _lock_dandiset_for_publishing(*, user: User, dandiset: Dandiset) -> None:
+    """
+    Prepare a dandiset to be published by locking it and setting its status to PUBLISHING.
+
+    This function MUST be called before _publish_dandiset is called.
+    """
+    if (
+        not user.has_perm('owner', dandiset)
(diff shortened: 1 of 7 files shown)


mvandenburgh:
whoops I forgot to check if tests were passing before opening this...

dandiapi/api/services/publish/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+def publish_dandiset(*, user: User, dandiset: Dandiset) -> None:
+    from dandiapi.api.tasks import publish_dandiset_task
+
+    if (
+        not user.has_perm('owner', dandiset)
+        or dandiset.embargo_status != Dandiset.EmbargoStatus.OPEN
+    ):
+        raise NotAllowed()
+    if dandiset.zarr_archives.exists() or dandiset.embargoed_zarr_archives.exists():
+        raise NotAllowed('Cannot publish dandisets which contain zarrs.')
```
AlmightyYakob, line 132: It seems this exact code block exists at the beginning of `_lock_dandiset_for_publishing`, which immediately follows here. So it seems this check is redundant? I'd understand if it had to stay to preserve order of operations, but it seems that wouldn't be impacted by its removal.

AlmightyYakob (changes requested):
One comment but overall looks good. The structure does bug me a bit (bouncing between the publish service and tasks files), but I'm not sure of a good way to address that.

dandiapi/api/tests/test_version.py
```diff
@@ -654,8 +654,8 @@ def test_version_rest_publish_zarr(
         f'/api/dandisets/{draft_version.dandiset.identifier}'
         f'/versions/{draft_version.version}/publish/'
     )
-    assert resp.status_code == 400
-    assert resp.json() == ['Cannot publish dandisets which contain zarrs']
+    assert resp.status_code == 403
```
danlamanna, line 657: Just a note to back out this one change we discussed.

mvandenburgh: :+1:

dandiapi/api/services/publish/__init__.py
mvandenburgh: [`ac75bf5` (#1363)](https://github.com/dandi/dandi-archive/pull/1363/commits/ac75bf5dd61958debc0ac32d50e4be81b39be850)

mvandenburgh:
> The structure does bug me a bit (bouncing between the publish service and tasks files), but I'm not sure of a good way to address that.

Yeah I'm still not completely happy with how that turned out :(
Definitely open to suggestions if you think of any.

AlmightyYakob (approved):

Issue #1355, opened 2022-11-07
assets can still be missing sha256


yarikoptic:
Ran into it while trying to force backup of 000350 to overcome outstanding issue of not updated modified timestamp on dandiset, see https://github.com/dandi/dandisets/issues/289

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ flock -E 0 -e -n /home/dandi/.run/backup2datalad-cron.lock bash -c '/mnt/backup/dandi/dandisets/tools/backups2datalad-update-cron --mode force 000350'
2022-11-07T09:00:47-0500 [ERROR   ] backups2datalad: Dandiset 000350: sub-20170228-4/sub-20170228-4_ses-20170228T185002_ophys.nwb: Asset created more than a day ago but SHA256 digest has not yet been computed
2022-11-07T09:00:47-0500 [ERROR   ] backups2datalad: Dandiset 000350: sub-20161109-2/sub-20161109-2_ses-20161109T211950_ophys.nwb: Asset created more than a day ago but SHA256 digest has not yet been computed
add .dandi/assets.json (non-large file; adding content to git repository) ok
(recording state in git...)
2022-11-07T09:00:47-0500 [ERROR   ] backups2datalad: Job failed on input <Dandiset 000350/draft>:
Traceback (most recent call last):
...
 File "/mnt/backup/dandi/dandisets/tools/backups2datalad/asyncer.py", line 105, in check
    raise RuntimeError(
RuntimeError: Errors occurred while downloading: 2 assets on server had no SHA256 hash despite advanced age
Traceback (most recent call last):
...
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/datasetter.py", line 97, in update_from_backup
    raise RuntimeError(
RuntimeError: Backups for 1 Dandiset failed
```

and indeed

```
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/dandisets/000350/versions/draft/assets/?metadata=true' -H 'accept: application/json' |  jq '.results[] | (.path, .modified, .metadata.digest)'
"sub-20170113-2/sub-20170113-2_ses-20170113T124907_ophys.nwb"
"2022-11-07T13:20:39.119231Z"
{
  "dandi:sha2-256": "86670927191b2e726d271ebaba38e5ae47457982efce7a13c6ab1fdb4cdacf70",
  "dandi:dandi-etag": "cdf34cc88fb19310f7d9e9712a525d4b-4976"
}
[...]

PR #1366, opened 2022-11-14
Save draft versions during asset metadata validation

AlmightyYakob:
This fixes a bug that was discovered in [this](https://github.com/dandi/dandi-archive/issues/1355) issue.

The previous behavior was to save the asset without also saving any draft versions, resulting in a mismatched `modified` timestamp between the asset and the version.


yarikoptic:
From what I know/see -- looks LGTM to me ;) thank you @AlmightyYakob

mvandenburgh (approved):

yarikoptic:
is there some quick way to identify dandisets where `modified` is earlier than latest `modified` of its assets so they could be "touched" ? I guess I could manually touch some I know had such assets (see https://github.com/dandi/dandisets/issues/289) but would prefer automated way if possible to be done quickly/uniformly.

PR #1367, opened 2022-11-15
Add embargo service

AlmightyYakob:
This refactors the embargo/unembargo workflow into a service.

### DIFF
--- a/dandiapi/api/services/embargo/__init__.py
+++ b/dandiapi/api/services/embargo/__init__.py
@@ -1,8 +1,10 @@
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.db.models import QuerySet
 
 from dandiapi.api.copy import copy_object_multipart
 from dandiapi.api.models import Asset, AssetBlob, Dandiset, Upload, Version
+from dandiapi.api.services.asset.exceptions import DandisetOwnerRequired
 from dandiapi.api.tasks import unembargo_dandiset_task
 
 from .exceptions import AssetNotEmbargoed, DandisetNotEmbargoed
@@ -67,9 +69,12 @@ def _unembargo_dandiset(dandiset: Dandiset):
     dandiset.save()
 
 
-def unembargo_dandiset(dandiset: Dandiset):
+def unembargo_dandiset(*, user: User, dandiset: Dandiset):
     """Unembargo a dandiset by copying all embargoed asset blobs to the public bucket."""
     if dandiset.embargo_status != Dandiset.EmbargoStatus.EMBARGOED:
         raise DandisetNotEmbargoed
+
+    if not user.has_perm('owner', dandiset):
+        raise DandisetOwnerRequired()
 
     unembargo_dandiset_task.delay(dandiset.id)
--- a/dandiapi/api/services/embargo/__init__.py
+++ b/dandiapi/api/services/embargo/__init__.py
@@ -1,0 +1,75 @@
+from django.conf import settings
+from django.db.models import QuerySet
+
+from dandiapi.api.copy import copy_object_multipart
+from dandiapi.api.models import Asset, AssetBlob, Dandiset, Upload, Version
+from dandiapi.api.tasks import unembargo_dandiset_task
+
+from .exceptions import AssetNotEmbargoed, DandisetNotEmbargoed
+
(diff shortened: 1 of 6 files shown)


dandiapi/api/tasks/__init__.py
```diff
@@ -163,27 +162,11 @@ def delete_doi_task(doi: str) -> None:
...
-    """Unembargo a dandiset by copying all embargoed asset blobs to the public bucket."""
-    dandiset: Dandiset = Dandiset.objects.get(id=dandiset_id)
-
-    # Only the draft version is needed, since embargoed dandisets can't be published
-    draft_version: Version = dandiset.draft_version
-    embargoed_assets: QuerySet[Asset] = draft_version.assets.filter(embargoed_blob__isnull=False)
-
-    # Unembargo all assets
-    for asset in embargoed_assets.iterator():
-        asset.unembargo()
```
danlamanna, line 176: Is there a reason you're not removing this model method?

dandiapi/api/services/embargo/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        {'schemaKey': 'AccessRequirements', 'status': 'dandi:OpenAccess'}
+    ]
+    draft_version.save()
+
+    # Set access on dandiset
+    dandiset.embargo_status = Dandiset.EmbargoStatus.OPEN
+    dandiset.save()
+
+
+def unembargo_dandiset(dandiset: Dandiset):
```
danlamanna, line 70: I think this should still take a user and check for permissions.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+from django.db.models import QuerySet
+
+from dandiapi.api.copy import copy_object_multipart
+from dandiapi.api.models import Asset, AssetBlob, Dandiset, Upload, Version
+from dandiapi.api.tasks import unembargo_dandiset_task
+
+from .exceptions import AssetNotEmbargoed, DandisetNotEmbargoed
+
+
+def unembargo_asset(asset: Asset):
```
danlamanna, line 11: I might `_` this for now, otherwise it's inviting people to call it without going through a method that does the proper permission checking. It also makes sense from the business logic perspective, we don't allow a user to unembargo an individual asset.

danlamanna (changes requested):

dandiapi/api/tasks/__init__.py
AlmightyYakob: Nope, thanks for catching that.

dandiapi/api/services/embargo/exceptions.py
```diff
@@ -0,0 +1,18 @@
...
+    http_status_code = status.HTTP_400_BAD_REQUEST
+    message = 'Only embargoed assets can be unembargoed.'
+
+
+class DandisetNotEmbargoed(DandiException):
+    http_status_code = status.HTTP_400_BAD_REQUEST
+    message = 'Dandiset not embargoed'
+
+
+class EmbargoedDandisetOwnerRequired(DandiException):
```
danlamanna, line 16: I think we can just use `DandisetOwnerRequired` for now.

danlamanna (approved):
LGTM!

dandiapi/api/services/embargo/exceptions.py
AlmightyYakob: There's an existing `DandisetOwnerRequired` in the asset service
https://github.com/dandi/dandi-archive/blob/6f2276d0c57e422008b7ebc80eceabde0a4827ba/dandiapi/api/services/asset/exceptions.py#L6-L7

so I was trying to avoid naming conflicts / confusion, but I'd also prefer that name. If you think it's not a concern I'll change it.

danlamanna: Oh, I was suggesting we use that one. It's a little odd but I think this exception is really a global one that just happened to get used in the asset module first.

AlmightyYakob: Sounds good to me. Maybe if we ever have more shared exceptions in the future we can factor them out into a common exceptions module.

PR #1368, opened 2022-11-15
Properly handle conflicting asset paths

### LINKED ISSUE #1109: Archive allows the same asset path to be both a directory and a file

AlmightyYakob:
Closes #1109

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -42,20 +42,45 @@ def get_root_paths(version: Version) -> QuerySet[AssetPath]:
     )
 
 
-def get_path_children(path: AssetPath) -> QuerySet[AssetPath]:
-    """Get all direct children from an existing path."""
-    qs = AssetPath.objects.select_related(
-        'asset',
-        'asset__blob',
-        'asset__embargoed_blob',
-        'asset__zarr',
+def get_path_children(path: AssetPath, depth: int | None = 1) -> QuerySet[AssetPath]:
+    """
+    Get all children from an existing path.
+
+    By default, returns only the direct children.
+    If depth is `None`, all children will be returned, regardless of depth.
+    """
+    relation_qs = AssetPathRelation.objects.filter(parent=path).exclude(child=path)
+    if depth is not None:
+        relation_qs = relation_qs.filter(depth=depth)
+
+    path_ids = relation_qs.values_list('child', flat=True).distinct()
+    return (
+        AssetPath.objects.select_related(
+            'asset',
+            'asset__blob',
+            'asset__embargoed_blob',
+            'asset__zarr',
+        )
+        .filter(id__in=path_ids)
+        .order_by('path')
     )
-    path_ids = (
-        AssetPathRelation.objects.filter(parent=path, depth=1)
-        .values_list('child', flat=True)
-        .distinct()
+
+
+def get_conflicting_paths(path: str, version: Version) -> list[str]:
+    """Return a list of existing paths that conflict with the given path."""
+    # Check that this path isn't already occupied by a "folder"
+    # Database constraints ensure this queryset could return at most 1 result
+    folder = AssetPath.objects.filter(version=version, path=path, asset__isnull=True).first()
+    if folder:
+        return list(get_path_children(folder, depth=None).values_list('path', flat=True))
+
+    # Check that this path doesn't conflict with any existing "files"
+    nodepaths = extract_paths(path)[:-1]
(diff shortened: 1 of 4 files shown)


dandiapi/api/asset_paths.py
```diff
@@ -16,6 +18,14 @@
 ####################################################################
 
 
+class ConflictingAssetPath(DandiException):
+    http_status_code = status.HTTP_409_CONFLICT
+
+    def __init__(self, new_path: str, existing_paths: list[str]) -> None:
+        message = f'Path of new asset "{new_path}" conflicts with existing assets: {existing_paths}'
+        super().__init__(message)
```
AlmightyYakob, line 26: @danlamanna I'm a bit conflicted about what to do here. This is something we want to handle and return a valid API error for, so using the service exceptions makes sense, but asset paths aren't currently a service. Do you think it's okay to be using the service layer exceptions here?

danlamanna:
Adding reference to #1109.

I think this is a problem with the paths of assets, not asset paths, right :confused:? I think the validation should occur outside of asset_paths, ideally in `add_asset_to_version` [next to the duplicate asset path check](https://github.com/dandi/dandi-archive/blob/be8559682490845ac07cffddfebf0691b72345d5/dandiapi/api/services/asset/__init__.py#L114-L116). Otherwise the asset_paths code is responsible for some portions of asset/version validation in addition to creating a performant data structure for asset lookups. I'm not as familiar with this code though so let me know if I got something wrong.

AlmightyYakob:
> I think this is a problem with the paths of assets, not asset paths, right confused? I think the validation should occur outside of asset_paths, ideally in `add_asset_to_version` [next to the duplicate asset path check](https://github.com/dandi/dandi-archive/blob/be8559682490845ac07cffddfebf0691b72345d5/dandiapi/api/services/asset/__init__.py#L114-L116). Otherwise the asset_paths code is responsible for some portions of asset/version validation in addition to creating a performant data structure for asset lookups. I'm not as familiar with this code though so let me know if I got something wrong.

Good point, I think it probably is best if the error is raised in `add_asset_to_version`, although I think the best way to perform such a check would be to use the asset paths model, via some utility function.

danlamanna (approved):

PR #1372, opened 2022-11-21
Increase asset validation celery task timeout

mvandenburgh:
Now that we're also saving every associated version, this might take longer than 10 seconds.

The main contributor to the potentially long validation time is the fact that we loop over every `Asset` in a `Version` when it's being saved https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py#L237-L242.


mvandenburgh:
Superseded by #1373

PR #1379, opened 2022-11-28
Move metadata validation tasks into the service layer

mvandenburgh:
This is the start of a metadata service layer module. For this first PR, I've just moved the `validate_asset_metadata` and `validate_version_metadata` tasks into a `metadata` service.

The goal is to eventually have all metadata changes go through the service layer, so there's definitely other metadata-related functionality that can be moved into this service; but, I wanted to start with what are hopefully the least controversial changes.

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -1,0 +1,107 @@
+from celery.utils.log import get_task_logger
+import dandischema.exceptions
+from dandischema.metadata import validate
+from django.db import transaction
+from django.utils import timezone
+import jsonschema.exceptions
+
+from dandiapi.api.models import Asset, Version
+from dandiapi.api.services.metadata.exceptions import AssetHasBeenPublished, VersionHasBeenPublished
+
+logger = get_task_logger(__name__)
+
+
+def _encode_pydantic_error(error) -> dict[str, str]:
+    return {'field': error['loc'][0], 'message': error['msg']}
+
+
+def _encode_jsonschema_error(error: jsonschema.exceptions.ValidationError) -> dict[str, str]:
+    return {'field': '.'.join([str(p) for p in error.path]), 'message': error.message}
+
+
+def _collect_validation_errors(
+    error: dandischema.exceptions.ValidationError,
+) -> list[dict[str, str]]:
+    if type(error) is dandischema.exceptions.PydanticValidationError:
+        encoder = _encode_pydantic_error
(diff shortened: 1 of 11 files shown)


dandiapi/api/services/metadata/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        raise error
+    return [encoder(error) for error in error.errors]
+
+
+def validate_asset_metadata(*, asset_id: int) -> None:
+    logger.info('Validating asset metadata for asset %s', asset_id)
+    asset: Asset = Asset.objects.get(id=asset_id)
+
+    # Published assets are immutable
+    assert not asset.published
```
waxlamp, line 38: Is a runtime assertion the right way to do this? Why not raise a customized exception or something else that can carry some semantic information?

dandiapi/api/services/metadata/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+    with transaction.atomic():
+        version.status = Version.Status.VALIDATING
+        version.save()
+
+        try:
+            publish_version = version.publish_version
+            metadata = publish_version.metadata
+
+            # Inject a dummy DOI so the metadata is valid
+            metadata['doi'] = '10.80507/dandi.123456/0.123456.1234'
```
waxlamp, line 85: Is there instead a way to signal to the `validate()` function to ignore a missing DOI?

waxlamp (commented):
Just some questions that are meant for discussion/education. Looks like a good improvement overall.

dandiapi/api/services/metadata/__init__.py
mvandenburgh: Not in its current form.

I didn't attempt to change any internal behavior of the validation functions for this PR, but eventually I think it's worth revisiting how we're validating version metadata. Right now we basically "fake publish" versions and validate against what the published metadata would be; maybe we should have a way to explicitly validate draft versions.

waxlamp: Ok, something to think about later. Moving things to the service layer makes this kind of consideration easier to analyze in the future, so right now I do agree with your minimal-change approach. Thanks!

dandiapi/api/services/metadata/__init__.py
mvandenburgh: Yes, we should have a customized exception here, this was a temporary solution until we determined whether or not this check makes sense

danlamanna (approved):

PR #1385, opened 2022-11-30
Prevent locking dandisets during zarr upload

danlamanna:
Zarr upload steps unintentionally lock the dandiset it's operating on because of a `select_related`. This could cause unnecessary blocking for parallel processes that want to perform zarr uploads on the same dandiset.


AlmightyYakob (approved):
Good catch! LGTM

mvandenburgh (approved):

PR #1387, opened 2022-11-30
Optimize `ingest_zarr_archive` task

AlmightyYakob:
This optimization is achieved in two main ways:

1. Using a priority queue (heap) in the `ZarrChecksumModificationQueue` class, instead of a normal queue
2. Storing all s3 files in the queue before doing any processing (see below for memory usage).

The main benefit this provides is being able to update the checksum tree strictly from the bottom up, removing duplicate file updates.

Previously, a batch of files were fetched, immediately processed, and the next batch fetched, etc. This resulted in all parent checksum files being updated (up to the root checksum file) in every batch. The optimized method only updates each checksum file once.

#### Results
The outcome of this optimization is an almost 2x speed improvement. For a zarr archive with ~37k files (52.5GB), I observed an average runtime of ~83s, compared to a previous runtime of ~147s.

In regards to memory usage, it seems the max usage was ~411kb for the same zarr with ~37k files, decreasing as the ingestion goes on. This was the largest zarr I was able to test on, since this is the largest zarr in staging. Also, to clarify, this testing was done by using a local shell pointed at the staging database and s3 bucket, so these times should be representative of what we would observe in production.

### DIFF
--- a/dandiapi/zarr/checksums.py
+++ b/dandiapi/zarr/checksums.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
+import heapq
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -132,6 +133,9 @@ class ZarrChecksumModification:
     directories_to_update: list[ZarrChecksum] = field(default_factory=list)
     paths_to_remove: list[str] = field(default_factory=list)
 
+    def __lt__(self, other):
+        return str(self.path) < str(other.path)
+
 
 class ZarrChecksumModificationQueue:
     """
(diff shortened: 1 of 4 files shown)


dandiapi/zarr/checksums.py
```diff
@@ -143,38 +147,44 @@ class ZarrChecksumModificationQueue:
...
-        self._queue[key].directories_to_update.append(checksum)
+        self._get_path(key).directories_to_update.append(checksum)
 
     def queue_removal(self, key: Path, path: Path | str):
-        if key not in self._queue:
-            self._queue[key] = ZarrChecksumModification(path=key)
-        self._queue[key].paths_to_remove.append(str(path))
+        self._get_path(key).paths_to_remove.append(str(path))
 
     def pop_deepest(self):
```
mvandenburgh, line 178: Not strictly needed, but fwiw mypy will not type-check functions that don't have any type annotations for their arguments or return values.
```suggestion
    def pop_deepest(self) -> ZarrChecksumModification:
```

dandiapi/zarr/tasks/__init__.py
```diff
@@ -131,7 +132,8 @@ def ingest_zarr_archive(
 
         # Instantiate updater and add files as they come in
         empty = True
-        updater = SessionZarrChecksumUpdater(zarr_archive=zarr)
+        queue = ZarrChecksumModificationQueue()
+        print(f'Fetching files for zarr {zarr.zarr_id}...')
```
mvandenburgh, line 136: If we want to log this, I think `logger.info` is more appropriate than a print statement
```suggestion
        logger.info(f'Fetching files for zarr {zarr.zarr_id}...')
```

AlmightyYakob:
FYI it seems that django-stubs was updated last night, and so caused a small mypy error which [9848539](https://github.com/dandi/dandi-archive/pull/1387/commits/984853900195e752953b19cf19df2d06994c233b) addresses. It's a pretty minor change, so I've included it here.

PR #1391, opened 2022-12-01
Simplify ingest_zarr_archive task

AlmightyYakob:
Depends on #1387 and #1390

This PR simplifies the zarr ingest task in the following ways:
* Don't explicitly count files/size, just retrieve from the checksum afterwards
* Remove `no_checksum`, `no_count`,  `no_size` flags

### DIFF
--- a/dandiapi/zarr/checksums.py
+++ b/dandiapi/zarr/checksums.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


dandiapi/zarr/checksums.py
```diff
@@ -21,6 +22,18 @@
...
+
+
+def parse_checksum_string(checksum: str) -> tuple[int, int]:
+    """Return a tuple of (file count, total size)."""
+    match = re.match(ZARR_CHECKSUM_REGEX, checksum)
+    if match is None:
+        raise Exception('Invalid zarr checksum provided.')
+
+    count, size = match.groups()
+    return (int(count), int(size))
```
mvandenburgh, line 35: This is subjective, but I wonder if we should just have two `@property`s on the Zarr model, `file_count` and `total_size`, instead of a function that we call exactly once. I think there's a lot of places we use properties incorrectly on our models, but these two values seem like a perfect use case as they are derived purely from the `checksum` field (usually I use this rationale https://github.com/HackSoftware/Django-Styleguide#properties for determining when something should be a model property)

mvandenburgh (approved):

dandiapi/zarr/checksums.py
AlmightyYakob: Very good point, I agree. I'll merge this PR as is and open another one to address this.

Issue #1378, opened 2022-11-22
Zarr 2a613dd6-73c7-45f7-9b17-eae495c26277 manifest is out of sync with S3


jwodder:
Various entries were recently deleted from Zarr 2a613dd6-73c7-45f7-9b17-eae495c26277, yet the Archive's API still reports them as part of the Zarr.

Specifically, running the following script:

```python
from dandi.dandiapi import DandiAPIClient, RemoteZarrEntry
import boto3
from botocore import UNSIGNED
from botocore.client import Config
import sys

s3bucket = "dandiarchive"
zarr_id = sys.argv[1]
s3prefix = f"zarr/{zarr_id}/"

with DandiAPIClient.for_dandi_instance("dandi") as client:
    root = RemoteZarrEntry(client=client, zarr_id=zarr_id, parts=(), _known_dir=True)
    entries = {str(e): e.get_digest().value for e in root.iterfiles()}
print("Found", len(entries), "entries in Zarr on Archive")

s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
for page in s3.get_paginator("list_objects_v2").paginate(Bucket=s3bucket, Prefix=s3prefix):
    for obj in page["Contents"]:
        path = obj["Key"][len(s3prefix):]
        digest = obj["ETag"].strip('"')
        if path not in entries:
            print(f"Path {path} is only in S3")
        elif entries[path] != digest:
            print(f"Digest mismatch for {path}: Archive says {entries[path]}, S3 says {digest}")
        entries.pop(path, None)

for e in entries:
    print(f"Path {e} is only on Archive")
```

outputs the following:

```
Found 37230 entries in Zarr on Archive
Path 0/0/0/15/12/105 is only on Archive
Path 0/0/0/15/12/116 is only on Archive
Path 0/0/0/15/12/124 is only on Archive
Path 0/0/0/15/12/137 is only on Archive
Path 0/0/0/15/12/140 is only on Archive
Path 0/0/0/15/12/161 is only on Archive
Path 0/0/0/15/12/172 is only on Archive
Path 0/0/0/15/12/178 is only on Archive
Path 0/0/0/15/12/185 is only on Archive
Path 0/0/0/15/12/19 is only on Archive
Path 0/0/0/15/12/196 is only on Archive
Path 0/0/0/15/12/206 is only on Archive
Path 0/0/0/15/12/21 is only on Archive
Path 0/0/0/15/12/215 is only on Archive
[...]

PR #1394, opened 2022-12-07
Use flat file listing in zarr file browser

AlmightyYakob:
This is a simplification, and will as a result fix any zarrs with out of sync file listings (any other zarrs affected by the underlying issue to https://github.com/dandi/dandi-archive/issues/1378).

@jwodder This is a breaking change and will require a corresponding CLI update. If anything here seem infeasible / unreasonable, please let me know.

Notable breaking changes:

* The url is now `zarr/{zarr_id}/files`
* To filter by prefix, use the `prefix` query param
* To limit the number of files returned, use the `limit` query param (default 1000)
* To perform pagination, provide the `after` query param, setting it to the key of the last item in the previous listing
* To be redirected to file download (requires providing file name in `prefix`), set the `download` flag to `true`
* The asset download endpoint no longer redirects to the zarr file explorer view if the specified asset contains a zarr. Instead, it returns a 400, instructing the user to use the zarr file listing endpoint directly. I prefer this than to blindly redirecting to zarr file listing endpoint, as the "download" operation on an entire zarr isn't something we support.

Head requests are still supported, and will use the provided `prefix` query param to return a presigned head URL. If the prefix is not the entire path of a file, the HEAD request against the presigned url will return a 404.

### DIFF
--- a/dandiapi/zarr/views/__init__.py
+++ b/dandiapi/zarr/views/__init__.py
@@ -3,24 +3,21 @@ from __future__ import annotations
 import logging
 from pathlib import Path
 
-from django.conf import settings
 from django.db import IntegrityError, transaction
 from django.db.models import Exists, OuterRef
 from django.db.models.query import QuerySet
-from django.http.response import HttpResponseRedirect
-from django.urls import reverse
-from drf_yasg import openapi
+from django.http import HttpResponseRedirect
 from drf_yasg.utils import no_body, swagger_auto_schema
 from rest_framework import serializers, status
-from rest_framework.decorators import action, api_view
+from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied, ValidationError
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 from rest_framework.viewsets import ReadOnlyModelViewSet
 
 from dandiapi.api.models.dandiset import Dandiset
+from dandiapi.api.storage import get_boto_client
 from dandiapi.api.views.common import DandiPagination
-from dandiapi.zarr.checksums import ZarrChecksumFileUpdater
 from dandiapi.zarr.models import ZarrArchive, ZarrArchiveStatus, ZarrUploadFile
 from dandiapi.zarr.tasks import cancel_zarr_upload, ingest_zarr_archive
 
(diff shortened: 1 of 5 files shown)


jwodder:
@AlmightyYakob What is the format of the `zarr/{zarr_id}/files` response?

> To perform pagination, provide the `after` query param, setting it to the key of the last item in the previous listing

Is there a reason that pagination doesn't work like other endpoints, with `"results"` and `"next"` keys in the response?

AlmightyYakob:
> @AlmightyYakob What is the format of the `zarr/{zarr_id}/files` response?

My bad. The format is as such:
```json
[
  {
    "Key": "foo/bar",
    "LastModified": "2022-12-06T21:46:34.331000Z",
    "Etag": "f0c0b28da13bec676d3a348d6f11dabe",
    "Size": 100
  }
]
```

> Is there a reason that pagination doesn't work like other endpoints, with `"results"` and `"next"` keys in the response?

Yes, S3 doesn't provide any mechanism for limit/offset pagination when listing objects, only providing a `StartAfter` parameter, to indicate the last element of the previous "page". There's also no way to get the total count of objects in the bucket (or under a certain prefix, as we would need), so it would be impossible to easily provide the `count` field, which is the standard for our other pagination endpoints.

danlamanna:
I think we should still provide a `next` link where we can, this way it's easier for consumers and (hopefully) prevents them from hand-constructing URLs.

dandiapi/zarr/views/__init__.py
```diff
@@ -97,13 +94,11 @@ class Meta:
...
-    directories = serializers.ListField(child=serializers.URLField())
-    files = serializers.ListField(child=serializers.URLField())
-    checksums = serializers.DictField(
-        child=serializers.RegexField('^[0-9a-f]{32}(-[0-9]+--[0-9]+)?$')
-    )
-    checksum = serializers.RegexField('^[0-9a-f]{32}-[0-9]+--[0-9]+$')
+class ZarrExploreQuerySerializer(serializers.Serializer):
+    after = serializers.CharField(default='')
+    prefix = serializers.CharField(default='')
+    limit = serializers.IntegerField(default=1000)
```
mvandenburgh, line 100: What happens if the client provides an extremely large value or a negative value here? Do we need `max_value` and `min_value`?

AlmightyYakob:
@jwodder The endpoint response has been updated to the following

```json
{
  "next": "http://localhost:8000/api/zarr/357a0699-4f08-4a3b-83a7-923353379e28/files/?after=subdir%2Fa.txt&download=false&limit=1&prefix=subdir",
  "results": [
    {
      "Key": "subdir/a.txt",
      "LastModified": "2022-12-07 17:46:51.962000+00:00",
      "ETag": "c560516266eb28040362e72ec278b4a2",
      "Size": 8
    }
  ]
}
```

jwodder:
@AlmightyYakob Does the `DELETE /zarr/:id/files/` endpoint for deleting Zarr entries still work the same way?

mvandenburgh (approved):
LGTM from the perspective of the API.

jwodder:
@AlmightyYakob There was one failure in the test against https://github.com/dandi/dandi-cli/pull/1175, regarding the test of `get_content_url()`.  Has the contents of the `contentUrl` metadata field changed for Zarr assets?

jwodder:
@AlmightyYakob The integration tests with my PR are passing now.

AlmightyYakob:
> @AlmightyYakob The integration tests with my PR are passing now.

@jwodder So there's no changes needed on this PR, in reference to your previous comment?

jwodder:
@AlmightyYakob Right.

PR #1399, opened 2022-12-10
Move asset publishing to publish service

mvandenburgh:
Moves asset publishing from a model method into a dedicated service layer function within the publish service.


AlmightyYakob (approved):
LGTM. My only concern was changing the publish operation from non-saving to saving, but it seems that the non-saving behavior wasn't used anywhere in the code.

PR #1402, opened 2022-12-12
Manage CI dependencies with dependabot

mvandenburgh:
We keep having issues with CI/testing dependencies introducing breaking changes, causing CI to fail. 

#1359 
#1380 
https://github.com/dandi/dandi-archive/pull/1387#issuecomment-1343052208
etc...

I think we should just pin all testing dependencies and let dependabot manage updating them for us. Since we'd only be pinning testing dependencies, any dependabot PRs that pass CI can just be merged without a second thought.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,9 @@ updates:
       include: scope
     labels:
       - internal
+  - package-ecosystem: pip
+    directory: /
+    schedule:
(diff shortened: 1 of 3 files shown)


danlamanna (approved):

AlmightyYakob (approved):
Great idea!

waxlamp (approved):

PR #1415, opened 2022-12-21
Add design doc for Zarr upload process simplification

waxlamp:
@satra, @yarikoptic, and @jwodder, please take a look at this design doc ([rendered](https://github.com/dandi/dandi-archive/blob/zarr-overhaul-dd/doc/design/zarr-performance-redesign.md)).

This simpler design should solve the remaining performance problems with Zarr upload without compromising correctness, only requiring minor changes in how the archive and the CLI coordinate to accomplish it (see the doc itself for details).

### DIFF
--- a/doc/design/zarr-performance-redesign.md
+++ b/doc/design/zarr-performance-redesign.md
@@ -121,8 +121,8 @@ sequenceDiagram
 step can actually be taken any time before step 7; it is listed here
 arbitrarily.) `dandi-cli` asks the server to create a new Zarr archive, which is
 put into the `PENDING` state (steps 2 and 3). **`dandi-cli` will request a
-presigned upload URL from the server for each Zarr chunk file.** (steps 4 and
-5). `dandi-cli` uses these URLs to upload the files **using S3's `Content-MD5`
+presigned upload URL from the server for each Zarr chunk file** (steps 4 and
+5). Note: For an existing zarr archive, this is where the upload process begins, as requesting a signed url for upload will always place the zarr archive into a `PENDING` state. `dandi-cli` uses these URLs to upload the files **using S3's `Content-MD5`
 header to verify the uploaded file's integrity** (step 6). **Instead of
 finalizing a batch (since there is no longer a batch concept), `dandi-cli`
 repeats these steps until all files are uploaded (repeating steps 4, 5, and
--- a/doc/design/zarr-performance-redesign.md
+++ b/doc/design/zarr-performance-redesign.md
@@ -100,8 +100,8 @@ sequenceDiagram
     Server-->>-Client: PENDING Zarr Archive
 
     loop for each file
-        Client->>+Server: Request signed URL
-        Server-->>-Client: A signed URL
+        Client->>+Server: Request signed URLs
+        Server-->>-Client: A list of signed URLs
         Client->>+S3: Upload individual file using signed URL
     end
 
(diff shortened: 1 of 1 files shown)


jwodder:
Questions so far:

* How is the wait for the Zarr to reach "COMPLETE" handled when uploading files to a pre-existing Zarr?  Does requesting presigned upload URLs on a pre-existing Zarr cause it to enter a non-COMPLETE state?
* The new procedure states "dandi-cli uses these URLs to upload the files using S3's Content-MD5 header to verify the uploaded file's integrity".  Is this header provided by the Archive server when requesting presigned URLs or by the client when uploading to a presigned URL?  Either way, if a malicious or buggy client uploads data that does not match the expected digest, what effect will this have on the upload?

waxlamp:
> Questions so far:
> 
> * How is the wait for the Zarr to reach "COMPLETE" handled when uploading files to a pre-existing Zarr?  Does requesting presigned upload URLs on a pre-existing Zarr cause it to enter a non-COMPLETE state?

Good question. We would want the Zarr in a non-COMPLETE state before new files can be uploaded to it (or existing files removed). We could either make this explicit by requiring the client to send a `POST` request to "open" the Zarr again, or (as you suggested) having it done upon a request for a new presigned URL. I think I like the former option better because deleting Zarr chunks would require a step like that anyway.

> * The new procedure states "dandi-cli uses these URLs to upload the files using S3's Content-MD5 header to verify the uploaded file's integrity".  Is this header provided by the Archive server when requesting presigned URLs or by the client when uploading to a presigned URL?  Either way, if a malicious or buggy client uploads data that does not match the expected digest, what effect will this have on the upload?

The client would supply the header from its own computation of the MD5 checksum.

Using an incorrect digest results in a failed S3 upload. A buggy client that ignores this failure would omit that file from the uploaded Zarr, leading to non-matching Zarr checksums in the last steps of the upload. (It is relatively straightforward to recover from this situation, as sketched out in the `Error Handling` section.)

jwodder:
To be clear, after the Zarr is finalized, the client is supposed to compare the server's Zarr checksum with the local Zarr checksum, and if they don't match, then what?  Is the client supposed to notify the server, and is this what leads to the "asking S3 for a list of all objects stored at the Zarr prefix in question" error handling step?  Wouldn't it be better for the client to send the locally-computed Zarr checksum to the server in the "finalize" request in case, say, the client gets killed while waiting for finalization?  Either way, it seems that a buggy or malicious client could easily report a checksum mismatch where there is none, leading to wasted cycles in communicating with S3.

jwodder:
Another thing: How exactly does this handle upload cancellation?  Say the client requests a presigned URL for a Zarr entry, but before the actual upload takes place, the user hits Ctrl-C.  Then, the user locally deletes the Zarr entry and starts another upload.  When the client fetches the list of entries already in the Zarr from the server, I assume that the cancelled entry will not be in the list, so the client won't do anything about it, but will the server's checksumming procedure assume that the entry was uploaded, leading to a checksum mismatch and fallback to the "Error handling" procedure?

AlmightyYakob:
> Another thing: How exactly does this handle upload cancellation? Say the client requests a presigned URL for a Zarr entry, but before the actual upload takes place, the user hits Ctrl-C. Then, the user locally deletes the Zarr entry and starts another upload. When the client fetches the list of entries already in the Zarr from the server, I assume that the cancelled entry will not be in the list, so the client won't do anything about it, but will the server's checksumming procedure assume that the entry was uploaded, leading to a checksum mismatch and fallback to the "Error handling" procedure?

The server doesn't store any information about zarr entries, it simply reads all of the entries from the storage medium (S3) during the checksumming processing and produces an output. So in your scenario, the original entry for which the presigned URL was requested would **not** be counted in the final checksum.

Generally speaking, in this design, there is no concept of "upload cancelling". So if you uploaded a bunch of files that you actually _didn't_ want included in the zarr, you'd need to use the `DELETE` endpoint to remove them. However if you simply requested presigned URLs for entries, but didn't upload to them, no explicit action is needed to "cancel" those uploads.

jwodder:
@AlmightyYakob 

> The server doesn't store any information about zarr entries, it simply reads all of the entries from the storage medium (S3) during the checksumming processing and produces an output.

The doc implies that reading from S3 would only be done as part of error handling when there's a checksum mismatch.  If the server always fetches all checksums from S3, what is done differently for error handling?

AlmightyYakob:
> The doc implies that reading from S3 would only be done as part of error handling when there's a checksum mismatch. If the server always fetches all checksums from S3, what is done differently for error handling?

I think for brevity, the diagram isn't showing what the worker itself is doing to compute the zarr checksum, since it's supposed to be a view of the process from the client's perspective. But yes, the zarr checksum task and file listing endpoint both start by listing objects from s3 (prefixed to the zarr location in the bucket).

The zarr file listing endpoint (used for error handling) does some processing to the returned objects list so that it appears relative to the zarr root, and returns that list of objects.

The zarr checksumming task lists all the objects within the prefix from S3, and adds them to a "tree", which is then processed bottom-up.

jwodder:
I think the document should somewhere address uploading to a pre-existing Zarr as described in [this comment](https://github.com/dandi/dandi-archive/pull/1415#issuecomment-1362965574).

AlmightyYakob:
> I think the document should somewhere address uploading to a pre-existing Zarr as described in [this comment](https://github.com/dandi/dandi-archive/pull/1415#issuecomment-1362965574).

@jwodder This has been done in [078b544](https://github.com/dandi/dandi-archive/pull/1415/commits/078b544875c0cf03860f02f455f36fd04e8af4d1). The upload process for an existing zarr is merely a subset of the upload process for a new zarr, so I've just added a note that describes this.

jwodder (approved):

satra:
@AlmightyYakob  and @jwodder - in the new schema, steps 4 and 5 does not seem to be based on batches. do we know what the performance penalty is for moving to a single file request-upload option?

AlmightyYakob:
> @AlmightyYakob and @jwodder - in the new schema, steps 4 and 5 does not seem to be based on batches. do we know what the performance penalty is for moving to a single file request-upload option?

This step is still done in bulk, it's just no longer limited to the concept of an "upload batch" (where an "upload batch" requires not being able to request any more presigned urls until the current "batch" is fully uploaded) . So you could request as few, or as many (up to the server limit, currently 255 _per request_) presigned upload urls as you'd like from the API.

satra:
thanks @AlmightyYakob - i didn't see that mentioned in the doc. may be useful to add.

PR #1416, opened 2022-12-22
Add 12 hour cache to stats endpoint

### LINKED ISSUE #1412: Optimize/cache stats endpoint
Currently there's a few things that make the stats endpoint inefficient:

1. When called, all asset blobs, embargoed asset blobs, and zarr archives are aggregated and summed up to produce a "total size" of the archive. This is inefficient, as the aggregation is already taken care of by the asset paths code, and so can just be summed over the root nodes of all versions.
2. This endpoint is not cached at all, so every call performs this aggregation, even though these metrics don't need to be known in real time, and could realistically be cached for the next few hours or so
3. This endpoint is called every time the home page for https://dandiarchive.org is loaded, further compounding the previous two issues.

I think the best way to address this would be to:
1. Update the aggregation code to make use of asset paths
2. Cache the endpoint response for some number of hours

AlmightyYakob:
Closes #1412 

Notably, this doesn't actually change how the stats are computed, it just adds a cache. Using the asset paths models to compute this results in a slightly different value than we have now. Rather than try to determine what course of action we should take (a behavior change), I just wanted to get the difference making optimization in.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -107,6 +107,9 @@ jobs:
       - name: Apply migrations to API server
         run: python manage.py migrate
 
+      - name: Create any cache tables
+        run: python manage.py createcachetable
+
       - name: Install E2E tests
         run: yarn install --frozen-lockfile
         working-directory: web/test
--- a/dandiapi/api/views/stats.py
+++ b/dandiapi/api/views/stats.py
@@ -1,10 +1,13 @@
 from django.contrib.auth.models import User
+from django.views.decorators.cache import cache_page
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
 
 from dandiapi.api.models import Asset, Dandiset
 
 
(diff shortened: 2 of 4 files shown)


AlmightyYakob:
@waxlamp It's now good to go

danlamanna (commented):
Is there already a configured cache backend?

AlmightyYakob:
> Is there already a configured cache backend?

The default is local memory caching. Do you think we should use a DB cache instead?

AlmightyYakob:
Note: Since we're using a db cache, we'll need to run `python manage.py createcachetable` manually once, after this is deployed.

danlamanna (approved):

danlamanna (approved):

AlmightyYakob:
> Note: Since we're using a db cache, we'll need to run `python manage.py createcachetable` manually once, after this is deployed.

This has been successfully performed.

PR #1431, opened 2023-01-13
Compute asset summary asynchronously for draft versions

danlamanna:
For some background: modifying assets requires modifying the draft version to update the asset summary. This creates write contention on the version when assets that belong to the same version are being modified concurrently. The problem is complicated because the asset summary can take several seconds on larger dandisets and the time grows linearly with the asset count. To date this has only been a problem with our largest dandiset. 

This PR moves the asset summary computation to an asynchronous task, similar to version metadata validation. 

@mvandenburgh Can you take a look at this PR closely since it touches some of your recently modified publishing code? 

This is a draft for right now, I want to look closer into the version lifecycle before I merge.

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -2,9 +2,7 @@ from __future__ import annotations
 
 import datetime
 import logging
-from typing import TYPE_CHECKING
 
-from dandischema.metadata import aggregate_assets_summary
 from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.validators import RegexValidator
@@ -15,9 +13,6 @@ from django_extensions.db.models import TimeStampedModel
 from dandiapi.api.models.metadata import PublishableMetadataMixin
 
 from .dandiset import Dandiset
-
-if TYPE_CHECKING:
-    from .asset import Asset
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


mvandenburgh (commented):
I don't see any issues here, looks good. At some point we'll also need to update the frontend to indicate when assets summary is being calculated.

waxlamp:
> I don't see any issues here, looks good. At some point we'll also need to update the frontend to indicate when assets summary is being calculated.

Ideally we'd add that change to this PR, but (1) it's a separate (but intimately related) change and (2) I wouldn't want to hold up merging this PR, but it's important that the frontend changes are done soon after if not alongside this PR.

@mvandenburgh, I'd like to be involved with designing the frontend changes. Please schedule a meeting with me and/or Dan to discuss.

waxlamp:
@danlamanna, what's the current status of this PR? You had wanted to hold off before merging, but I see the PR is out of draft mode now.

@mvandenburgh, did you have a chance to do a full review? (You left a comment-review rather than an approval, so just checking in with your status on this PR as well.)

mvandenburgh:
@waxlamp we discussed this PR earlier this week, and Dan made some changes based on that, he's just waiting on my review now (which I will do today).

mvandenburgh (approved):

PR #1433, opened 2023-01-16
ENH: explicitly describe both cases of not allowed dandiset deletion

yarikoptic:
User reported being unable to delete a dandiset and it was not clear why it was denied.  If we do channel those messages in responses - this should make it clear why request was denied.

### DIFF
--- a/dandiapi/api/services/dandiset/__init__.py
+++ b/dandiapi/api/services/dandiset/__init__.py
@@ -48,7 +48,7 @@ def create_dandiset(
 
 def delete_dandiset(*, user, dandiset: Dandiset) -> None:
     if not user.has_perm('owner', dandiset):
-        raise NotAllowed('Cannot delete dandsets which you do not own.')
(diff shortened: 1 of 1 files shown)


dandiapi/api/services/dandiset/__init__.py
```diff
@@ -47,12 +47,10 @@ def create_dandiset(
...
 
 def delete_dandiset(*, user, dandiset: Dandiset) -> None:
-    if (
-        not user.has_perm('owner', dandiset)
-        or dandiset.embargo_status != Dandiset.EmbargoStatus.OPEN
-    ):
-        raise NotAllowed()
-
+    if not user.has_perm('owner', dandiset):
+        raise NotAllowed('Cannot delete dandsets which you do not own.')
```
satra, line 51:
```suggestion
        raise NotAllowed('Cannot delete dandisets which you do not own.')
```

danlamanna (approved):

PR #1442, opened 2023-01-23
Fix publishing UI bug

mvandenburgh:
To replicate this bug, go to a publishable dandiset and click publish:

https://gui-staging.dandiarchive.org/dandiset/204686?page=6&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=41

https://gui-staging.dandiarchive.org/dandiset/204685?page=6&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=42

https://gui-staging.dandiarchive.org/dandiset/204677?page=6&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=45

https://gui-staging.dandiarchive.org/dandiset/204681?page=6&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=44

After clicking publish, the UI immediately starts bugging out and rerendering the DLP every few seconds. This is caused by some reactivity bug with the `loading` variable in the store. This particular variable is only used in a single component, so I refactored it to a local variable within that component, and now the publish UI behaves as expected.

### DIFF
--- a/web/src/stores/dandiset.ts
+++ b/web/src/stores/dandiset.ts
@@ -11,7 +11,6 @@ import { draftVersion } from '@/utils/constants';
 interface State {
   dandiset: Version | null;
   versions: Version[] | null,
-  loading: boolean,
   owners: User[] | null,
   schema: any,
 }
@@ -20,7 +19,6 @@ export const useDandisetStore = defineStore('dandiset', {
   state: (): State => ({
     dandiset: null,
     versions: null,
-    loading: false,
     owners: null,
     schema: null,
   }),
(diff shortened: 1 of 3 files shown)


AlmightyYakob (commented):
I can confirm the bug is fixed, although when you finally confirm the publish, the dialog box stays open, with a snackbar at the bottom that says "Dandiset published, go to published dandiset", taking you to the published dandiset. However, the old dialog still says up, allowing for someone to immediately hit publish again, resulting in error. This could be easily done if they simply don't see the bottom snackbar, as it's not extremely visible.

Is this intended behavior? If so, I'm happy to approve this as fixed and file a new issue to address that design, just wanted to make sure there's been no unintended behavior change.

mvandenburgh:
Good catch, no that wasn't intended - I only tested this change on dandisets I owned, so I never saw that admin dialog. I'll fix that here.

PR #1443, opened 2023-01-23
Require rejection reason when rejecting a new user

### LINKED ISSUE #1441: Make reason mandatory to reject a new user?
#1301 added rejection reason.  I often click on Reject for some reason assuming it would somehow give me a dialog to provide a reason... sure thing it doesn't and thus I reject without stating a reason.  I think it is reasonable to expect to provide reason for every rejection, thus should we make Reject disabled unless reason is specified or add some check to not actually do the action and give error message if no reason is given?

mvandenburgh:
Closes #1441

### DIFF
--- a/dandiapi/api/templates/dashboard/user_approval.html
+++ b/dandiapi/api/templates/dashboard/user_approval.html
@@ -81,15 +81,15 @@
           {% csrf_token %}
           <input type="hidden" name="status" value="REJECTED">
           <button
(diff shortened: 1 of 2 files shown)


mvandenburgh:
I noticed we're pulling out query params/POST body parameters manually in these endpoints, and it gets kind of messy sometimes. I opened an issue #1444 to address this.

danlamanna (approved):

PR #1452, opened 2023-01-30
Update DLP UI for asynchronous assets summary

mvandenburgh:
async_asset_summary_ui

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -109,11 +109,20 @@
     </MetadataCard>
 
     <v-card
-      v-if="assetSummary && Object.keys(assetSummary) && Object.keys(assetSummary).length"
+      v-if="assetSummary"
       outlined
     >
       <v-card-title class="font-weight-regular">
-        <v-icon class="mr-3 grey--text text--lighten-1">
+        <v-progress-circular
+          v-if="assetSummaryBeingComputed"
+          class="mr-3"
+          size="20"
+          indeterminate
+        />
+        <v-icon
+          v-else
+          class="mr-3 grey--text text--lighten-1"
+        >
           mdi-clipboard-list
         </v-icon>
         Assets Summary
@@ -123,7 +132,20 @@
         class="px-3 ml-2"
       >
         <div
+          v-if="assetSummaryBeingComputed"
+          class="text-subtitle-2"
+        >
+          The assets summary is being computed, please wait.
+        </div>
+        <div
+          v-else-if="!assetSummary || !Object.keys(assetSummary).length"
+          class="font-italic font-weight-bold"
+        >
+          This Dandiset does not contain any assets.
+        </div>
+        <div
           v-for="([type, items], i) in Object.entries(assetSummary)"
+          v-else
           :key="i"
         >
           <div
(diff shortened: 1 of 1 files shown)


web/src/components/DLP/OverviewTab.vue
```diff
@@ -290,19 +316,33 @@ export default defineComponent({
...
+    onMounted(() => {
+      setInterval(async () => {
+        if (!currentDandiset.value || !assetSummaryBeingComputed.value) {
+          return;
+        }
+        const { identifier } = currentDandiset.value.dandiset;
+        const { version } = currentDandiset.value;
+        await store.fetchDandiset({ identifier, version });
+      }, 5000);
+    });
```
AlmightyYakob, line 336: We should probably add a stop condition where we call `clearInterval`, so we're not needlessly making the same request every 5 seconds after the fact.

AlmightyYakob (changes requested):

web/src/components/DLP/OverviewTab.vue
mvandenburgh: Once the assets summary is computed, `assetSummaryBeingComputed` will be `true`, so the request won't happen due to the if guard.

It's also possible that `assetSummaryBeingComputed` is initially `true` on page load, but then gets set to `false` due to something the user does on the DLP (for example, editing metadata in the meditor), so I don't think we'd want to do a `clearInterval` at any point here.

AlmightyYakob (approved):

web/src/components/DLP/OverviewTab.vue
```diff
@@ -122,8 +131,21 @@
...
+          v-if="assetSummaryBeingComputed"
+          class="text-subtitle-2"
+        >
+          The assets summary is being computed, please wait.
+        </div>
+        <div
+          v-else-if="!assetSummary || !Object.keys(assetSummary).length"
+          class="text-subtitle-2"
+        >
+          Assets summary computation failed. This dandiset does not contain any assets.
```
waxlamp, line 144:
```suggestion
      <span class="font-italic font-weight-bold">
          This Dandiset does not contain any assets.
      </span>
```

I don't think this condition means the asset summary *failed*: if there are no assets, then there is no summary.

OTOH, if the asset summary actually fails for any reason, we ought to capture that as a specific status to report (e.g., so the user can click something to report the problem to us).

Finally, I added the bold and italic to match with "special" messages such as the "no funding information" message visible in your example.

web/src/components/DLP/OverviewTab.vue
waxlamp: Just so I understand: the request wouldn't fire, but the interval callback would continue to run every 5 seconds, right? Would we still not want to call `clearInterval()` just before the `return` statement on line 330?

waxlamp (changes requested):
One wording change to discuss, plus a followup question about `setInterval()`. Looks great though!

web/src/components/DLP/OverviewTab.vue
mvandenburgh: > Just so I understand: the request wouldn't fire, but the interval callback would continue to run every 5 seconds, right? 

Correct.

> Would we still not want to call clearInterval() just before the return statement on line 330?

No, because it's possible that the user will make some change on the DLP that will cause the assets summary to become outdated. In that case, if we had called clearInterval prior, we would need to call `setInterval` again. We _could_ do that, but that would require figuring out all the ways the user can retrigger assets summary calculation from the DLP, and calling setInterval from each of those places. Then that introduces more issues, like we'd need some way to tell if the setInterval call is already running to avoid duplicating it, etc. So I think just leaving it as-is is the cleanest way to do this.

danlamanna (approved):

waxlamp (approved):

PR #1454, opened 2023-01-30
Display error message for invalid dandiset URL

### LINKED ISSUE #1020: Must not silently ignore incorrect version of a dandiset in URL
Was not sure if I should add to the WiP design doc PR https://github.com/dandi/dandi-archive/pull/1017 (thus attn @AlmightyYakob)  but just noted that behavior, which is also present with current main instance still using `#` in URLs.

If version in the URL is not correct, app just ignores it so that https://gui.dandiarchive.org/#/dandiset/000167/0.220329.0721 silently leads to https://gui.dandiarchive.org/#/dandiset/000167/0.220329.0720 version of the dataset.

In comparison, the redirector is *rightfully* not that forgiving: going to https://dandiarchive.org/dandiset/000167/0.220329.0721 leads to `dandi:000167/0.220329.0721 not found.`

```shell
$> wget https://dandiarchive.org/dandiset/000167/0.220329.0721
--2022-04-08 18:13:45--  https://dandiarchive.org/dandiset/000167/0.220329.0721
Resolving dandiarchive.org (dandiarchive.org)... 3.12.21.28
Connecting to dandiarchive.org (dandiarchive.org)|3.12.21.28|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2022-04-08 18:13:45 ERROR 404: Not Found.
```

and web app should behave similarly.

https://dandiarchive.org/dandiset/000167/0.220329.0720

mvandenburgh:
Closes #1020 #1470

dlp_error

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -10,6 +10,24 @@
       />
     </v-toolbar>
     <v-container
+      v-if="dandisetDoesNotExist"
+      class="d-flex justify-center align-center"
+      style="height: 50vh;"
+    >
+      <div class="d-block">
+        <span class="text-h5">
+          <v-icon>mdi-alert</v-icon>
+          Error: Dandiset does not exist.
+        </span>
+        <br><br>
+        <span class="text-body-2">
+          Proceed back to the <a href="/dandiset">Public Dandisets page</a>
+          or use the search bar to find a valid Dandiset.
+        </span>
+      </div>
+    </v-container>
+    <v-container
+      v-else
       fluid
       class="grey lighten-4 pa-0"
     >
@@ -117,6 +135,7 @@ const store = useDandisetStore();
 
 const currentDandiset = computed(() => store.dandiset);
 const loading = ref(false);
+const dandisetDoesNotExist = ref(false);
 const schema = computed(() => store.schema);
 const userCanModifyDandiset = computed(() => store.userCanModifyDandiset);
 
@@ -143,6 +162,11 @@ watch(() => props.identifier, async () => {
     loading.value = true;
     await store.initializeDandisets({ identifier, version });
     loading.value = false;
+
+    // Render an error message if currentDandiset is still null after fetching
+    if (!currentDandiset.value) {
+      dandisetDoesNotExist.value = true;
+    }
   }
 }, { immediate: true });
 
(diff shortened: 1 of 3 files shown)


web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -10,6 +10,24 @@
       />
     </v-toolbar>
     <v-container
+      v-if="dandisetDoesNotExist"
```
AlmightyYakob, line 13: I don't think there's actually a need to use a new ref. If you're going for clarity, I'd think a comment would suffice, but I won't block on it.

```suggestion
      v-if="!(loading || currentDandiset)"
```

AlmightyYakob (changes requested):

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -10,6 +10,24 @@
...
+      class="d-flex justify-center align-center"
+      style="height: 50vh;"
+    >
+      <div class="d-block">
+        <span class="text-h5">
+          <v-icon>mdi-alert</v-icon>
+          Error: Dandiset does not exist.
```
waxlamp, line 20:
```suggestion
          Error: Dandiset does not exist
```

web/src/views/DandisetLandingView/DandisetLandingView.vue
waxlamp: I think I agree with @AlmightyYakob, fwiw. Though (and now we're at real risk of bikeshedding) I would probably express it as `!loading && !currentDandiset` only because that reads a little bit clearer to me ("ok so I'm done loading, but there's still no dandiset...").

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -10,6 +10,24 @@
...
+        </span>
+        <br><br>
+        <span class="text-body-2">
+          Proceed back to the <a href="/dandiset">Public Dandisets page</a>
```
waxlamp, line 24:
```suggestion
          Proceed to the <a href="/dandiset">Public Dandisets page</a>
```

since they may not have come from the public dandisets page.

waxlamp (approved):
Looks great overall!

I had a couple of wording updates (very minor), and I second Jake's suggestion. If you address those things, this is good to go from me.

web/src/views/DandisetLandingView/DandisetLandingView.vue
mvandenburgh: Agreed. For clarity, I went with a `computed` property instead of inlining `!loading && !currentDandiset` into the `v-if`, since you can't leave comments inside HTML tags.

AlmightyYakob (approved):

PR #1485, opened 2023-02-08
Add locking to asset update/delete at API level

### LINKED ISSUE #1469: Lock assets at API level for update/delete
We should add a lock on the asset specific update/delete endpoints, to ensure that error encountered in #1463 doesn't occur. This should be done at the API level, as this allows for the earliest holding of a lock by one of the two assets, preventing both from getting far into the operation. This will handle race conditions externally, however if race conditions become an issue internally as well, we can  move this locking into the service layer.

AlmightyYakob:
Closes #1469

~~Opening this as a draft, as the locking has been added, but I haven't figured out a good way to test this, since we'd need to test against a race condition.~~

This adds a lock at the API level for asset update and delete, as well as updates the asset paths implementation to raise an `AssetAlreadyExists` error, if the `unique-version-path` unique constraint fails on asset path creation (i.e., if an asset with a conflicting path is attempted to be added to a version).

Since we can't effectively test the API level race condition, I've done testing locally _without_ the included changes, confirmed I was able to reproduce the error, and then tested locally _with_ the included changes, and confirmed that the API behaves as expected.

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 
-from django.db import transaction
+from django.db import IntegrityError, transaction
 from django.db.models import Count, F, QuerySet, Sum
 from django.db.models.functions import Coalesce
 from tqdm import tqdm
@@ -101,8 +101,23 @@ def search_asset_paths(query: str, version: Version) -> QuerySet[AssetPath] | No
 
 def insert_asset_paths(asset: Asset, version: Version):
     """Add all intermediate paths from an asset and link them together."""
-    # Get or create leaf path
-    leaf, created = AssetPath.objects.get_or_create(path=asset.path, asset=asset, version=version)
+    try:
+        # Get or create leaf path
+        leaf, created = AssetPath.objects.get_or_create(
+            path=asset.path, asset=asset, version=version
+        )
+    except IntegrityError as e:
+        from dandiapi.api.services.asset.exceptions import AssetAlreadyExists
+
+        # If there are simultaneous requests to create the same asset, this check constraint can
+        # fail, and should be handled directly, rather than be allowed to bubble up
+        if 'unique-version-path' in str(e):
+            raise AssetAlreadyExists()
+
+        # Re-raise original exception otherwise
+        raise
+
+    # If the asset was not created, return early, as the work is already done
     if not created:
         return leaf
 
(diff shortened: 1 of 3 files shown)


waxlamp:
Can you spam your local instance with O(dozens) of requests to create an asset?

AlmightyYakob:
> Can you spam your local instance with O(dozens) of requests to create an asset?

We could do that, but that would:

1. Involve multithreading in a test
2. Make the test stochastic, which is very undesirable.

If you just meant doing locally outside of our testing suite as a sanity check, I'll probably do that yes. I'm not sure if the dev server is multi-threaded or not, but if not I could just run it with gunicorn.

waxlamp:
> If you just meant doing locally outside of our testing suite as a sanity check, I'll probably do that yes. I'm not sure if the dev server is multi-threaded or not, but if not I could just run it with gunicorn.

This is what I meant, yes, but I see you're (rightly) more concerned with automated tests, not just "seeing if it works in dev".

> > Can you spam your local instance with O(dozens) of requests to create an asset?
> 
> We could do that, but that would:
> 
> 1. Involve multithreading in a test

To make the test deterministic, we would want some way to execute the first request and then suspend it during that critical section where the race condition becomes possible, then execute a second request to completion, and then finally resume the first request. This sounds like an interesting systems programming type challenge but perhaps not easy or natural to accomplish in Python, and probably not worth the effort.

> 2. Make the test stochastic, which is very undesirable.

Unpopular opinion incoming: if you can *virtually* guarantee that the race condition can be hit (by spamming an unreasonably large number of requests), and the test passes when the proper 400 error is raised, then I don't think the stochasticity is too bad. It's sort of like how we don't observe quantum effects at macroscopic scales--the effects exist but are virtually undetectable.

But, I agree that going to the trouble of doing this may not be worth it.

Maybe this is a case where we test in dev to make sure this largely works as intended, and then deploy to production to see if the recurring Sentry errors stop. If crafting a test is prohibitively difficult, perhaps that would be good enough at least for now.

dandiapi/api/asset_paths.py
```diff
@@ -101,8 +101,22 @@ def search_asset_paths(query: str, version: Version) -> QuerySet[AssetPath] | No
...
-    leaf, created = AssetPath.objects.get_or_create(path=asset.path, asset=asset, version=version)
+    try:
+        # Get or create leaf path
+        leaf, created = AssetPath.objects.get_or_create(
+            path=asset.path, asset=asset, version=version
+        )
+    except IntegrityError as e:
+        from dandiapi.api.services.asset.exceptions import AssetAlreadyExists
+
+        # Check for version/path constraint violation
```
danlamanna, line 112: I think it would be better if these comments were more about why than what, I think the what is clear. Something about parallel asset creation resulting in already reserved asset paths would help me a few months from now when I forget why this is here :sweat_smile:.

danlamanna (approved):

Issue #1197, opened 2022-07-15
dashboard: Separate out "Non-valid" per each "status", improve reporting of errors with summaries


yarikoptic:
ATM dashboard shows only 10 first non-valid assets, selecting them here;
https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/api/dashboard.py#L29

but those which are "Pending" might be not that "interesting" as the ones which are actually "Invalid".  I would say it would be useful to report per each of the "status"es including the counts, and then "up to 10" within each one of those types.

Also should be useful to provide listing of uniq 10 "Validation errors" so admins could immediately assess what is 
the most common error, e.g. now it could be (numbers are "fake"):

<h2>Non-valid Assets</h2>
There are currently 95 Assets that are not valid.

<h3>Pending: 20</h3>

.... here table of first (sorted by modified time from oldest) 10 of those ...

<h3>Invalid: 32</h3>

<h4>Unique Validation Errors: 11</h4>

Showing only first 5:
<ul>
<li> 10: <pre>[{'field': '', 'message': 'Metadata version None is not allowed. Allowed are: 0.6.3, 0.4.4, 0.5.1, 0.5.2, 0.6.0, 0.6.1, 0.6.2.'}]</pre></li>
<li> 2: <pre>...smth else </pre></li>
<li>...</li>
</ul>

.... here table of first (sorted by modified time from oldest) 10 of those ...

I can implement this but I would need guidance how to test it against our main instance so I could observe it in action. @dandi/archive-maintainers ?

PR #1489, opened 2023-02-13
Skip computing assets summary for invalid assets

danlamanna:
This part of the code should have been ported over during #1431.

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -73,9 +73,14 @@ def version_aggregate_assets_summary(version: Version):
     if version.version != 'draft':
         raise VersionHasBeenPublished()
 
-    version.metadata['assetsSummary'] = aggregate_assets_summary(
-        version.assets.values_list('metadata', flat=True).iterator()
-    )
+    try:
+        version.metadata['assetsSummary'] = aggregate_assets_summary(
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):

satra:
i think this needs to be a more critical response than an avoidance. i.e. we need to fix when this happens. so i would suggest not merging this, but instead using something like this to add to the admin dashboard, so that we know what to fix. the silent suppression that this does is not going to be helpful.

danlamanna:
I'm merging this for now because all it does is restore the status quo that existed up until a few days ago. 

> so i would suggest not merging this, but instead using something like this to add to the admin dashboard, so that we know what to fix.

It seems like we already track invalid assets, we just need to improve the dashboard for it (#1197). Are you asking for something different? In general, I would like it if dandi-schema provided a better way of distinguishing between genuinely exceptional events and problems with client provided metadata (https://github.com/dandi/dandi-schema/issues/127#issuecomment-1183501908).

satra:
i think this would result, for any landing page to not show anything in the asset summary. this would only be detected when someone goes on the landing page, rather than fixing the core issue of assets. hence i wanted a more severe restriction that when this happens, we should, as dandi, fix the issue as fast as possible. in most cases this is not a user issue. more often a dandi issue.

Issue #1450, opened 2023-01-25
updating metadata on all draft assets


satra:
i'm starting this issue so that we can do this within the next two weeks.

requires re-extracting metadata from all draft assets and uploading it to the draft version of the dataset. we won't touch published versions here. this is to ensure that all draft datasets are compliant with current schema.

i think this is best done at an admin process level. let's start with all nwb datasets. it would involve running a script on assets from all draft versions to update it. 

@jwodder - could you please provide a dandi api-based function to re-extract metadata from a remote nwb file (i.e. a dandi asset url)? this is not getting the asset metadata but to reextract it from the nwb file itself.

cc/ @yarikoptic

satra:
> Are you expecting some new metadata being extracted from nwb

yes. our metadata extraction routines have changed several times, so the older assets are definitely out of sync with our current metadata extraction routines. to start with we can limit updates to assets with versions lower than current.

yarikoptic:
support for that was added already in dandi-cli .  I do not think it is an issue to keep in dandi-archive since I don't think code should be developed here now - just a matter of running that command/script.  but I can be wrong ;)

waxlamp:
I believe this issue was filed as a task reminder, not as something to change in the archive codebase. If this can be carried out with dandi-cli, that's fantastic.

I think we still need some clarity on exactly what to do, how to do it, and who will do it. Can we get those things decided so we can keep this moving forward?

yarikoptic:
or just finally expose dandiset locking as admin-only operation in API, and adjust already coded in implementation since it seems that we would need similar mechanism nearly for any archive wide operations which might interfere with original owner's actions.

AlmightyYakob:
Running the numbers against the current draft assets in prod (ignoring zarrs), I believe this will involve the processing of ~700GB in order to accomplish. We might be able to bring this down if we can define which assets are considered "old" and which are okay.

satra:
how did you get that number? it shouldn't actually download all the data per asset to do this. the updated code works on cached/remote streaming to do this.

AlmightyYakob:
> how did you get that number?

Summing the size of all assets in draft versions, excluding zarr assets. 

> it shouldn't actually download all the data per asset to do this. the updated code works on cached/remote streaming to do this.

[...]

PR #1490, opened 2023-02-13
Remove overly aggressive exception handling

danlamanna:
Related #1489 

@satra I think I understand what you're getting at and this PR is what I'm hoping satisfies requirements. I'll recap my understanding of these changes and rationale for confirmation/discussion.

1. Stop catching ValueErrors when validating assets
	We shouldn't be marking assets invalid for what are likely bugs in our dandi-archive/dandi-schema software. Instead, an exception will be unhandled and thus tracked in Sentry. The asset will be unable to be validated until we intervene to fix the bug. Since we've been masking this error and storing the results, we can say that all ValueErrors that have been caught to date are of the form `Metadata version x.y.z is not allowed. Allowed are: x.y.z, ...`. (Maybe this is related to https://github.com/dandi/dandi-cli/issues/935).
2. Stop catching blanket exceptions when aggregating assets summary
	This could also hide genuine bugs. These are currently logged but not alerted, which is problematic. Historically this has resulted in masking this error:
```
Traceback (most recent call last):                          
    File "/app/dandiapi/api/models/version.py", line 237, in _populate_metadata                             
    summary = aggregate_assets_summary(                     
    File "/app/.heroku/python/lib/python3.10/site-packages/dandischema/metadata.py", line 334, in aggregate_assets_summary                                    
    _add_asset_to_stats(meta, stats)                        
    File "/app/.heroku/python/lib/python3.10/site-packages/dandischema/metadata.py", line 319, in _add_asset_to_stats                                         
    if "nwb" in assetmeta["encodingFormat"]:                
KeyError: 'encodingFormat'                                  
```

Does this sound right? And if so, is the idea to merge a PR like this and then address #1450?

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -73,14 +73,9 @@ def version_aggregate_assets_summary(version: Version):
     if version.version != 'draft':
         raise VersionHasBeenPublished()
 
-    try:
-        version.metadata['assetsSummary'] = aggregate_assets_summary(
-            version.assets.values_list('metadata', flat=True).iterator()
-        )
-    except Exception:
-        # The assets summary aggregation may fail if any asset metadata is invalid, skip
-        # updating it if it fails.
-        logger.info('Error calculating assetsSummary', exc_info=True)
+    version.metadata['assetsSummary'] = aggregate_assets_summary(
+        version.assets.values_list('metadata', flat=True).iterator()
+    )
 
     Version.objects.filter(id=version.id, version='draft').update(
         modified=timezone.now(), metadata=version.metadata


satra:
@danlamanna - thank you. i feel #2 is appropriate, but #1 i think should perhaps be caught and reported/flagged to admin's at least. 

as far as i can tell those value errors won't happen on published dandisets, if they were run through the same checker. could we do that as an admin side script, just for sanity?

mvandenburgh (approved):

danlamanna:
> as far as i can tell those value errors won't happen on published dandisets, if they were run through the same checker. could we do that as an admin side script, just for sanity?

I double checked this and confirmed that published assets revalidate without error.

Issue #1494, opened 2023-02-20
Unable to upload Zarrs to staging because "There were headers present in the request which were not signed"


jwodder:
Currently, attempting to upload a Zarr to staging fails when PUTing a file to S3 with a 403 error and a response of the form:

```xml
<Error>
  <Code>AccessDenied</Code>
  <Message>There were headers present in the request which were not signed</Message>
  <HeadersNotSigned>content-md5</HeadersNotSigned>
  <RequestId>S086HFMAEH43CNF6</RequestId>
  <HostId>9Uenhku5cje+7L4e4+h3y0aBwv4CsC4+d/01j/57xuBTeF5RdGtJ8jTtThUFAJ6EpcsjXTMXLi8=</HostId>
</Error>
```

Upload succeeds when testing against a Dockerized environment.  I have not tried uploading to production.

waxlamp:
@AlmightyYakob, can you look into this?

In case it helps, I found [this SO question](https://stackoverflow.com/questions/45653850/there-were-headers-present-in-the-request-which-were-not-signed) but it is a few years old and I don't know if it applies to our situation.

AlmightyYakob:
@danlamanna and I discussed this offline, and he's going to be taking it on.

jwodder:
Resolved.

AlmightyYakob:
Resolved via #1497, #1498 and https://github.com/dandi/dandi-cli/pull/1215

PR #1497, opened 2023-02-21
Add back md5 to zarr object url signing

danlamanna:
Related #1494 

The MD5 parameter needs to be part of the original signed URL, otherwise S3 will reject the upload URL. This means the MD5 header used by the client is mandatory.

The CLI will need to update the `/files` step to take a `[{path: foo, md5: bar}, ...]` structure instead of a plain list of paths.

### DIFF
--- a/dandiapi/api/storage.py
+++ b/dandiapi/api/storage.py
@@ -143,13 +143,14 @@ class VerbatimNameS3Storage(VerbatimNameStorageMixin, TimeoutS3Boto3Storage):
                 return etag[1:-1]
             return etag
 
-    def generate_presigned_put_object_url(self, blob_name: str) -> str:
+    def generate_presigned_put_object_url(self, blob_name: str, md5: str) -> str:
         return self.connection.meta.client.generate_presigned_url(
             ClientMethod='put_object',
             Params={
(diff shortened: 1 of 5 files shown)


AlmightyYakob (approved):

AlmightyYakob:
@jwodder This will require another update from the CLI

jwodder:
Will the CLI still need to supply a `Content-MD5` header when PUTing file contents?

danlamanna:
> Will the CLI still need to supply a `Content-MD5` header when PUTing file contents?

Yes, that's what allows S3 to refuse it at time of upload if it's incorrect.

jwodder:
PR to CLI: https://github.com/dandi/dandi-cli/pull/1215

danlamanna:
Merging this side of things first.

jwodder:
I tried to use my CLI PR to upload to staging, but it failed with a 403 and the message "The request signature we calculated does not match the signature you provided. Check your key and signing method."

danlamanna:
Yeah, sorry about this. We're in an unfortunate situation where our local object store behaves differently than S3 in this specific case, so we're developing a bit blind. I'll update when I figure out what's going on.

PR #1501, opened 2023-02-24
Fix max length for name/description on Create Dandiset page

### LINKED ISSUE #1474: dandiset title character limit agreement
The title field of the new dandiset form has a character limit of 120

image

The title field of the dandiset metadata editor has a character limit of 150

image

It looks like the proper number might be 150

https://github.com/dandi/dandi-schema/blob/71a8ea3f8231cc8bf4fadcc51ea76db86171925f/dandischema/models.py#L313

mvandenburgh:
There was an inconsistency because we were hardcoding the max lengths for name and description, while the meditor was pulling them from the schema. This PR updates the create dandiset page to pull these values from the schema as well.

Fixes #1474

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -10,7 +10,7 @@
           label="Name*"
           hint="Provide a title for this dataset"
           persistent-hint
-          :counter="120"
+          :counter="nameMaxLength"
           required
           outlined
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):

AlmightyYakob:
My only concern here is if the schema is guaranteed to be present when this dialog is opened, but the licenses field does the same thing, so it's not creating a new issue.

PR #1503, opened 2023-02-24
Optimize Dandiset List Endpoint

AlmightyYakob:
This PR does a couple of things:

1. Updates the function for retrieving a version's root paths to use the `path` field, instead of a count of parent links
2. Updates the dandiset list endpoint to make use of already aggregated asset paths information, instead of summing and counting all assets in a version on every request.


dandiapi/api/views/dandiset.py
```diff
@@ -163,47 +164,59 @@ def get_object(self):
...
+        for path in root_paths:
+            if path.version_id not in version_stats:
+                version_stats[path.version_id] = {'total_size': 0, 'num_assets': 0}
+            version_stats[path.version_id]['total_size'] += path.aggregate_size
+            version_stats[path.version_id]['num_assets'] += path.aggregate_files
+
+        # Create a map from dandiset IDs to their draft and published versions
         # Because of above query, a max of 1 of each (per dandiset) will be present.
         dandisets_to_versions = {}
         for version in versions.iterator():
```
mvandenburgh, line 204: Doesn't necessarily have to be this PR, but removing `.iterator()` here might also help performance (although I'm not sure if we have enough versions in the database for it to actually matter yet).

dandiapi/api/views/dandiset.py
```diff
@@ -163,47 +164,59 @@ def get_object(self):
...
+        drafts = relevant_versions.filter(version='draft')
+
+        # Union published with drafts
         versions = published.union(drafts).order_by('dandiset_id', '-version')
 
-        # Organize draft and most recently published for each dandiset.
+        # Map version IDs to their stats
+        version_stats = {}
+        root_paths = get_root_paths_many(versions=relevant_versions)
+        for path in root_paths:
```
mvandenburgh, line 195: Won't introducing this loop over this queryset makes this endpoint's run time grow linearly with the number of root-level assets in each dandiset? i.e. if someone uploads a dandiset with 10,000 assets at the root level, will that add 10,000 iterations to this loop?

dandiapi/api/views/dandiset.py
AlmightyYakob: Good point, since `versions` is restricted to at most 1000 dandisets, and so is also limited in size.

dandiapi/api/views/dandiset.py
AlmightyYakob: I believe that's correct, but it's not a problem due to the following:

1. While it may be linear, the "coefficient" of that is _extremely low_, since we're dealing with simple postgres queries
2. Based on the data we've seen so far, it's unlikely for a dandiset to contain 10,000 assets at the root level (most I've seen is ~1100). In fact, at the time of writing this, the entire production archive has barely more than 10,000 root paths across all versions
3. Even if we run this loop over ~10k root paths, it takes ~200ms to run, which I think is acceptable, and certainly better than what we had

So for those reasons, I don't see a reason to be concerned about this yet.

mvandenburgh: That makes sense to me 👍

dandiapi/api/models/asset_paths.py
```diff
@@ -24,8 +24,11 @@ class AssetPath(models.Model):
 
     class Meta:
         constraints = [
-            # Disallow slashses in path
-            models.CheckConstraint(check=~models.Q(path__endswith='/'), name='consistent-slash'),
+            # Disallow slashses at the beginning or end of path
```
mvandenburgh, line 27:
```suggestion
            # Disallow slashes at the beginning or end of path
```
😄 I'm surprised our codespell check didn't catch this

mvandenburgh (approved):

dandiapi/api/models/asset_paths.py
waxlamp: (all that to say: `slashses` probably is not a "common misspelling" 🤷🏾)

mvandenburgh: Ah, that makes a lot more sense

PR #1505, opened 2023-02-27
Only allow selecting one license on create dandiset page

### LINKED ISSUE #1476: selecting multiple licenses
The new dandiset form allows you to select both CC0 and CC-BY licenses. What would this mean? I don't see how a dandiset could be under both licenses at the same time.

image

mvandenburgh:
Closes #1476

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -81,7 +81,6 @@
             :items="dandiLicenses"
             label="License*"
             class="my-4"
-            multiple
             outlined
             dense
           />
@@ -114,7 +113,7 @@ import { useDandisetStore } from '@/stores/dandiset';
 import { useRouter } from 'vue-router/composables';
 
 import type { ComputedRef } from 'vue';
-import type { IdentifierForAnAward, LicenseType, License } from '@/types';
+import type { IdentifierForAnAward, LicenseType } from '@/types';
 
 // Regular expression to validate an NIH award number.
 // Based on https://era.nih.gov/files/Deciphering_NIH_Application.pdf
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -159,7 +158,14 @@ if (!loggedIn()) {
 }
 
 async function registerDandiset() {
-  const metadata = { name: name.value, description: description.value, license: license.value };
+  const metadata: Record<string, any> = {
```
waxlamp, line 161: Shouldn't this be `Record<"name" | "description", string>`?

AlmightyYakob: I'm guessing the reason this was done was to prevent the type checker from complaining when the `license` field is added below.

danlamanna: Can you do
```
const metadata: {name: string, description: string, license?: LicenseType} = ...
```
:man_shrugging:

mvandenburgh: > Can you do
> 
> ```
> const metadata: {name: string, description: string, license?: LicenseType} = ...
> ```
> 
> man_shrugging


Yeah, that syntax is much better. And my use of `any` here also inadvertently hid a bug :disappointed: 

Screenshot from 2023-02-27 19-49-06

I thought I tested it on the deploy preview and it worked, but clearly I messed something up

mvandenburgh: Follow up on this - #1510

waxlamp: Typechecking for the win!

PR #1523, opened 2023-03-06
Add button to copy id to clipboard

### LINKED ISSUE #1194: add copy button for id
can we have a copy icon next to the ID: 
image

that will copy the id into a clipboard?

### LINKED ISSUE #1451: make doi clear on a published DLP
at present there is no clear indication that there is a doi associated with a published dandiset. we either need to add a specific element (see any zenodo page). or improve the dropdown for cite to show the full citation including the doi. also share does not show the doi for a published dandiset.

### LINKED ISSUE #1525: Add a button to copy dandiset DOI
See https://github.com/dandi/dandi-archive/pull/1523#issuecomment-1457166421 for original discussion.

marySalvi:
closes #1194 
closes #1525
closes #1451

Adds the ability to copy the ID to clipboard

image

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetMain.vue
+++ b/web/src/views/DandisetLandingView/DandisetMain.vue
@@ -6,17 +6,54 @@
       outlined
     >
       <v-row class="mx-2 my-2 mb-0">
-        <v-col>
+        <v-col
+          class="d-flex align-center"
+        >
           <h1 :class="`font-weight-light ${$vuetify.breakpoint.xs ? 'text-h6' : ''}`">
             <ShareDialog />
             {{ meta.name }}
           </h1>
+          <v-chip
+            v-if="meta.doi"
+            outlined
+            class="mx-2 pl-1"
+          >
+            <v-tooltip
+              top
+            >
+              <template #activator="{ on, attrs }">
+                <v-btn
+                  v-if="currentDandiset !== null"
+                  icon
+                  small
+                  v-bind="attrs"
+                  v-on="on"
+                  @click="copy('DOI')"
+                >
+                  <v-icon
+                    small
+                  >
+                    mdi-content-copy
+                  </v-icon>
+                </v-btn>
+              </template>
+              <span>Copy DOI URL to clipboard</span>
+            </v-tooltip>
+            <span>
+              DOI:
+            </span>
+            <v-divider
+              vertical
+              class="mx-2"
+            />
+            {{ meta.doi }}
+          </v-chip>
         </v-col>
       </v-row>
       <v-row class="mx-1">
         <v-col :cols="$vuetify.breakpoint.xs ? 12 : 3">
           <v-chip
-            class="text-wrap py-1"
+            class="text-wrap py-1 pl-1"
             style="text-align: center;"
             outlined
           >
@@ -346,9 +383,7 @@ export default defineComponent({
     }
 
     function copy(value:string) {
-      if (value === 'dandiID') {
-        navigator.clipboard.writeText(`DANDI:${currentDandiset.value?.dandiset.identifier}`);
-      }
+      navigator.clipboard.writeText(value === 'dandiID' ? meta.value?.identifier as string : `https://doi.org/:${meta.value?.doi}`);
     }
 
     return {
--- a/web/src/views/DandisetLandingView/DandisetMain.vue
+++ b/web/src/views/DandisetLandingView/DandisetMain.vue
@@ -31,9 +31,7 @@
                   v-on="on"
                   @click="currentDandiset ? copy(currentDandiset.dandiset.identifier) : ''"
                 >
-                  <v-icon
-                    color="white"
-                  >
+                  <v-icon>
                     mdi-content-copy
                   </v-icon>
                 </v-btn>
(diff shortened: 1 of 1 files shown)


satra:
it would be nice if this would copy the dandi accession id for draft dandisets, but DOIs for published dandisets.

AlmightyYakob:
> it would be nice if this would copy the dandi accession id for draft dandisets, but DOIs for published dandisets.

I'd think in that case we should add the DOI as a first hand field on the DLP, with a copy button.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -20,9 +20,30 @@
...
+              <template #activator="{ on, attrs }">
+                <v-btn
+                  v-if="currentDandiset !== null"
+                  icon
+                  v-bind="attrs"
+                  v-on="on"
+                  @click="currentDandiset ? copy(currentDandiset.dandiset.identifier) : ''"
+                >
+                  <v-icon
+                    color="white"
```
AlmightyYakob, line 35: Are you running a dark mode extension in your browser? In the preview, this copy button being white makes it hard to see, and seems like there's nothing there

marySalvi: ah yes. I have it turned off for 8080 but didn't think about it since we are on 8085. I'll fix

marySalvi: fixed in f173382357ff401eb728606139da8570b486d49b

waxlamp:
> > it would be nice if this would copy the dandi accession id for draft dandisets, but DOIs for published dandisets.
> 
> I'd think in that case we should add the DOI as a first hand field on the DLP, with a copy button.

Seconded. I have opened #1525 to track this separately.

mvandenburgh (approved):

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -20,9 +20,31 @@
...
+              top
+            >
+              <template #activator="{ on, attrs }">
+                <v-btn
+                  v-if="currentDandiset !== null"
+                  icon
+                  small
+                  v-bind="attrs"
+                  v-on="on"
+                  @click="currentDandiset ? copy(currentDandiset.dandiset.identifier) : ''"
```
satra, line 33: i think we want to do this for copying the identifer, as that is the accession id for the dandiset.

```suggestion
                  @click="currentDandiset ? copy('DANDI:' + currentDandiset.dandiset.identifier) : ''"
```

AlmightyYakob: Furthermore,  I don't like the idea of using a ternary expression as an if statement to perform an action. I think we should just have `@click="copy"` here, and move the if condition into that function.

marySalvi: updated in 6d3b2bfa463bce605d92904392f99d25f6d6ddf1

marySalvi:
I decided to combine these three issues, #1194 #1525 and #1451, into 1 pr because they are so similar.

satra:
thank you @marySalvi - for the published dandiset, the id should include the version (so `DANDI:<dandiset id>/<version>` see the id tag over here: https://api.dandiarchive.org/api/dandisets/000402/versions/0.230307.2132/

marySalvi:
> thank you @marySalvi - for the published dandiset, the id should include the version (so `DANDI:<dandiset id>/<version>` see the id tag over here: https://api.dandiarchive.org/api/dandisets/000402/versions/0.230307.2132/

Thank you for the clarification. I'll update.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -323,6 +382,11 @@ export default defineComponent({
       return moment(date).format('LL');
     }
 
+    function copy(value:string) {
+      const version = meta.value?.version === 'draft' ? meta.value?.identifier as string : meta.value?.id as string;
+      navigator.clipboard.writeText(value === 'dandiID' ? version : `https://doi.org/:${meta.value?.doi}`);
+    }
```
AlmightyYakob, line 388: In this situation, if `meta.value` is 	`undefined`, `undefined` or `https://doi.org/:undefined` will be copied to the clipboard. Since this is essentially conveying that the metadata needs to be present, why don't we just explicitly enforce that:

```suggestion
    function copy(value: string) {
      if (!meta.value) {
        throw new Error('metadata is undefined!');
      }

      const version = meta.value.version === 'draft' ? meta.value.identifier : meta.value.id;
      navigator.clipboard.writeText(value === 'dandiID' ? version : `https://doi.org/:${meta.value.doi}`);
    }
```

AlmightyYakob (changes requested):
Looks good overall, one suggestion about the copy function logic

web/src/views/DandisetLandingView/DandisetMain.vue
marySalvi: added here: 92c257ba17bc8d31626e86a4b7f28767c51975ea

AlmightyYakob (approved):

PR #1529, opened 2023-03-10
Fix zarr publish error bug

mvandenburgh:
`params` must be used to pass query parameters for a GET request with axios. Since we weren't using that, no dandiset id was being sent to the zarr endpoint, causing the zarr endpoint to return the count of zarrs in the entire system, which in turn caused dandisets with no zarrs to be flagged as having zarrs.

### DIFF
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -116,13 +116,13 @@ const dandiRest = new Vue({
       }
     },
     async zarr({ dandiset }: { dandiset?: string }) : Promise<Paginated<Zarr>> {
-      const data: { dandiset?: string } = {};
+      const params: { dandiset?: string } = {};
       if (dandiset !== undefined) {
-        data.dandiset = dandiset;
+        params.dandiset = dandiset;
       }
 
       const resp = await client.get('zarr/', {
-        data,
+        params,
       });
 
       return resp.data;


AlmightyYakob (approved):
LGTM, as long as you've tested locally I approve

mvandenburgh:
@waxlamp fwiw this was an oversight from #1517, although a really weird set of circumstances caused it to go unnoticed in testing for me, and I suspect you too. Because I had no zarrs in my local environment, it still appeared to work ok for me when I tested it on a non-zarr dandiset. And then when I tested it on the Zarr dandiset using the netlify deploy preview, it correctly flagged it as non-publishable (but not because there were zarrs in that dandiset, but because staging in general contains zarrs :sweat_smile: )

waxlamp:
> @waxlamp fwiw this was an oversight from #1517, although a really weird set of circumstances caused it to go unnoticed in testing for me, and I suspect you too. Because I had no zarrs in my local environment, it still appeared to work ok for me when I tested it on a non-zarr dandiset. And then when I tested it on the Zarr dandiset using the netlify deploy preview, it correctly flagged it as non-publishable (but not because there were zarrs in that dandiset, but because staging in general contains zarrs sweat_smile )

Yikes! Thank you for tracking this down and fixing so quickly.

PR #1542, opened 2023-03-20
Boost minimal client version to be 0.51.0 since needed for fresh dandischema

### LINKED ISSUE #1408: update minimal client version
ATM we have `"cli-minimal-version": "0.14.2",` whenever dandi-cli already `0.46.6-215-g8e674820` and we added better validation. So we should

- [x] release new dandi-cli
- [x] test it a little in the real
- [x] provide PR here boosting `cli-minimal-version` to that new version released

yarikoptic:
and supported zarr upload mechanisms etc

Closes #1408

### DIFF
--- a/dandiapi/api/tests/test_info.py
+++ b/dandiapi/api/tests/test_info.py
@@ -11,7 +11,7 @@ def test_rest_info(api_client):
         'schema_version': settings.DANDI_SCHEMA_VERSION,
         'schema_url': schema_url,
         'version': __version__,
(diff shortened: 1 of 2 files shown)


yarikoptic:
yes, as long as they are to work with the archive instance. Then should still be able to use local commands (`ls`, `organize` etc) without upgrading.

yarikoptic:
@satra should we?

satra:
sure.

PR #1545, opened 2023-03-20
Add command for asset metadata re-extraction

### LINKED ISSUE #1118: Add a mechanism to recompute asset metadata for uploaded assets
For instance, when the asset metadata schema changes, this kind of recomputation would become necessary.

`dandi-cli` currently performs this computation; abstracting it out into the `dandi` package (or making use of the functionality if it is already so abstracted) would enable us to write a serverside task which performs the same function.

Also needed: a way to trigger this recomputation for all assets in the system. This might be a special purpose button we add to the admin console views, or we could possibly work with the existing admin app to trigger an action on a selection of assets.

@yarikoptic, @satra: does this mean that `dandi-cli` should no longer perform this function at all, leaving it for the server to do during upload and finalization?

### LINKED ISSUE #1450: updating metadata on all draft assets

jjnesbitt:
Closes #1450
Closes #1118

### DIFF
--- a/dandiapi/api/management/commands/extract_metadata.py
+++ b/dandiapi/api/management/commands/extract_metadata.py
@@ -1,0 +1,121 @@
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+from dandi.dandiapi import RemoteReadableAsset
+from dandi.metadata.nwb import nwb2asset
+from dandi.misctypes import Digest, DigestType
+from dandischema.models import get_schema_version
+from django.contrib.auth.models import User
+from django.db import transaction
+import djclick as click
+from tqdm import tqdm
+
+from dandiapi.api.models import Asset, Dandiset, Version
+from dandiapi.api.services.asset import change_asset
+
+logger = logging.getLogger(__name__)
+
+
+@click.group()
+def group():
+    pass
+
+
+def get_asset_digest(asset: Asset) -> Digest:
+    if asset.zarr is not None:
+        return Digest(algorithm=DigestType.dandi_zarr_checksum, value=asset.zarr.checksum)
+    if asset.sha256 is not None:
+        return Digest(algorithm=DigestType.sha2_256, value=asset.sha256)
+
+    # Default to Etag
+    if asset.blob is not None:
+        return Digest(algorithm=DigestType.dandi_etag, value=asset.blob.etag)
+
+    raise Exception('Unsupported asset type')  # noqa: TRY002
+
+
+def extract_asset_metadata(asset: Asset, draft_version: Version):
+    readable_asset = RemoteReadableAsset(
+        asset.s3_url, size=asset.size, mtime=asset.modified, name=Path(asset.path).name
+    )
+
+    if not asset.path.lower().endswith('.nwb'):
+        logger.info('Asset %s: Not an NWB file, skipping...', asset.path)
+        return
+
+    new_metadata = nwb2asset(
+        readable_asset, digest=get_asset_digest(asset), schema_version=get_schema_version()
+    ).json_dict()
+
+    # Use dandiset owner, default to some admin user
+    user = (
+        draft_version.dandiset.owners.first()
+        or User.objects.filter(is_superuser=True, is_staff=True).first()
+    )
+
+    # Lock asset and version to ensure they don't change while performing this operation
+    with transaction.atomic():
+        locked_asset = Asset.objects.select_for_update().get(id=asset.id)
+        locked_version = Version.objects.select_for_update().get(id=draft_version.id)
+
+        # Ensure asset hasn't already been removed from the version
+        if not locked_version.assets.filter(id=locked_asset.id).exists():
+            raise Exception(f'Asset {locked_asset} no longer exists in version {locked_version}')  # noqa: TRY002
+
+        # Replace old asset with new asset containing updated metadata
+        change_asset(
+            user=user,
+            asset=locked_asset,
+            version=locked_version,
+            new_asset_blob=asset.blob,
+            new_zarr_archive=asset.zarr,
+            new_metadata=new_metadata,
+        )
+
+
+def extract_dandiset_assets(dandiset: Dandiset):
+    # Only update NWB assets which are out of date and do not belong to a published version
+    assets = dandiset.draft_version.assets.filter(
+        published=False,
+        path__iendswith='.nwb',
+        metadata__schemaVersion__lt=get_schema_version(),
+    ).select_related('blob', 'zarr')
+    if not assets:
+        logger.info('No old draft NWB assets found in dandiset %s. Skipping...', dandiset)
+        return
+
+    for asset in tqdm(assets):
+        extract_asset_metadata(asset=asset, draft_version=dandiset.draft_version)
+
+
+@group.command(help='Re-extracts the metadata of this asset')
+@click.argument('asset_id')
+def asset(asset_id: str):
+    asset = Asset.objects.get(asset_id=asset_id)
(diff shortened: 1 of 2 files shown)


yarikoptic:
What's your plan @AlmightyYakob here - wasn't it close to completion?

AlmightyYakob:
> What's your plan @AlmightyYakob here - wasn't it close to completion?

I unfortunately fell behind on this, as other things took priority. There are still logistical issues to work out on this PR, and at the moment my current priority is the embargo re-design. After that, I could take another pass at this.

yarikoptic:
this might be relevant in the scope/to address
- https://github.com/dandi/dandi-archive/issues/1970

> my current priority is the embargo re-design. After that, I could take another pass at this.

embargo is a curse! ;)  we should see how to move this from the stalled point.  attn @waxlamp

yarikoptic:
in principle even

- https://github.com/dandi/dandi-archive/issues/2024

is related here since we might want to redo extraction upon dandischema upgrade.

@jjnesbitt do you have plans to get back to this PR?

jjnesbitt:
I believe this is ready to go. @jwodder or @yarikoptic can you verify that the `nwb2asset` function is being used correctly? As far as I'm aware from local testing it seems to be working correctly, but I don't exactly have a very good example asset to test the before and after. If you have a good example asset for this, that would also be helpful.

dandiapi/api/management/commands/extract_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+def extract_asset_metadata(asset: Asset, draft_version: Version):
+    readable_asset = RemoteReadableAsset(
+        asset.s3_url, size=asset.size, mtime=asset.modified, name=Path(asset.path).name
+    )
+
+    if not asset.path.lower().endswith('.nwb'):
+        logger.info('Asset %s: Not an NWB file, skipping...', asset.path)
+        return
+
+    new_metadata = nwb2asset(readable_asset, schema_version=get_schema_version()).json_dict()
```
yarikoptic, line 34: if would be close to what we use (for nwb only) in [NWBAsset.get_metadata](https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files/bases.py#L475)

notable differences:
- we provide digest (not schema_version) . I guess should be added here too?
- we fallback to default record if extraction fails for some reason
- we assign .path explicitly to be posix path from the top of the dandiset, but I think here it would be matching fine the one of RemoteReadableAsset (worth checking on example)

waxlamp:
> I believe this is ready to go. @jwodder or @yarikoptic can you verify that the `nwb2asset` function is being used correctly? As far as I'm aware from local testing it seems to be working correctly, but I don't exactly have a very good example asset to test the before and after. If you have a good example asset for this, that would also be helpful.

@yarikoptic, @jwodder: could you address Jake's question?

dandiapi/api/management/commands/extract_metadata.py
jjnesbitt: I've updated my script to provide the digest as well. I chose not to automatically handle errors as I'm not sure that's the behavior we want. Since this is a management command to be run manually, it's alright to manually intervene in that failure mode.

yarikoptic:
Looks good to me.  I would have tried on some asset on staging and checked how/if metadata got changed.

waxlamp:
> Looks good to me. I would have tried on some asset on staging and checked how/if metadata got changed.

Could you point out a good asset on staging for this kind of test?

yarikoptic:
sorry, I don't have any specific one in mind. I could only suggest some random sensible  one, e.g. what about this one
https://api-staging.dandiarchive.org/api/dandisets/200560/versions/draft/assets/ad5d6b54-9974-4559-b400-88b406be983d/
from 
https://gui-staging.dandiarchive.org/dandiset/200560/draft/files?location=for_jeremy&page=1
?  it is quite large so would also ensure that online access works fine

dandiapi/api/management/commands/extract_metadata.py
```diff
@@ -0,0 +1,121 @@
...
+    assets = dandiset.draft_version.assets.filter(
+        published=False,
+        path__iendswith='.nwb',
+        metadata__schemaVersion__lt=get_schema_version(),
+    ).select_related('blob', 'zarr')
+    if not assets:
+        logger.info('No old draft NWB assets found in dandiset %s. Skipping...', dandiset)
+        return
+
+    for asset in tqdm(assets):
```
mvandenburgh, line 89: For large dandisets, we could potentially see some excessive memory usage here.

```suggestion
    for asset in tqdm(assets.iterator(), total=assets.count()):
```

dandiapi/api/management/commands/extract_metadata.py
```diff
@@ -0,0 +1,121 @@
...
+)
+@click.argument('dandiset_id')
+def dandiset(dandiset_id: str):
+    dandiset = Dandiset.objects.get(id=int(dandiset_id))
+    extract_dandiset_assets(dandiset)
+
+
+@group.command(name='all', help='Re-extracts the metadata of all assets in all draft versions')
+def all_dandisets():
+    for dandiset in Dandiset.objects.all():
```
mvandenburgh, line 119:
```suggestion
    for dandiset in Dandiset.objects.iterator():
```

dandiapi/api/management/commands/extract_metadata.py
```diff
@@ -0,0 +1,121 @@
...
+def asset(asset_id: str):
+    asset = Asset.objects.get(asset_id=asset_id)
+    draft_versions = asset.versions.filter(version='draft')
+    if not draft_versions.exists():
+        raise click.ClickException(
+            'Cannot re-extract metadata of asset that has no associated draft versions.'
+        )
+
+    # Re-extract for every draft version
+    for version in draft_versions:
```
mvandenburgh, line 104: This isn't likely necessary, but technically the number of versions is unbounded and we're running inside a management command, so there's not really any downsides
```suggestion
    for version in draft_versions.iterator():
```

mvandenburgh (changes requested):
I can't speak to the NWB/dandi CLI code correctness, but the rest of the logic looks fine to me. Just some suggestions to use `.iterator()` when iterating over querysets when possible.

dandiapi/api/management/commands/extract_metadata.py
```diff
@@ -37,8 +41,12 @@ def get_asset_digest(asset: Asset) -> Digest:
 
 
 def extract_asset_metadata(asset: Asset, draft_version: Version):
+    # Test
+    s3_url = asset.s3_url.replace(
+        'http://localhost:9000/dandi-dandisets', 'https://dandiarchive.s3.amazonaws.com'
+    )
```
mvandenburgh, line 47: Is this leftover from debugging?

jjnesbitt: Whoops yes, thanks. That's the only good way I have to test the re-extraction.

mvandenburgh (approved):
If you rebase, the playwright tests should pass (see #2042)

mvandenburgh (approved):

PR #1547, opened 2023-03-21
Improve the admin interface for versions

danlamanna:
This just adds some basic text searching and status filtering to the version admin.


mvandenburgh (approved):
Nice! I'll also add release labels here so that we can use this in production immediately after merging.

waxlamp (commented):
Pretty cool!

PR #1548, opened 2023-03-21
Optimize dandiset/versions and dandiset/versions/info endpoints

AlmightyYakob:
These endpoint were making an excessive number of SQL queries and aggregations. This may have gone unnoticed due to this behavior stemming from properties on the `Version` model. I noticed this when viewing the DLP of dandiset `000026`, which takes several seconds to load due to these queries. This PR should greatly reduce that time.

This PR also removes the `publish_status` property from the `Version` model. This property is funtionally equivalent to the `valid` property, and causes unnecessary API slowdown, as the same information is already included in the `asset_validation_errors` field.

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -1,14 +1,21 @@
 from __future__ import annotations
 
 import os
+from typing import TYPE_CHECKING
 
 from django.db import IntegrityError, transaction
 from django.db.models import F, QuerySet, Sum
 from django.db.models.functions import Coalesce
 from tqdm import tqdm
 
-from dandiapi.api.models import Asset, AssetPath, AssetPathRelation, Version
-from dandiapi.zarr.models import ZarrArchive
+# Import only the asset paths models
+from dandiapi.api.models.asset_paths import AssetPath, AssetPathRelation
+
+# Import models for type checking only (prevent cyclic imports)
+if TYPE_CHECKING:
+    from dandiapi.api.models import Asset, Version
+    from dandiapi.zarr.models import ZarrArchive
+
 
 ####################################################################
 # Dandiset and version deletion will cascade to asset path deletion.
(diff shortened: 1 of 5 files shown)


AlmightyYakob:
I've checked with @jwodder and confirmed that the CLI doesn't use the Version `status` field as a way to check if any contained assets are Valid/Invalid.

danlamanna:
> I've checked with @jwodder and confirmed that the CLI doesn't use the Version `status` field as a way to check if any contained assets are Valid/Invalid.

Does the GUI use `status` anywhere?

AlmightyYakob:
> Does the GUI use `status` anywhere?

Yes, but that question is too broad in the context of this PR imo. I think the real question is "Does the GUI use the `status` field in a way that depends on the assets themselves?"

To that I'd say yes, right here: https://github.com/dandi/dandi-archive/blob/5413b8e4ce7d466e79bfcfbae5f345802d5d3aaa/web/src/components/DLP/OverviewTab.vue#L313

However, I don't think this change really affects that, since I don't see how that field is accurately used anyway. If we look at what `publish_status` is currently returning as the `status` field:

```python
def publish_status(self) -> Version.Status:
    if self.status != Version.Status.VALID:
        return self.status

    # Import here to avoid dependency cycle
    from .asset import Asset

    invalid_asset = self.assets.exclude(status=Asset.Status.VALID).first()
    if invalid_asset:
        return Version.Status.INVALID

    return Version.Status.VALID
```

This wouldn't result in a `status` of `PENDING` being changed.

---

Since it seems that any other blockers are already caught by directly checking the `version_validation_errors` and `asset_validation_errors` fields, my opinion is that we should leave those fields alone on the front end, and if we observe any unintended behavior, we can step in then. This seems like another situation where our confusing metadata lifecycle is biting us.

danlamanna (approved):

PR #1554, opened 2023-04-03
Add license info to dandiset creation page + do a style overhaul

### LINKED ISSUE #758: Improve dandiset creation screen with more information for end user
* the License (i) icon should offer more helpful hints about the different license types

* ~provide brief instruction on the top:~
	~Red is required~
	~To add funding information, …~ (Roni: I don't think this is still valid; all fields in the form are required, and funding information in particular must be added at a later time; @bendichter, please explain in a new comment if you disagree)

* ~Protocol (i): protocol.io -> protocols.io~ (Roni: this was fixed at some point)

waxlamp:
- closes #758
- depends on https://github.com/dandi/handbook/pull/82 being merged and deployed

Some background: my intent here was to add a link to a new handbook page including information about licenses (see the first commit in this PR, https://github.com/dandi/dandi-archive/commit/4a42b189a3caeeb446f2ff7511df721ecc099751). But due to the way I had to add that information (without using a `hint` prop since the v-select component lacks it), this created a style dissonance. Then I looked closer and discovered several other inconsistencies in the page, which I addressed over the course of the remaining commits.

So I'd be ok with reformulating some of the changes into follow-on PRs, but I also think the improvements are of enough value that we could just accept them in this PR. As for reviews, I would request that @bendichter focus on the new content (i.e. the link to the handbook page and the text surrounding it) while @mvandenburgh focus on the overall changes I've made (as well as Vuetify style, etc.).

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -5,38 +5,9 @@
     </v-card-title>
     <v-card-text class="my-3">
       <v-form>
-        <h1>Dandiset Title</h1>
-        <div>
-          Provide a title for this Dandiset. The title will appear in search
-          results and at the top of the home page for this Dandiset, so make it
-          concise and descriptive.
-        </div>
-        <v-text-field
-          v-model="name"
-          label="Title"
-          :counter="nameMaxLength"
-          required
-          outlined
-        />
-
-        <h1>Description</h1>
-        <div>
-          Provide a description for this Dandiset. This will appear prominently
-          under the title in the home page for this Dandiset.
-        </div>
-        <v-textarea
-          v-model="description"
-          label="Description"
-          :counter="descriptionMaxLength"
-          required
-          outlined
-          class="my-4"
-        />
         <div>
           <v-checkbox
             v-model="embargoed"
-            hide-details
-            class="shrink mr-2 mt-0"
           >
             <template #label>
               Embargo this Dandiset
@@ -67,21 +38,6 @@
               </v-tooltip>
             </template>
           </v-checkbox>
-          <v-text-field
-            v-if="embargoed"
-            v-model="awardNumber"
-            label="Award number*"
-            hint="Provide an NIH award number for this embargoed dataset.
-                Note: this can be changed at any time and additional award
-                numbers can be added later."
-            persistent-hint
-            :counter="120"
-            :required="embargoed"
-            outlined
-            class="mt-4 shrink"
-            style="width: 20vw;"
-            :rules="awardNumberRules"
-          />
         </div>
         <h1>Title</h1>
         <div>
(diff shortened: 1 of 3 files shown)


web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -51,41 +30,82 @@
...
-            :counter="120"
-            :required="embargoed"
-            outlined
-            class="mt-4 shrink"
-            style="width: 20vw;"
-            :rules="awardNumberRules"
-          />
+          </v-switch>
+        </div>
+        <h1>Title</h1>
```
mvandenburgh, line 42: The "vuetify way" to do this is to use the `text-h1` class instead of delegating styling choices to HTML with the `h#` elements.
```suggestion
        <div class="text-h1">Title</div>
```

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -51,41 +30,82 @@
...
+          v-model="name"
+          label="Title"
+          :counter="nameMaxLength"
+          required
+          outlined
+          dense
+          class="my-4"
+        />
+
+        <h1>Description</h1>
```
mvandenburgh, line 58: Same as above comment.
```suggestion
        <div class="text-h1">Description</div>
```

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -51,41 +30,82 @@
...
+          v-model="description"
+          label="Description"
+          :counter="descriptionMaxLength"
+          required
+          outlined
+          dense
+          class="my-4"
+        />
         <div v-if="!embargoed">
+          <h1>License</h1>
```
mvandenburgh, line 73: Same as above comment.
```suggestion
          <div class="text-h1">License</div>
```

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -51,41 +30,82 @@
...
-            label="License*"
+            label="License"
             class="my-4"
             outlined
             dense
           />
         </div>
-        <small class="float-right font-weight-bold">*indicates required field</small>
+        <div v-else>
+          <h1>NIH Award Number</h1>
```
mvandenburgh, line 91: Same as above comment.
```suggestion
          <div class="text-h1">NIH Award Number</div>
```

mvandenburgh (changes requested):

web/src/views/CreateDandisetView/CreateDandisetView.vue
waxlamp: I did notice that, but also the `text-h1` class gives a _very_ different outcome than just using an `h1` element. I'll have to figure out what the right class to use is.

waxlamp: `text-h4` does the trick. See f1cef20b8d9db90d0a24220dcbc721393749f8eb.

web/src/views/CreateDandisetView/CreateDandisetView.vue
waxlamp: See f1cef20b8d9db90d0a24220dcbc721393749f8eb.

web/src/views/CreateDandisetView/CreateDandisetView.vue
waxlamp: See f1cef20b8d9db90d0a24220dcbc721393749f8eb.

web/src/views/CreateDandisetView/CreateDandisetView.vue
waxlamp: See f1cef20b8d9db90d0a24220dcbc721393749f8eb.

mvandenburgh (approved):
lgtm!

PR #1557, opened 2023-04-03
Avoid clobbering version metadata when calculating assets summary

danlamanna:
Changes to version metadata that came in while `version_aggregate_assets_summary` is running can be overwritten. This change will only update the assetsSummary if the metadata is the same at save time, otherwise, it's a noop and the summary aggregation is retried.

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -81,16 +81,20 @@ def version_aggregate_assets_summary(version: Version) -> None:
     if version.version != 'draft':
         raise VersionHasBeenPublishedError
 
-    version.metadata['assetsSummary'] = aggregate_assets_summary(
+    assets_summary = aggregate_assets_summary(
         asset.full_metadata
         for asset in version.assets.filter(status=Asset.Status.VALID)
         .select_related('blob', 'zarr')
         .iterator()
     )
 
-    Version.objects.filter(id=version.id, version='draft').update(
-        modified=timezone.now(), metadata=version.metadata
+    updated_metadata = {**version.metadata, **{'assetsSummary': assets_summary}}
+
+    updated_count = Version.objects.filter(id=version.id, metadata=version.metadata).update(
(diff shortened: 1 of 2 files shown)


mvandenburgh (approved):

danlamanna:
This is dependent on #1638.

yarikoptic:
#1638 was merged in July.  If this PR is still pertinent -- please re-base/resolve conflicts.

waxlamp:
This was approved but never merged. Was there something blocking it, @danlamanna?

Additionally, it's now out of date, so if this is still a valid thing to merge (I believe it is) let's get it in shape to merge. @danlamanna do you want to do that yourself, or can you delegate to someone else?

danlamanna:
> This was approved but never merged. Was there something blocking it, @danlamanna?
> 
> Additionally, it's now out of date, so if this is still a valid thing to merge (I believe it is) let's get it in shape to merge. @danlamanna do you want to do that yourself, or can you delegate to someone else?

I had thought this was replacing a `Version.save` with a `Version.update`, but it's not. So it's still "broken" in the sense that we're missing the hooks called in `.save`, but it fixes the potential data loss bug with version changes occurring simultaneously with aggregate summaries, so we should merge it.

@mvandenburgh Did you see something else wrong with it?

mvandenburgh:
> > This was approved but never merged. Was there something blocking it, @danlamanna?
> > Additionally, it's now out of date, so if this is still a valid thing to merge (I believe it is) let's get it in shape to merge. @danlamanna do you want to do that yourself, or can you delegate to someone else?
> 
> I had thought this was replacing a `Version.save` with a `Version.update`, but it's not. So it's still "broken" in the sense that we're missing the hooks called in `.save`, but it fixes the potential data loss bug with version changes occurring simultaneously with aggregate summaries, so we should merge it.
> 
> @mvandenburgh Did you see something else wrong with it?

Oh, i thought the same thing :facepalm: 

No, I agree this should be merged

mvandenburgh (approved):

Issue #444, opened 2021-07-23
allow for exact path matching in dandisets_versions_assets_list and unify `path*` options


yarikoptic:
somewhat relates to https://github.com/dandi/dandi-api/issues/422 for `dandisets_versions_assets_paths_read` where situation is somewhat opposite: there is both `path` and `path_prefix` which are used as a `path` really (which should end with `/`, but as I just commented on there -- it accepts without now and causes different behavior) and not `path_prefix`.

in `dandisets_versions_assets_list` there is only `path` but it is used really as `path_prefix`. So to check for an asset path, users (like dandi-cli) need to loop through all assets which start with that prefix as done here: https://github.com/dandi/dandi-cli/blob/457c1e519c3daa75f729f686e963e637e5a78e91/dandi/dandiapi.py#L510 to select the one (if any) which matches that exact path.  Not too difficult to see that it might be a VERY expensive for heavy datasets while looking at some specific path which is common prefix to other directories (e.g. an asset under path `sub` at top level).

I think we should harmonize behavior/options across end points.  One choice:

- harmonize to have `path` (exact) and `path_prefix` (the prefix) across end points
- that would require then
  - changing `path` to be `path_prefix` in `dandisets_versions_assets_list` for current logic (**breaking change**)
  - adding handling for exact `path` matching
  - adding errorring out if both `path` and `path_prefix` provided (we are still weak in error reporting/handling)
 
an alternative:

- have `path` (only) and add `path_format={prefix,exact}`
- I kinda like this one more, since 
  - expandable: we could even add later `regex` (limitless capabilities ;) ) , `suffix` (find all .nwb assets?!)
  - backward compatible for `dandisets_versions_assets_list`
  - no need for checking combinatorics of options

waxlamp:
> somewhat relates to #422 for `dandisets_versions_assets_paths_read` where situation is somewhat opposite: there is both `path` and `path_prefix` which are used as a `path` really (which should end with `/`, but as I just commented on there -- it accepts without now and causes different behavior) and not `path_prefix`.
> 
[...]

PR #1566, opened 2023-04-06
Embed glob-derived regex inside a ^/$ pair

### LINKED ISSUE #1563: Tighten implementation of `glob` parameter
The internal usage of `glob` parameter of the [`dandisets_versions_assets_list` endpoint](https://github.com/dandi/dandi-archive/blob/83d3a10432f6554969542a2f8fc209bc5f9c269f/dandiapi/api/views/asset.py#L388) should be made more precise by embedding the resulting regex between a `^`/`$` pair, "closing" it. This would have the following effects:
- Entering a literal phrase would cause the backend to perform an exact match, something that is almost but not quite possible now.
- The existing behavior of an unrestricted string match would still be possible by supplying `*` characters around the search term (i.e., changing `foobar` to `*foobar*`).

This would be a breaking change, although the `glob` functionality is not used in the web client nor the CLI, and so it is unlikely it's being used anywhere else either.

See #444 for original discussion and context.

waxlamp:
Closes #1563

This makes the `glob` parameter of the asset search endpoint more useful. Now, when supplying a string to the parameter, the search will be for exact matches to the string. The original behavior can be recovered by prepending and appending `*`s to cause the search to allow for an arbitrary prefix/suffix to the supplied string.

This is a breaking change, but my suspicion is that no one is using the `glob` parameter at present; furthermore, this fix adds a new capability to the system of finding assets by exact path, something that was requested in #444.

We should discuss whether a warning to the userbase is necessary because of this change.

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -1457,6 +1457,9 @@ def test_asset_direct_info(api_client, asset):
         ('a/b/d/*', ['a/b/d/e.txt']),
         ('*b*e.txt', ['a/b/c/e.txt', 'a/b/d/e.txt']),
         ('.*[a|b].txt', []),  # regexes shouldn't be evaluated
+        ('a/b/c', []),
+        ('a/b/c*', ['a/b/c.txt', 'a/b/c/d.txt', 'a/b/c/e.txt']),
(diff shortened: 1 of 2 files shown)


AlmightyYakob (approved):
LGTM, I don't think a warning to the userbase is necessary.

Issue #1424, opened 2023-01-06
AWS S3 file download logging functionality


micheleangelofabbri:
There is a potential that International Brain Lab data in NWB format will soon be housed on dandi. A nice to have from the IBL perspective is to know how many times a particular dataset is downloaded. We are currently doing as much with our public data and would like to offer any assistance you may require setting this up on your end with the hopes that this information will one day also be shared with us. Please let me know if there is anything I can do to help.

Link for AWS S3 documentation on how to set up the logs - https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html

satra:
thanks @micheleangelofabbri - i think this is a feature that's been on the roadmap for a while and it would be great to add to the landing page, and perhaps in the listing records as well. i'll leave this to @waxlamp to plan an incorporation on the server side. but the logging should be relatively easy to setup. also given the structure of dandi the atomicity would be at the asset level, so dataset level stats would simply be an aggregate on top of that.

satra:
@waxlamp - just a ping here as this also came up in a conversation with aws that they would like to see some of the access logs. and it does look like a relatively easy setup.

waxlamp:
Got it - prioritizing this.

waxlamp:
@satra: we've come up with a preliminary plan to move on this:

We'll look into turning on the S3 bucket access logs as a first, easy step, and see how that works.

Then, we'll need to understand a bit more deeply exactly how download counts would work. This will be a combination of assessing how we actually tie S3 object downloads back to the Assets (and therefore Versions and Dandisets) they belong to and how to therefore count/report downloads in those entities, and gathering requirements from you (and other stakeholders) on what functionality you actually want to see in the archive.

satra:
thanks @waxlamp - since we use a content-based store the reverse mapping may be difficult unless the log stores where the redirect request came from.

danlamanna:
I think the trickiest part here is tracking zarr downloads, because we only track the top level archive and the requests are essentially range requests. There are 2 approaches that come to mind given that it's prohibitively expensive to store rows related to every single zarr subfile:

1. Consider a zarr download count by bandwidth cost
    We could use some form of S3 object tagging to tag objects with their zarr-id and query usage statistics on some interval. Then the download count is approximate and is computed by `bandwidth / file_size`.
2. Consider a zarr download count when it's "touched"
    We could consider any request to a zarr to be a "download", and again have an approximate download count.

[...]

PR #1570, opened 2023-04-11
Add download count tracking to asset blobs

danlamanna:
Related #1424.

This PR adds S3 server-side log processing to track the number of downloads per asset blob. The plan is to do an initial ingest and then set up a scheduled beat task to run this at some interval. For now, this increments the counters in the database and avoids exposing it in the GUI or aggregating these figures anywhere.

### DIFF
--- a/dandiapi/analytics/tasks/__init__.py
+++ b/dandiapi/analytics/tasks/__init__.py
@@ -1,0 +1,85 @@
+from collections import Counter
+from pathlib import Path
+from typing import Generator
+
+from celery.app import shared_task
+from django.conf import settings
+from django.db import transaction
+from django.db.models.aggregates import Max
+from django.db.models.expressions import F
+from s3logparse import s3logparse
+
+from dandiapi.analytics.models import ProcessedS3Log
+from dandiapi.api.models.asset import AssetBlob, EmbargoedAssetBlob
+from dandiapi.api.storage import get_boto_client, get_embargo_storage, get_storage
+
+# should be one of the DANDI_DANDISETS_*_LOG_BUCKET_NAME settings
+LogBucket = str
+
+
+def _bucket_objects_after(bucket: str, after: str | None) -> Generator[dict, None, None]:
+    assert bucket in [
+        settings.DANDI_DANDISETS_LOG_BUCKET_NAME,
+        settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME,
+    ]
+    embargoed = bucket == settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME
+    s3 = get_boto_client(get_storage() if not embargoed else get_embargo_storage())
+    kwargs = {}
+    if after:
+        kwargs['StartAfter'] = after
+
+    paginator = s3.get_paginator('list_objects_v2')
+    for page in paginator.paginate(Bucket=bucket, **kwargs):
+        yield from page.get('Contents', [])
+
+
+@shared_task(soft_time_limit=60, time_limit=80)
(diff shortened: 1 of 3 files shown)


dandiapi/analytics/tests/test_download_counts.py
```diff
@@ -0,0 +1,65 @@
...
+                '-',
+                '[06/Feb/2019:00:00:38 +0000]',
+                '-',
+                '-',
+                '-',
+                'REST.GET.OBJECT',
+                asset_blob.blob.name,
+            ]
+            + ['-'] * 10
+        ),
```
AlmightyYakob, line 40: Not sure what all of the null columns are, is it just to keep the formatting consistent with what we'd get from S3?

AlmightyYakob (commented):
Looks good overall, just a few comments

dandiapi/analytics/tests/test_download_counts.py
danlamanna: Yes, I added a comment to clarify.

dandiapi/analytics/models.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ProcessedS3Log(models.Model):
+    class Meta:
+        unique_together = ['name', 'embargoed']
+
+    name = models.CharField(
+        max_length=36,
+        validators=[
+            # https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html#server-log-keyname-format
+            RegexValidator(r'^\d{4}-(\d{2}-){5}[A-F0-9]{16}$')
```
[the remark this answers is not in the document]
danlamanna: Correct, our logs are just stored in the log bucket at the root.

mvandenburgh (approved):

AlmightyYakob (approved):

dandiapi/analytics/tasks/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+def _bucket_objects_after(bucket: str, after: str | None) -> Generator[dict, None, None]:
+    assert bucket in [
+        settings.DANDI_DANDISETS_LOG_BUCKET_NAME,
+        settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME,
+    ]
+    embargoed = bucket == settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME
+    s3 = get_boto_client(get_storage() if not embargoed else get_embargo_storage())
+    kwargs = {}
+    if after:
+        kwargs['StartAfter'] = after
```
yarikoptic, line 29: that is great -- I forgot or didn't know that it is possible to do that since we have millions of keys/files there already.  

With that in mind -- are you planing to run it first fetching all those millions of logs first?   as I have mentioned on slack - it would be millions and TBs of log files. Would such job have enough time?

danlamanna: If we do an initial migration I was planning on doing that out of band, this was just for the incremental job. I haven't yet looked at the price implications of the initial migration, if you're already downloading the historical logs for backup purposes it might make sense to have you run a one liner against them instead.

dandiapi/api/models/asset.py
```diff
@@ -56,6 +56,7 @@ class BaseAssetBlob(TimeStampedModel):
     )
     etag = models.CharField(max_length=40, validators=[RegexValidator(f'^{ETAG_REGEX}$')])
     size = models.PositiveBigIntegerField()
+    download_count = models.PositiveBigIntegerField(default=0)
```
yarikoptic, line 59: where in the API could I see/get those?

Issue #1214, opened 2022-07-25
Dandiset info endpoint large response sizes


mvandenburgh:
The dandiset `/info` endpoint currently returns every metadata validation error in a list. For dandisets with large amounts of assets, this causes abnormally large response payload sizes (example which will probably freeze your browser: https://api-staging.dandiarchive.org/api/dandisets/200026/versions/draft/info/)

The validation errors need to be paginated, either within the `/info` endpoint or as their own endpoint.

AlmightyYakob:
We could paginate them, but I don't see how that would realistically be used in practice. Like, we'd need to deal with that pagination on the client somehow. It might be enough to just truncate the errors after like 10 or something, and pass a boolean flag alongside that, indicating that there are more errors after these ones are dealt with. Sort of saying "fix these errors and then you'll see what you have to fix next".

We also could keep this endpoint as a fallback for utility purposes under `/info/errors`. Then pagination could be implemented there (if we deem it necessary). 

What do you think?

mvandenburgh:
I think truncation would be fine for the client, but yeah we'd likely still want some sort of `/info/errors` endpoint at least so that the full list of errors could be inspected via the API if needed.

waxlamp:
> > I think truncation would be fine for the client, but yeah we'd likely still want some sort of `/info/errors` endpoint at least so that the full list of errors could be inspected via the API if needed.
> 
> I think this would work nicely actually, with the list being initially truncated, and then we could provide a "view all errors" button that would show the rest in a separate page/view/etc.

Agreed, with the addition that "show the rest" would likely be in a paginated list (again, so as not to overwhelm either the server or the client).

I think we should also consider codifying errors in a more structured way than just slamming an error message into the DB. @AlmightyYakob, we did some work along these lines in the Multinet project, and I like how we did it there. Perhaps a similar design (heavily adapted to DANDI's needs) could work here.

mvandenburgh:
Coming back to this, I've noticed another issue with the way we do things that (I think) makes pagination with our current data model impossible. The serializer that returns these validation errors calls [this property method](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py#L101) to retrieve the asset validation errors for each `Asset` in that `Version`.
The two major issues here are 1) we're iterating through every `Asset` in a `Version` within an endpoint, and 2) we're not only iterating through every validation error for each of those assets, we're building up a list of all of them, which will blow up memory usage if there's a lot. 
[...]

PR #1579, opened 2023-04-24
Improve validation error display

AlmightyYakob:
This displays validation errors in a modal, instead of a small menu next to the sidebar. The version metadata validation errors remain largely unchanged. The asset validation errors have been improved to both display the path of the asset, as well as group the errors of an asset, if an asset contains multiple errors.

This is how it looks now:

image

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -81,7 +81,7 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         return not self.assets.exclude(status=Asset.Status.VALID).exists()
 
     @property
-    def asset_validation_errors(self) -> dict[str, list[dict[str, str]]]:
+    def asset_validation_errors(self) -> list[dict[str, str]]:
         # Import here to avoid dependency cycle
         from .asset import Asset
 
@@ -95,19 +95,22 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
             'message': 'asset is currently being validated, please wait.',
         }
 
+        def inject_path(asset: dict, err: dict):
+            return {**err, 'path': asset['path']}
+
         # Aggregate errors, ensuring the path of the asset is included
-        asset_error_map = {}
+        errors = []
         for asset in invalid_assets:
             # Must be pending or validating
             if asset['status'] != Asset.Status.INVALID:
-                asset_error_map[asset['path']] = [asset_validating_error]
+                errors.append(inject_path(asset, asset_validating_error))
                 continue
 
             # Must be invalid, only add entry in map if it has any errors
             if asset['validation_errors']:
-                asset_error_map[asset['path']] = asset['validation_errors']
+                errors.extend([inject_path(asset, err) for err in asset['validation_errors']])
 
-        return asset_error_map
+        return errors
 
     @staticmethod
     def datetime_to_version(time: datetime.datetime) -> str:
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -81,7 +81,7 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         return not self.assets.exclude(status=Asset.Status.VALID).exists()
 
     @property
-    def asset_validation_errors(self) -> list[str]:
+    def asset_validation_errors(self) -> dict[str, list[dict[str, str]]]:
         # Import here to avoid dependency cycle
         from .asset import Asset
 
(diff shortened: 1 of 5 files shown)


AlmightyYakob:
I'm aware of the test failures and will fix them soon.

jwodder:
CLI PR: https://github.com/dandi/dandi-cli/pull/1288

yarikoptic:
@jwodder was proactive and prepared a PR for the change in API return values, but I wonder if such change is really needed?  Ideally API (or ABI as the schema of returned structures) should not change without a significant reason since it would break prior versions of client(s) etc.  
@AlmightyYakob would you be so kind to provide a summary of the change in returned data structure and why it cannot be avoided.

AlmightyYakob:
> @AlmightyYakob would you be so kind to provide a summary of the change in returned data structure and why it cannot be avoided.

Previously for asset validation errors, the API returned a flat list of errors (dicts), with no distinction as to which errors belonged to which assets. Since each asset can have more than one error, the format of `dict[str, list[dict]]` seemed the most efficient way to convey this.

yarikoptic:
> > @AlmightyYakob would you be so kind to provide a summary of the change in returned data structure and why it cannot be avoided.
> 
> Previously for asset validation errors, the API returned a flat list of errors (dicts), with no distinction as to which errors belonged to which assets. Since each asset can have more than one error, the format of `dict[str, list[dict]]` seemed the most efficient way to convey this.

ATM it does feel like that indeed. But in the long run, it might be "too restrictive": an error might relate to multiple assets, or to some entity over assets ("sub-X/ folder is not listed among participants.tsv" in BIDS), etc. Why not to just add `path` field and keep otherwise returning a list?  
If it is desired for a viewer to group by path, it would remain a trivial operation, and could have any flexibility desired.  FWIW, it would also relate nicely to our [`ValidationResult` records](https://github.com/dandi/dandi-cli/blob/7a86c283ab926f4c53b791d4afc32f797f08e189/dandi/validate_types.py#L29) which we might want to group by path or by error id or by some other criteria.

waxlamp:
> > > @AlmightyYakob would you be so kind to provide a summary of the change in returned data structure and why it cannot be avoided.
> > 
> > 
> > Previously for asset validation errors, the API returned a flat list of errors (dicts), with no distinction as to which errors belonged to which assets. Since each asset can have more than one error, the format of `dict[str, list[dict]]` seemed the most efficient way to convey this.
> 
> ATM it does feel like that indeed. But in the long run, it might be "too restrictive": an error might relate to multiple assets, or to some entity over assets ("sub-X/ folder is not listed among participants.tsv" in BIDS), etc. Why not to just add `path` field and keep otherwise returning a list? If it is desired for a viewer to group by path, it would remain a trivial operation, and could have any flexibility desired. FWIW, it would also relate nicely to our [`ValidationResult` records](https://github.com/dandi/dandi-cli/blob/7a86c283ab926f4c53b791d4afc32f797f08e189/dandi/validate_types.py#L29) which we might want to group by path or by error id or by some other criteria.

In the interest of reducing the blast radius from the changes we're making to support better validation error UX, we'll maintain a flat list structure, adding a `path` property. This will likely allow for closing https://github.com/dandi/dandi-cli/pull/1288 while allowing the UI to add incremental value.

Your other arguments about restrictiveness point to a different issue, that it is becoming more and more difficult to mediate both the CLI's and web app's needs through a single public API. I have a bold idea here, but I need to think about it a bit more before I share my thoughts (perhaps in a design doc). Stay tuned.

yarikoptic:
> Your other arguments about restrictiveness point to a different issue, that it is becoming more and more difficult to mediate both the CLI's and web app's needs through a single public API. I have a bold idea here, but I need to think about it a bit more before I share my thoughts (perhaps in a design doc).

I would like to hear some of such issues first before we even contemplate a possible redesigns etc.  IMHO the proposed here idea of changing the Model (if we think in terms of MVC design pattern) was not really needed/desired regardless either it is for Web or CLI.

yarikoptic:
ping: conflicts

AlmightyYakob:
@mvandenburgh @marySalvi This PR is ready for review.

jwodder:
@AlmightyYakob Please summarize the changes to the API that this PR currently makes.

AlmightyYakob:
> @AlmightyYakob Please summarize the changes to the API that this PR currently makes.

The only API change is that now the objects in the `asset_validation_errors` list include an extra `path` field, which represents the path of the asset that the error belongs to.

dandiapi/api/models/version.py
```diff
@@ -90,18 +90,25 @@ def asset_validation_errors(self) -> list[str]:
...
+            'field': '',
+            'message': 'asset is currently being validated, please wait.',
+        }
+
+        def inject_path(asset: dict, err: dict):
+            return {**err, 'path': asset['path']}
+
+        # Aggregate errors, ensuring the path of the asset is included
         errors = []
         for asset in invalid_assets:
```
mvandenburgh, line 103: I think this will be really slow and/or memory intensive if a dandiset has hundreds of thousands of invalid assets. In general, I don't like how we go about storing asset validation errors in the database, but this PR is a strict improvement over the current situation imo. Perhaps related to #1214.

mvandenburgh (approved):
LGTM!

PR #1583, opened 2023-04-25
Call `clearInterval` after `setInterval` when needed

mvandenburgh:
This fixes a bug where the `setInterval` would continue firing after the user navigates from one dandiset to a new dandiset. This would result in unexpected behavior, such as the overwriting of the new dandiset with the metadata from the old one retrieved by the `setInterval` call.

One note about the changes - I changed the use of `setInterval` to `window.setInterval`. I couldn't get TypeScript to play nicely with the return value of `setInterval` for some reason; it was trying to get me to import a type from NodeJS, but since we're in a browser environment it was erroring out. They are functionally identical though.

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -224,7 +224,10 @@
 </template>
 
 <script setup lang="ts">
-import { computed, getCurrentInstance, onMounted } from 'vue';
+import {
+  computed, getCurrentInstance, onMounted, ref, onUnmounted,
+} from 'vue';
+
 import MetadataCard from '@/components/DLP/MetadataCard.vue';
 import { useDandisetStore } from '@/stores/dandiset';
 
@@ -318,8 +321,9 @@ const contactPeople = computed(
     .map((contributor) => contributor.name)),
 );
 
+const timer = ref<number| null>(null);
 onMounted(() => {
-  setInterval(async () => {
+  timer.value = window.setInterval(async () => {
     if (!currentDandiset.value || !assetSummaryBeingComputed.value) {
       return;
     }
@@ -327,6 +331,12 @@ onMounted(() => {
     const { version } = currentDandiset.value;
     await store.fetchDandiset({ identifier, version });
   }, 5000);
+});
+onUnmounted(() => {
+  if (timer.value === null) {
+    throw Error('Invalid timer value');
+  }
+  window.clearInterval(timer.value);
 });
 
 </script>
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -225,7 +225,7 @@
 
 <script setup lang="ts">
 import {
-  computed, getCurrentInstance, onMounted, ref, onUnmounted,
+  computed, getCurrentInstance, onMounted, onUnmounted,
 } from 'vue';
 
 import MetadataCard from '@/components/DLP/MetadataCard.vue';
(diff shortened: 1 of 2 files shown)


mvandenburgh:
Vue docs explicitly mention `onUnmounted` as the correct lifecycle hook for this btw https://vuejs.org/api/composition-api-lifecycle.html#onunmounted

> Use this hook to clean up manually created side effects such as timers, DOM event listeners or server connections.

AlmightyYakob (approved):

web/src/components/DLP/OverviewTab.vue
```diff
@@ -318,8 +321,9 @@ const contactPeople = computed(
     .map((contributor) => contributor.name)),
 );
 
+const timer = ref<number| null>(null);
```
waxlamp, line 324: Does this need to be a `ref`?

web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -563,24 +567,33 @@ const publishDisabledMessage: ComputedRef<string> = computed(() => {
   return '';
 });
 
-setInterval(async () => {
+const timer = ref<number | null>(null);
```
waxlamp, line 570: Same question.

waxlamp (commented):
When you changed the function from `setInterval()` to `window.setInterval()`, did the typechecking errors go away? This [Stack Overflow question](https://stackoverflow.com/questions/51376589/typescript-what-type-is-setinterval) (which you may already have seen) talks about the difference between the web and Node versions of `setInterval()` and suggests a TypeScript idiom that would work for both environments (and allow you not to include the `window.` prefix).

mvandenburgh:
> When you changed the function from `setInterval()` to `window.setInterval()`, did the typechecking errors go away? This [Stack Overflow question](https://stackoverflow.com/questions/51376589/typescript-what-type-is-setinterval) (which you may already have seen) talks about the difference between the web and Node versions of `setInterval()` and suggests a TypeScript idiom that would work for both environments (and allow you not to include the `window.` prefix).

Yeah, I saw that SO answer. I couldn't get it to work, it causes this error for me -
`Property '__promisify__' is missing in type 'Timer' but required in type 'typeof setInterval'.ts(2741)`

web/src/views/DandisetLandingView/DandisetPublish.vue
mvandenburgh: [`42594c0` (#1583)](https://github.com/dandi/dandi-archive/pull/1583/commits/42594c0416229151e3c0a4808f2a44bd3ba17f75)

web/src/components/DLP/OverviewTab.vue
```diff
@@ -328,5 +332,11 @@ onMounted(() => {
     await store.fetchDandiset({ identifier, version });
   }, 5000);
 });
+onUnmounted(() => {
+  if (timer === undefined) {
+    throw Error('Invalid timer value');
+  }
```
waxlamp, line 338: To the extent that we think it's actually an error condition if the timer is not set at this point, this is fine. If it ever happens that we want to cancel the interval timer for some other reason, then I think we would invert this conditional, remove the `throw` statement, and not worry about a non-existing timer.

waxlamp (commented):
If others are happy with this, then it LGTM.

web/src/components/DLP/OverviewTab.vue
mvandenburgh: I'll also add, the `throw` statement has the advantage of causing a sentry error if it somehow ever did get called.

PR #1584, opened 2023-04-25
Restore metadata editing

mvandenburgh:
Reverts dandi/dandi-archive#1582


mvandenburgh:
Merging this to staging only so we can test the fix in #1583 more thoroughly.

PR #1590, opened 2023-05-15
Fix GUI linting warnings

AlmightyYakob:
This fixes some linting warnings that have been hanging around for a while.


mvandenburgh (approved):
I think we should also figure out why these aren't causing CI failures. I recall seeing this before where an eslint `ERROR` results in a non-zero exit code, but an eslint `WARNING` does not.

PR #1598, opened 2023-05-22
New Dandiset Search Portal

mvandenburgh:
Adds the new search portal interface we demoed last week.


dandiapi/search/tests/test_views.py
```diff
@@ -0,0 +1,14 @@
...
+@pytest.mark.django_db
+def test_genotype_autocomplete_rest(api_client: APIClient) -> None:
+    resp = api_client.get('/api/search/genotypes/')
+    assert resp.status_code == 200
+
+
+@pytest.mark.django_db
+def test_species_autocomplete_rest(api_client: APIClient) -> None:
+    resp = api_client.get('/api/search/species/')
+    assert resp.status_code == 200
```
mvandenburgh, line 14: I wanted to add testing here for embargo permission checking, etc. but couldn't get the `REFRESH MATERIALIZED VIEW` query to work in pytest-django. So I added these tests just so we have _some_ parts of these endpoints touched in our test suite.

danlamanna (approved):

PR #1605, opened 2023-05-25
Fix S3 log file processing

danlamanna:
This reverts commit e286ef8, and is follow on work from #1570.

The comments should explain the main points, but this should remove the deadlocking issues and silence the duplicate log file exceptions that were bubbling up to Sentry (while preserving legitimate errors).


dandiapi/analytics/models.py
```diff
@@ -4,7 +4,12 @@
 
 class ProcessedS3Log(models.Model):
     class Meta:
-        unique_together = ['name', 'embargoed']
```
mvandenburgh, line 7: I didn't know `models.UniqueConstraint` is advised to be used over `unique_together`. https://docs.djangoproject.com/en/4.1/ref/models/options/#unique-together. We should remove this from the other places it's used.

mvandenburgh (approved):

PR #1607, opened 2023-05-29
Serve pure JSON from /server-info frontend endpoint

### LINKED ISSUE #1593: /server-info leads to "unavailble" page for pure JSON
pointing browser to https://dandiarchive.org/server-info leads to not just a json record but

image

where if I click on `json` in the top - leads to https://dandiarchive.org/api/info/?format=json which is pretty much 404 -- need to go to https://dandiarchive.org/server-info?format=json . Not sure how it is all setup there and might relate to #1200

waxlamp:
By default, the /api/info endpoint (to which the frontend /server-info endpoint performs a rewrite) serves a response with `format=api` which includes framing content, including CSS, etc., and a dropdown to switch formats, all of which is inappropriate for this frontend endpoint.

This PR changes the target of the rewrite to the `format=json` URL, sidestepping the problems reported in #1593.

Closes #1593.

### DIFF
--- a/web/netlify/plugins/server-info-build/index.js
+++ b/web/netlify/plugins/server-info-build/index.js
@@ -10,7 +10,7 @@ module.exports = {
     apiUrl = apiUrl.endsWith('/') ? apiUrl.slice(0, -1) : apiUrl;
     netlifyConfig.redirects.unshift({
       from: '/server-info',
-      to: `${apiUrl}/info/`,
+      to: `${apiUrl}/info/?format=json`,
       status: 200,
     });
   },


waxlamp:
@AlmightyYakob, I don't know how to test this short of deploying it to staging and just seeing if it works. From reading [the Netlify docs](https://docs.netlify.com/routing/redirects/redirect-options/#query-parameters) I believe the query parameters I added will be respected by the rewrite, but I think the only way to really know is to see if it works. LMK what you think.

yarikoptic (approved):
FWIW: going https://deploy-preview-1607--gui-staging-dandiarchive-org.netlify.app/server-info works for me nicely now

waxlamp:
> Although since we only ever want to display this as JSON, would it not be better to update the endpoint itself to only ever render as json?

I think it's just a user-agent thing. The endpoint is https://api.dandiarchive.org/api/info/; if you go there in your web browser, you'll get the "API" view by default, and then there's a (query parameter mediated) option to get a JSON response, but from a client like cURL, you will just get JSON. So I think this is the right approach: rig the proxied frontend endpoint to forcibly request JSON, while leaving the backend API endpoint as it is.

PR #1613, opened 2023-06-05
Load search results on page load

mvandenburgh:
Currently, the search page is empty when a user navigates to it, and only gets populated with results after they interact with the search controls. This updates it to instead fetch a list of initial results.


mvandenburgh:
See https://deploy-preview-1613--gui-staging-dandiarchive-org.netlify.app/search

PR #1621, opened 2023-06-14
Fix user approval workflow

### LINKED ISSUE #1619: dashboard/user/ says approved but DB doesn't have "Active"
@JordanNechvatal whom we had to "approve" after "rejecting" contacted stating that "Account Inactive".  I have checked the dashboard/user/ page -- says "Active" although DJANGO interface had indeed "Active" not marked!  
- Are there multiple "places" where activation of the user is reflected?
- Do we have a record of failure to activate that user from the dashboard interface?

edit: I did activate user manually in django interface now

waxlamp:
The user approval form handler was relying on "active" being the default state for a Django user, and therefore users who were approved after being rejected retained their "inactive" status. This PR fixes that by explicitly setting users to active whenever they are approved.

Closes #1619.

### DIFF
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -96,6 +96,8 @@ def user_approval_view(request, username: str):
 
         if user.metadata.status == UserMetadata.Status.APPROVED:
             send_approved_user_message(user, social_account)
+            user.is_active = True
+            user.save()
         elif user.metadata.status == UserMetadata.Status.REJECTED:
             send_rejected_user_message(user, social_account)
             user.is_active = False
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
LGTM. I noticed some other issues with this endpoint that may also be causing problems - they are likely out of scope for this PR so I filed an issue https://github.com/dandi/dandi-archive/issues/1623

yarikoptic:
is there any way to figure out all possibly effected users since may be there were some others who just gave up and didn't complain last the last one?

mvandenburgh:
> is there any way to figure out all possibly effected users since may be there were some others who just gave up and didn't complain last the last one?

Yes, it appears there are some other users in this situation - I'll share them privately

PR #1625, opened 2023-06-15
Use 1080p resolution browser window for e2e tests

mvandenburgh:
This makes the puppeteer browser window bigger and increases the amount of elements jest/puppeteer can "see" without needing to scroll the window.


waxlamp:
I don't think this is a bad idea but it attempts to solve the visibility problem in a partial way ("increases the amount of elements jest/puppeteer can 'see'") rather than addressing the underlying issue, which is that hidden elements in general need a way to be made visible in the same way a human tester would do. I'd prefer to fix the tests by making use of "scroll into view" functionality in the framework (if that exists).

LMK what you think.

mvandenburgh:
Closing in favor of the solution @waxlamp mentioned above.

Issue #1626, opened 2023-06-15
Feature request: play videos in archive


bendichter:
![image](https://github.com/dandi/dandi-archive/assets/844306/bc607bcb-34ef-46f2-b7c1-cb3c06607207)

For .avi and other similar files, it would be great if there were a button to stream the video in browser without having to download it. The video player should have standard video controls: play, pause, scan, etc.

bendichter:
@CodyCBakerPhD yes, we should support this in the NWB Widgets (the I agree it's not highest priority). It also might be nice to have an implementation working for stand-alone video files. I know for myself, when I see an .avi file in an archive, my first impulse is to try to watch it directly. Having this implemented outside of NWB Widgets could have a number of advantages:
* The DANDI Archive may have video files that are not associated with NWB files.
* The user may only be interested in the video and not the rest of the NWB file.
* We might be able to implement in a serverless fashion where the data is streamed to the browsers and the browser plays the video. This could save us some overhead on spinning up servers for the widgets.

waxlamp:
This is a regression: the eye icon is supposed to do exactly what you're looking for, @bendichter. I'll look into it.

yarikoptic:
I beat you for a feature request  @bendichter in #1389 .  browser does play e.g https://api.dandiarchive.org/api/assets/bceab4dc-f921-444c-9a85-d26fa5e9b1c6/download/?content_disposition=inline from https://dandiarchive.org/dandiset/000231/0.220904.1554/files?location=sub-219CR%2Fsub-219CR_ses-20190403T123013_behavior%2Bimage for me. But indeed doesn't play https://api.dandiarchive.org/api/assets/942b0806-2c8b-4289-a072-9e965884fcb6/download?content_disposition=inline you mention.  I guess those two have different encodingFormat.

- works:
```bash
❯ curl --silent https://api.dandiarchive.org/api/assets/bceab4dc-f921-444c-9a85-d26fa5e9b1c6/ | jq .encodingFormat
"video/x-matroska"
```
- not works: 
```bash
❯ curl --silent https://api.dandiarchive.org/api/assets/942b0806-2c8b-4289-a072-9e965884fcb6/ | jq .encodingFormat
"video/x-msvideo"
```

not a regression. I will propose a quick fix shortly.

waxlamp:
[...]

PR #1627, opened 2023-06-15
Assume that browser can play any "video/" content type.

yarikoptic:
x-matroska is not the only one which browsers should be able to play. E.g. video/mp4, video/webm, and video/ogg are the other ones I (and chat gpt) knows to generally be supported.  To avoid needing to list them all explicitly and thus always playing catch up game of assumptions, I would just allow browser to try on any video/ one. Worse could happen it would not work, and might ask to download I guess.

Might close (edited by @waxlamp: it doesn't) #1626 but I didn't check if it would work for those files:

    ❯ curl --silent https://api.dandiarchive.org/api/assets/942b0806-2c8b-4289-a072-9e965884fcb6/ | jq .encodingFormat
    "video/x-msvideo"

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -156,7 +156,7 @@ class AssetViewSet(DetailSerializerMixin, GenericViewSet):
                 content_type,
             )
 
-            if content_type == 'video/x-matroska':
+            if content_type.startswith('video/'):
                 return HttpResponse(
                     f"""
                     <video autoplay muted controls>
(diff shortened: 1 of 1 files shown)


waxlamp:
Just a note: I've been experimenting on this PR--will have some thoughts to share soon (regarding this PR and also #1626).

waxlamp (approved):
This is a strict improvement over the status quo, but it exposes some problems with this approach to displaying video in-browser. See the discussion thread at #1626 for more information.

yarikoptic:
@dandi/archive-maintainers please advise either this could be merged by me or should I wait for a click from you.

Note that test-e2e has failed ... seeing `    > 114 |   await expect(page).toClickXPath(vListItem('spdx:CC0-1.0'));` as the place of an error suggests that unrelated, but just to make sure , I will rebase and force push

PR #1629, opened 2023-06-19
Limit aggregate asset summaries to valid assets

danlamanna:
Fixes https://github.com/dandi/dandi-schema/issues/183.

This should remove the `encodingFormat` errors we've been getting for a while. After this I'll manually trigger a re-aggregation for draft versions.


satra:
@AlmightyYakob - the assets summary only includes metadata, not size, so it should not change either filecount or size.

PR #1632, opened 2023-06-21
add scrollIntoView for UI tests

marySalvi:
This PR adds core-js to frontend testing allowing us to use `scrollIntoView` on the puppeteer DOM.

It is good to note that this is a workaround. Ideally we would use the `scrollIntoView` function provided by puppeteer but that was not introduced until `v19.9.0`. We are currently on `v17.1.3`. There are quite a few breaking changes between these releases though.

https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v19.9.0

### DIFF
--- a/web/test/package.json
+++ b/web/test/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@faker-js/faker": "^7.5.0",
+    "core-js": "^3.31.0",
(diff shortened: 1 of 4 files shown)


waxlamp (approved):
Looks great, but we need to suppress the linting errors (which are spurious in this case afaict).

Issue #1479, opened 2023-02-07
Assets stuck in `Pending` validation state


AlmightyYakob:
The most recent example of this is #1473, but this has been occurring for a while. For some reason, assets find themselves stuck in a `Pending` state, never being updated. The only place this is updated is the [`validate_asset_metadata`](https://github.com/dandi/dandi-archive/blob/725db83403156956c7351251e3cdaf1ebff9a56f/dandiapi/api/services/metadata/__init__.py#L36-L69) function, which is run within a task, and _that_ is only ever called at the end of the [`calculate_sha256`](https://github.com/dandi/dandi-archive/blob/725db83403156956c7351251e3cdaf1ebff9a56f/dandiapi/api/tasks/__init__.py#L19-L47) task. 

I'm not sure exactly what the cause of the disconnect is, nor if it's feasible to address consistently. One thing we could do to make sure assets are eventually validated is to run a scheduled task which would run `validate_asset_metadata` on any `PENDING` assets.

satra:
another example: https://dandiarchive.org/dandiset/000243 (seems to also be a zarr asset).

satra:
+1 to scheduled tasks (several use cases of those that i can think of).

dandibot:
:rocket: Issue was released in [`v0.3.48`](https://github.com/dandi/dandi-archive/releases/tag/v0.3.48) :rocket:

satra:
this is still happening - see:  https://dandiarchive.org/dandiset/000108/

satra:
while we are at it, it would be good to know when asset started validating in the message.

AlmightyYakob:
> this is still happening - see: [dandiarchive.org/dandiset/000108](https://dandiarchive.org/dandiset/000108/)

I checked out Dandiset 108. It seems there are a total of 62 zarr files that have a `null` checksum, 53 of which have an `INVALID` status, and 9 of which (the ones we're seeing "stuck") have a `PENDING` status. Is it possible that the `finalize` API call was not made to those 9 assets? If that seems unlikely, it's possible there was a momentary blip in our infra uptime that would have caused the messages to finalize those 9 zarrs to be lost.

satra:
i wouldn't be able to tell as most of those files were done a long time back. would it be possible to remove them and redo a validation of the dandiset? or trigger a finalization.

[...]

PR #1634, opened 2023-06-22
Move asset validation to a scheduled task

danlamanna:
Addresses #1479 

This addresses a few potential race conditions and uses a scheduled task for asset validation instead of the previous flow where it could be triggered on asset creation or by completion of a sha256.

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -41,9 +41,6 @@ def validate_asset_metadata(*, asset: Asset) -> None:
         raise AssetHasBeenPublished()
 
     with transaction.atomic():
-        asset.status = Asset.Status.VALIDATING
-        asset.save()
-
         try:
             metadata = asset.published_metadata()
             validate(metadata, schema_key='PublishedAsset', json_validation=True)
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -69,7 +69,7 @@ def validate_asset_metadata(*, asset: Asset) -> None:
         asset.versions.filter(version='draft').update(modified=timezone.now())
 
 
-def version_aggregate_assets_summary(version: Version):
+def version_aggregate_assets_summary(version: Version) -> None:
     if version.version != 'draft':
         raise VersionHasBeenPublished()
 
@@ -82,8 +82,6 @@ def version_aggregate_assets_summary(version: Version):
     Version.objects.filter(id=version.id, version='draft').update(
         modified=timezone.now(), metadata=version.metadata
     )
-    version.refresh_from_db()
-    return version
 
 
 def validate_version_metadata(*, version: Version) -> None:
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -59,11 +59,17 @@ def validate_asset_metadata(*, asset: Asset) -> None:
             asset.status = Asset.Status.INVALID
             asset.validation_errors = [{'field': '', 'message': str(e)}]
 
-        # Save asset
-        asset.save()
-
-        # Update modified timestamps on all draft versions this asset belongs to
-        asset.versions.filter(version='draft').update(modified=timezone.now())
+        updated_asset = Asset.objects.filter(
+            id=asset.id, status=Asset.Status.PENDING, metadata=asset.metadata, published=False
+        ).update(
+            status=asset.status,
+            validation_errors=asset.validation_errors,
+            # include metadata in update since we're bypassing .save()
+            metadata=asset._populate_metadata(),
+        )
+        if updated_asset:
+            # Update modified timestamps on all draft versions this asset belongs to
+            asset.versions.filter(version='draft').update(modified=timezone.now())
 
 
 def version_aggregate_assets_summary(version: Version) -> None:
--- a/dandiapi/api/tasks/__init__.py
+++ b/dandiapi/api/tasks/__init__.py
@@ -1,6 +1,5 @@
 from celery import shared_task
 from celery.utils.log import get_task_logger
-from django.db import transaction
 from django.db.transaction import atomic
 
 from dandiapi.api.doi import delete_doi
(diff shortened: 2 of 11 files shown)


dandiapi/api/services/metadata/__init__.py
```diff
@@ -62,14 +59,20 @@ def validate_asset_metadata(*, asset: Asset) -> None:
...
+            id=asset.id, status=Asset.Status.PENDING, metadata=asset.metadata, published=False
+        ).update(
+            status=asset.status,
+            validation_errors=asset.validation_errors,
+            # include metadata in update since we're bypassing .save()
+            metadata=asset._populate_metadata(),
+        )
+        if updated_asset:
+            # Update modified timestamps on all draft versions this asset belongs to
+            asset.versions.filter(version='draft').update(modified=timezone.now())
```
AlmightyYakob, line 72: What's the reason for bypassing `asset.save()` here?

dandiapi/api/services/metadata/__init__.py
```diff
@@ -41,9 +41,6 @@ def validate_asset_metadata(*, asset: Asset) -> None:
         raise AssetHasBeenPublished()
 
     with transaction.atomic():
-        asset.status = Asset.Status.VALIDATING
-        asset.save()
```
AlmightyYakob, line 45: This seems to remove all use of the `VALIDATING` status then, correct? I'm actually in favor of this change, but if so, we should probably just remove that status from the model altogether.

AlmightyYakob (commented):
Really liking the change, just have a few questions

dandiapi/api/tasks/scheduled.py
```diff
@@ -28,6 +35,29 @@ def aggregate_assets_summary_task(version_id: int):
...
+            )
+        )
+        .values_list('id', flat=True)
+    )
+    validatable_assets_count = validatable_assets.count()
+    if validatable_assets_count > 0:
+        logger.info('Found %s assets to validate', validatable_assets_count)
+        for asset_id in validatable_assets.iterator():
+            # TODO: throttle?
+            validate_asset_metadata_task.delay(asset_id)
```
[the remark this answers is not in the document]
danlamanna: CloudAMQP throttles the amount of messages that can be placed on queues per second (https://www.cloudamqp.com/plans.html). Our plan is 500/s which we could easily hit here. We've had to do it on [other projects](https://github.com/imagemarkup/isic/blob/640734ed866a9fb3ca29f7f7cdd1a900d5960d9b/isic/ingest/tasks.py#L24-L28).

dandiapi/api/services/metadata/__init__.py
danlamanna: In general I think we need to move away from `.save` unless we're using `select_for_update`. Otherwise modifications could come in between the time this function starts and completes and they would be immediately overwritten, similar to #1557. Here I'm using filter criteria of `id,status,metadata,published` to give us a version of optimistic locking that's pretty robust to that failure mode. The update would no-op and the next tick of the cron would attempt it again.

dandiapi/api/tasks/__init__.py
```diff
@@ -65,8 +50,9 @@ def write_manifest_files(version_id: int) -> None:
 def validate_asset_metadata_task(asset_id: int) -> None:
     from dandiapi.api.services.metadata import validate_asset_metadata
 
-    asset: Asset = Asset.objects.get(id=asset_id)
-    validate_asset_metadata(asset=asset)
+    asset: Asset = Asset.objects.filter(id=asset_id, status=Asset.Status.PENDING).first()
+    if asset:
+        validate_asset_metadata(asset=asset)
```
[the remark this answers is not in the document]
danlamanna: Correct. This doesn't add any correctness it's strictly an optimization to avoid re-running the entire `validate` method before detecting that there's nothing to update. I'll add a comment.

dandiapi/api/services/metadata/__init__.py
mvandenburgh: > In general I think we need to move away from `.save` unless we're using `select_for_update`. 

1000% agree on this ^^

dandiapi/api/services/metadata/__init__.py
```diff
@@ -82,8 +85,6 @@ def version_aggregate_assets_summary(version: Version):
     Version.objects.filter(id=version.id, version='draft').update(
         modified=timezone.now(), metadata=version.metadata
     )
-    version.refresh_from_db()
-    return version
```
mvandenburgh, line 86: Is the idea here that there's no value in doing a refresh from db because the version could become stale again immediately after? Because the call to `refresh_to_db` _would_ update the `version` instance with the new changes from the prior update query, right?

danlamanna: > Is the idea here that there's no value in doing a refresh from db because the version could become stale again immediately after?

Yes. I think the goal long-term should be to use `refresh_from_db` only when staleness isn't a concern: testing, returning from an endpoint, etc. Plus, this function doesn't have any callers that store the return value, so I removed it entirely :P.

dandiapi/api/services/metadata/__init__.py
AlmightyYakob: > In general I think we need to move away from .save unless we're using select_for_update. Otherwise modifications could come in between the time this function starts and completes and they would be immediately overwritten, similar to https://github.com/dandi/dandi-archive/pull/1557.

Great point, yeah I agree with that. 

> Here I'm using filter criteria of id,status,metadata,published to give us a version of optimistic locking that's pretty robust to that failure mode. The update would no-op and the next tick of the cron would attempt it again.

There's something about this that doesn't sit quite right with me. This function used to accept an asset in any state, run validation on it, and save the result. Now, it will accept an asset in any state, run validation on it, but then only save the result if the status is `PENDING`. I feel like the silent failure mode is asking for trouble. Maybe we should raise an error at the beginning of the function, similar to what we do for `asset.published`? Or maybe just an if statement that logs the fact that we're skipping this asset because it's pending. What do you think?

dandiapi/api/tasks/scheduled.py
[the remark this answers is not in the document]
AlmightyYakob: Oh then yeah, I agree we should throttle here.

dandiapi/api/services/metadata/__init__.py
danlamanna: > This function used to accept an asset in any state, run validation on it, and save the result. Now, it will accept an asset in any state, run validation on it, but then only save the result if the status is PENDING.

This is just a mistake on my part and overfitting this to the cron use case. The `revalidate` script as it is would silently fail right now. I tweaked this to accept an asset in any state. I also added a log message about the asset being modified while it was validating. 

I think part of problem here is that we need to overly specify our `filter` because we're using it as a proxy to mean "only update the asset if it hasn't been modified". If we were using a revision column similar to how https://github.com/saxix/django-concurrency works, then the update query would only filter on the `id,revision` pair plus any additional requirements (in this case `published`).

Even with the changes made, it's still possible for a caller to get a log message and nothing else to indicate that the call failed. Maybe this is as simple as returning a boolean. In the case of the celery beat, we ignore it because we don't need it, but something like the revalidate script might want to retry, abort, or something else. What do you think?

dandiapi/api/services/metadata/__init__.py
danlamanna: Will you allow me to do this in a follow on PR :pray:?

AlmightyYakob: Sure, no issue from me.

waxlamp (commented):
Looks good overall.

dandiapi/api/services/metadata/__init__.py
AlmightyYakob: I like that pattern, and it looks like you've already done that here, so this looks good to me.

AlmightyYakob (approved):

PR #1635, opened 2023-06-26
Consolidate two DB queries into one

mvandenburgh:
We're evaluating two querysets here (which produces two SQL queries), when we really only need to evaluate one.

### DIFF
--- a/dandiapi/api/services/publish/__init__.py
+++ b/dandiapi/api/services/publish/__init__.py
@@ -106,8 +106,9 @@ def _publish_dandiset(dandiset_id: int) -> None:
 
     Calling `_lock_dandiset_for_publishing()` is a precondition for calling this function.
     """
-    old_version: Version = (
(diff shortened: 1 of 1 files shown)


dandiapi/api/services/publish/__init__.py
```diff
@@ -106,11 +106,15 @@ def _publish_dandiset(dandiset_id: int) -> None:
 
     Calling `_lock_dandiset_for_publishing()` is a precondition for calling this function.
     """
-    old_version: Version = (
-        Dandiset.objects.get(id=dandiset_id).versions.select_for_update().get(version='draft')
-    )
-
     with transaction.atomic():
+        old_version: Version = (
+            Version.objects.all()
```
AlmightyYakob, line 111: I don't believe the `.all()` is needed, correct?

mvandenburgh: Fixed.

AlmightyYakob (approved):

PR #1638, opened 2023-06-28
Don't store dynamic fields in Asset.metadata

AlmightyYakob:
Depends on #1634, so please evaluate changes starting from [c8ef518](https://github.com/dandi/dandi-archive/pull/1638/commits/c8ef5186ecad0800cad9530fee6f167847446f54).

According to the [schema](https://github.com/dandi/schema/tree/master), assets have certain fields that are [required](https://github.com/dandi/schema/blob/1866f4d60ec12efbd03c0fabe467e7887238d89a/releases/0.6.4/asset.json#L313-L322), all of which are derivatives of data that already exists in our database. Currently, to ensure that these required fields are present in an asset's metadata, `Asset._populate_metadata` is called any time the asset is saved, injecting the required fields into the asset's `metadata` field (which is stored directly in the database). However, this design decision has the following consequences:

1. Rather than just ensuring that the metadata is populated any time the asset is saved, it results in _needing_ to use the `.save` method when updating an asset, to ensure these fields don't go out of sync. This can cause race conditions if not done properly, and if done properly, can result in unnecessary row locking, which hampers the overall system. It also prevents the full use of `.update`, which is a safer and often times more efficient way to update records.
2. The populated fields are all derived from existing data in the `Asset` model, which means the data in these fields are de-normalized. This is both a maintenance burden and a technical liability, as if the metadata isn't kept consistent across updates, it will become out of sync with the model, resulting in bugs. When it comes to concurrent access of assets, this can be particularly problematic.

---
This PR changes the way these required fields are handled. Rather than computing and saving them to the database whenever anything in the asset is updated, they are computed and returned along with the metadata at _runtime_. This means that the underlying fields of the asset can be updated without worrying about keeping the derived fields in sync, and any de-normalized data within the asset's metadata is removed.

The exception to this is published assets. When an asset is published, all required metadata fields (along with any extra fields required for publishing) are injected into the asset's metadata, and saved. While this does still result in de-normalized data, published assets cannot be changed, so there is no risk of the data becoming out of sync.

**Note: This PR contains no API changes.**


dandiapi/api/models/asset.py
```diff
@@ -253,7 +253,8 @@ def is_different_from(
     def dandi_asset_id(asset_id: str | uuid.UUID):
         return f'dandiasset:{asset_id}'
 
-    def _populate_metadata(self):
+    @property
```
zachmullen, line 275: 👍

danlamanna (approved):

PR #1651, opened 2023-07-11
Minor tuneups to footer

yarikoptic:
to mention in-kind amazon similarly to what was done on https://www.dandiarchive.org/ . But also to do proper casing of DANDI and GitHub names

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -18,7 +18,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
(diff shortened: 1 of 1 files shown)


web/src/components/DandiFooter.vue
```diff
@@ -28,21 +28,28 @@
           >{{ version }}</a>
         </v-col>
         <v-col>
-          Funding:<br>
+          Funding / In-Kind:<br>
```
waxlamp, line 31:
```suggestion
          Funding / In-Kind Support:<br>
```

waxlamp (approved):

PR #1654, opened 2023-07-13
adjust neurosift endpoint

magland:
See https://github.com/flatironinstitute/neurosift/issues/47

Note that the old version will still redirect to the newer URL.

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -337,7 +337,7 @@ const EXTERNAL_SERVICES = [
     name: 'Neurosift',
     regex: /\.nwb$/,
     maxsize: Infinity,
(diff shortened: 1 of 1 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -337,7 +337,7 @@ const EXTERNAL_SERVICES = [
     name: 'Neurosift',
     regex: /\.nwb$/,
     maxsize: Infinity,
-    endpoint: 'https://flatironinstitute.github.io/neurosift/#/nwb?url=',
+    endpoint: 'https://flatironinstitute.github.io/neurosift?p=/nwb&url=',
```
satra, line 340: i was about to comment that i had not seen a pattern with two `?` :)

satra (approved):

PR #1655, opened 2023-07-15
enh: add dandi-asset as an additional parameter to external services

satra:
to address passing additional info this adds a new parameter for external services.

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -377,6 +377,10 @@ const isOwner = computed(() => !!(
 ));
 const itemsNotFound = computed(() => items.value && !items.value.length);
 
+function assetMetadataURI(asset_id: string) {
+  return dandiRest.assetMetadataURI(props.identifier, props.version, asset_id);
+}
+
 function getExternalServices(path: AssetPath) {
   const servicePredicate = (service: Service, _path: AssetPath) => (
     new RegExp(service.regex).test(path.path)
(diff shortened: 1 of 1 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -388,7 +392,7 @@ function getExternalServices(path: AssetPath) {
     .filter((service) => servicePredicate(service, path))
     .map((service) => ({
       name: service.name,
-      url: `${service.endpoint}${trimEnd((path.asset as AssetFile).url, '/')}`,
+      url: `${service.endpoint}${trimEnd((path.asset as AssetFile).url, '/')}&dandi-asset=${assetMetadataURI((path.asset as AssetFile).asset_id)}`,
```
yarikoptic, line 395: why not just provide as a separate kwarg to be used in definition of the url to jump to in the service? i.e.

```suggestion
      url: `${service.endpoint}${trimEnd((path.asset as AssetFile).url, '/')}`,
      asset_id: `${assetMetadataURI((path.asset as AssetFile).asset_id)}`,
```
?

yarikoptic: AH -- I just now saw that we didn't follow the original design idea here and had only prefix for the url:

```
  {
    name: 'Neurosift',
    regex: /\.nwb$/,
    maxsize: Infinity,
    endpoint: 'https://flatironinstitute.github.io/neurosift?p=/nwb&url=',
  },
```

my bad -- we ideally would need to RF it to provide all those "format" kwargs, so here it would become 

```
  {
    name: 'Neurosift',
    regex: /\.nwb$/,
    maxsize: Infinity,
    endpoint: 'https://flatironinstitute.github.io/neurosift?p=/nwb&url={url}&dandi-asset={asset_id}',
  },
```

or some other way java script allows for such embeddings...

yarikoptic: according to chatgpt JS/vue has similar to python f-strings so we can easily do that.

```
const struct = { field1: "value" };
const v = `something ${struct.field1}`;
```

PR #1661, opened 2023-07-17
Unpin dev-only dependencies + disable dependabot

mvandenburgh:
#1402 was a useful experiment to see if we could avoid having to deal with CI failures due to breaking changes in flake8, black, mypy, etc. by pinning them and managing upgrades via dependabot. But as we discussed, the large volume of dependabot PRs that it resulted in ends up consuming more time to approve and merge than it is to deal with the (very) occasional breakage. Thus, this PR effectively reverts #1402.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,12 +12,3 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-patch"]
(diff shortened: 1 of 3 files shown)


mvandenburgh:
Setting this to a release so that we can get #1652 released.

danlamanna (approved):

PR #1663, opened 2023-07-18
Set up throttling for validating asset metadata

danlamanna:
This is to prevent putting too many messages on the queue too quickly. Our current CloudAMQP plan allows us to put 500msg/s on the queue.

### DIFF
--- a/dandiapi/api/tasks/scheduled.py
+++ b/dandiapi/api/tasks/scheduled.py
@@ -4,6 +4,8 @@ Define and register any scheduled celery tasks.
 This module is imported from celery.py in a post-app-load hook.
 """
 from datetime import timedelta
+import time
+from typing import Iterable
 
 from celery import shared_task
 from celery.app.base import Celery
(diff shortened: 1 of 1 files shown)


dandiapi/api/tasks/scheduled.py
```diff
@@ -29,6 +32,19 @@
...
+    """
+    Yield items from an iterable, with a sleep between chunks.
+
+    This is useful for putting messages on a queue, where you want to avoid
+    overwhelming the queue with too many messages at once.
+    """
+    for chunk in ichunked(iterable, chunk_size):
+        for item in chunk:
+            yield item
+            time.sleep(sleep_time / chunk_size)
```
AlmightyYakob, line 45: I'm confused by this. As chunk size increases, sleep time decreases? Shouldn't it be the other way around, where as your chunks get bigger, you sleep for longer?

AlmightyYakob: Oops, nevermind, I misread this initially and thought that the sleep was happening between chunks. I can see now it's ensuring that each chunk takes at least `sleep_time` number of seconds to complete.

AlmightyYakob (approved):

Issue #1644, opened 2023-07-03
user registered and no github account linked


yarikoptic:
There is a follow up email from @satra on 30 Jun 2023 on that matter.  The email mentioned in the registration approval page is `@gmail.com` whenever I see some logged interactions from `.edu` account.  May be user changed email interim registration/approval process and that somehow broke correspondence?  github api for that user shows .edu email and 

```
  "created_at": "2017-03-06T22:31:33Z",
  "updated_at": "2023-07-01T11:28:05Z"

```

In DB I see two accounts for the name -- one with `@gmail.com` (PENDING, that is where we lack association with github account) and another one with `.edu` - APPROVED and joined 2 minutes after the prior one.

@xmival00 , do you remember any detail from you registering on dandi
- did you switch your email in your github account shortly after after registering at DANDI archive website?
- did you request registration twice?

waxlamp:
Thanks for the background info, @xmival00.

@yarikoptic, I will remove the pending gmail.com application from the system as suggested.

PR #1670, opened 2023-07-31
Add new Django setting to restrict allauth endpoints that are exposed

mvandenburgh:
Currently, we use allauth's regular "account" provider for user management in dev, and its github "socialaccount" provider for prod. However, we expose endpoints for both of these auth flows in prod and dev. This means that someone can sign up using the regular "account" provider in production, bypassing github entirely. In practice, this hasn't happened before because we don't advertise these URLs publicy; however if you know the URL, a user could navigate to https://api.dandiarchive.org/accounts/signup/ directly and sign up.

This PR adds a new setting `ENABLE_GITHUB_OAUTH`. If enabled, only the github socialaccount provider endpoints are exposed. If disabled, only the "account" provider endpoints are exposed.

### DIFF
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -39,11 +39,14 @@ class DandiMixin(ConfigMixin):
             'dandiapi.zarr.apps.ZarrConfig',
         ] + configuration.INSTALLED_APPS
 
-        # Install additional apps
-        configuration.INSTALLED_APPS += [
-            'guardian',
-            'allauth.socialaccount.providers.github',
-        ]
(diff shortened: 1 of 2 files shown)


AlmightyYakob (approved):
Nice work! Tested this locally and can confirm I don't see any github oauth URLs, while manually changing the setting shows only the github oauth login flow.

I haven't tested the full Github login flow with this change, but to my knowledge it should remain unchanged.

danlamanna:
Related #1644

danlamanna (approved):

PR #1674, opened 2023-08-28
Design doc for "undelete" feature

mvandenburgh:
This PR contains a design document proposing how we plan to go about implementing recoverability of deleted files, or "undelete".


doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
...
+
+    # Expire objects with delete markers after 1 day
+    noncurrent_version_expiration {
+      noncurrent_days = 1
+    }
+
+    # Also delete any delete markers associated with the expired object
+    expiration {
+      expired_object_delete_marker = true
+    }
```
yarikoptic, line 40: oh, that is neat! I didn't know that is possible.
Just to make sure: Would it be in effect also for old/already existing DeleteMarkers or somehow works only on the new ones which it detects being placed after adding this lifecycle rule?

doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
...
+
+Our ultimate goal is to prevent data loss from application programming errors. With protection such as an undelete capability, we will be safer in implementing application features that involve intentional deletion of data. Any bugs we introduce while doing so are far less likely to destroy data that was not supposed to be deleted.
+
+The original GitHub issue around this feature request can be found at [https://github.com/dandi/dandi-archive/issues/524](https://github.com/dandi/dandi-archive/issues/524). Although the issue asks for a Deep Glacier storage tier, the design in this document solves the underlying problem differently (and in a more robust way). Below we address the possible usage of a Deep Glacier tiered bucket as a solution to the orthogonal problem of data ******backup******, which addresses a different problem than the undelete capability described in this document.
```
yarikoptic, line 9: why here 6 `*`s for bold, instead of e.g. just 2?  below you also used 4 `*`s -- any particular reason for extra emphasis?

Note, although we I think have not introduced such convention to design docs but we use it in other projects already: It would be nice to start using "one sentence a line" paragraph formatting.  In this case it would have helped to break down a possible patch to the paragraph (now 3 long sentences) for the formatting into a much more manageable unit - a sentence.

doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
...
+        {
+            "Sid": "PreventDeletionOfDeleteMarkers",
+            "Effect": "Deny",
+            "Principal": "*",
+            "Action": "s3:DeleteObjectVersion",
+            "Resource": "arn:aws:s3:::dandi-s3-experiment-bucket/*"
+        }
+    ]
+}
+```
```
yarikoptic, line 61: just to make sure: wouldn't "terraform" rule above be "above" this S3 policy and thus in effect tripping terraform on this rule as well when it would try to expire the delete marker?

doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
...
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "PreventDeletionOfDeleteMarkers",
+            "Effect": "Deny",
+            "Principal": "*",
+            "Action": "s3:DeleteObjectVersion",
+            "Resource": "arn:aws:s3:::dandi-s3-experiment-bucket/*"
```
yarikoptic, line 57: I guess we could try all this goodness first on the staging one, right?

doc/design/s3-undelete.md
mvandenburgh: It would apply to existing delete markers as well. The docs mention this briefly, and I also tested it myself to confirm.

https://aws.amazon.com/s3/faqs/
> S3 Lifecycle policies apply to both existing and new S3 objects, helping you optimize storage and maximize cost savings for all current data and any new data placed in S3 without time-consuming manual data review and migration.

doc/design/s3-undelete.md
mvandenburgh: Ah that's weird, I used a markdown editor and then exported it to a `.md` file. I guess I didn't look closely enough at the markdown it generated, I'll normalize these.

doc/design/s3-undelete.md
mvandenburgh: Yes, we will deploy this to the staging bucket before production.

doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Distinction from Data Backup
+
+It’s important to note that the “undelete” implementation proposed above does not cover backup of the data in the bucket. While backup is out of scope for this design document, nothing proposed here *prevents* backup from being implemented, and features such as [S3 Replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication.html) may be useful for that.
+
+A possible backup implementation would simply allocate a new bucket in a separate S3 region set to use the Deep Glacier storage tier (to control costs), then use S3 Replication as mentioned above to maintain an object-for-object copy of the production bucket as a pure backup. This backup solution would ***not*** defend against application-level bugs deleting data, but would instead protect the production bucket against larger-scale threats such as destruction of Amazon data centers, etc.
```
yarikoptic, line 69:
```suggestion
A possible backup implementation would simply allocate a new bucket in a separate S3 region set to use the Deep Glacier storage tier (to control costs), then use S3 Replication as mentioned above to maintain an object-for-object copy of the production bucket as a pure backup. This backup solution would **not** defend against application-level bugs deleting data, but would instead protect the production bucket against larger-scale threats such as destruction of Amazon data centers, etc.
```

for consistency.  Apparently (https://www.markdownguide.org/basic-syntax/) `***` are used for emphasis wi***th***in a word

doc/design/s3-undelete.md
```diff
@@ -0,0 +1,69 @@
+# **S3 Undelete**
```
yarikoptic, line 1:
```suggestion
# **S3 Trailing Delete**
```

Per my comment on slack, I think that "S3 Undelete" is a bit of mis-leading title for this development since it is not really about "undelete" feature, which is already similarly possible in current S3 versioned setup without GC, but rather about establishing policy for delayed pruning/GC for keys on the level of S3.
Since "trailing delete" was used below (although I found no good hits googling for it) -- I am suggesting it here.  But may be even it could be something like

```suggestion
# **S3 Delayed GC Setup**
```

which would
- better describe what is presented in the document
- accent on relation to GC

doc/design/s3-undelete.md
waxlamp: I'm not sure I understand your question. Are you asking whether this policy would prevent the lifecycle rule from deleting delete markers?

doc/design/s3-undelete.md
waxlamp: According to https://www.markdownguide.org/basic-syntax/#bold-and-italic, triple stars are used to apply both ***bold and italic*** at the same time. I assume that's what Mike meant to do here.

doc/design/s3-undelete.md
yarikoptic: yes

doc/design/s3-undelete.md
yarikoptic: oh, right! getting blind. fine as is then. thanks @waxlamp and sorry for the noise

doc/design/s3-undelete.md
mvandenburgh: I agree that "trailing delete" is a better description for this. While the primary motivator for this feature is doing garbage collection, I think the title of "S3 Delayed GC Setup" is too low-level, especially considering that I don't provide a precise definition for "Garbage Collection" anywhere in this doc. I'll go ahead and change it to "S3 Trailing Delete", but let me know if you have any further thoughts on this.

doc/design/s3-trailing-delete.md
```diff
@@ -0,0 +1,69 @@
...
+```
+  # Terraform-encoded lifecycle rule.
+  # Based on https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html#lifecycle-config-conceptual-ex7
+  rule {
+    id = "ExpireOldDeleteMarkers"
+    filter {}
+
+    # Expire objects with delete markers after 1 day
+    noncurrent_version_expiration {
+      noncurrent_days = 1
```
yarikoptic, line 34: resurrecting now outdated comment with 

```suggestion
    # Expire objects with delete markers after 30 days
    noncurrent_version_expiration {
      noncurrent_days = 30
```

to align with the **Requirements** section.

doc/design/s3-trailing-delete.md
```diff
@@ -0,0 +1,69 @@
+# **S3 Trailing Delete**
```
yarikoptic, line 1: I guess IDE did another misservice -- added not needed `**` around every section header. 
for consistency with other doc/design files and typical use of markdown - let's get rid of them.  I will push a commit

yarikoptic (approved):

PR #1679, opened 2023-09-11
Fix possible race condition in deployment actions

mvandenburgh:
We don't enforce any `concurrency` rule on the github actions that deploy our Django app to Heroku. This opens us up to potential issues, such as a scenario where an older deploy job ends up running after a newer job, causing the older version of the app to "overwrite" the newer one on Heroku.

### DIFF
--- a/.github/workflows/backend-production-deploy.yml
+++ b/.github/workflows/backend-production-deploy.yml
@@ -4,6 +4,13 @@ on:
   release:
     types: [released]
 
+concurrency:
+  # If this workflow is already running, cancel it to avoid a scenario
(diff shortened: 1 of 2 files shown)


AlmightyYakob (approved):

waxlamp:
Neat. @mvandenburgh how did you know to fix this problem?

mvandenburgh:
> Neat. @mvandenburgh how did you know to fix this problem?

We ran into this issue on SMART, where we also use GitHub Actions for deployment. On that project deploys usually take about 30 mins, so this issue was somewhat common since merging multiple PRs in the same 30 minute window would potentially cause it to happen.

PR #1683, opened 2023-09-21
Add "Designs" table to the documentation

### LINKED ISSUE #1675: Annotate design documents with the "state"
We are accumulating design documents under https://github.com/dandi/dandi-archive/tree/master/doc/design which is great.  But as time goes it becomes unclear on what is the destiny of each design document -- was it just proposed, being implemented superseded, etc.  I propose to add some header structure, could be just a markdown list of keys like

- State: proposed (other values: work-in-progress, partially-implemented,implemented, abandoned, superseded; could be mixed, e.g., "implemented,superseded")
- Released-In: `<version-of-dandi-archive>`
- Superseded-By: `<another-design-doc>`
- Superseded-In: `<version-of-dandi-archive>`
- Implemented-In: `<list of PRs>`

WDYT?  I know that it would be a bit more of work, but would bring clarity since ATM not clear in what state we are in for each of those designed docs.

yarikoptic:
which would provide a summary over the designs

TODOs if decide to proceed forward

- [ ] do fill out the table with current state of things. Closes #1675 (somewhat of an alternative to have it inside the docs, could be moved there later)
- [ ] (later or now) improve the script to just add new rows whenever new files appear

### DIFF
--- a/doc/design/index.md
+++ b/doc/design/index.md
@@ -1,0 +1,20 @@
+# Design documents
+
+
+| File                         | Originating Date | Version        | Implementation State | Date             | Version        | Superseded by |
+| ---------------------------- | ---------------- | -------------- | -------------------- | ---------------- | -------------- | ------------- |
+| blob-uuid-1.md               | 2021-03-09       | v0.0.1~15^2~29 |                      |                  |                |               |
+| doi-generation-1.md          | 2021-04-05       | v0.1.0~22^2~14 |                      |                  |                |               |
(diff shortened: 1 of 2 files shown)


waxlamp:
Unfortunately I think this complicates our design documentation process without really adding significant value.

yarikoptic:
@waxlamp - did you read the https://github.com/dandi/dandi-archive/issues/1675 ? with this PR I tried to address it - ATM we have no clue what is implemented, what replaced etc.  So even if there would be some extra burden, I think it is warranted, if not significant.

PR #1690, opened 2023-09-27
use substitutions for external service endpoints

magland:
In light of this issue on neurosift
https://github.com/flatironinstitute/neurosift/issues/112

it is necessary to use the DANDI API asset download URL rather than the s3 bucket URL for the endpoint of the neurosift external service.

For example, here's an s3 bucket URL
https://dandiarchive.s3.amazonaws.com/blobs/d64/0c3/d640c378-7025-4c3b-a334-169ec3df4f3f

which corresponds to the following DANDI API download URL:
https://api.dandiarchive.org/api/assets/34a21ebe-c756-4da4-a30b-f1a838a6430b/download/

As described in the above quoted issue: In the cases of embargoed dandisets, it is necessary for neurosift to receive the API download URL so that it can use proper authentication to retrieve the signed download URL. The direct s3 bucket URL will never work.

This PR enables this, while still utilizing the s3 URL for the other services.

A detail about the modified code: While the s3 bucket url is directly available on the path.asset object, the api download url is not. So I needed to construct the url, and that required knowing whether or not it's the staging site.

In short, this PR is needed to allow Neurosift to be able to view assets from embargoed dandisets.

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -309,35 +309,35 @@ const EXTERNAL_SERVICES = [
     name: 'Bioimagesuite/Viewer',
     regex: /\.nii(\.gz)?$/,
     maxsize: 1e9,
-    endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=$s3_url$',
+    endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=$asset_s3_download_url$',
   },
 
   {
     name: 'MetaCell/NWBExplorer',
     regex: /\.nwb$/,
     maxsize: 1e9,
-    endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=$s3_url$',
+    endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=$asset_s3_download_url$',
   },
 
   {
     name: 'VTK/ITK Viewer',
     regex: /\.ome\.zarr$/,
     maxsize: Infinity,
-    endpoint: 'https://kitware.github.io/itk-vtk-viewer/app/?gradientOpacity=0.3&image=$s3_url$',
+    endpoint: 'https://kitware.github.io/itk-vtk-viewer/app/?gradientOpacity=0.3&image=$asset_s3_download_url$',
   },
 
   {
     name: 'OME Zarr validator',
     regex: /\.ome\.zarr$/,
     maxsize: Infinity,
-    endpoint: 'https://ome.github.io/ome-ngff-validator/?source=$s3_url$',
+    endpoint: 'https://ome.github.io/ome-ngff-validator/?source=$asset_s3_download_url$',
   },
 
   {
     name: 'Neurosift',
     regex: /\.nwb$/,
     maxsize: Infinity,
-    endpoint: 'https://flatironinstitute.github.io/neurosift?p=/nwb&url=$api_download_url$',
+    endpoint: 'https://flatironinstitute.github.io/neurosift?p=/nwb&url=$asset_api_download_url$',
   },
 ];
 type Service = typeof EXTERNAL_SERVICES[0];
@@ -402,8 +402,9 @@ function getExternalServices(path: AssetPath) {
     ? 'https://api-staging.dandiarchive.org'
     : 'https://api.dandiarchive.org';
   const substitutions = path.asset ? {
-    $s3_url$: trimEnd(path.asset.url, '/'),
-    $api_download_url$: `${baseApiUrl}/api/assets/${path.asset.asset_id}/download/`,
+    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
+    $asset_api_download_url$: `${baseApiUrl}/api/assets/${path.asset.asset_id}/download/`,
+    $asset_id$: path.asset.asset_id,
   } : undefined;
 
   return EXTERNAL_SERVICES
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -389,8 +389,6 @@ function replaceStringsInEndpoint(
   return result;
 }
 
-const isStaging = process.env.VUE_APP_SENTRY_ENVIRONMENT === 'staging';
-
 function getExternalServices(path: AssetPath) {
   const servicePredicate = (service: Service, _path: AssetPath) => (
     new RegExp(service.regex).test(path.path)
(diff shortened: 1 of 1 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -377,18 +377,40 @@ const isOwner = computed(() => !!(
...
           && _path.asset !== null
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = isStaging
+    ? 'https://api-staging.dandiarchive.org'
+    : 'https://api.dandiarchive.org';
+  const substitutions = path.asset ? {
+    $s3_url$: trimEnd(path.asset.url, '/'),
+    $api_download_url$: `${baseApiUrl}/api/assets/${path.asset.asset_id}/download/`,
```
yarikoptic, line 406: Those are both URLs to the asset download. So I think we could/should make them a little more uniform, what about

```suggestion
    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
    $asset_api_download_url$: `${baseApiUrl}/api/assets/${path.asset.asset_id}/download/`,
```
so in the future we might expand the listing here without fear of collision?

yarikoptic: BTW , in https://github.com/dandi/dandi-archive/pull/1655 (does this PR supersedes it?) we seems also wanted to expose `asset_id`.  I think it would be safe to include it here as well but I guess we could add more as demand comes.

yarikoptic (commented):
That's GREAT! Thank you @magland ! I left some comments/recommendations.

web/src/views/FileBrowserView/FileBrowser.vue
magland: Thanks @yarikoptic !

I have adjusted the strings as you suggested and added a $asset_id$ (which of course is unused at this point)

Yes, I think this is accomplishing the idea laid out in #1655

magland:
BTW, do I have the correct way of testing whether it's the staging site?

```typescript
const isStaging = process.env.VUE_APP_SENTRY_ENVIRONMENT === 'staging';
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -377,18 +377,41 @@ const isOwner = computed(() => !!(
...
 function getExternalServices(path: AssetPath) {
   const servicePredicate = (service: Service, _path: AssetPath) => (
     new RegExp(service.regex).test(path.path)
           && _path.asset !== null
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = isStaging
+    ? 'https://api-staging.dandiarchive.org'
+    : 'https://api.dandiarchive.org';
```
yarikoptic, line 403: Let's avoid harcoding any url:

```suggestion

function getExternalServices(path: AssetPath) {
  const servicePredicate = (service: Service, _path: AssetPath) => (
    new RegExp(service.regex).test(path.path)
          && _path.asset !== null
          && _path.aggregate_size <= service.maxsize
  );

  const baseApiUrl = process.env.VUE_APP_DANDI_API_ROOT;
```
and adjust below according to suggestion?

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -377,18 +377,41 @@ const isOwner = computed(() => !!(
...
+  const substitutions = path.asset ? {
+    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
+    $asset_api_download_url$: `${baseApiUrl}/api/assets/${path.asset.asset_id}/download/`,
```
yarikoptic, line 406:
```suggestion
    $asset_api_download_url$: `${baseApiUrl}/assets/${path.asset.asset_id}/download/`,
```

magland:
Thanks @yarikoptic I have made your suggested changes.

magland:
@yarikoptic @waxlamp I'm not sure if this is feasible, but I was hoping this PR could be merged and released sometime this week, because I am giving a presentation on the 10th, and it would be really nice to have this in place. Happy to make adjustments if needed.

waxlamp:
@magland I'm going to try something today that may solve the problem. Stand by.

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -377,18 +377,38 @@ const isOwner = computed(() => !!(
...
     new RegExp(service.regex).test(path.path)
           && _path.asset !== null
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = process.env.VUE_APP_DANDI_API_ROOT;
+
+  const substitutions = path.asset ? {
+    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
+    $asset_api_download_url$: `${baseApiUrl}assets/${path.asset.asset_id}/download/`,
```
yarikoptic, line 403: minor suggestion:

```suggestion
  const baseApiUrl = trimEnd(process.env.VUE_APP_DANDI_API_ROOT, '/');

  const substitutions = path.asset ? {
    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
    $asset_api_download_url$: `${baseApiUrl}/assets/${path.asset.asset_id}/download/`,
```
to avoid assumption of having trailing `/`?

waxlamp:
@magland, please check out #1692. I believe that PR (inspired by this one) is a simpler way to solve the problem you are facing.

I would prefer to defer this "API" creation in the external services integrations to a real design round so that we don't hastily get stuck with suboptimal choices. That is to say, I'd like to solve your problem in as direct as a mode as possible while addressing the ideas of #1655 with more deliberation (which I invite you to collaborate on if you're interested).

Let me know what you think.

magland:
Thanks @waxlamp . I think that will do the trick!

From the perspective of neurosift, I believe it would be better to get the same type of URL from dandi-archive regardless of whether or not it is embargoed. But if #1692 could be released soon, then that would satisfy my short-term need.

yarikoptic:
> I would prefer to defer this "API" creation in the external services integrations to a real design round so that we don't hastily get stuck with suboptimal choices.

note that in this PR we are changing *internal* supplementary code so we would be able to as easily fix it up to some more *optimal* solution happen we need it -- it is not an external registry ATM etc.  But I think it is moving in the right direction and would allow for better basis for any possible design round if it ever to happen.

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -377,18 +377,38 @@ const isOwner = computed(() => !!(
...
     new RegExp(service.regex).test(path.path)
           && _path.asset !== null
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = trimEnd(process.env.VUE_APP_DANDI_API_ROOT, '/');
+
+  const substitutions = path.asset ? {
+    $asset_s3_download_url$: trimEnd(path.asset.url, '/'),
+    $asset_api_download_url$: `${baseApiUrl}/assets/${path.asset.asset_id}/download/`,
```
satra, line 403: @waxlamp - actually one of the things i wanted in the other PR is the dandiset specific asset download url, not the general one. that would allow in essence for someone to fetch relevant dandiset info or for externallinks such as videos, the video url. also since this is clicked not from a global search but a view within the dandiset, i think it's appropriate to have the other url.

waxlamp: > the other PR

Do you mean #1655 or #1692?

> dandiset specific asset download url, not the general one

When you say "dandiset specific", do you mean the longer URL that is "nested" under a `/dandisets/{dandiset_id}/...` URL?

satra: correct - sorry was in meetings. yes, i would like that asset url to change to the one that is nested under /dandisets/

PR #1692, opened 2023-10-02
Use correct asset download URL for external services

waxlamp:
When an external service is called from the file browser, the asset URL used is always the S3 URL, even in the case of an owner browsing an embargoed Dandiset. In the latter case, the S3 URL, which is _not_ presigned, does not allow general access to the asset, causing the external service to fail when it tries to retrieve the data.

This PR modifies the generation of that asset URL to emit the S3 URL for open Dandiset assets, and the internal DANDI API asset URL (which redirects to the correct presigned S3 URL) for embargoed Dandiset assets. (Note that the target external service must still authenticate to the DANDI server in order for this redirection to succeed; that in turn requires a design that supports using DANDI API keys, etc.)

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -370,6 +370,7 @@ const updating = ref(false);
 // Computed
 const owners = computed(() => store.owners?.map((u) => u.username) || null);
 const currentDandiset = computed(() => store.dandiset);
+const embargoed = computed(() => currentDandiset.value && currentDandiset.value.dandiset.embargo_status === 'EMBARGOED');
 const splitLocation = computed(() => location.value.split('/'));
 const isAdmin = computed(() => dandiRest.user?.admin || false);
 const isOwner = computed(() => !!(
@@ -384,11 +385,16 @@ function getExternalServices(path: AssetPath) {
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = process.env.VUE_APP_DANDI_API_ROOT;
+  const assetURL = () => (embargoed.value
+    ? `${baseApiUrl}assets/${path.asset?.asset_id}/download/`
+    : trimEnd((path.asset as AssetFile).url, '/'));
+
   return EXTERNAL_SERVICES
     .filter((service) => servicePredicate(service, path))
     .map((service) => ({
       name: service.name,
-      url: `${service.endpoint}${trimEnd((path.asset as AssetFile).url, '/')}`,
+      url: `${service.endpoint}${assetURL()}`,
     }));
 }
 
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -370,7 +370,7 @@ const updating = ref(false);
 // Computed
 const owners = computed(() => store.owners?.map((u) => u.username) || null);
 const currentDandiset = computed(() => store.dandiset);
-const embargoed = computed(() => currentDandiset.value && currentDandiset.value.dandiset.embargo_status === 'EMBARGOED');
+const embargoed = computed(() => currentDandiset.value?.dandiset.embargo_status === 'EMBARGOED');
 const splitLocation = computed(() => location.value.split('/'));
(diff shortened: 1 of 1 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -370,6 +370,7 @@ const updating = ref(false);
 // Computed
 const owners = computed(() => store.owners?.map((u) => u.username) || null);
 const currentDandiset = computed(() => store.dandiset);
+const embargoed = computed(() => currentDandiset.value && currentDandiset.value.dandiset.embargo_status === 'EMBARGOED');
```
AlmightyYakob, line 373: This might just be a bit more succinct, as if `currentDandiset` is not set the comparison will be `undefined === 'EMBARGOED'` which will still evaluate to `false`.

```suggestion
const embargoed = computed(() => currentDandiset.value?.dandiset.embargo_status === 'EMBARGOED');
```

web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -384,11 +385,16 @@ function getExternalServices(path: AssetPath) {
           && _path.aggregate_size <= service.maxsize
   );
 
+  const baseApiUrl = process.env.VUE_APP_DANDI_API_ROOT;
+  const assetURL = () => (embargoed.value
+    ? `${baseApiUrl}assets/${path.asset?.asset_id}/download/`
+    : trimEnd((path.asset as AssetFile).url, '/'));
```
AlmightyYakob, line 391: Is there some particular reason to remove trailing slashes here? Shouldn't we just use whatever URL comes back from the API?

AlmightyYakob (changes requested):
A couple questions/suggestions but otherwise looks good

web/src/views/FileBrowserView/FileBrowser.vue
waxlamp: I think I was maybe satisfying the typechecker? As you point out, the type of the succinct version will be `boolean | undefined`, whereas the longer one is just `boolean`.

web/src/views/FileBrowserView/FileBrowser.vue
waxlamp: I am not sure. I was just copying over the `trimEnd()` expression into this more expansive function. So if we want to change that, I'd say let's do it in a followup where we can properly analyze the effect of doing so. (For example, will it cause one of these dreaded and annoying redirects to the version with the trailing slash? etc.)

web/src/views/FileBrowserView/FileBrowser.vue
AlmightyYakob: The succinct version would still return just a `boolean`, as the check of `undefined === 'EMBARGOED'` is `false` (and at least for me locally, results in just a boolean computed ref). In general I'd like to avoid the pattern of `x && x.y` if possible, as I think it's excessively verbose and complicates things.

web/src/views/FileBrowserView/FileBrowser.vue
AlmightyYakob: My bad, I didn't notice that it existed before the change.

yarikoptic:
what is "correct" in the context of a URL?

web/src/views/FileBrowserView/FileBrowser.vue
waxlamp: Good analysis, @AlmightyYakob.

waxlamp:
> what is "correct" in the context of a URL?

I updated the description to answer this question in better detail (I should have done that upon marking this PR as ready for review--sorry!).

AlmightyYakob (approved):

magland:
I noticed this PR has been approved. Do you have an estimate for when it will be deployed? Thanks

waxlamp:
> I noticed this PR has been approved. Do you have an estimate for when it will be deployed? Thanks

@magland, I just merged this PR and it is deploying as we speak! Please check in a little while to see if the new feature is working as intended, and follow up here if there are any lingering problems.

magland:
@waxlamp it's working, thanks so much!

PR #1696, opened 2023-10-03
Add a note to registration page that no account is necessary to access public data

yarikoptic:
I think we start to get an increasing number of registration requests from people who are not intending to upload data to DANDI.  This explicit message might help to avoid those registration requests.

### DIFF
--- a/dandiapi/api/templates/api/account/questionnaire_form.html
+++ b/dandiapi/api/templates/api/account/questionnaire_form.html
@@ -2,6 +2,7 @@
 
 {% block body %}
   <h5>Welcome to DANDI! Please take a moment to fill out this form.</h5>
+  <h6><b>Note:</b> no account is necessary to access public data!</h6>
   <form method="POST" class="col s12">
     {% csrf_token %}
     {{ form }}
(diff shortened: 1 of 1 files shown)


dandiapi/api/templates/api/account/questionnaire_form.html
```diff
@@ -2,6 +2,7 @@
 
 {% block body %}
   <h5>Welcome to DANDI! Please take a moment to fill out this form.</h5>
+  <h6><b>Note:</b> no account is necessary to access public data!</h6>
```
satra, line 5:
```suggestion
  <h6><b>Note:</b> No account is necessary to access public data!</h6>
```

satra:
Yes, this is a trivial thing that should also be released asap.

waxlamp (approved):
This doesn't hurt, but it may not really be enough: by the time they see this message, they've already created their account (albeit, without yet getting it approved, but they've already jumped through the hoop of possibly creating a GitHub account).

Ultimately, the problem here is with an accidental dark pattern we've introduced, where the "log in with github" button doesn't deter people at that stage, and then they are basically on a waterslide that takes them into the kiddie pool of account creation. We should discuss further what we might do about this.

In the meantime, maybe this PR will help.

PR #1697, opened 2023-10-04
Align Postgres tables/indexes/constraints with Django

AlmightyYakob:
While working on a separate PR, I encountered an error when trying to update a constraint on the `ZarrArchive` model, which led me to realize that there was a mismatch between what Django _thinks_ the name of these constraints are, and what they actually are in Postgres. It appears the renaming of these constraints slipped through the cracks in #1256, specifically the `*-consistent-checksum-status` and `*-unique-name` constraints.

This PR aligns those names with reality in Postgres, as well as renames the tables and all other indexes/constraints, to be consistent with what we expect.


dandiapi/zarr/migrations/0003_zarr_rename_tables_indexes_constraints.py
```diff
@@ -0,0 +1,79 @@
...
+            '"zarr_zarrarchive_dandiset_id_68510762";'
+        ),
+        migrations.RunSQL(
+            'ALTER INDEX "api_embargoedzarrarchive_dandiset_id_b61f0a08" RENAME TO '
+            '"zarr_embargoedzarrarchive_dandiset_id_b61f0a08";'
+        ),
+        # dandiset_id foreign key constraint
+        migrations.RunSQL(
+            'ALTER TABLE zarr_zarrarchive RENAME CONSTRAINT '
+            '"api_zarrarchive_dandiset_id_68510762_fk" TO '
```
danlamanna, line 52: Will this work on a fresh checkout of dandi? I don't think it will always make the constraints with these names.

AlmightyYakob: I believe it will. I tore down my database and re-ran migrations several times to test this.

danlamanna: Wow, I guess the "unique" string is just a truncated md5 of the name: https://github.com/django/django/blob/stable/4.0.x/django/db/backends/base/schema.py#L1229.

danlamanna (approved):

dandiapi/zarr/migrations/0003_zarr_rename_tables_indexes_constraints.py
AlmightyYakob: That makes sense, although now I'm wondering if this is correct. Using that `names_digest` method, I can confirm the following behavior:
```
In [3]: names_digest('api_zarrarchive', 'dandiset_id', length=8)
Out[3]: '68510762'
```

However, since this PR is changing the underlying table name, that should also change:
```
In [4]: names_digest('zarr_zarrarchive', 'dandiset_id', length=8)
Out[4]: '2fdb93be'
```

Does that seem right? Should I change the hash to avoid some weird behavior in the future?

AlmightyYakob: Afaik no (at least, not in a concise way). I don't think Django has any idea of these lower level constraints/indexes within its state. I was originally trying to do this with the Django API directly, but wasn't able to.

However, I did some experimenting. First I updated the hashes for the zarr and embargoed zarr indexes / fk constraints with the "correct" values, and Django applied that without issue. Then I set `db_index=False` on `ZarrArchive.dandiset`, created a migration (to delete the existing index), reverted the change, and created a new migration (to add the index back). After running those migrations, the new index it came up with had the same hash as mentioned above.

As a final test, I did the same experiment again, but in migration `0003`, I inserted random characters into the index/constraint names, instead of the correct ones. Django applied the migrations successfully, and the end result was the same.

So I guess it doesn't really matter what they're named? It might just be more of a consistency thing. Either way, I'd prefer for them to be named "correctly", if that's alright with you.

danlamanna (approved):

PR #1698, opened 2023-10-04
Add UPLOADED state to Zarr models

AlmightyYakob:
**This is a breaking API change.**

This adds an `UPLOADED` state to the Zarr models, to delineate between when a user calls "finalize", and when the zarr begins ingestion.

The use of the Zarr status is now as follows:

1. User creates a Zarr, it defaults to `PENDING`
2. User uploads data to a zarr
3. User is done uploading data and calls the `finalize` endpoint, placing the Zarr in an `UPLOADED` state and kicking off the ingestion task
4. The ingestion task is begun on that zarr, placing it in an `INGESTING` state
5. Ingestion is finished, placing it in a `COMPLETE` state
6. If the User wants to upload more data, they request files from the Zarr file creation endpoint, placing the Zarr into a `PENDING` state
7. Go-to bullet 2
---

@jwodder The changes regarding the CLI are as follows:

1. When calling `finalize` on a zarr, the returned status of the zarr will now be `UPLOADED`, instead of `PENDING`
2. The zarr file create and delete endpoints will now return the `ZarrArchive.INGEST_ERROR_MSG` if the zarr in question is in either the `UPLOADED` _or_ `INGESTING` state
3. The `ZarrArchive.INGEST_ERROR_MSG` has been changed from `"Zarr archive already ingested"` to `"Zarr archive is currently ingesting or has already ingested"`


waxlamp (commented):
Could you (in the PR description) sketch out the intended progression through the different statuses after this change?

dandiapi/zarr/views/__init__.py
```diff
@@ -170,6 +170,11 @@ def finalize(self, request, zarr_id):
             if zarr_archive.status != ZarrArchiveStatus.PENDING:
                 return Response(ZarrArchive.INGEST_ERROR_MSG, status=status.HTTP_400_BAD_REQUEST)
 
+            # Indicate that the user is done uploading to this zarr
+            # TODO: If locking above is removed, use an update query
```
danlamanna, line 174: I think we can remove this line and the above TODO around locking.

danlamanna (approved):

yarikoptic:
We returned dandi-cli into green state again.  Do you think you @AlmightyYakob or @danlamanna can workout a needed patch for it for this API change?  would be appreciated. If not -- that is ok, just let me know.

AlmightyYakob:
> We returned dandi-cli into green state again. Do you think you @AlmightyYakob or @danlamanna can workout a needed patch for it for this API change? would be appreciated. If not -- that is ok, just let me know.

It seems no change is needed to the CLI, since it awaits the `COMPLETE` state to ensure there is a valid checksum.

PR #1689, opened 2023-09-27
Enable path-only changes to metadata to trigger asset change

waxlamp:
The logic for PUT on asset metadata includes a check to see if the asset is "different" enough to warrant an actual change. This works in part by stripping the supplied metadata of all computed fields, of which `path` is one, and then comparing what remains. This had the effect of barring any "move" type operations (by changing just the `path` value of a metadata and performing a PUT), as changes to the path did not "count" towards being different enough. The workaround was to _also_ change a _non_-computed metadata field; this made the asset different enough to trigger a change, which included the changed path.

The fix is to teach the "is different" function about path-only changes.

Thanks to @AlmightyYakob for his help in figuring this out.

TODO:
- [x] add a test (@AlmightyYakob)

Closes #1682.

Issue #1682, opened 2023-09-15
PUT "silently" (returns 200) doesn't rename and doesn't update metadata (path remains old)


yarikoptic:
User reported that `dandi move` doesn't work : https://github.com/dandi/dandi-cli/issues/1325 .
I have checked `test_move` tests there -- work fine.
I thought it is zarr specific, have tested on sample files including nwb -- indeed `dandi move` doesn't do good.
I have tried on a sample file via copy/pasting metadata in swagger -- seems worked.
Trying in pdb session of dandi-cli -- nope. Here trying to rename from `bar-renamed` to `bar-renamed2` -- see that result contains `bar-renamed` although provided record `bar-renamed2`:

```
(Pdb) p path
'/dandisets/000029/versions/draft/assets/e4032a47-141d-4c4c-84c1-2be3ac5aeb33/'

*(Pdb) p json.dumps(kwargs['json'])
[...]

Issue #1686, opened 2023-09-27
make asset PUT api consistent


satra:
this endpoint returns `blob` and `zarr` as dictionary keys (which is the same as asset `INFO` endpoint) but expects `zarr_id` and `blob_id` as parameters. the `_id` suffix should be removed from the parameters.

PR #1700, opened 2023-10-05
Specify asset path at top level in API

### LINKED ISSUE #1687: Specify asset path at top level in asset update endpoint
An asset's path is a top level field, and is stored as a column on the asset table directly. Yet, in the asset update REST endpoint, path is only allowed to be specified inside the `metadata` field. This breaks the API model, and is in essence the underlying cause of #1682.

AlmightyYakob:
Closes #1687, follow-up to #1689

**This is a breaking API change.**

Previously, for asset creation/update, the path of the asset was specified within the `metadata` field. This is confusing, as an asset's path is stored directly on the model itself (_not_ in  the metadata), and was the leading cause of #1682. This PR changes that behavior, by requiring an asset's path to be specified directly at the top level, alongside `metadata`, `blob_id`, and `zarr_id`.

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -182,6 +182,7 @@ class AssetViewSet(DetailSerializerMixin, GenericViewSet):
 
 
 class AssetRequestSerializer(serializers.Serializer):
+    path = serializers.CharField()
     metadata = serializers.JSONField()
     blob_id = serializers.UUIDField(required=False)
     zarr_id = serializers.UUIDField(required=False)
@@ -213,12 +214,14 @@ class AssetRequestSerializer(serializers.Serializer):
             raise serializers.ValidationError(
                 {'blob_id': 'Exactly one of blob_id or zarr_id must be specified.'}
             )
-        if 'path' not in data['metadata'] or not data['metadata']['path']:
-            raise serializers.ValidationError({'metadata': 'No path specified in metadata.'})
+        if 'path' in data['metadata']:
+            raise serializers.ValidationError(
+                {'metadata': 'Path must only be specified at top level.'}
+            )
 
         # Validate the asset path. If this fails, it will raise a django ValidationError, which
         # will be caught further up the stack and be converted to a DRF ValidationError
-        validate_asset_path(data['metadata']['path'])
+        validate_asset_path(data['path'])
 
         data['metadata'].setdefault('schemaVersion', settings.DANDI_SCHEMA_VERSION)
         return data
@@ -312,6 +315,7 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
             version=version,
             asset_blob=serializer.get_blob(),
             zarr_archive=serializer.get_zarr_archive(),
+            path=serializer.validated_data['path'],
             metadata=serializer.validated_data['metadata'],
         )
 
@@ -353,6 +357,7 @@ class NestedAssetViewSet(NestedViewSetMixin, AssetViewSet, ReadOnlyModelViewSet)
                 version=version,
                 new_asset_blob=serializer.get_blob(),
                 new_zarr_archive=serializer.get_zarr_archive(),
+                new_path=serializer.validated_data['path'],
                 new_metadata=serializer.validated_data['metadata'],
             )
 
--- a/dandiapi/api/management/commands/create_dev_dandiset.py
+++ b/dandiapi/api/management/commands/create_dev_dandiset.py
@@ -42,10 +42,14 @@ def create_dev_dandiset(name: str, email: str):
         'schemaVersion': settings.DANDI_SCHEMA_VERSION,
         'encodingFormat': 'text/plain',
         'schemaKey': 'Asset',
-        'path': 'foo/bar.txt',
     }
+    asset_path = 'foo/bar.txt'
     asset = add_asset_to_version(
-        user=owner, version=draft_version, asset_blob=asset_blob, metadata=asset_metadata
+        user=owner,
+        version=draft_version,
+        asset_blob=asset_blob,
+        metadata=asset_metadata,
+        path=asset_path,
     )
 
     calculate_sha256(blob_id=asset_blob.blob_id)
--- a/dandiapi/api/services/asset/__init__.py
+++ b/dandiapi/api/services/asset/__init__.py
@@ -44,6 +44,7 @@ def change_asset(
     version: Version,
     new_asset_blob: AssetBlob | EmbargoedAssetBlob | None = None,
     new_zarr_archive: ZarrArchive | None = None,
+    new_path: str,
     new_metadata: dict,
 ) -> tuple[Asset, bool]:
     """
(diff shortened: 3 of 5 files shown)


dandiapi/api/views/asset.py
```diff
@@ -213,12 +214,14 @@ def validate(self, data):
             raise serializers.ValidationError(
                 {'blob_id': 'Exactly one of blob_id or zarr_id must be specified.'}
             )
-        if 'path' not in data['metadata'] or not data['metadata']['path']:
-            raise serializers.ValidationError({'metadata': 'No path specified in metadata.'})
+        if 'path' in data['metadata']:
+            raise serializers.ValidationError(
+                {'metadata': 'Path must only be specified at top level.'}
+            )
```
waxlamp, line 220: Is this the part that makes this a breaking change?

Related question: it would be reasonable under the general usage of PUT endpoints to copy the metadata from a GET call and then paste it into the PUT call; if someone does that, will the presence of the `path` entry cause this error?

waxlamp (commented):
Looks good overall, but I had one question about the user-facing interface changes.

satra:
it may really be helpful to document the lifecycle of an asset somewhere (can point to code pieces as needed). this should include what is injected asset metadata and when.

also if we are making API changes related to the asset, can we also address the blob/zarr vs blob_id/zarr_id difference between API returns and API POST/PUT endpoints and make them consistent.

AlmightyYakob:
> it may really be helpful to document the lifecycle of an asset somewhere (can point to code pieces as needed). this should include what is injected asset metadata and when.

Good point, I agree.

> also if we are making API changes related to the asset, can we also address the blob/zarr vs blob_id/zarr_id difference between API returns and API POST/PUT endpoints and make them consistent.

You're referring to #1686, correct? I think we can do that, but I'd prefer to do it in a separate PR, to keep things clean. We should delay the merging/deployment of these two PRs until the CLI has the corresponding changes ready to go.

dandiapi/api/views/asset.py
danlamanna: Assuming we'll disallow path: does this check belong in the serializer or the service layer? Since we would want to prevent it regardless of where  it came from, I think it belongs in the service layer. And if that's the case, is `path` still a necessary element in `ASSET_COMPUTED_FIELDS`?

AlmightyYakob: I think we'd still need to keep `path` as an entry in `ASSET_COMPUTED_FIELDS`, since it's still returned as a part of the computed metadata (and as such we'd want to prevent storing it in the metadata directly).

The validation error I added here is more a signal externally to prevent confusion / false bug reporting (e.g. "I changed path in the metadata, why does nothing change?"). If we assume we're keeping `path` as a field in `ASSET_COMPUTED_FIELDS`, I don't see what the benefit of performing a check in the service layer is, since the field will be stripped out anyway.

What do you think? It's possible I'm missing something.

danlamanna: FYI Jake and I discussed this and settled on generalizing the "path can't be specified as metadata" requirement to all reserved (computed) fields. As it stands, the same bug as #1682 exists for the other fields.

satra: but path is not a computed field, the others are. so why is path treated as the same?

AlmightyYakob: In this context computed means "computed from other properties of the asset or asset metadata". In the case of `path`, since it's stored directly on an asset, it's "computed" by just injecting that field into the metadata.

satra: thanks @AlmightyYakob - i still find it weird that the API requires this as a separate element as opposed to being able to pull it out internally from the metadata as it used to. locally the CLI validates the asset with these fields inside, and hence seems easiest from a programmer standpoint using the API to have one point of entry. since this is a breaking API change, just want to use the opportunity to both reflect that dandi objects are the metadata and ensuring a good pattern for outside clients, which are growing beyond us, interacting with the API.

AlmightyYakob: > i still find it weird that the API requires this as a separate element as opposed to being able to pull it out internally from the metadata as it used to. locally the CLI validates the asset with these fields inside, and hence seems easiest from a programmer standpoint using the API to have one point of entry. since this is a breaking API change, just want to use the opportunity to both reflect that dandi objects are the metadata and ensuring a good pattern for outside clients, which are growing beyond us, interacting with the API.

The intent of this change is to make it more clear the distinction between modifiable and un-modifiable metadata. The issue with the current state of things is as follows:

You have an asset and it's metadata, and you want to change a field in that asset's metadata. Depending on what field you want to modify, it will actually result in no change when the request is made to the API, since some fields are computed/derived/etc, and some are actually modifiable. Things like `contentUrl` and `digest` are examples of fields that you _cannot_ modify directly. They are computed/derived from the asset, or its associated asset blob / zarr.

The `path` field is a unique example, where it's a field that _is_ computed (since that data is stored on the asset itself and is then injected into the metadata), yet prior to this proposed change, was _also_ specified in the metadata, and then stripped out by the server. This is counter-intuitive behavior, and is the reason I originally intended on barring `path` from being specified in the metadata with this PR. Dan and I also discussed extending this to all computed fields, so that there is a clear one-way data flow for any computed fields (top level -> metadata).

Leaving things the way they are now, how would a consumer of the API know which fields can be modified and which can't? The only way they _could_ know (aside from reading the source code), is through trial and error, which seems undesirable from a user-facing point of view. If we were to update the API to make it explicit the fields which can and can't be modified, I think that would result in a better experience for consumers of the API.

I do see your point, in the sense that the format of the API and that of the Dandi schema would be diverging somewhat. However, regardless of the uniformity between the two formats, the dichotomy between metadata that can be updated, and computed metadata, still exists.

satra: @AlmightyYakob - sorry for the delay. 

> The path field is a unique example, where it's a field that is computed (since that data is stored on the asset itself and is then injected into the metadata), yet prior to this proposed change, was also specified in the metadata, and then stripped out by the server. 

regarding `path`, i don't know why it was being stripped out or inserted by the server. `path` is indeed special as it is the unique key to an asset. i.e. the same path cannot point to multiple blobs.

> The only way they could know (aside from reading the source code), is through trial and error, which seems undesirable from a user-facing point of view.

indeed this is the case right now, and is really because of the API service not trusting clients to determine what those fields are. however, the exception is still for `path` - which is definitely a client prerogative to set, not for the server to decide. the API should have nothing to say about paths except regarding "our" policies on clashes with either `POST`/`PUT`. 

can we better understand why path should be computed?

> the dichotomy between metadata that can be updated, and computed metadata, still exists.

let's figure out a way to add this info to dandischema itself.

yarikoptic: I also do not grasp why it is exactly computed, besides may be that it is first extracted from metadata for the asset record and then placed back into it (why?).

yarikoptic:
I feel that this a "good to have" and not really "has to have" change, and that is why not sure if it is worth the API breakage at this point.  Should we may be better

- establish "dandi-api-2.0" (or whatever we want to call it, e.g. `dandi-api-0.2` to signify that it is not yet stable despite all the years; or `dandi-api-1.0` - as the API we would approach as "finally a stable one to aim") project where we start collecting the issues/PRs which would require API breakage and bundle them all.
- do get back to documenting "lifecycle" and formalize then how lifecycle would need to change in that "new" one

WDYT?

AlmightyYakob:
I'm closing this PR, and will be making a new PR containing the generalized solution discussed [here](https://github.com/dandi/dandi-archive/pull/1700#discussion_r1357161216). That will still constitute a breaking API change, but will be less disruptive, as it'll be contained in one PR.

PR #1703, opened 2023-10-11
Upgrade `dandischema` dependency

### LINKED ISSUE #1702: bad doi inserted into dandiset metadata
there are several dandisets, including a couple of published ones that have a doi field in the metadata: `"doi": "10.80507/dandi.123456/0.123456.1234"`

this should never happen. something is inserting this or not removing the "doi" field for a draft dandiset.

mvandenburgh:
Fixes #1702. 

#1702 was fixed in `dandischema` a while back with https://github.com/dandi/dandi-schema/pull/167 and a new version of `dandischema` was released, but we never bumped the version of `dandischema` in dandi-archive. 

Note this does not solve the underlying fake doi injection issue, but it will fix DOI creation on dandiarchive.org. As part of this, I will query for all published dandisets that do not have a DOI due to this bug and re-run the DOI generation on them.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
     include_package_data=True,
     install_requires=[
         'celery',
-        'dandischema~=0.8.2',
+        'dandischema~=0.8.4',
         'django~=4.1.0',
         'django-admin-display',
         # TODO: unpin allauth when https://github.com/girder/django-composed-configuration/pull/189


danlamanna (approved):

AlmightyYakob (approved):

waxlamp (approved):

mvandenburgh:
@yarikoptic @satra any ideas why the dandi-cli tests are failing here? It seems to be unrelated to this PR, because I re-ran them on `master` and they fail there as well with the same error messages. 
- [Here's the initial run on `master`](https://github.com/dandi/dandi-archive/actions/runs/6423622802/attempts/1) that succeeded on 10/5
- [Here's a re-run that I did today on `master`](https://github.com/dandi/dandi-archive/actions/runs/6423622802/attempts/2), which fails despite being run on the same commit

satra:
@yarikoptic - just a ping to see why dandi-cli is failing with this PR.

yarikoptic:
seems a test started to fail in dandi-cli about a week or so ago - the same as here, so changes here do not break any other test and thus should be ok. Filed https://github.com/dandi/dandi-cli/issues/1332 , will look into it,

PR #1713, opened 2023-10-16
Remove unused `modified` field on `Upload` model

mvandenburgh:
I noticed this while investigating what new fields (if any) we need to add for garbage collection of `Uploads`. We don't use this anywhere at this time, and since we've already decided to rework how we do `modified` (not using `TimeStampedModel`), I think it makes sense to just remove this.


dandiapi/api/migrations/0044_remove_embargoedupload_modified_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+        migrations.AlterField(
+            model_name='embargoedupload',
+            name='created',
+            field=django_extensions.db.fields.CreationDateTimeField(auto_now_add=True),
+        ),
+        migrations.AlterField(
+            model_name='upload',
+            name='created',
+            field=django_extensions.db.fields.CreationDateTimeField(auto_now_add=True),
+        ),
```
mvandenburgh, line 30: Note: this migration got created because `TimeStampedModel` defines `created` internally a bit differently using `gettext_lazy` https://github.com/django-extensions/django-extensions/blob/main/django_extensions/db/models.py#L17

AlmightyYakob (approved):
LGTM

mvandenburgh:
#1714 will fix the failing type check here.

PR #1717, opened 2023-10-17
Remove pinning on django-storages[boto3]

mvandenburgh:
The bug that mandated pinning this dependency (#1691) has been fixed in the latest release of `django-storages` ([here's the changelog entry](https://github.com/jschneier/django-storages/blob/master/CHANGELOG.rst#s3)), so we can remove the pinning.

Additionally, the latest minor release of `django-storages` also deprecates `storages.backends.s3boto3.S3Boto3Storage` in favor of `storages.backends.s3.S3Storage`, as well as changes the install extra from `[boto3]` to `[s3]`, ([changelog entry](https://github.com/jschneier/django-storages/blob/master/CHANGELOG.rst#s3-2)), so I went ahead and made those changes here as well.

### DIFF
--- a/dandiapi/api/copy.py
+++ b/dandiapi/api/copy.py
@@ -8,10 +8,10 @@ from django.conf import settings
 from dandiapi.api.storage import get_boto_client
 
 try:
-    from storages.backends.s3boto3 import S3Boto3Storage
+    from storages.backends.s3 import S3Storage
 except ImportError:
     # This should only be used for type interrogation, never instantiation
-    S3Boto3Storage = type('FakeS3Boto3Storage', (), {})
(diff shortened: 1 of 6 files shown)


danlamanna (approved):

mvandenburgh:
> While everything here seems to make sense, have you been able to properly test it? Since locally it wouldn't actually change anything, correct?

Right. I'll make sure to do a test upload in staging once this is deployed there.

mvandenburgh:
(just rebasing so the dandi-cli tests pass)

mvandenburgh:
I uploaded some assets to staging using the CLI after deploying this and everything appears to work fine, so this should be fine to deploy to production the next time we cut a release.

PR #1720, opened 2023-10-18
Remove unnecessary atomic decorators from tasks

mvandenburgh:
I don't think the `atomic` decorator actually does anything for these task functions, but please correct me if I'm missing something @danlamanna @AlmightyYakob. 
Any complex operations that do need to be atomic are already wrapped in an `atomic` block inside the service layer functions that the tasks are calling.

### DIFF
--- a/dandiapi/api/services/publish/__init__.py
+++ b/dandiapi/api/services/publish/__init__.py
@@ -103,12 +103,12 @@ def _publish_dandiset(dandiset_id: int) -> None:
 
     Calling `_lock_dandiset_for_publishing()` is a precondition for calling this function.
     """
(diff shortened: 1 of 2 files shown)


danlamanna (approved):

mvandenburgh:
Looks like the `atomic` decorator was actually doing something on the publish task :sweat_smile:  - this select_for_update call is erroneously made outside of the `atomic` block, but it went unnoticed until now since the celery task had the `atomic` decorator. https://github.com/dandi/dandi-archive/pull/1720/commits/a5b5786f28881d70b94a161055de1f9a0f733813

danlamanna (approved):

PR #1721, opened 2023-10-18
Clean up handling of missing `DOI_API_PREFIX` setting

mvandenburgh:
This setting should never be unset in a real deployment of DANDI, so silently replacing it with a default if it's missing is not ideal. 
Instead, this PR changes it so that things fail loudly and raise and error if this setting is not supplied in an env var. In dev/testing configurations it is still hardcoded to the placeholder value, but this is done at the Django setting layer instead of directly in the application code.


mvandenburgh:
Closing this. I'll re-evaluate whether this refactor makes sense as part of redesigning the DOI creation process.

PR #1730, opened 2023-10-26
Optimize dandiset listing endpoint

AlmightyYakob:
This reduces the load time of the dandiset listing endpoint by optimizing the `_get_dandiset_to_version_map` function. This means the search endpoint should also experience a slight performance improvement, since it also uses that function.

The optimization is accomplished with the following changes:
1. Removes an unnecessary query being made in the `get_root_paths_many` function
2. Shifts the version stats aggregation to be done in postgres, making it more efficient
3. Optimize the query for retrieving the latest version. Similar to the other optimized query, this leverages postgres aggregation to efficiently return the latest published versions for all relevant dandisets
4. Use two smaller queries instead of one union-ed query when retrieving drafts and published versions, to take advantage of the last published version optimization.

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -1,7 +1,7 @@
 from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Count, F, OuterRef, Subquery, Sum
+from django.db.models import Count, Max, OuterRef, Subquery, Sum
 from django.db.models.functions import Coalesce
 from django.db.models.query_utils import Q
 from django.http import Http404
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -163,64 +163,55 @@ def get_object(self):
...
-            version: Version
+        # This query sums the size and file count for root paths, and groups by the version_id,
+        # ensuring that the queryset is unique w.r.t the version_id. For some reason, the
+        # `order_by` clause is necessary to ensure this grouping
+        version_stats = {
+            entry['version_id']: entry
+            for entry in get_root_paths_many(versions=relevant_versions)
+            .values('version_id')
+            .annotate(total_size=Sum('aggregate_size'), num_assets=Sum('aggregate_files'))
+            .order_by('version_id')
```
danlamanna, line 181:
```suggestion
            .order_by()
```

I think you're getting bitten by the model `ordering`. WIthout the `order_by` it groups by `version_id, path` and orders by `path`. With `order_by('version_id')` it groups and orders by `version_id`. Using just `order_by()` ignores the model ordering and omits an `order by` clause and just groups by `version_id` which is what I think you want.

danlamanna (approved):
This LGTM, it seems like a performance improvement and a code simplification which is nice.

dandiapi/api/views/dandiset.py
AlmightyYakob: Good catch, I didn't realize you could omit the parameters and still obtain the group clause.

danlamanna (approved):

PR #1733, opened 2023-11-03
Add upload/asset blob garbage collection design doc

mvandenburgh:


doc/design/garbage-collection-uploads-asset-blobs.md
```diff
@@ -0,0 +1,26 @@
+# Upload and Asset Blob Garbage Collection
+
+## Background
+
+Now that the [design for S3 trailing delete](https://github.com/dandi/dandi-archive/blob/master/doc/design/s3-trailing-delete.md) is deployed to staging, we are ready to implement garbage collection. [This older design document](https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-1.md#uploads) is still relevant, and summarizes the various types of garbage collection we want to implement. This document will present a design for garbage collection of uploads and asset blobs, i.e. garbage that accumulates due to improper uploads done by users. A design for garbage collection of orphaned “Assets” (i.e. files that have been properly uploaded, have metadata, etc. but are no associated with any dandisets) is more complex and is left for a future design document.
```
satra, line 5: i think it would be good to add this type of asset below under the next section. in the dandi model this would be the biggest source of garbage as we mint a new asset each time an asset or blob is modified. 

on this note, can we run a quick analysis of the numbers in each category? i.e. does skipping this for a future iteration really address the issue.

doc/design/garbage-collection-uploads-asset-blobs.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Why do we need garbage collection?
+
+When a user creates an asset, they send a request to the API and the API returns a series of presigned URLs for the user to perform a multipart upload to. Then, an `Upload` database row is created to track the status of the upload. When the user is done uploading their data to the presigned URLs, they must “finalize” the upload by sending a request to the API to create an `AssetBlob` out of that `Upload`. Finally, they must make one more request to actually associate this new `AssetBlob` with an `Asset`.
+
+### Orphaned Uploads
```
satra, line 11: do the following sections address zarr trees? i see multipart mentioned, which doesn't apply to zarr at this point.

doc/design/garbage-collection-uploads-asset-blobs.md
mvandenburgh: > i think it would be good to add this type of asset below under the next section. in the dandi model this would be the biggest source of garbage as we mint a new asset each time an asset or blob is modified.

The intention of this document is to only cover asset blob/upload garbage collection, since these are more straightforward to implement than `Asset` garbage collection (with most of the complexity there surrounding (1) are there cases where we want to retain old assets in the DB, and (2) the fundamental incompatibility the current `Asset` data model has with garbage collection due to the `Asset.previous` pointer). 

> on this note, can we run a quick analysis of the numbers in each category? i.e. does skipping this for a future iteration really address the issue.

Yes, I can get these numbers and add them here.

yarikoptic, line 5: inspired me for https://github.com/dandi/handbook/pull/99 -- chime in there. Meanwhile -- no need (and undesired) for absolute urls AFAIK to link nearby docs, I think this should work but frankly I didn't check:

```suggestion
Now that the [design for S3 trailing delete](./s3-trailing-delete.md) is deployed to staging, we are ready to implement garbage collection. [This older design document](./garbage-collection-1.md#uploads) is still relevant, and summarizes the various types of garbage collection we want to implement. This document will present a design for garbage collection of uploads and asset blobs, i.e. garbage that accumulates due to improper uploads done by users. A design for garbage collection of orphaned “Assets” (i.e. files that have been properly uploaded, have metadata, etc. but are no associated with any dandisets) is more complex and is left for a future design document.
```

yarikoptic:
```suggestion
Now that the [design for S3 trailing delete](https://github.com/dandi/dandi-archive/blob/master/doc/design/s3-trailing-delete.md) is deployed to staging, we are ready to implement garbage collection. [This older design document](https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-1.md#uploads) is still relevant, and summarizes the various types of garbage collection we want to implement. This document will present a design for garbage collection of uploads and asset blobs, i.e. garbage that accumulates due to improper uploads done by users. A design for garbage collection of orphaned “Assets” (i.e. files that have been properly uploaded, have metadata, etc. but are no longer associated with any version of a dandiset) is more complex and is left for a future design document.
```

doc/design/garbage-collection-uploads-asset-blobs.md
yarikoptic:
```suggestion
### Orphaned Asset Uploads
```
to clarify?

doc/design/garbage-collection-uploads-asset-blobs.md
```diff
@@ -0,0 +1,26 @@
...
+
+If the user cancels a multipart upload partway through, or completes the multipart upload to S3 but does not “finalize” the upload, then the upload becomes “orphaned”, i.e. the associated `Upload` record and S3 object remain in the database/bucket indefinitely.
+
+### Orphaned AssetBlobs
+
+In this case, assume that the user properly completes the multipart upload flow and “finalizes” the `Upload` record such that it is now an `AssetBlob`, but they do not send a request to associate the new blob with an `Asset`. That `AssetBlob` record and associated S3 object will remain in the database/bucket indefinitely.
```
yarikoptic, line 17:
```suggestion
In this case, assume that the user properly completes the multipart upload flow and “finalizes” the `Upload` record such that it is now an `AssetBlob`, but they do not send a request to associate the new blob with an `Asset`. That `AssetBlob` record and associated S3 object will remain in the database/bucket indefinitely.
Also future `Asset` garbage collection implementation can be designed independently of logic to also clean up AssetBlobs, and we might end up with such orphaned `AssetBlobs` too to be picked up by this implementation. 
```

feel free to ignore, but might hint on it being a good idea for future designs.

doc/design/garbage-collection-uploads-asset-blobs.md
mvandenburgh: I added the suggested title change, and also added a sentence explicitly mentioning that zarr files are not covered in this document https://github.com/dandi/dandi-archive/pull/1733/commits/11942e10dee6a8369f9f732ff535a59f2fdbfdf6.

doc/design/garbage-collection-uploads-asset-blobs.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Implementation
+
+We will introduce a new celery-beat task that runs daily. This task will
+
+- Query for and delete any uploads that are older than the multipart upload presigned URL expiration time (this is currently 7 days).
+- Query for and delete any AssetBlobs that are (1) not associated with any Assets, and (2) older than 7 days.
```
yarikoptic, line 24: might be worth adding a note on implementation that first it should delete from S3 and then delete from DB, since otherwise we might end up in inconsistent state.

Might still warrant `fsck` style (I have noted other cases, which could be [searched up](https://github.com/search?q=repo%3Adandi%2Fdandi-archive+fsck&type=issues)) operation on the archive -- check if all keys on S3 in `blobs/` have AssetBlob entries in DB.  ATM our django admin says we have `168737 asset blobs` but listing our bucket I get `168940`.  Small difference but it is there -- something for `fsck` to check I guess.

doc/design/garbage-collection-uploads-asset-blobs.md
mvandenburgh: Nice, I didn't know you can use relative paths like that in github.

mvandenburgh: I added these numbers as of today to the design doc (https://github.com/dandi/dandi-archive/pull/1733/commits/5f1301a478a471e818a916cee29bda639002b5bb).

doc/design/garbage-collection-uploads-asset-blobs.md
mvandenburgh: I tweaked the wording a bit to fit in with that section - https://github.com/dandi/dandi-archive/pull/1733/commits/eef69762a8fc12ddc4b21a06a00cd7667b75dcaf

mvandenburgh: I tweaked the wording a bit to fit in with that section - https://github.com/dandi/dandi-archive/pull/1733/commits/23c55ad6727985c98a8f7d18bfb05548c44f55c7

satra:
i'll add my thoughts here on linked assets. the primary reason for that was to maintain complete history so we could have an audit trail of changes. since a `PUT` generates a new asset id. the secondary reason was provenance. there was supposed to be a a provenance record added on a `PUT`. i don't know if that was done. 

the audit trail is still key to have given that we have not exposed an audit trail to our users and we really should. and that can follow the same principles as the trailing delete. we only keep audit trails up to some specified number of days. 

i suspect there are lots of assets that don't belong to any dandiset and are also not part of any chain to an asset that belongs to a dandiset. can we break up assets into those categories: asset chains that don't belong to any dandiset? and asset chains that belong to some? @mvandenburgh - would it be possible to get these numbers (total numbers of assets in these two categories)?

yarikoptic:
@mvandenburgh ping on above

waxlamp (approved):
This design doc looks great. I would like to ratify and merge this, while handling the slightly complex issue tree surrounding `Asset` garbage collection ("chains", what to do about audit/provenance, and how to reliably delete dead assets while maintaining audit/provenance information) separately.

@satra, @yarikoptic if you are ok with that, please leave an approval as well.

satra (approved):
i'm approving this and wanted to comment why i felt adding those numbers were important. this started with garbage collection, but the amount of garbage collected by this is less than 1%.

waxlamp:
> i'm approving this and wanted to comment why i felt adding those numbers were important. this started with garbage collection, but the amount of garbage collected by this is less than 1%.

I sympathize with your reaction here, but the tradeoff is that this was the easiest phase of GC to design and develop, which of course correlates with its lower level of "cleaning power". I guess Mike and I are following the snowball strategy here, knocking out the simplest phases in order, to try to achieve faster incremental results. We'll only see the full power after everything is in.

Thanks for the approval.

PR #1719, opened 2023-10-17
Update backend to Python 3.12

mvandenburgh:

PR #1736, opened 2023-11-06
Require Python 3.11

brianhelba:


mvandenburgh:
Thanks @brianhelba. I made an attempt to go to 3.12 here #1719, but it seems like the removal of `distutils` (among other things) makes that not as easy.

mvandenburgh (approved):
FYI I don't think the `--keep-runtime-typing` to pyupgrade does anything on 3.11. The pyupgrade docs aren't really clear on the exact functionality of it, but Ruff also implements that rule and has an [explanation in its docs for it](https://docs.astral.sh/ruff/settings/#pyupgrade-keep-runtime-typing).

brianhelba:
> FYI I don't think the `--keep-runtime-typing` to pyupgrade does anything on 3.11. The pyupgrade docs aren't really clear on the exact functionality of it, but Ruff also implements that rule and has an [explanation in its docs for it](https://docs.astral.sh/ruff/settings/#pyupgrade-keep-runtime-typing).

Yeah, good point. Thanks for the tip.

PR #1737, opened 2023-11-06
Optimize dandiset owner PUT endpoint

mvandenburgh:
Now that django-allauth uses a regular JSONField for `SocialAccount.extra_data`, we can optimize the dandisets owner PUT endpoint to be more efficient and do less SQL queries. See release notes for allauth here https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0580-2023-10-26

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -362,7 +362,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
     # TODO: move these into a viewset
     @action(methods=['GET', 'PUT'], detail=True)
     def users(self, request, dandiset__pk):
-        dandiset = self.get_object()
+        dandiset: Dandiset = self.get_object()
         if request.method == 'PUT':
             # Verify that the user is currently an owner
             response = get_40x_or_None(request, ['owner'], dandiset, return_403=True)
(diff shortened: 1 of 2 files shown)


AlmightyYakob (approved):
LGTM

mvandenburgh:
Looking at this again, the original ORM expression I wrote can be simplified - see https://github.com/dandi/dandi-archive/pull/1737/commits/5cb05e6a227c3305743e002ec3a2d39f4807149a. I tested it and it works just as well as the previous query.

AlmightyYakob:
@mvandenburgh I noticed that we're still calling the `get_user_or_400` query `n` many times, which can be avoided. I pushed [`f3b362c`](https://github.com/dandi/dandi-archive/pull/1737/commits/f3b362c08f305c84701a4418a1c2da3acaea2028) which does it in one query (two if you include the `count` check). It also removes the check about the empty `owners` list, since I'm not even sure how that could be triggered (a 400 would be raised before that in any case).

Let me know what you think. Sorry for the trampling, it wasn't possible to achieve the whole diff with github suggestions.

dandiapi/api/views/dandiset.py
```diff
@@ -372,25 +371,33 @@ def users(self, request, dandiset__pk):
...
+            # Ensure not all owners removed
+            if not serializer.validated_data:
                 raise ValidationError('Cannot remove all draft owners')
 
+            # Get all owners that have the provided username in one of the two possible locations
+            usernames = [owner['username'] for owner in serializer.validated_data]
+            user_owners = list(User.objects.filter(username__in=usernames))
+            socialaccount_owners = list(
+                SocialAccount.objects.select_related('user').filter(
+                    Q(extra_data__login__in=usernames)
```
mvandenburgh, line 383: This Q object should no longer be needed now that there's only one condition.

AlmightyYakob (approved):

PR #1741, opened 2023-11-09
Fix style issues found by Ruff

brianhelba:
None of these changes should affect logical behavior.

Arbitrary style decisions (like using parentheses in Pytest decorators, but not in raised exceptions) are determined by accepting Ruff's default settings. These are generally auto-fixable by (future usage of) Ruff, so it shouldn't impose any burden for developers to change their habits.


mvandenburgh (approved):
LGTM

PR #1744, opened 2023-11-09
Fix DJ001 Disallow null CharField on BaseZarrArchive.checksum

brianhelba:
See: https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/


dandiapi/zarr/models.py
```diff
@@ -39,22 +39,22 @@ class Meta:
...
-                    checksum__isnull=True,
+                    checksum='',
                     status__in=[
                         ZarrArchiveStatus.PENDING,
                         ZarrArchiveStatus.UPLOADED,
                         ZarrArchiveStatus.INGESTING,
                     ],
                 )
-                | models.Q(checksum__isnull=False, status=ZarrArchiveStatus.COMPLETE),
+                | (~models.Q(checksum='') & models.Q(status=ZarrArchiveStatus.COMPLETE)),
```
brianhelba, line 49: I'm not sure if there's a less awkward way to combine these conditions?

dandiapi/zarr/tests/test_ingest_zarr_archive.py
```diff
@@ -88,7 +88,7 @@ def test_ingest_zarr_archive_assets(zarr_archive_factory, zarr_file_factory, dra
     # Assert asset size, metadata
     assert asset.size == 0
     assert asset.full_metadata['contentSize'] == 0
-    assert asset.full_metadata['digest']['dandi:dandi-zarr-checksum'] is None
+    assert asset.full_metadata['digest']['dandi:dandi-zarr-checksum'] == ''
```
brianhelba, line 91: Note, this change does leak out via the HTTP API, as shown. I'm not sure if that's a compatibility problem with anything.

AlmightyYakob: This will most likely require a corresponding change in the dandi CLI

brianhelba:
Closing. Will just `#noqa` this issue.

PR #1748, opened 2023-11-16
Fix issues found by Ruff

brianhelba:


brianhelba:
Note, on f8855f29effe3cc963e2d1787d4d91e0ae9f3be1, I had to make some judgement calls as to whether the newly raised exception was sufficiently related to the surrounding exception. Feel free to comment if you disagree anywhere.

dandiapi/api/asset_paths.py
```diff
@@ -240,7 +240,7 @@ def update_asset_paths(old_asset: Asset, new_asset: Asset, version: Version):
 @transaction.atomic
 def add_version_asset_paths(version: Version):
     """Add every asset from a version."""
-    print('\t Leaves...')
+    # Leaves
```
mvandenburgh, line 243: @AlmightyYakob is there value in printing this out? Should we replace this with a `logger` call?

brianhelba: I don't know. I'm happy to change it to a `debug` or `info` log if you want.

AlmightyYakob: I'm kind of surprised those print statements are still there. I think they were originally added when the asset paths stuff was new, to make sure things were working correctly. You can remove them.

mvandenburgh (approved):

PR #1750, opened 2023-11-16
Fix G004 Logging statement uses f-string

brianhelba:
See https://docs.astral.sh/ruff/rules/logging-f-string/ for the basic rationale.

However, this is particularly relevant to the use of Sentry. By ensuring variable parts of the log message are always passed as parameters, it is much more likely that Sentry consolidate multiple events with different parameters into the same issue.


brianhelba:
I'm making this a separate PR because it's not a particularly intuitive "fix", and I want to give additional visibility in case others (@danlamanna ?) want to chime in.

yarikoptic:
FWIW, with @jwodder we fought this battle for awhile, at least on the aspect of positional args instead of `f-strings` but kinda lost it ;-)  but since Ruff helps to formalize this practice -- why not? ;)

mvandenburgh (approved):

PR #1751, opened 2023-11-16
Fix S113 Probable use of requests call without timeout

brianhelba:


brianhelba:
This may not capture all use of `requests`, but it's at least an improvement.

dandiapi/api/doi.py
```diff
@@ -44,6 +44,7 @@ def create_doi(version: Version) -> str:
                     settings.DANDI_DOI_API_USER,
                     settings.DANDI_DOI_API_PASSWORD,
                 ),
+                timeout=5,
```
yarikoptic, line 47: do we have any timing measurements so far for this endpoint?
5 seconds sounds like a lot, but not really to me, and I would hate for us to fail to mint a DOI due to some "slow processing" by datacite. I would have made it 60 or so or do you see this process blocking some other legitimate load?

yarikoptic: What about

```suggestion
                timeout=30,
```

yarikoptic:
FTR similar pr in client: https://github.com/dandi/dandi-cli/pull/1369

dandiapi/api/doi.py
brianhelba: Good suggestion, thanks. I'm updating this code and the scripts to timeout in 30 seconds to be conservative, but I'll keep the tests at 5 seconds, so they fail faster.

mvandenburgh (approved):

PR #1752, opened 2023-11-16
Fix RET505 Unnecessary `else` / `elif` after `return` statement

brianhelba:
I'm not sure if I like this rule or not. Most of the time, I think it probably is the right thing for readability, but sometimes it seems worse. Maybe I'm just not used to it?

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -203,37 +203,33 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
     def size(self):
         if self.is_blob:
             return self.blob.size
-        elif self.is_embargoed_blob:
+        if self.is_embargoed_blob:
             return self.embargoed_blob.size
-        else:
-            return self.zarr.size
+        return self.zarr.size
 
     @property
     def sha256(self):
         if self.is_blob:
             return self.blob.sha256
-        elif self.is_embargoed_blob:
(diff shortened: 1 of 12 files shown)


dandiapi/api/models/asset.py
```diff
@@ -203,35 +203,32 @@ def is_zarr(self):
     def size(self):
         if self.is_blob:
             return self.blob.size
-        elif self.is_embargoed_blob:
+        if self.is_embargoed_blob:
             return self.embargoed_blob.size
-        else:
-            return self.zarr.size
+        return self.zarr.size
```
danlamanna, line 208: I think this is a good example of this rule being kind of bad 🤷‍♂️.

yarikoptic: I think overall the design pattern above is suboptimal regardless of the diff which is overall "potato vs potato" to me. Neither original nor new code is provisioning to have some case which is no longer fulfilling the design (e.g. when new types of assets would be added...).  Should have been likely a sequence of `if/elif/else: raise AssertionError(f"Do not know how to get size for the {self}")`.

AlmightyYakob:
I support this rule and would be happy to have this merged. I think 99% of the time this vastly improves readability, and in the other 1% you can disable the rule.

AlmightyYakob (approved):

PR #1755, opened 2023-11-17
Update to preferred form of NWBE url

pgleeson:
The new form of this URL has been tested, and it would be great to get this updated on your side.

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -316,7 +316,7 @@ const EXTERNAL_SERVICES = [
     name: 'MetaCell/NWBExplorer',
     regex: /\.nwb$/,
     maxsize: 1e9,
-    endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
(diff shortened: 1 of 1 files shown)


yarikoptic:
note: would conflict with https://github.com/dandi/dandi-archive/pull/1706 but easy to resolve.

pgleeson:
Thanks @yarikoptic!

yarikoptic:
@pgleeson would you please be kind to update pr too resolve conflicts, now that we merged that other pr

waxlamp:
The URL looks like what is intended, but I wasn't able to test it (first I got a page that directed me to launch a server; after that I tried again and got a 503).

attn: @pgleeson

pgleeson:
Thanks @yarikoptic @waxlamp. Actually we've discovered an issue with our url construction which causes strange behaviour across subsequent loadings of the same url: https://github.com/OpenSourceBrain/OSBv2/issues/820, which may suggest that the original url you're using is the preferred one. We'll have a new deployment of osbv2 in the next few days and can test more, but closing this for now - ideally no change will be required. 

CC @filippomc

PR #1757, opened 2023-11-20
Fix race condition in version PUT endpoint

mvandenburgh:
If multiple concurrent calls to this endpoint occur, it's possible that some data will be clobbered. This is fixed by using select_for_update to lock the version before doing the diff and potential update of metadata.

### DIFF
--- a/dandiapi/api/views/version.py
+++ b/dandiapi/api/views/version.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from django.utils.decorators import method_decorator
 from django_filters import rest_framework as filters
 from drf_yasg.utils import no_body, swagger_auto_schema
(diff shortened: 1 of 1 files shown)


dandiapi/api/views/version.py
```diff
@@ -99,16 +100,21 @@ def update(self, request, **kwargs):
...
-        # Only save version if metadata has actually changed
-        if (name, new_metadata) != (version.name, old_metadata):
-            version.name = name
-            version.metadata = new_metadata
-            version.status = Version.Status.PENDING
-            version.save()
+        with transaction.atomic():
+            # Re-query for the version, this time using a SELECT FOR UPDATE to
+            # ensure the object doesn't change out from under us.
+            locked_version = Version.objects.select_for_update().get(id=version.id)
```
danlamanna, line 107: Did you verify this isn't already using select for update? My understanding was that `get_object` called `get_queryset` which already selects for update?

mvandenburgh: Yes, it doesn't use select for update. This is the query `get_object` runs -

```sql
SELECT "api_version"."id",
       "api_version"."created",
       "api_version"."modified",
       "api_version"."dandiset_id",
       "api_version"."name",
       "api_version"."metadata",
       "api_version"."version",
       "api_version"."doi",
       "api_version"."status",
       "api_version"."validation_errors",
       "api_dandiset"."id",
       "api_dandiset"."created",
       "api_dandiset"."modified",
       "api_dandiset"."embargo_status"
  FROM "api_version"
 INNER JOIN "api_dandiset"
    ON ("api_version"."dandiset_id" = "api_dandiset"."id")
 WHERE ("api_version"."dandiset_id" = 625 AND "api_version"."version" = 'draft' AND "api_version"."version" = 'draft')
 LIMIT 21
```

mvandenburgh: Ideally there'd be a way to tell `get_object` to do a select for update, but that doesn't seem possible. And since `get_object` does some implicit DRF validation/permission checks, making a separate query was the best thing I could think of here.

Somewhat related, upon further examination I discovered this endpoint executes 11 SQL queries :fearful:

danlamanna: I mistook `select_related` for `select_for_update` 🤦‍♂️.

danlamanna (approved):

PR #1765, opened 2023-11-30
Fix FBT001 Boolean-typed positional argument in function definition

brianhelba:
This explicitly expects Django signals and Click options to be called with keyword arguments.

### DIFF
--- a/dandiapi/api/services/version/metadata.py
+++ b/dandiapi/api/services/version/metadata.py
@@ -5,7 +5,7 @@ from dandiapi.api.models.version import Version
 
 
 def _normalize_version_metadata(
-    raw_version_metadata: dict, embargo: bool, name: str, email: str
+    raw_version_metadata: dict, name: str, email: str, *, embargo: bool
 ) -> dict:
     """
     Take raw version metadata and convert it into something suitable to be used in a formal Version.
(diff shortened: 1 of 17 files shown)


dandiapi/api/services/version/metadata.py
```diff
@@ -5,7 +5,7 @@
 
 
 def _normalize_version_metadata(
-    raw_version_metadata: dict, embargo: bool, name: str, email: str
+    raw_version_metadata: dict, embargo: bool, name: str, email: str  # noqa: FBT001
```
waxlamp, line 8: Why exclude this one? Because these arguments (including `embargo`) constitute a uniform and required description of the data to be normalized?

brianhelba: @waxlamp Good point. I was just being conservative. Changing this would either require making multiple other arguments keyword-only or changing the order of the arguments. It's an internal API, so it wouldn't break anything to change it, however.

Now they you mentioned it, I'm inclined to change this one too, particularly to have another example of the "right" (really just "safer") way to have boolean arguments.

brianhelba: Changed.

mvandenburgh:
I agree with this. I think scrutinizing any usage of boolean function arguments is a good thing, and requiring boolean args to be keyword-only is a good middle-ground.

mvandenburgh (approved):
LGTM

PR #1767, opened 2023-11-30
Fix DJ008 Model does not define `__str__` method

brianhelba:


brianhelba:
I'm not sure if any of these are ideal string representations, so feedback is welcome.

mvandenburgh (approved):
These all seem logical to me.

waxlamp:
Out of curiosity, why is this considered an error? More specifically, when are these models coerced to strings generally -- in a Python shell?

brianhelba:
@waxlamp The Django Admin pages use it in a number of places, particularly when showing a foreign key reference to another object.

See: https://docs.astral.sh/ruff/rules/django-model-without-dunder-str/#why-is-this-bad

PR #1770, opened 2023-12-01
[FIX] serviceUrl replacement

magland:
I overlooked this in #1706. The replacements on service url aren't being done correctly.


yarikoptic (approved):
Looks LGTM to me

waxlamp (approved):

waxlamp:
Adding a test for this is a good idea, but in the interest of expediency, I'll merge this as-is and @magland or someone else can come through and add tests.

magland:
Thanks!

PR #1771, opened 2023-12-01
Fix invalid contributors causing crash

### LINKED ISSUE #1769: Incorrect Metadata Specification Broke dandiset/my Request
Related to https://github.com/dandi/dandi-archive/issues/1640, I've been using a custom JavaScript API to make POST requests to create new Dandisets. However, I recently specified `contributor` as a dictionary rather than a list on a new Dandiset I was trying to set to be embargoed—resulting in the inability to make a request to https://gui-staging.dandiarchive.org/dandiset/my.

Screenshot 2023-11-30 at 2 25 45 PM

Because of this, I have no way to personally manage the issue—and, in fact, I don't even know the Dandiset ID for the problem Dandiset ("Embargo Test" as a title)

Beyond this personal problem, this points to a general issue coming from a (seeming) reliance on client-side validation—usually handled by the Python CLI—to ensure that DANDI receives the appropriate metadata to complete a Dandiset. Although it didn't break anything, I was also able to specify an invalid License field as well as create dandisets with only a Title using my JS API.

mvandenburgh:
Currently, we assume that `Version.metadata['contributors']` is a list of JSON objects. This holds true for any metadata uploaded by the DANDI CLI, but there's nothing preventing users from uploading arbitrary JSON. Invalid metadata will be marked as invalid, but there's some places in the code where we assume it's valid anyway which results in a crash. This PR adds some additional checks to those places to avoid outright crashes.

Closes #1769

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -137,7 +137,7 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         version = metadata['version']
         # If we can't find any contributors, use this citation format
         citation = f'{name} ({year}). (Version {version}) [Data set]. DANDI archive. {url}'
-        if 'contributor' in metadata and metadata['contributor']:
+        if 'contributor' in metadata and isinstance(metadata['contributor'], list):
             cl = '; '.join(
                 [
                     val['name']
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -1010,3 +1010,19 @@ def test_dandiset_rest_search_identifier(api_client, draft_version):
 
     assert results[0]['draft_version']['version'] == draft_version.version
     assert results[0]['draft_version']['name'] == draft_version.name
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize(
+    'contributors',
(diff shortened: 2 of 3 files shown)


mvandenburgh:
> My only question is, is the only place in our application where these assumptions are made? If you feel confident the answer is yes then I think this is good.

I'm not confident that this is the only place. I initially only noticed the issue with `extract_contact_person`, and the test revealed the `citation` issue. So yeah we should definitely follow up on this.

waxlamp:
Dropping a thought here so it's recorded somewhere: this seems like further evidence for a need for different schemata for "draft" and "published" phases of things. In this case, we choose not to validate anything at all until we're ready for publishing, but in reality we shouldn't allowed broken formats for things like collaborators, while we *should* allow, e.g., missing values for certain required-at-publishing-time values (like, e.g., license). This is something that should be codified in different schemata that are applied at different times. Then, we might be able to catch people sending bad metadata etc.

bendichter:
I second this. I think all contributions to the archive should be validated against an (appropriately weakened) schema and a simple error like using a dict instead of a list for the contributors field should result in a error response with an appropriate error message. It will be way easier for us if we have some semblance of order for data we receive instead of trying to handle all edge cases in query and display logic.

PR #1772, opened 2023-12-01
Add embargo re-design doc

AlmightyYakob:
This PR adds the design doc that outlines the new embargo design, with unified bucket storage.

### DIFF
--- a/doc/design/embargo-redesign.md
+++ b/doc/design/embargo-redesign.md
@@ -1,0 +1,118 @@
+# Embargo Redesign
+
+Author: Jacob Nesbitt
+
+The current embargo infrastructure setup is both inefficient and prone to error, as we store embargoed data in one bucket, and regular data in another. This means that to unembargo data, it needs to be copied from one bucket to another, which not only costs money, but time as well. Plus, when it comes to the intersection of Zarrs and embargo (”Zarrbargo”), this approach is a non-starter. Therefore, a new approach is required.
+
+## Problems with the Existing Approach
+
+### Inefficiency
+
+Embargoed data is currently uploaded to a bucket separate from the main sponsored bucket. Unembargoing involves copying data from that bucket into the sponsored bucket. Performing this copy by managing individual copy-object commands proved to show major performance issues due to the need for monitoring errors and from the sheer count of objects that need to be copied.
+
+### Error-proneness
+
+As Dandisets grow in size, comprising of more and more assets, the probability rises that the unembargo process—consisting of individually copying objects from bucket to bucket—will fail. While such failures are recoverable, that in turn requires further engineering efforts to make the process self-healing. This brings complexity and thus continued risk of failures for unembargo.
+
+### Zarrbargo non-starterness
+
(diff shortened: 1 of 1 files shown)


satra:
@AlmightyYakob - a few folks have asked for unembargoing a subset of files. they haven't gone into discussing whether this becomes a new dandiset (probably initial approach) or whether a dandiset contains both open and embargoed files (future approach, also related to restricted data in dandiset). also, in the current implementation, one can add a public file to an embargoed dandiset and it won't copy it over to the embargoed bucket. 

and this doesn't need to be implemented right away, but this use case may help think about design and ReBAC that we will likely need to implement in the future.

AlmightyYakob:
> @AlmightyYakob - a few folks have asked for unembargoing a subset of files. they haven't gone into discussing whether this becomes a new dandiset (probably initial approach) or whether a dandiset contains both open and embargoed files (future approach, also related to restricted data in dandiset). also, in the current implementation, one can add a public file to an embargoed dandiset and it won't copy it over to the embargoed bucket.
> 
> and this doesn't need to be implemented right away, but this use case may help think about design and ReBAC that we will likely need to implement in the future.

As far as I can tell, this design allows for that use case, as public read access is permitted on a per-object basis.

waxlamp (approved):

waxlamp:
> @AlmightyYakob - a few folks have asked for unembargoing a subset of files. they haven't gone into discussing whether this becomes a new dandiset (probably initial approach) or whether a dandiset contains both open and embargoed files (future approach, also related to restricted data in dandiset). also, in the current implementation, one can add a public file to an embargoed dandiset and it won't copy it over to the embargoed bucket.
> 
> and this doesn't need to be implemented right away, but this use case may help think about design and ReBAC that we will likely need to implement in the future.

@satra, I assume these ideas are not blocking development of Jake's proposal. If you would like, we can make mention of these things in Jake's doc as future directions to address, especially if that lets Jake get moving on this proposed work.

satra:
no, these are non-blocking.

the only comment i have is on the worker job for unembargoing. is there something that catches failure modes there, and what happens when it fails/partially fails?

AlmightyYakob:
> the only comment i have is on the worker job for unembargoing. is there something that catches failure modes there, and what happens when it fails/partially fails?

The S3 Batch job returns a report on if there were any failures per object, and what the reason was (after the auto-retrying of trivial errors). So we would be able to see if there was a problem with anything.

AlmightyYakob:
> perhaps it should return those in some form to the user in the validation page.

I don't think so. Failures in this process indicate an error internally, and should be hidden from the user (we could just say "there was an error un-embargoing this data). There's nothing the user can do with the details of the failure to resolve the situation, and would just lead to more clutter, confusion, etc.

If it _were_ to fail, we'd get a sentry error about it, or in the worst case scenario, the user could say "hey this failed", and we'd look at the details of it to see what happened.

satra:
> Failures in this process indicate an error internally

isn't this the same thing we are seeing with some of the validation logic, and nothing informs us or the user really when failures exist (e.g. stuck in pending validation). hence it would be good to have some kind of timeout or process so the users don't have to manually check when the unembargoing is done. i.e. something that monitors the process and has some sanity checks in it based on size/number of objects being unembargoed.

the human thing hasn't been very robust given the number of issues i have myself filed about a weird state.

i agree the user shouldn't have to bother with this, but the system should have something that notifies/handles errors or stuck states for us.

PR #1774, opened 2023-12-08
Convert `rest.ts` to composition API

### LINKED ISSUE #1267: Migrate `rest.ts` to Composition API
[`rest.ts`](https://github.com/dandi/dandi-archive/blob/master/web/src/rest.ts) should be rewritten using the Composition API to be consistent with the rest of the codebase.

mvandenburgh:
The `rest` module seems to be blocking the upgrade to Vue 3; in Vue 2, it was possible to initialize a Vue object as follows:
```javascript
const dandiRest = new Vue({
	data() {
		return { user: null }
	},
	...
});
```

And then access a data object like this:

```javascript
console.log(dandiRest.user)
```

But, the equivalent Vue 3 code doesn't work:

```javascript
const dandiRest = Vue.createApp({
	data() {
		return { user: null }
	},
	...
});

console.log(dandiRest.user) // doesn't work
```

I'm not sure what the issue is or how it was changed, but I can't even find the docs that specified this behavior for Vue 2. I'd be curious to hear if anyone has any ideas why that is. But regardless, converting to Composition API avoids this issue and is also something we've planned for a long time, so I just went ahead and did that.

Fixes #1267

### DIFF
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -1,12 +1,23 @@
-import type { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
+import type { AxiosRequestConfig, AxiosResponse } from 'axios';
 import axios from 'axios';
-import Vue from 'vue';
-import { mapState } from 'pinia';
+import { ref } from 'vue';
 import OAuthClient from '@girder/oauth-client';
 import type {
-  Asset, Dandiset, Paginated, User, Version, Info, AssetPath, Zarr, DandisetSearchResult,
+  Asset,
+  Dandiset,
+  Paginated,
+  User,
+  Version,
+  Info,
+  AssetPath,
+  Zarr,
+  DandisetSearchResult,
 } from '@/types';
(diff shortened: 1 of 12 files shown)


web/src/rest.ts
```diff
@@ -278,12 +295,12 @@ client.interceptors.request.use((config) => ({
   },
 }));
 
-const user = () => dandiRest.user;
```
mvandenburgh, line 281: Note, this function is redundant now that we're using composition API. I made the necessary updates to the rest of the codebase to account for the fact that this is a `Ref` instead of a function.

AlmightyYakob:
Is there a particular reason to _keep_ the `dandiRest` object? Or is it just a matter of PR scope? I think we could remove the wrapping entirely and just have all those functions at the top level of `rest.ts`.

AlmightyYakob (approved):
Looks good, tested locally. My only concern would be my other comment, but I still think we could merge this aside from that.

PR #1777, opened 2023-12-08
Squash migrations for "api", "analytics", and "zarr" apps

brianhelba:


danlamanna (approved):

brianhelba:
@danlamanna helped me to test this against local copies of the production and staging databases. The migrations successfully applied on both.

PR #1782, opened 2023-12-13
Don't use filesystem APIs to manipulate URLs

brianhelba:
Following #1743, this removes the remaining uses of the filesystem APIs `os.path` and `pathlib` (which are platform-dependent and contain methods for local filesystem manipulation) to process URLs.

The `zarr_checksum` package still has this confusion in its `ZarrArchiveFile` class, but attempts to refactor some of that code failed due to other issues with it. So, the `ZarrArchiveFile.path` field is just cast to a `str` when interfacing with local application code.

### DIFF
--- a/dandiapi/analytics/tasks/__init__.py
+++ b/dandiapi/analytics/tasks/__init__.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from collections import Counter
-from pathlib import Path
 from typing import TYPE_CHECKING
 
 from celery.app import shared_task
@@ -78,7 +77,7 @@ def process_s3_log_file_task(bucket: LogBucket, s3_log_key: str) -> None:
 
     # short circuit if the log file has already been processed. note that this doesn't guarantee
     # exactly once processing, that's what the unique constraint on ProcessedS3Log is for.
-    if ProcessedS3Log.objects.filter(name=Path(s3_log_key).name, embargoed=embargoed).exists():
+    if ProcessedS3Log.objects.filter(name=s3_log_key.split('/')[-1], embargoed=embargoed).exists():
         return
 
     s3 = get_boto_client(get_storage() if not embargoed else get_embargo_storage())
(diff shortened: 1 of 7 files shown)


dandiapi/zarr/tests/test_zarr.py
```diff
@@ -334,7 +334,7 @@ def test_zarr_rest_delete_multiple_files(
 
     # Assert not found
     for file in zarr_files:
-        assert not zarr_archive.storage.exists(zarr_archive.s3_path(file))
```
brianhelba, line 337: Based on the object types, it seems the old use of `file` instead of `file.path` here was a bug.

dandiapi/analytics/tasks/__init__.py
```diff
@@ -70,7 +69,7 @@ def process_s3_log_file_task(bucket: LogBucket, s3_log_key: str) -> None:
 
     # short circuit if the log file has already been processed. note that this doesn't guarantee
     # exactly once processing, that's what the unique constraint on ProcessedS3Log is for.
-    if ProcessedS3Log.objects.filter(name=Path(s3_log_key).name, embargoed=embargoed).exists():
+    if ProcessedS3Log.objects.filter(name=s3_log_key.split('/')[-1], embargoed=embargoed).exists():
```
danlamanna, line 72: These aren't really URLs, more like `PurePath`s, no?

brianhelba: It follows the structure of a `PurePosixPath`, but not necessarily the semantics (so using methods like [PurePath.as_uri](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_uri) would be wrong).

It might still make sense to change the type of all the variables holding S3 paths (to `PurePosixPath` or maybe [yarl](https://github.com/aio-libs/yarl)), but that's beyond the scope of this change. Most usages are already strings, so this makes them internally consistent and stops using inappropriate methods on them.

danlamanna (approved):

mvandenburgh (dismissed):
Looks ok from a read through, but something is clearly changing an API response or something as the CLI tests are failing. I'll take a closer look at this.

PR #1783, opened 2023-12-14
Fix S101 Use of `assert` detected

brianhelba:
In non-test parts of the codebase, `assert` shouldn't be used. See https://docs.astral.sh/ruff/rules/assert/ for rationale.

### DIFF
--- a/dandiapi/analytics/tasks/__init__.py
+++ b/dandiapi/analytics/tasks/__init__.py
@@ -22,10 +22,11 @@ LogBucket = str
 
 
 def _bucket_objects_after(bucket: str, after: str | None) -> Generator[dict, None, None]:
-    assert bucket in [
+    if bucket not in {
         settings.DANDI_DANDISETS_LOG_BUCKET_NAME,
         settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME,
-    ]
+    }:
+        raise ValueError(f'Non-log bucket: {bucket}')
     embargoed = bucket == settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME
     s3 = get_boto_client(get_storage() if not embargoed else get_embargo_storage())
     kwargs = {}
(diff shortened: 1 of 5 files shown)


dandiapi/analytics/tasks/__init__.py
```diff
@@ -22,10 +22,11 @@
 
 
 def _bucket_objects_after(bucket: str, after: str | None) -> Generator[dict, None, None]:
-    assert bucket in [
+    if bucket not in {
         settings.DANDI_DANDISETS_LOG_BUCKET_NAME,
         settings.DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME,
-    ]
+    }:
+        raise ValueError(f'Non-log bucket: {bucket}')
```
mvandenburgh, line 29: Why did you choose a `ValueError` here, but `RuntimeError` for the same error below?

Also, why the change from a `list` to a `set`?

danlamanna: I understand that the order of the buckets is irrelevant, which I'm guessing is why it was converted, but I've come to really dislike the braces-style set syntax. I hardly see it used in the wild and it doesn't work for empty sets, which just makes it confusing IMO.

brianhelba: For those cases, since the problem was directly caused by the given value of a parameter, [`ValueError` seemed like a better fit](https://docs.python.org/3/library/exceptions.html#ValueError):
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

The more [generic exception is `RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError):
> Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

brianhelba: I made it a set because that's a more appropriate collection for things which have no ordering and where membership testing is a primary consideration. Certainly here, it makes no real difference either way.

PR #1784, opened 2023-12-14
Use Ruff for Python static analysis and formatting

brianhelba:
This replaces the entire Flake8 / isort / Black toolchain, while adding many new checks and improving performance.

---

These must be merged first, for tests here to pass:
* [x] #1778
* [ ] #1782
* [x] #1783


pyproject.toml
```diff
@@ -28,3 +14,80 @@ exclude = [
...
+  "ISC001", # single-line-implicit-string-concatenation
+  "ISC002", # multi-line-implicit-string-concatenation
+  "Q", # flake8-quotes
+  "W191", # tab-indentation
+
+  "A003", # Class attribute is shadowing a Python builtin
+  "ANN", # flake8-annotations
+  "ARG001", # Unused function argument
+  "ARG002", # Unused method argument
+  "D100", # Missing docstring in public module
```
brianhelba, line 41: Could change to `D1`.

pyproject.toml
```diff
@@ -28,3 +14,80 @@ exclude = [
...
+  "RUF012", #  Mutable class attributes should be annotated with `typing.ClassVar`
+]
+
+[tool.ruff.per-file-ignores]
+"scripts/**" = [
+  "INP001", # File is part of an implicit namespace package
+]
+"**/migrations/**" = [
+  "N806", # Variable in function should be lowercase
+  "RUF012", # Mutable class attributes should be annotated with `typing.ClassVar`
```
brianhelba, line 67: Could be removed after upstream fix?

pyproject.toml
```diff
@@ -28,3 +14,80 @@ exclude = [
...
+]
+"**/management/commands/**" = [
+  "INP001", # File is part of an implicit namespace package
```
brianhelba, line 70: Maybe require this to be importable, so commands can be overridden.

pyproject.toml
```diff
@@ -28,3 +14,80 @@ exclude = [
...
+  "PLR0913", # Too many arguments to function call
+  "PLR2004", # Magic value used in comparison
+  "S101", # Use of assert detected
+  "S105", # Possible hardcoded password
+  "SLF001", # Private member accessed
+]
+
+[tool.ruff.format]
+quote-style = "single"
+
```
brianhelba, line 84: Add https://docs.astral.sh/ruff/settings/#isort-required-imports

dandiapi/api/apps.py
```diff
@@ -14,7 +14,7 @@ class PublishConfig(AppConfig):
     verbose_name = 'DANDI: Publish'
 
     @staticmethod
-    def _get_sentry_performance_sample_rate(*args, **kwargs) -> float:
+    def _get_sentry_performance_sample_rate(*args, **kwargs) -> float:  # noqa: ARG004
```
brianhelba, line 17: Maybe ignore overall.

pyproject.toml
```diff
@@ -28,3 +14,80 @@ exclude = [
...
+  "EM102", # Exception must not use an f-string literal, assign to variable first
+  "ERA001", # Found commented-out code
+  "FIX", # flake8-fixme
+  "TD002", # Missing author in TODO
+  "TD003", # Missing issue link on the line following this TODO
+  "TRY003", # Avoid specifying long messages outside the exception class
+
+  # Try to fix upstream
+  "RUF012", #  Mutable class attributes should be annotated with `typing.ClassVar`
+]
```
mvandenburgh, line 59:
```suggestion
  "PTH119", # TODO: re-enable this when it's fixed
]
```

@brianhelba we discussed this yesterday, and decided that a somewhat significant change like #1782 would be better to decouple from the Ruff work. Do you have any objections to just temporarily ignoring this rule so we can merge this PR?

mvandenburgh (approved):

pyproject.toml
mvandenburgh: Added in https://github.com/dandi/dandi-archive/pull/1784/commits/71225896f95576c9ebfdebd43d3c55b7f2eb734a

PR #1786, opened 2023-12-14
Approval view: automatically redirect back to page after authentication

### LINKED ISSUE #1701: Add authentication to /dashboard server side pages (e.g. for user review etc)
ATM procedure requires first to visit the `/swagger` page and click there "Django login" to get authenticated, and only then `dashboard/user` page becomes available.  Since it is just a matter of automated login, it would be great if such login was either performed automatically or at least prompted me for it to just click a button instead of

- upon clicking in email a url like  https://api-staging.dandiarchive.org/dashboard/user/... get confronted with `403 Forbidden` page
- if on the phone: click to open in browser
- go edit url manually to remove everything and replace with `/swagger` endpoint -- note that it could be staging or main instance... later even more
- click there on 'Django login'
- go back in history to finally review the user

mvandenburgh:
Fixes #1701 

This is a follow on to #1781. After that PR, accessing the user approval pages while unauthenticated results in a redirect to the login endpoint, which subsequently redirects to github; but, github redirects the user back to the root path `/` because it doesn't know where the user came from. They allow specifying an optional `next` query parameter, which github will respect and redirect the user to if it's supplied, so I updated the redirect to do that here.

### DIFF
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -4,7 +4,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.db.models import Exists, OuterRef
-from django.http.response import HttpResponseRedirect
+from django.http import HttpRequest, HttpResponseRedirect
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
Look good, tested locally without any issue

PR #1789, opened 2023-12-14
Add an e2e test for the FileBrowser

### LINKED ISSUE #1760: Add rudimentary end-to-end test for meditor and file browser from DLP
As we all failed to catch an obvious bug while reviewing #1706 and that required hotfix  #1759 . Could be prevented with a basic test of opening file browser view.  I looked at https://github.com/dandi/dandi-archive/tree/master/web/test/src/tests but didn't find any for meditor or file browser. Even if not thorough/sophisticated, I think such smoke tests could be already of notable value.

marySalvi:
Adds simple e2e test for the FileBrowser 
closes #1760

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -113,6 +113,13 @@ jobs:
       - name: Create any cache tables
         run: python manage.py createcachetable
 
+        # Populate the database with a dandiset for FileBrowser tests
+      - name: Create a superuser
+        run: python manage.py createsuperuser --noinput --email admin@admin.com
+
+      - name: Create Dandiset
+        run: python manage.py create_dev_dandiset --owner admin@admin.com
+
       - name: Install E2E tests
         run: yarn install --frozen-lockfile
         working-directory: e2e
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -115,10 +115,10 @@ jobs:
 
         # Populate the database with a dandiset for FileBrowser tests
       - name: Create a superuser
-        run: python manage.py createsuperuser --noinput --email admin@admin.com
+        run: python manage.py createsuperuser --noinput --email admin@example.com
 
       - name: Create Dandiset
-        run: python manage.py create_dev_dandiset --owner admin@admin.com
+        run: python manage.py create_dev_dandiset --owner admin@example.com
 
       - name: Install E2E tests
         run: yarn install --frozen-lockfile
--- a/e2e/src/tests/fileBrowserPage.test.js
+++ b/e2e/src/tests/fileBrowserPage.test.js
@@ -1,0 +1,21 @@
+import { vBtn, vListItemSubtitle  } from 'jest-puppeteer-vuetify';
+import {
+    waitForRequestsToFinish,
+  } from '../util';
+
+describe('file browser page', () => {
+    it('view file browser', async () => {
+
+        await expect(page).toClickXPath(vBtn('Public Dandisets'));
+        await waitForRequestsToFinish();
+
+        // Assert that the page has loaded with a list of dandisets
+        await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
+        await expect(page).toClickXPath(vBtn('Files'));
+        await waitForRequestsToFinish();
+
+        // Ensure that the page has loaded with a folder
+        await expect(page).toMatch('foo');
+
+    });
+});
--- a/e2e/src/tests/fileBrowserPage.test.js
+++ b/e2e/src/tests/fileBrowserPage.test.js
@@ -1,21 +1,19 @@
-import { vBtn, vListItemSubtitle  } from 'jest-puppeteer-vuetify';
+import { vBtn, vListItemSubtitle } from 'jest-puppeteer-vuetify';
 import {
-    waitForRequestsToFinish,
-  } from '../util';
+  waitForRequestsToFinish,
+} from '../util';
 
 describe('file browser page', () => {
-    it('view file browser', async () => {
+  it('view file browser', async () => {
+    await expect(page).toClickXPath(vBtn('Public Dandisets'));
+    await waitForRequestsToFinish();
 
-        await expect(page).toClickXPath(vBtn('Public Dandisets'));
-        await waitForRequestsToFinish();
+    // Assert that the page has loaded with a list of dandisets
+    await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
+    await expect(page).toClickXPath(vBtn('Files'));
+    await waitForRequestsToFinish();
 
-        // Assert that the page has loaded with a list of dandisets
-        await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
-        await expect(page).toClickXPath(vBtn('Files'));
-        await waitForRequestsToFinish();
-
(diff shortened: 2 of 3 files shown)


e2e/src/tests/fileBrowserPage.test.js
```diff
@@ -0,0 +1,21 @@
...
+        await expect(page).toClickXPath(vBtn('Public Dandisets'));
+        await waitForRequestsToFinish();
+
+        // Assert that the page has loaded with a list of dandisets
+        await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
+        await expect(page).toClickXPath(vBtn('Files'));
+        await waitForRequestsToFinish();
+
+        // Ensure that the page has loaded with a folder
+        await expect(page).toMatch('foo');
```
marySalvi, line 18: I feel as this is not ideal since the name for the folder could change. Preferably I would select the element with js but that would require us to install `jsdom` for jest and I don't know if that is wise for just this test.

.github/workflows/frontend-ci.yml
```diff
@@ -113,6 +113,13 @@ jobs:
       - name: Create any cache tables
         run: python manage.py createcachetable
 
+        # Populate the database with a dandiset for FileBrowser tests
+      - name: Create a superuser
+        run: python manage.py createsuperuser --noinput --email admin@admin.com
```
waxlamp, line 118: Since admin.com is a real domain, I'd change this to example.com

```suggestion
        run: python manage.py createsuperuser --noinput --email admin@example.com
```

.github/workflows/frontend-ci.yml
```diff
@@ -113,6 +113,13 @@ jobs:
...
+
+      - name: Create Dandiset
+        run: python manage.py create_dev_dandiset --owner admin@admin.com
```
waxlamp, line 121:
```suggestion
        run: python manage.py create_dev_dandiset --owner admin@example.com
```

e2e/src/tests/fileBrowserPage.test.js
waxlamp: How does `.toMatch()` work on the `page` object? Can you use the `page` object's [`$()` method](https://pptr.dev/api/puppeteer.page._) to select the element?

You would still have the problem of hard-coding the name of the folder, but I don't think that's a big deal. If we ever change the folder name in the test dandisets, this test will fail and we'll just fix it.

waxlamp (changes requested):
This is looking good. Minor change suggested to the test data domain, and an external (to this review) suggestion to use `Page.$()` to select an element for test assertions.

e2e/src/tests/fileBrowserPage.test.js
marySalvi: `.toMatch()` is just regex, you can't use jquery with it:
https://jestjs.io/docs/expect#tomatchregexp--string

To use js/jquery we would need to set the jest environment to `jsdom` which does not come with jest out of the box and would be an additional package:
https://jestjs.io/docs/configuration#testenvironment-string
https://www.npmjs.com/package/jest-environment-jsdom

In that case it would be something along the lines of `Page.(document.getElementsByCLassName('v-list')`

marySalvi: After a bit more searching its likely to complete this using xpath ie `path.(//div[contains(@class,'v-list'))`. I'll give it a shot.

marySalvi: implemented in 1258b94bb37a266cc6262f55b302201715c28300

.github/workflows/frontend-ci.yml
marySalvi: fixed in 1258b94bb37a266cc6262f55b302201715c28300

.github/workflows/frontend-ci.yml
marySalvi: fixed in 1258b94bb37a266cc6262f55b302201715c28300

.github/workflows/frontend-ci.yml
```diff
@@ -113,6 +113,13 @@ jobs:
       - name: Create any cache tables
         run: python manage.py createcachetable
 
+        # Populate the database with a dandiset for FileBrowser tests
```
yarikoptic, line 116: What was then tested before if there were not a single dandiset?

e2e/src/tests/fileBrowserPage.test.js
```diff
@@ -0,0 +1,19 @@
...
+    await expect(page).toClickXPath(vBtn('Public Dandisets'));
+    await waitForRequestsToFinish();
+
+    // Assert that the page has loaded with a list of dandisets
+    await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
+    await expect(page).toClickXPath(vBtn('Files'));
+    await waitForRequestsToFinish();
+
+    // Ensure that the page has loaded with a folder
+    await expect(page).toContainXPath(vList());
```
mvandenburgh, line 17: I _think_ something like this should work, although I haven't tested it:
```suggestion
    // Ensure that the page has loaded with a folder
    await expect(page).toContainXPath(vListItem('foo'));
```

That way there's a stricter test for the presence of the folder, rather than assuming that the presence of a `v-list` element means the folder component is there.

e2e/src/tests/fileBrowserPage.test.js
```diff
@@ -0,0 +1,19 @@
...
+  waitForRequestsToFinish,
+} from '../util';
+
+describe('file browser page', () => {
+  it('view file browser', async () => {
+    await expect(page).toClickXPath(vBtn('Public Dandisets'));
+    await waitForRequestsToFinish();
+
+    // Assert that the page has loaded with a list of dandisets
+    await expect(page).toClickXPath(vListItemSubtitle('DANDI:000001'));
```
mvandenburgh, line 12: The assumption that dandiset 000001 is on the first page of the Public Dandisets page is only true for certain test execution orderings. In practice, it seems jest is picking it as one of the first tests it runs. But if it got scheduled after, say, the [search page test](https://github.com/dandi/dandi-archive/blob/master/e2e/src/tests/dandisetsPage.test.js#L31), it would end up on the second or third page at least.

One option could be to manually specify an order so that this test runs first (https://jestjs.io/docs/configuration#testsequencer-string), or just modify the code to navigate directly to `/dandisets/000001`/ instead of clicking through the UI to get there. Personally I think the second option sounds cleaner and also has the benefit of decreasing the scope of this test (since this test is only focused on the file browser, and doesn't need to concern itself with clicking through the Public Dandiset page)

e2e/src/tests/fileBrowserPage.test.js
marySalvi: changed in 21e7b0f81fed2209d602daa987c6657bce2d1bf0

e2e/src/tests/fileBrowserPage.test.js
marySalvi: changed in 21e7b0f81fed2209d602daa987c6657bce2d1bf0

mvandenburgh (approved):
LGTM

mvandenburgh:
It looks like the one CLI test is failing due to https://github.com/dandi/dandi-cli/pull/1385 not being released yet. The tests that run on dandi-cli `master` are passing, so I think it's fine to merge this despite that particular check failing.

waxlamp (approved):

PR #1790, opened 2023-12-14
Return 401 on unauthenticated embargo dandiset list

### LINKED ISSUE #1788: API should respond with 401 not just silently not list embargoed if embargoed are requested
ATM we have odd behavior that even if I request embargoed dandisets explicitly, I would simply not get them listed in the output listing all dandisets, instead of what would be more appropriate IMHO to provide some feedback that authentication is needed to get that information.

```shell
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/dandisets/?page_size=1000&draft=true&empty=true&embargoed=true' -H 'accept: application/json' -H "Authorization: token $DANDI_API_KEY" | jq . | grep -e embargo_status | sort | uniq -c
     81       "embargo_status": "EMBARGOED",
    350       "embargo_status": "OPEN",
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/dandisets/?page_size=1000&draft=true&empty=true&embargoed=true' -H 'accept: application/json' | jq . | grep -e embargo_status | sort | uniq -c
    350       "embargo_status": "OPEN",
```

NB not to say that there is no option in swagger to provide examples with proper header field for that, just some insufficient X-CSRFToken.

IMHO [401 with WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) would be most DX friendly to respond if I do request explicitly `embargoed=true`.  If `embargoed=false` or not explicitly provided - then it would be fine to not list them.

WDYT @satra @jwodder and @dandi/dandiarchive  ?

jjnesbitt:
Fixes #1788

This slightly changes the way the API behaves, aside from the stated goal. Previously, we defaulted to a value of `true` when `embargoed` was not specified. This PR changes that default to `false`, as otherwise, any un-authenticated users would need to always explicitly pass `embargoed=false` when calling the dandiset list endpoint.

cc @yarikoptic

### DIFF
--- a/dandiapi/api/services/dandiset/exceptions.py
+++ b/dandiapi/api/services/dandiset/exceptions.py
@@ -5,3 +5,10 @@ from dandiapi.api.services.exceptions import DandiError
 
 class DandisetAlreadyExistsError(DandiError):
     http_status_code = status.HTTP_400_BAD_REQUEST
+
+
+class UnauthorizedEmbargoAccess(DandiError):
+    http_status_code = status.HTTP_401_UNAUTHORIZED
+    message = (
+        'Authentication credentials must be provided when attempting to access embargoed dandisets'
+    )
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -173,7 +173,7 @@ class DandisetDetailSerializer(DandisetSerializer):
 class DandisetQueryParameterSerializer(serializers.Serializer):
     draft = serializers.BooleanField(default=True)
     empty = serializers.BooleanField(default=True)
-    embargoed = serializers.BooleanField(default=True)
+    embargoed = serializers.BooleanField(default=False)
     user = serializers.CharField(required=False)
 
 
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -63,7 +63,8 @@ def test_dandiset_manager_visible_to(
 
 
 @pytest.mark.django_db()
-def test_dandiset_rest_list(api_client, dandiset):
+def test_dandiset_rest_list(api_client, user, dandiset):
+    # Test un-authenticated request
     assert api_client.get('/api/dandisets/', {'draft': 'true', 'empty': 'true'}).json() == {
         'count': 1,
         'next': None,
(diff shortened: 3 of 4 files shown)


dandiapi/api/services/dandiset/exceptions.py
```diff
@@ -5,3 +5,10 @@
 
 class DandisetAlreadyExistsError(DandiError):
     http_status_code = status.HTTP_400_BAD_REQUEST
+
+
+class UnauthorizedEmbargoAccess(DandiError):
+    http_status_code = status.HTTP_401_UNAUTHORIZED
+    message = (
+        "Authentication credentials must be provided when attempting to access embargoed dandisets"
```
yarikoptic, line 13: while troubleshooting this I realized we nowhere show how to authenticate for such requests.  I wonder if it would make sense to add here or not? e.g.

```suggestion
        "Authentication credentials must be provided in the 'Authorization' header field with 'token API_TOKEN' value when attempting to access embargoed dandisets"
```

or alike. WDYT?

dandiapi/api/views/serializers.py
```diff
@@ -173,7 +173,7 @@ class Meta(DandisetSerializer.Meta):
 class DandisetQueryParameterSerializer(serializers.Serializer):
     draft = serializers.BooleanField(default=True)
     empty = serializers.BooleanField(default=True)
-    embargoed = serializers.BooleanField(default=True)
+    embargoed = serializers.BooleanField(default=False)
```
yarikoptic, line 176: I think it might be insufficient to retain current webui behavior, since when authenticated we do see embargoed ones "by default". May be there should be need to implement handling of default None, which would maintain current behavior

dandiapi/api/services/dandiset/exceptions.py
jjnesbitt: The response of any one API endpoint isn't really the place for documentation about authentication to our API. That would go in some kind of specific documentation, like the [handbook](https://www.dandiarchive.org/handbook/40_development/).

If what you're saying is that we need better documentation around API authentication, that's a fair point, but is separate from the concerns of this PR.

dandiapi/api/views/serializers.py
jjnesbitt: I'll be including the appropriate webui changes in this PR as well, so that everything is consistent.

dandiapi/api/services/dandiset/exceptions.py
yarikoptic: agreed!

yarikoptic: filed https://github.com/dandi/handbook/issues/109

jjnesbitt:
> I'll be including the appropriate webui changes in this PR as well, so that everything is consistent.

It turns out no webui changes are necessary, as `embargoed=true` isn't included in the UI unless the user is already logged in.

waxlamp (approved):

PR #1795, opened 2023-12-19
Update + Reconfigure Vue Sentry SDK

mvandenburgh:
I noticed this error log appearing in each Sentry issue for the frontend -
```
[@sentry/vue]: Misconfigured SDK. Vue specific errors will not be captured.
Update your `Sentry.init` call with an appropriate config option:
`app` (Application Instance - Vue 3) or `Vue` (Vue Constructor - Vue 2).
```

So, I updated the sentry SDK to the latest version and reconfigured it using the latest docs. I also enabled [Sentry Replay](https://docs.sentry.io/product/session-replay/) so we can try that out.


mvandenburgh:
Hmm, the deploy preview is still logging that SDK error message in the console :thinking:

mvandenburgh:
It looks like [this](https://github.com/dandi/dandi-archive/pull/1795/commits/2dbb39e12fdcc59f4916bb41c6f4f91f71481a19) was the issue :facepalm:

danlamanna (approved):

Issue #1653, opened 2023-07-13
Remove use of Asset `VALIDATING` status if not used


AlmightyYakob:
Follow up from https://github.com/dandi/dandi-archive/pull/1634#discussion_r1238884164

PR #1800, opened 2023-12-30
Don't use .save() in validate_version_metadata

### LINKED ISSUE #1734: asset metadata not appearing on DLC
https://dandiarchive.org/dandiset/000691/draft

image

Does anyone know why asset metadata might not be appearing here?

jjnesbitt:
We're seeing a very prevalent race condition, resulting in version asset summary aggregations being skipped (this can be observed via copious amounts of `Skipped updating assetsSummary for version ***` in the papertrail logs).

Here's what I believe is happening:

1. A version is updated in some way, and status set to `PENDING`
2. The scheduled task [`validate_draft_version_metadata`](https://github.com/dandi/dandi-archive/blob/65eca4a4586bfbad820d4b44ba40f97ce487a1ae/dandiapi/api/tasks/scheduled.py#L75) is run, dispatching the [`validate_version_metadata`](https://github.com/dandi/dandi-archive/blob/65eca4a4586bfbad820d4b44ba40f97ce487a1ae/dandiapi/api/services/metadata/__init__.py#L100) and [`aggregate_assets_summary`](https://github.com/dandi/dandi-archive/blob/65eca4a4586bfbad820d4b44ba40f97ce487a1ae/dandiapi/api/services/metadata/__init__.py#L80) tasks at the same time
3. `validate_version_metadata` finishes before `aggregate_assets_summary`, calling `Version.save()`, which calls [`Version._populate_metadata`](https://github.com/dandi/dandi-archive/blob/65eca4a4586bfbad820d4b44ba40f97ce487a1ae/dandiapi/api/models/version.py#L173), potentially modifying the metadata
4. `aggregate_assets_summary` finishes generating the summary and reaches [this](https://github.com/dandi/dandi-archive/blob/65eca4a4586bfbad820d4b44ba40f97ce487a1ae/dandiapi/api/services/metadata/__init__.py#L93) line, which will only update the version if the `metadata` field hasn't changed from when the task was dispatched. Since the above step modifies the metadata, this filter fails and the update is skipped
5. The version assets summary remains outdated/incorrect

This PR changes the behavior of `validate_version_metadata` to use `.update` when updating the `status` and `validation_errors` of a Version, subverting the underlying call to `Version._populate_metadata`, and thus any metadata changes.

@danlamanna I'd like to hear your input on this, particularly to make sure I'm not violating any intended behavior. I believe I've largely kept in line with what was already there, but I'm unfamiliar with the reason/need for using the `metadata=version.metadata` filter when updating a version's asset summary.

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -140,43 +140,42 @@ def validate_version_metadata(*, version: Version) -> None:
     with transaction.atomic():
         # validating version metadata needs to lock the version to avoid racing with
         # other modifications e.g. aggregate_assets_summary.
-        version = (
-            Version.objects.filter(id=version.id, status=Version.Status.PENDING)
-            .select_for_update()
-            .first()
-        )
 
         # It's possible for this version to get deleted during execution of this function.
         # If that happens *before* the select_for_update query, return early.
-        if version is None:
+        version_qs = Version.objects.filter(id=version.id).select_for_update()
+        if not version_qs.filter(status=Version.Status.PENDING).exists():
             logger.info('Version %s no longer exists, skipping validation', version_id)
             return
 
-        version.status = Version.Status.VALIDATING
-        version.save()
-
+        # Set to validating and continue
+        version_qs.update(status=Version.Status.VALIDATING)
         try:
             validate(
-                _build_validatable_version_metadata(version),
+                _build_validatable_version_metadata(version_qs.first()),
                 schema_key='PublishedDandiset',
                 json_validation=True,
             )
         except dandischema.exceptions.ValidationError as e:
             logger.info('Error while validating version %s', version.id)
-            version.status = Version.Status.INVALID
-
-            validation_errors = _collect_validation_errors(e)
-            version.validation_errors = validation_errors
-            version.save()
+            version_qs.update(
+                status=Version.Status.INVALID,
+                validation_errors=_collect_validation_errors(e),
+                modified=timezone.now(),
+            )
             return
         except ValueError as e:
             # A bare ValueError is thrown when dandischema generates its own exceptions, like a
             # mismatched schemaVersion.
-            version.status = Version.Status.INVALID
-            version.validation_errors = [{'field': '', 'message': str(e)}]
-            version.save()
+            logger.info('Error while validating version %s', version.id)
+            version_qs.update(
+                status=Version.Status.INVALID,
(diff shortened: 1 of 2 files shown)


dandiapi/api/services/metadata/__init__.py
```diff
@@ -129,36 +129,33 @@ def _build_validatable_version_metadata(version: Version) -> dict:
...
-        version = (
-            Version.objects.filter(id=version.id, status=Version.Status.PENDING)
-            .select_for_update()
-            .first()
-        )
-        version.status = Version.Status.VALIDATING
-        version.save()
+        version_qs = Version.objects.filter(id=version.id).select_for_update()
+        if not version_qs.filter(status=Version.Status.PENDING).exists():
+            # TODO: Should raise an error here?
```
jjnesbitt, line 134: I'd be happy to just insert a log statement and return here, but I wasn't sure if this warranted raising a more concrete error.

jjnesbitt:
I just realized I may need to also include an update to the version `modified` field.

dandiapi/api/services/metadata/__init__.py
```diff
@@ -140,43 +160,23 @@ def _build_validatable_version_metadata(version: Version) -> dict:
...
-            version.validation_errors = [{'field': '', 'message': str(e)}]
-            version.save()
             return
-        logger.info('Successfully validated version %s', version.id)
-        version.status = Version.Status.VALID
-        version.validation_errors = []
-        version.save()
+
+        # Set to validating and continue
+        version_qs.update(status=Version.Status.VALIDATING)
```
mvandenburgh, line 180: This won't ever get committed to the DB, right? Since this code is wrapped in `transaction.atomic`, and the status gets updated again immediately after this.

jjnesbitt: That's true, although that problem already existed in the prior implementation. I'm not sure it makes sense to try to solve that here and now, since it might change the implementation significantly. As far as I can tell the `VALIDATING` status doesn't really get used much, if at all.

waxlamp: FWIW, a similar thing (for asset `VALIDATING` status) is being tracked in #1653. Perhaps we should have a parallel issue for version `VALIDATING` as well.

mvandenburgh: Sounds good. Considering it was an issue prior to this PR as well, I think this is fine as-is.

mvandenburgh (approved):

Issue #565, opened 2021-10-15
Add migrated version metadata endpoint


dchiquito:
Implement https://github.com/dandi/dandi-api/blob/master/doc/design/draft-metadata-migration.md

Add an endpoint `GET api/dandisets/{dandiset_pk}/versions/{version}/migrated/` that attempts to migrate the version metadata to the current schemaVersion. If the metadata is already at the current schemaVersion, nothing is done. If the migration fails for whatever reason, simply return the current metadata.

The meditor will use this new endpoint to get the metadata for a version, ensuring that whenever a user saves the meditor content, the schemaVersion is as up to date as possible.

satra:
@AlmightyYakob - just to clarify that this particular issue is explicitly about the DLP/dandiset metadata, which is only editable through the meditor (or if someone uses the API directly). It was also there to provide a route where the DLP does not need to consider metadata versions.  it is possible that migration fails, but that should be something we should handle as part of some CI.

while i would want to ensure that our current schemas are always applicable, this particular scenario is about dandiset metadata, while the other publishing issue was more related to asset metadata, which is a different issue, not this one.

AlmightyYakob:
> @AlmightyYakob - just to clarify that this particular issue is explicitly about the DLP/dandiset metadata, which is only editable through the meditor (or if someone uses the API directly). It was also there to provide a route where the DLP does not need to consider metadata versions. it is possible that migration fails, but that should be something we should handle as part of some CI.
> 
> while i would want to ensure that our current schemas are always applicable, this particular scenario is about dandiset metadata, while the other publishing issue was more related to asset metadata, which is a different issue, not this one.

I don't understand your point, I was not referring to asset metadata. I was referring to this [design](https://github.com/dandi/dandi-archive/blob/master/doc/design/draft-metadata-migration.md) doc, which exclusively refers to dandiset/version metadata.

yarikoptic:
In general I like the idea of explicit auto-metadata migration @AlmightyYakob upon user request! It could visualize some kind of a diff to the user etc.  Also it should be then a record in the perspective audit DB (https://github.com/dandi/dandi-archive/issues/525).  We have the design doc you refer to and I am not entirely clear on what is the discussion exactly about here ;)

satra:
[...]

PR #1801, opened 2024-01-02
Design doc for Audit MVP

### LINKED ISSUE #525: Write design docs for internal and external audit functionality
In addition to a design for the audit system as a whole, it should include specifics on the following features:

- external
  - who made what changes (create/add/delete/modify/publish/change ownership - includes data and metadata - even better includes known apps such as web ui, cli )
- internal
  - how many blobs not associated with assets (or delta between total blob size and repo size)
  - how much to garbage collect
  - uploaders by week
  - API requests by origin
  - data or metadata last accessed by dandiset 
  - number of assets not accessed in a given time delta

waxlamp:
Closes #525.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,7 @@
 [codespell]
-skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log,./doc
+skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log
 # Disabled until https://github.com/codespell-project/codespell/issues/2727
 # got answer/re-solution
 # dictionary = .codespell_dict
+builtin = clear
 # ignore-words-list = 
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,5 +1,5 @@
 [codespell]
-skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log
+skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log,./doc
 # Disabled until https://github.com/codespell-project/codespell/issues/2727
 # got answer/re-solution
 # dictionary = .codespell_dict
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -52,7 +52,7 @@ additional, event-specific data each will carry with it.
 | Add owner | date_time, user_id, dandiset_id | owner_id |
 | Remove owner | date_time, user_id, dandiset_id | owner_id |
 | Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
-| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id |
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id |
 | Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata |
 | Remove asset | date_time, user_id, dandiset_id | asset_path |
 | Publish Dandiset | date_time, user_id, dandiset_id | version |
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -46,17 +46,18 @@ The following is the list of actions performed on Dandisets that are considered
 "significant" (i.e., require an audit record to be generated), along with any
 additional, event-specific data each will carry with it.
 
-|  | General data | Specific data |
-| --- | --- | --- |
-| Create Dandiset | date_time, user_id, dandiset_id | metadata |
-| Add owner | date_time, user_id, dandiset_id | owner_id |
-| Remove owner | date_time, user_id, dandiset_id | owner_id |
-| Update Version metadata | date_time, user_id, dandiset_id | metadata |
-| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id |
-| Update asset | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
-| Remove asset | date_time, user_id, dandiset_id | asset_path, asset_id |
-| Publish Dandiset | date_time, user_id, dandiset_id | version |
-| Delete Dandiset | date_time, user_id, dandiset_id | none |
+| Action                  | General data                    | Specific data                                                                   |
+| ---                     | ---                             | ---                                                                             |
+| Create Dandiset         | date_time, user_id, dandiset_id | metadata, embargoed                                                             |
+| Add owner               | date_time, user_id, dandiset_id | owner_id                                                                        |
+| Remove owner            | date_time, user_id, dandiset_id | owner_id                                                                        |
+| Update Version metadata | date_time, user_id, dandiset_id | metadata                                                                        |
+| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id                               |
+| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
+| Remove asset            | date_time, user_id, dandiset_id | asset_path, asset_id                                                            |
+| Unembargo Dandiset      | date_time, user_id, dandiset_id | (none)                                                                          |
+| Publish Dandiset        | date_time, user_id, dandiset_id | version                                                                         |
+| Delete Dandiset         | date_time, user_id, dandiset_id | (none)                                                                          |
 
 ## Asset chain analysis
 
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -52,8 +52,8 @@ additional, event-specific data each will carry with it.
 | Add owner               | date_time, user_id, dandiset_id | owner_id                                                                        |
 | Remove owner            | date_time, user_id, dandiset_id | owner_id                                                                        |
 | Update Version metadata | date_time, user_id, dandiset_id | metadata                                                                        |
-| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id                               |
-| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
+| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, zarr_id, asset_id                               |
+| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, zarr_id, metadata, asset_id, old_asset_id |
 | Remove asset            | date_time, user_id, dandiset_id | asset_path, asset_id                                                            |
 | Unembargo Dandiset      | date_time, user_id, dandiset_id | (none)                                                                          |
 | Publish Dandiset        | date_time, user_id, dandiset_id | version                                                                         |
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -11,6 +11,11 @@ when a given owner did not make the change and wishes to know who to speak to
 about those changes. In essence, it is a tool to support effective
 collaboration.
 
+Additionally, this audit feature can support other efforts that depend on the
+stream of events affecting a Dandiset, such as reconstructing the history of
+changes that are then reflected in the DataLad versions of Dandisets maintained
+at https://github.com/dandisets.
+
 ## Requirements
 
 1. **Persistent tracking of significant actions.** Whenever a *significant
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -54,7 +54,7 @@ additional, event-specific data each will carry with it.
 | Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
 | Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id |
 | Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
-| Remove asset | date_time, user_id, dandiset_id | asset_path |
+| Remove asset | date_time, user_id, dandiset_id | asset_path, asset_id |
 | Publish Dandiset | date_time, user_id, dandiset_id | version |
 | Delete Dandiset | date_time, user_id, dandiset_id | none |
 
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -1,0 +1,152 @@
+# Audit Trail MVP
+
+NIH has required that DANDI Archive generate audit logs while leaving undefined
+what that means. We are choosing to fulfill the requirement with a feature that
+indicates to Dandiset owners and admins the sequence of significant actions
+taken to modify Dandisets, along with identification of who performed each
+action.
+
+The major motivation is to inform owners about how their Dandisets have changed
+when a given owner did not make the change and wishes to know who to speak to
+about those changes. In essence, it is a tool to support effective
+collaboration.
+
+## Requirements
+
+1. **Persistent tracking of significant actions.** Whenever a *significant
+   action* (see list [below](#significant-actions)) occurs successfully, the archive will generate a
+   persistent record of that action, including the following data:
+    - The date and time
+    - The authorized user who triggered the action
+    - The ID of the Dandiset on which the action occurs
+    - Event-specific data about the action itself
+
+2. **Display of audit log (to owners and admins only).** The Dandiset landing
+   page will display a rendered (i.e. human-readable) version of the log event
+   data (as long as the viewing user is an owner of the Dandiset, or an admin).
+
+   A possible example of such a display follows (the final rendered form may
+   differ from this depending on best practices, etc.):
+    
+    ```
+    YYYYMMDDTHHMMSS.mmmm: X added asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X modified asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X deleted asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X modified dandiset metadata
+    ```
+    
+3. **Extensible significant action list.** The significant actions the system is
+   capable of tracking should be extensible--that is, in the future new actions
+   should be able to be added to the system, and existing actions should be able
+   to expand the specific data they carry.
+
+## Significant Actions
+
+The following is the list of actions performed on Dandisets that are considered
+"significant" (i.e., require an audit record to be generated), along with any
+additional, event-specific data each will carry with it.
+
+|  | General data | Specific data |
+| --- | --- | --- |
+| Create Dandiset | date_time, user_id, dandiset_id | metadata |
+| Add owner | date_time, user_id, dandiset_id | owner_id |
+| Remove owner | date_time, user_id, dandiset_id | owner_id |
+| Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id |
+| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata |
+| Remove asset | date_time, user_id, dandiset_id | asset_path |
+| Publish Dandiset | date_time, user_id, dandiset_id | version |
+| Delete Dandiset | date_time, user_id, dandiset_id | none |
+
+## Asset chain analysis
+
+This section provides some analysis of the historical data living in the
+"`Asset` chains" kept in the `Asset.previous` links. The immediate goal is to
+determine how much "significant action" data can be reconstructed therefrom;
+there is also some analysis of how this may affect provenance work in the
+future.
+
+For context: because most changes to an `Asset` model cause a new model instance
+to be created, and because that new `Asset` instance points to its "ancestor"
+via a `previous` property, the history of an asset is at least partially
+traceable through this "asset chain".
+
+### Audit
+
+| Action | Recoverable fields | Unrecoverable fields | Notes |
+| --- | --- | --- | --- |
+| Create Dandiset | date_time, dandiset_id | user_id, metadata | date_time is equal to the Dandiset’s created_at value. user_id and metadata (as they were at the beginning of the Dandiset’s lifecycle) cannot be known with certainty.<br><br> Dandisets that have already been deleted will not have any recoverable fields (though we believe it very unlikely that a Dandiset has ever actually been deleted).<br><br> Independent of asset chains. |
+| Add owner | none | date_time, user_id, dandiset_id, owner_id | These events are all "summed" into the current owners list, with no way to know the evolution thereof.<br><br> Independent of asset chains. |
+| Remove owner | none | date_time, user_id, dandiset_id, owner_id | These events are all "summed" into the current owners list, with no way to know the evolution thereof.<br><br> Independent of asset chains. |
+| Update Dandiset metadata | none | date_time, user_id, dandiset_id, metadata | These events are all "summed" into the current Dandiset metadata, with no way to know the evolution thereof.<br><br> Independent of asset chains. |
+| Add asset | date_time, dandiset_id, asset_path, checksum, metadata | user_id | The last Asset instance in the chain (i.e., the one that has a null previous pointer) represents the version of the asset that was first added to a Dandiset. Everything about this asset besides the user who uploaded it is available in the Asset instance, the AssetBlob it points to, and the association table between Dandisets and Assets that relates them.<br><br> The dandiset_id may not be recoverable for Asset instances that have been removed from a Dandiset, or for Dandisets that were deleted. |
+| Update asset | date_time, dandiset_id, asset_path, checksum, metadata | user_id | As above, the user_id information is missing, and the dandiset_id may be as well. |
+| Remove asset | asset_path | date_time, user_id, dandiset_id | A removed asset will be one at the head of an asset chain that has no association to a Dandiset. Unfortunately, without that association, the dandiset_id cannot be recovered, and because it is the absence of a relation that indicates removal, the date_time value is also unrecoverable. |
+| Publish Dandiset | date_time, dandiset_id, version | user_id | The date_time, dandiset_id, and version are all available as part of the Version instance’s metadata. |
+| Delete Dandiset | none | date_time, user_id, dandiset_id | Deleted Dandisets just don’t exist anymore, so no information (even the dandiset_id) is available. |
+
+Note that the majority of information describing these auditable actions is
+unrecoverable.
+
+### Provenance
+
+Provenance describes the heritage of domain artifacts in the system in terms of
+how they originated and where they originated from. Many of the audit actions
+described above may be thought of as simple provenance events as well: for
+example, when a user publishes a Dandiset, the published `Version` that results
+can be understood as originating from the draft version present in the Dandiset
+at a specific time. Taken in that sense, these simple provenance events can be
+identified with corresponding audit events.
+
+More complex provenance events do not currently exist in the system. For
+example, a reasonable action a user could take would be to manufacture a new
+Dandiset from the state of a published version in some other Dandiset. If this
+action were afforded to the user, we might think of recording a provenance event
+describing the specific way in which that derived Dandiset came to be.
+
+Because we do not currently afford such complex provenance events in the system,
+the operations to fulfill the basic audit system will be sufficient to support
+any provenance events in the future. See the [Operations](#operations) section below for
+more details.
+
+## Operations
+
+In addition to providing audit-style trails for Dandisets and assets, this plan
+fulfills a secondary requirement to cut the `previous` links, freeing up
+non-current assets for garbage collection. As this will destroy the asset chains
+from which we might draw some limited historical information, we need to
+carefully design an operations plan that preserves maximal information while
+empowering the garbage collection work to continue.
+
+Because it is broadly impossible to glean all of the information needed to
+reconstruct the historical audit trail for existing and deleted Dandisets, we
+plan to *begin* collecting complete information with the implementation
+of this design, while **preserving** a database snapshot at that
+same moment so that, if we choose in the future, we may wish to reconstruct that
+partial historical record. We would be essentially splitting time into two
+periods: before-audit-trail (represented by the DB snapshot) and
+after-audit-trail (represented by live, ongoing collection of complete audit
+records).
+
+This plan does not negatively impact future efforts at building
+provenance-related features. If we need to generalize provenance events from the
+audit trail, we can do so from the after-audit-trail events as needed, and the
+before-audit-trail DB snapshot remains available for that purpose as well. In
+essence, our plan retains the maximum possible amount of information while
+freeing the system to perform new features like audit, provenance, and
+comprehensive garbage collection.
+
+In sum, this plan allows us to (1) implement audit into perpetuity sooner; (2)
+preserve the available record at the moment that goes into effect; and (3)
+enable the completion of garbage collection design and implementation.
+
+The sequence of events looks like this:
+
+1. Implement the design in this document as a pull request, but do not
+   merge/release it yet (upon merge and release, the Archive will begin
+   generating audit records).
+2. Put the system into maintenance mode.
+3. Take a DB snapshot and store it in a safe place (sponsored S3 bucket, with
+   access restricted to admins).
+4. Merge and release the PR.
+5. Put the system into production mode.
--- a/doc/design/audit-trail-mvp.md
+++ b/doc/design/audit-trail-mvp.md
@@ -30,9 +30,9 @@ collaboration.
     
     ```
     YYYYMMDDTHHMMSS.mmmm: X added asset at path (/full/path) checksum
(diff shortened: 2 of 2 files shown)


doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+
+   A possible example of such a display follows (the final rendered form may
+   differ from this depending on best practices, etc.):
+    
+    ```
+    YYYYMMDDTHHMMSS.mmmm: X added asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X modified asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X deleted asset at path (/full/path) checksum
+    YYYYMMDDTHHMMSS.mmmm: X modified dandiset metadata
+    ```
```
bendichter, line 36: Would this be one row per file? That could generate a lot of row if a dandiset has many files. I don't think this is necessarily a bad thing, but if we do it this way we should plan for this, e.g. potentially paginating the rendered log

waxlamp: Pagination will definitely be part of the display. This snippet is meant to show what the log will be like (independent of how we will display it).

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+This section provides some analysis of the historical data living in the
+"`Asset` chains" kept in the `Asset.previous` links. The immediate goal is to
+determine how much "significant action" data can be reconstructed therefrom;
+there is also some analysis of how this may affect provenance work in the
+future.
+
+For context: because most changes to an `Asset` model cause a new model instance
+to be created, and because that new `Asset` instance points to its "ancestor"
+via a `previous` property, the history of an asset is at least partially
+traceable through this "asset chain".
```
bendichter, line 72: I don't really understand this section. I suppose you are talking about previous versions of assets that belong to published dandisets to these previous versions are still kept around even as the asset is modified? Is that correct?

bendichter: What sorts of "analysis" are you thinking of?

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Audit
+
+| Action | Recoverable fields | Unrecoverable fields | Notes |
+| --- | --- | --- | --- |
+| Create Dandiset | date_time, dandiset_id | user_id, metadata | date_time is equal to the Dandiset’s created_at value. user_id and metadata (as they were at the beginning of the Dandiset’s lifecycle) cannot be known with certainty.<br><br> Dandisets that have already been deleted will not have any recoverable fields (though we believe it very unlikely that a Dandiset has ever actually been deleted).<br><br> Independent of asset chains. |
+| Add owner | none | date_time, user_id, dandiset_id, owner_id | These events are all "summed" into the current owners list, with no way to know the evolution thereof.<br><br> Independent of asset chains. |
```
bendichter, line 79: Is this indicating that we wouldn't know the time at which an owner is added? Why is that?

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+
+|  | General data | Specific data |
+| --- | --- | --- |
+| Create Dandiset | date_time, user_id, dandiset_id | metadata |
+| Add owner | date_time, user_id, dandiset_id | owner_id |
+| Remove owner | date_time, user_id, dandiset_id | owner_id |
+| Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id |
+| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata |
+| Remove asset | date_time, user_id, dandiset_id | asset_path |
```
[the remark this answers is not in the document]
waxlamp: See 7080456d6d8de62a546f38d70c6afddc6d0fd4d9. I retained "delete" for Dandisets themselves, but assets should now be "remove" and "update".

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+future.
+
+For context: because most changes to an `Asset` model cause a new model instance
+to be created, and because that new `Asset` instance points to its "ancestor"
+via a `previous` property, the history of an asset is at least partially
+traceable through this "asset chain".
+
+### Audit
+
+| Action | Recoverable fields | Unrecoverable fields | Notes |
```
satra, line 76: not sure what unrecoverable means?

doc/design/audit-trail-mvp.md
waxlamp: This is more of a technical quirk that presaged this feature (and provenance concerns).

Essentially, we're doing something kind of odd whenever an `Asset` is changed: we manufacture an entirely new `Asset` object, then refer to the original `Asset` object through the new `Asset`'s `previous` value. This was done to attempt to preserve the historical record, and this section analyzes how well that has worked, and what we can do to shed those old `Asset` objects via garbage collection.

This is different from hanging onto assets that belong to published versions--those will not be affected by this design.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+about those changes. In essence, it is a tool to support effective
+collaboration.
+
+## Requirements
+
+1. **Persistent tracking of significant actions.** Whenever a *significant
+   action* (see list [below](#significant-actions)) occurs successfully, the archive will generate a
+   persistent record of that action, including the following data:
+    - The date and time
+    - The authorized user who triggered the action
```
satra, line 20: just a note here that the authorized user can be a backend task (i.e. the archive). for example when an asset summary is updated or checksum is computed.

doc/design/audit-trail-mvp.md
waxlamp: > What sorts of "analysis" are you thinking of?

The analysis is here in this design doc, not something that becomes a feature of the system (apologies if that wasn't clear). In particular, this analysis has led to the Operations plan detailed in the final section of the doc. The main upshot there is that if we take a snapshot of the database as it is at the moment we turn on the new audit feature, we can still, at any point in the future, reconstruct whatever history is recoverable from that snapshot--or we can just live with the fact that we don't have a complete history before that moment. It's up to us, since there's no loss of information and therefore no urgency.

doc/design/audit-trail-mvp.md
waxlamp: In this context, "unrecoverable" means "unrecoverable from the historical record".

I may not have made this section of the doc clear enough. I can try to clarify it, or we can try to resolve questions here and/or in a quick meeting. Let me know what you prefer.

doc/design/audit-trail-mvp.md
waxlamp: Is there a reason users (i.e., Dandiset owners) would want to know when these things are computed/updated?

satra: knowing that something has changed in the dandiset is important. as currently listed one should consider any modification a significant event (independent of agent). we can/should list examples of non-significant events if we currently think that a user wouldn't care, and we can discuss those to moved from one pile to another as needed. 

for the future, we will continue to add tasks that will make changes, some more important than others (e.g., generating a derived asset, altering metadata, etc.,.). hence i already want to start having a notion that an internal action is still an action that should be recorded.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+    - The ID of the Dandiset on which the action occurs
+    - Event-specific data about the action itself
+
+2. **Display of audit log (to owners and admins only).** The Dandiset landing
+   page will display a rendered (i.e. human-readable) version of the log event
+   data (as long as the viewing user is an owner of the Dandiset, or an admin).
```
kabilar, line 26: Thanks @waxlamp.  It might be useful for the audit logs to be public to all users.  For users performing downstream analysis, they might be interested in understanding the changes between versions.  Similar to being able to traverse a git commit history. 
 Is there a privacy reason that we wouldn't want to make the logs public?  Perhaps this suggestion would be beyond the MVP?

waxlamp: We have a feature in the design phase (well, pre-design phase, meaning it's on our minds) to enable owners to bundle release notes with a published Dandiset version.

I would argue that we need to keep this audit log private for now, until we have a chance to analyze the privacy considerations of making it public, as well as determining exactly why we would want to make this public. As conceived currently, this is a mechanism for enabling communication and collaboration between Dandiset owners, so conservatively we're keeping this private. Put another way, "might be useful" doesn't feel a strong enough reason to make this public at the moment.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+| Update asset | date_time, dandiset_id, asset_path, checksum, metadata | user_id | As above, the user_id information is missing, and the dandiset_id may be as well. |
+| Remove asset | asset_path | date_time, user_id, dandiset_id | A removed asset will be one at the head of an asset chain that has no association to a Dandiset. Unfortunately, without that association, the dandiset_id cannot be recovered, and because it is the absence of a relation that indicates removal, the date_time value is also unrecoverable. |
+| Publish Dandiset | date_time, dandiset_id, version | user_id | The date_time, dandiset_id, and version are all available as part of the Version instance’s metadata. |
+| Delete Dandiset | none | date_time, user_id, dandiset_id | Deleted Dandisets just don’t exist anymore, so no information (even the dandiset_id) is available. |
+
+Note that the majority of information describing these auditable actions is
```
aaronkanzer, line 88: Is there a reason that an audit log could not be preserved and provided for deleted dandisets?

Deletion seems important to know which user may have deleted the set when (assumingely downstream users of a dandiset would want to know if it no longer exists as well -- especially for scientific reproducibility)

Curious to know your thoughts here -- if there is a way of noting historical evidence of a given dandiset, even though the dandiset's data is no longer available.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+In sum, this plan allows us to (1) implement audit into perpetuity sooner; (2)
+preserve the available record at the moment that goes into effect; and (3)
+enable the completion of garbage collection design and implementation.
+
+The sequence of events looks like this:
+
+1. Implement the design in this document as a pull request, but do not
+   merge/release it yet (upon merge and release, the Archive will begin
+   generating audit records).
+2. Put the system into maintenance mode.
```
aaronkanzer, line 148: Is there a doc for what mechanism will be used to implement this system? (e.g. is the audit log maintained by some Django super.save() call? Inspecting the incoming auth token to infer the user, etc. etc.?)

doc/design/audit-trail-mvp.md
kabilar: Thanks @waxlamp.  I agree.

doc/design/audit-trail-mvp.md
waxlamp: My comments in this part of the doc are about recovering _historical_ events from the existing record. Indeed, it is possible, valuable, and in our plans to record deletion events and keep them in the audit record for a given Dandiset. That is, the Dandiset may be deleted, but we'll still have the full history of that Dandiset, even though the data itself is gone.

It may be worth noting that deleting Dandisets is not a generally available action: it is only possible for Dandisets that have not published any versions (with a major exception for emergency situations, such as a lab accidentally publishing something they should not have, etc.; in these situations we intervene to solve the problem out-of-band).

doc/design/audit-trail-mvp.md
waxlamp: There is domain knowledge about the codebase in terms of internal services we would use to identify, capture, and export this information to the database. It is too fine-grained to include in this doc, which largely is concerned with the concepts, requirements, and operations concerns involved.

If you are interested in these finer details, you can take a look at the eventual implementation, or else I would be happy to meet with you to discuss.

aaronkanzer: I'll stay tuned for the implementation then -- thanks Roni!

doc/design/audit-trail-mvp.md
waxlamp: @satra, since I've answered the question (and clarified what this section is supposed to be), I'll consider this resolved. You can unresolve and continue the thread if you disagree.

doc/design/audit-trail-mvp.md
waxlamp: See de07b152c51ffc139a9a61f65057b6cf3bd64c0a.

doc/design/audit-trail-mvp.md
waxlamp: (@bendichter, I'm going to resolve this conversation; please unresolve and continue the conversation if you want to discuss further.)

.codespellrc
```diff
@@ -1,5 +1,5 @@
 [codespell]
-skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log
+skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log,./doc
```
yarikoptic, line 2: why do we want to allow typos in `doc/`? https://github.com/dandi/dandi-archive/pull/1801/commits/e132b29d74ba12d76c9db51f4568dd90334d3a87 gives no reasoning.

yarikoptic: I see -- it was because of the hit on the rare "therefrom" as a single word. Ok, we can either exclude it explicitly via

```patch
diff --git a/.codespellrc b/.codespellrc
index 624b90fd..edc3ab29 100644
--- a/.codespellrc
+++ b/.codespellrc
@@ -3,4 +3,4 @@ skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md
 # Disabled until https://github.com/codespell-project/codespell/issues/2727
 # got answer/re-solution
 # dictionary = .codespell_dict
-# ignore-words-list = 
+ignore-words-list = therefrom
```

or just decide for us to not use "rare" dictionary and only the one with clear typos:

```patch
diff --git a/.codespellrc b/.codespellrc
index 624b90fd..87d72897 100644
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,7 @@
 [codespell]
-skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log,./doc
+skip = .git,node_modules,dist,venv,.venv,.mypy_cache,.tox,yarn.lock,CHANGELOG.md,./htmlcov,*.pyc,*.log
 # Disabled until https://github.com/codespell-project/codespell/issues/2727
 # got answer/re-solution
 # dictionary = .codespell_dict
+builtin = clear
 # ignore-words-list = 
```

I am ok with either with smaller preference to former as of being specific (I rarely run into those "rare" words)

waxlamp: codespell, from its own README, is "designed primarily for checking misspelled words in source code". And despite the claim that "it shouldn't generate false-positives when you use a niche term it doesn't know about", it complained about the word "therefrom", which is very much a properly spelled English word.

More generally, I don't really care if typos creep into design documents--that generally won't impede the use of design docs for the reasons we want to use them, and I think preventing merge for reasons of a misspelling is much more trouble than it's worth, especially when codespell is in fact giving us false positives like it did here.

waxlamp: codespell by default excludes "rare" words? Why?

If we use that `builtin = clear`, it will not flag words like "therefrom"? If that's the case, I am willing to try that, though I remain skeptical of the value of slowing down the writing of design docs because of CI-based spellcheck.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+collaboration.
+
+## Requirements
+
+1. **Persistent tracking of significant actions.** Whenever a *significant
+   action* (see list [below](#significant-actions)) occurs successfully, the archive will generate a
+   persistent record of that action, including the following data:
+    - The date and time
+    - The authorized user who triggered the action (note that this may be an actual user or the system itself for internal automated actions)
+    - The ID of the Dandiset on which the action occurs
```
yarikoptic, line 21:
```suggestion
    - The ID of the Dandiset on which the action occurs (if performed on a dandiset endpoint or has dandiset as the parameter)
```

since could potentially be some operation on a blob or asset without

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+    - Event-specific data about the action itself
+
+2. **Display of audit log (to owners and admins only).** The Dandiset landing
+   page will display a rendered (i.e. human-readable) version of the log event
```
yarikoptic, line 25:
```suggestion
   page will provide ability to display a rendered (i.e. human-readable) version of the log event
```

since I do not think we want to actually display the log, may be just the last action associated with the dandiset, clearly not more than that. But we would want some interface (e.g. a modal window) to display/navigate and potentially search

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+
+1. **Persistent tracking of significant actions.** Whenever a *significant
+   action* (see list [below](#significant-actions)) occurs successfully, the archive will generate a
+   persistent record of that action, including the following data:
+    - The date and time
+    - The authorized user who triggered the action (note that this may be an actual user or the system itself for internal automated actions)
+    - The ID of the Dandiset on which the action occurs
+    - Event-specific data about the action itself
+
+2. **Display of audit log (to owners and admins only).** The Dandiset landing
```
yarikoptic, line 24: Don't we want to gain API (hence a separate item before this one) to query the audit log and then "display" would be based on that?

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+"significant" (i.e., require an audit record to be generated), along with any
+additional, event-specific data each will carry with it.
+
+|  | General data | Specific data |
+| --- | --- | --- |
+| Create Dandiset | date_time, user_id, dandiset_id | metadata |
+| Add owner | date_time, user_id, dandiset_id | owner_id |
+| Remove owner | date_time, user_id, dandiset_id | owner_id |
+| Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id |
```
yarikoptic, line 55: I wonder if we should care about addition (upload) of blobs or zarrs as a separate item?

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata |
```
yarikoptic, line 56: in both above it could be `blob_id` or `zarr_id`...

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+| Remove asset | date_time, user_id, dandiset_id | asset_path |
+| Publish Dandiset | date_time, user_id, dandiset_id | version |
+| Delete Dandiset | date_time, user_id, dandiset_id | none |
```
yarikoptic, line 59: @jwodder do you think the list of actions above, and data for them would be sufficient for `backups2datalad` to perform synchronization to at least the level it does now, or some data would be missing?

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+The following is the list of actions performed on Dandisets that are considered
+"significant" (i.e., require an audit record to be generated), along with any
+additional, event-specific data each will carry with it.
+
+|  | General data | Specific data |
+| --- | --- | --- |
+| Create Dandiset | date_time, user_id, dandiset_id | metadata |
+| Add owner | date_time, user_id, dandiset_id | owner_id |
+| Remove owner | date_time, user_id, dandiset_id | owner_id |
+| Update Dandiset metadata | date_time, user_id, dandiset_id | metadata |
```
yarikoptic, line 54: if for 
image
should we also get `version` into "specific data"?

```suggestion
| Update Dandiset metadata | date_time, user_id, dandiset_id | version, metadata |
```

doc/design/audit-trail-mvp.md
yarikoptic: also if we are still doing CoW, and 
image
then as a result of "Update asset" we actually are creating a new asset UUID.  So I think it might be valuable here to have
```suggestion
| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
```
thus mimicing above for update in that `asset_id` is the id of the created asset.

doc/design/audit-trail-mvp.md
yarikoptic: Also since API for assets more about asset ids than paths (paths don't uniquely identify assets through time), I think it would be useful to gain 
```suggestion
| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id |
```
where `asset_id` is the id of the Add'ed asset

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id |
+| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata |
+| Remove asset | date_time, user_id, dandiset_id | asset_path |
```
yarikoptic, line 57:
```suggestion
| Remove asset | date_time, user_id, dandiset_id | asset_path, asset_id |
```

for the reason stated above

yarikoptic (changes requested):
Some initial changes, yet to finish reading.

doc/design/audit-trail-mvp.md
jwodder: The API doesn't appear to currently provide any mutating operations on blobs or assets without the context of a Dandiset.  I don't think it even could with the current architecture, as any change to an asset or blob produces a new asset with a new identifier, and in order for the operation to not be a waste of time, this new asset needs to be associated with one or more Dandisets.

doc/design/audit-trail-mvp.md
jwodder: `backups2datalad` also needs exact information on:

* Creation, replacement/modification, and deletion of individual entries within a Zarr
* Unembargoing of Dandisets

Also, although `backups2datalad` doesn't currently use this, I think deletions of Dandiset versions should be recorded as well.

doc/design/audit-trail-mvp.md
jwodder, line 54: Shouldn't this be "Update version metadata"?  Dandisets don't have metadata of their own.

doc/design/audit-trail-mvp.md
jwodder, line 56: If the update involves moving the asset, then the old asset path should be logged as well.

doc/design/audit-trail-mvp.md
waxlamp: Why do you think we don't want to display the log? If you read the motivation section, you'll see that we're conceiving of this as an aid to Dandiset owners (to see e.g. who made specific changes to the Dandiset, etc.), which would require displaying the information. This reflects Satra's comments and the sample log display he mocked up.

doc/design/audit-trail-mvp.md
waxlamp: That's not an absolute requirement. The API to deliver the display results could be private, for example. And admin users will, for instance, be able to see audit records using the admin console.

If we actually want *public* API to be a part of this, we should discuss why we want that, and then we'll need to add it to the requirements. But my understanding is that the admins want to be able to see the audit events for debugging etc., and the display could be offered to Dandiset owners as a feature.

To avoid extending this discussion (and the implementation phase) further, I would argue for adding a public API later if needed for some reason.

doc/design/audit-trail-mvp.md
waxlamp: > I wonder if we should care about addition (upload) of blobs or zarrs as a separate item?

Conceptually, no. It's a (rather dirty) implementation detail that we handle Zarrs through different code paths than other assets at the moment, and I don't think we should allow that to infect the concept of audit events.

doc/design/audit-trail-mvp.md
waxlamp: We do need to add embargo/unembargo events to the list so those operations can be part of the event log.

Though I said in a different thread that I don't think we should make a special case for Zarr operations, perhaps I am wrong (since, as you're pointing out, there are indeed special ways in which Zarrs are manipulated). Still, it doesn't seem quite right, and I want to be careful not to overspecify the audit feature to work precisely with `backups2datalad`, lest we lose sight of "audit", which is the real goal here.

If this is important, I think we should discuss it in a meeting.

doc/design/audit-trail-mvp.md
waxlamp: If you look at https://github.com/dandi/dandi-archive/blob/880a1022c590e4ef6606e0f6a00e16a98e90da9b/dandiapi/api/views/version.py#L90-L95 you'll see that only `draft` versions can have their metadata updated, hence no need to record the version.

waxlamp:
```suggestion
| Update Version metadata | date_time, user_id, dandiset_id | metadata |
```

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,156 @@
...
+additional, event-specific data each will carry with it.
+
+|  | General data | Specific data |
+| --- | --- | --- |
+| Create Dandiset | date_time, user_id, dandiset_id | metadata |
+| Add owner | date_time, user_id, dandiset_id | owner_id |
+| Remove owner | date_time, user_id, dandiset_id | owner_id |
+| Update Version metadata | date_time, user_id, dandiset_id | metadata |
+| Add asset | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id |
+| Update asset | date_time, user_id, dandiset_id | asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
```
waxlamp, line 56:
```suggestion
| Update asset | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
```

doc/design/audit-trail-mvp.md
waxlamp: See 5e7b1e286ba808f660ea0487ce32c213da72e11d.

doc/design/audit-trail-mvp.md
satra: given the second sentence above in the opening paragraph, wouldn't an implementation detail of this be an API end point? since owners only interact via the api.

waxlamp: Yes, I think that's what I was trying to get at: implementation details are not requirements; if we want to have a public API for audit data, that would be a requirement, and it would mean we need to do more design to get it right. I'm arguing for allowing whatever API is necessary for this MVP to remain private, and therefore not trigger any extra design before we can get to implementation.

If, in the future, we decide it needs to be a public API, we can do the work at that point to make it public. In the meantime, owners and admins will be able to browser the audit logs in-app, and admins will additionally be able to search/view/etc. from the admin console.

jwodder: While the question of whether the display of the audit log in the web interface should be based on some API is an implementation detail that has no place in a requirements doc, I think there should still be some API endpoint(s) for accessing ths logs (regardless of how that ends up relating to the web interface), as otherwise `backups2datalad` won't be able to benefit from this.

yarikoptic: I could argue that "Display of ... on DLP" is an implementation detail of a "Provide access to ".
I think that overall hiding the need for designing API access could foster implementing some solution which would workaround that need and thus becoming hard to change later on.  If the concern is not making API private, could be stated explicitly  that API in this MVP would remain "admin-only" but the point should be that there would be an API end-point and that "Display on DLP" should use that API as a requirement to avoid possible ad-hoc DLP-only solution.

doc/design/audit-trail-mvp.md
yarikoptic: because it would clutter the display and generally would not be of interest to anyone. For the same reason github repository page does not list you the log -- you need to click on "commits" or alike to get to it.  But GitHub does display the latest day an item was modified at to give user ability to assess how stale the project is.  Hence my motivation to not bother displaying the full log but rather state the last modification date and give ability to list entire log in some modal window.

But anyways -- feel free to ignore since it is a minor and UX aspect.

PS I missed the sample log display I guess.

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,157 @@
...
+
+| Action                  | General data                    | Specific data                                                                   |
+| ---                     | ---                             | ---                                                                             |
+| Create Dandiset         | date_time, user_id, dandiset_id | metadata, embargoed                                                             |
+| Add owner               | date_time, user_id, dandiset_id | owner_id                                                                        |
+| Remove owner            | date_time, user_id, dandiset_id | owner_id                                                                        |
+| Update Version metadata | date_time, user_id, dandiset_id | metadata                                                                        |
+| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, asset_id                               |
+| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, metadata, asset_id, old_asset_id |
```
yarikoptic, line 56: I thought I left the comment before but I fail to find resolved conversation, sorry if repeating but seems this does not acknowledge the fact that we have now a multitude of blob vs zarr. So either we need 
```suggestion
| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, blob_id, zarr_id, asset_id                               |
| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, blob_id, zarr_id, metadata, asset_id, old_asset_id |
```

or (more flexible/scalable for the future possible addition of some other beastie)

```suggestion
| Add asset               | date_time, user_id, dandiset_id | asset_path, checksum, metadata, object_id, object_type, asset_id                               |
| Update asset            | date_time, user_id, dandiset_id | asset_path, old_asset_path, checksum, object_id, object_type, metadata, asset_id, old_asset_id |
```

where `object_type` could be `blob` or `zarr` (not sure if we have semantic already for this)

doc/design/audit-trail-mvp.md
```diff
@@ -0,0 +1,157 @@
...
+| Remove asset            | date_time, user_id, dandiset_id | asset_path, asset_id                                                            |
+| Unembargo Dandiset      | date_time, user_id, dandiset_id | (none)                                                                          |
+| Publish Dandiset        | date_time, user_id, dandiset_id | version                                                                         |
+| Delete Dandiset         | date_time, user_id, dandiset_id | (none)                                                                          |
+
```
yarikoptic, line 61:
```suggestion

Note that we explicitly are not aiming in this MVP to audit individual Blob and Zarr uploads.

```

if that is the case... 

But my question is: from the DB or other logs -- would we be able to figure out which user_id has uploaded a blob with porn which we are now happily serving through our `/blobs/digest` endpoint?

doc/design/audit-trail-mvp.md
waxlamp: > generally would not be of interest to anyone

I am really surprised to see this--the entire motivation for this was to provide helpful information to Dandiset owners (again, see the motivation section of the doc). If that's not the case, why are we even implementing this?

Because I'm getting super mixed signals about this, I'd like to get @satra's take. If we don't want to include the display in the frontend, I'm happy to remove that from the requirements, but I really thought that was a goal of this feature.

> list entire log in some modal window

This would be the design approach. I was not planning on including the full log in the DLP or something like that.

yarikoptic: nevermind @waxlamp - we can treat my suggestion as a nit pick and UI design specific. I only argued about how it would be communicated, not that it would not communicated at all.

doc/design/audit-trail-mvp.md
yarikoptic: zarr: relates to the disclaimer I suggested to add that we do not want in this MVP to cover blob/zarr level information https://github.com/dandi/dandi-archive/pull/1801/files#r1467945249 . Without such explicit "scoping" statement MVP then IMHO should include blob/zarr level basic audit

unembargo: TL;DR -- should we actually remove "Unembargo Dandiset" and rely on changes to metadata?

Rationale:   given that we doing "asset level" audit and unembargoing is to happen (in [redesign](https://github.com/dandi/dandi-archive/pull/1802/files#diff-f3d72d358571f346f0a3e26e3bf309127d6af6aac1e39f60e291b947d86eeb19R132)) on assets level, it feels like we might want to add "Unembargo asset".  But "embargoed" is also a state within metadata, and we already have "Update asset" with metadata being recorded, so at least we will have information collected.

But then should we make it symmetric at Dandiset level where ATM we have two "Update Version metadata" and "Unembargo Dandiset". Should we remove "Unembargo Dandiset" and rely on discoverying that action through changes to its metadata?

doc/design/audit-trail-mvp.md
yarikoptic: ATM - yes. But I thought we had already cases where we - admins - either wanted or did update metadata.  We might have not gone through this endpoint though.  We have issues for updates of metadata on assets (https://github.com/dandi/dandi-archive/issues/1450) and discussed in the past aspects of metadata migration (https://github.com/dandi/dandi-archive/issues/565)...

But I guess we could keep as is, and get to change schema as we get to that point of need.

waxlamp: I would say let's keep this as-is for now. I'm pretty sure we didn't go through this endpoint to update metadata on published versions (since, as I showed above, that simply wouldn't have worked), and if we want to record a special audit event for metadata migration, we can add that in as part of the work on #565.

doc/design/audit-trail-mvp.md
waxlamp: We will likely want to record that an asset went from hidden to open, etc., but since that's not a capability in the system yet, we can wait until that lands to design its appropriate audit events.

I think keeping "unembargo dandiset" as its own significant action is important. We do indeed record that state in the metadata, but that doesn't mean the event needs to be swallowed up in the "update metadata" event -- it is special enough to warrant the event.

Issue #1787, opened 2023-12-14
dandi:EmbargoedAccess metadata can be silently inconsistent with the actual embargo status


yarikoptic:
https://dandiarchive.org/dandiset/000537/draft says

![image](https://github.com/dandi/dandi-archive/assets/39889/6d6f3057-8cc5-45ef-97f9-adcab8ecde4f)

but I can navigate and access data files without login

![image](https://github.com/dandi/dandi-archive/assets/39889/4acad058-cf29-4b51-a9dd-1f0691ae5360)

so either metadata is incorrect and that is silently ok (in validation errors there is only "access: An embargo end date is required for NIH awards to be in compliance with NIH resource sharing policy.") or embargoed dandiset is publicly accessible?

edit: good candidate for yet another fsck style check

yarikoptic:
FWIW, according to the "dandiset record" (not metadata) we seems to have no embargoed dandisets at all ATM

```shell
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/dandisets/?page_size=1000&draft=true&empty=true&embargoed=true' -H 'accept: application/json' | jq . | grep -e embargo_status | sort | uniq -c
    350       "embargo_status": "OPEN",
```

and for that one it does list as OPEN too... so must be a discrepancy between status and metadata -- should have not metadata been changed after it went public?  there definitely must be some WARNING if such incongruency is found, and it must be on dandi-archive level since `dandi-schema` itself would not know DB level status.

dandibot:
:rocket: Issue was released in [`v0.3.86`](https://github.com/dandi/dandi-archive/releases/tag/v0.3.86) :rocket:

PR #1802, opened 2024-01-02
Add quirks section to embargo redesign doc

waxlamp:
This PR extends the embargo redesign with some odd edge/corner cases that complicate the semantics of embargo. We need to have some discussion in order to resolve the questions of how the feature should behave in this situations.


doc/design/embargo-redesign.md
```diff
@@ -116,3 +116,49 @@ sequenceDiagram
...
+
+**Quirk.** Dandiset 2, though embargoed, contains an asset that is open and therefore world-readable. In spirit this contradicts the embargo on Dandiset 2.
+
+**Remediation.** None needed, despite the apparent contradiction. Though Dandiset 2 contains a world-readable asset, no one who is not authorized to view the contents of Dandiset 2 knows that fact; for all they know, any data at all, *including open assets stored in the archive*, could be in that Dandiset. Thus, no excess information about Dandiset 2 is leaked.
+
+### Open Dandiset includes embargoed asset
+
+**Scenario.** Roughly the opposite of the last scenario: someone creates an *embargoed* Dandiset 3, uploading asset B to it. They then create an *open* Dandiset 4, also uploading B to that.
+
+**Quirk.** Dandiset 4 now contains a hidden asset, which contradicts the openness of Dandiset 4 and seems to compromise the secrecy of Dandiset 3.
```
yarikoptic, line 136: after re-reading original design above I noted that although it describes technical aspect of the solution on S3 level, it does not cover at all annotation of embargoed "stuff" within archive DB, and might be worth clarifying on what  exactly (dandiset? asset? blob? zarr? all of that) and how (metadata? extra DB column? both) is annotated to be either open or embargoed?  I guess intention is to just keep it as is now (EmbargoedAssetBlob, EmbargoedZarrArchive + metadata annotation of embargoed dandiset), but feels like a missing detail here which instructs implementation logic and potentially these use-cases / quirks.

doc/design/embargo-redesign.md
```diff
@@ -116,3 +116,49 @@ sequenceDiagram
...
+
+**Remediation.** In this case, it seems that we need to allow for unembargo and fall back to the situation described above, where access to the asset is granted through presigned URLs, without otherwise disclosing the fact that the asset is still present in another embargoed Dandiset. A warning to the user on their attempt to unembargo may be warranted so they can decide if they want to expose that asset while it is still under embargo in the other Dandiset.
+
+Finally, this brings up a question about when to remove the `embargoed` tag from the object in the S3 bucket: we would insert logic to detect whether the embargoed asset is included in any other embargoed Dandisets, and only trigger the tag removal if there are none. This brings up one major question:
+
+- **Do these situations nullify the effect/need to keep data hidden in the bucket at all?**
+The answer to this is probably “no”, but some discussion may help to illuminate answers to the other questions and quirks above.
```
yarikoptic, line 164: those are great user stories!  Thank you @waxlamp .  I think we should just generalize them into some concise "design decision" which would inform on what is to happen in any of them as a result/remediation.  Echoing the first one I would say it could be

- Anyone with access to EMBARGOED content can make it OPEN
  - I think this would be the crucial statement which would offload responsibility to the owner (or someone with access) to embargoed data.  May be a suboptimal example, but it is similar to "leaking the data" -- as soon as someone leaks private data, it becomes known publicly.  The same here -- if someone with access to EMBARGOED data decides to upload it publicly, they can.
  - We indeed would want interface(s) to make such transitions from EMBARGOED to OPEN to be explicit. I see two "entry points"
     - Explicit "make content of a dandiset (or prefix within it? e.g. a participant)" OPEN -- that goes to "Object Tag Removal Workflow".
     -  Make content OPEN via upload.
  - If we do not assume that checksums are private data -- we would need to verify that the person indeed has access to EMBARGOED content to make it OPEN upon upload. We would need a (missing ATM) workflow here which instead of minting a new blob would just make EMBARGOED to become OPEN upon verification of upload digest.

- EMBARGOED dandiset can point to OPEN content
- EMBARGOED content can be made OPEN by any authorized user
- Only a dandiset without assets pointing to EMBARGOED content can be announced to be OPEN
   - so, explicit action: UNEMBARGOING of dandiset consists of unembargoing any still embargoed content first
   - UNEMBARGOING some content might make one or more dandisets available to transition to OPEN without unembargoing more conent
- Upload to EMBARGOED dandiset makes new to the archive content EMBARGOED
- Upload to OPEN dandiset makes content (new or already known to archive) OPEN
  - that is where we might want to indeed add some explicit option to allow for unembargoing of embargoed content, since otherwise people might get a little pissed after someone from their group would "leak" the embargoed data.

But in general it boils down to be able to inform user on the effects of potential unembargo -- would we be able to say how many dandisets would be affected? how many would become OPEN (as in content)?

yarikoptic (commented):
Some initial thoughts -- I think we might want just to formalize some rules/assumptions over embargoed content and dandisets and points of interaction with the user where UNEMBARGOing might happen (explicit request for known asset(s) for a given set of assets, or during upload of a new blob/zarr)

doc/design/embargo-redesign.md
waxlamp: I sort of agree with you, but I see it a little bit differently: I don't want to formalize what it means for "assets to be embargoed", because "embargo" is a very specific application-level concept that models *NIH-authorized embargo* for a very specific situation. In that sense, how we are keeping such embargoed Dandisets' contents hidden from public view is an implementation detail, and I think it would actually be good to use a different term than "embargoed" to describe the state of such assets.

Where we agree (I believe) is the need to formalize what "embargo" means (in my opinion, it is simply the state of a Dandiset which includes its contents and existence being hidden from unauthorized users).

As for implementation: yes, I think we are planning to keep the current models in place, only changing them if (1) the changes improve the code and (2) the requirements/behavior of the system are not changed. (This is essentially what it means to be an "implementation detail".)

yarikoptic: ref: https://github.com/dandi/dandi-archive/issues/1787 a somewhat related since details of implementation (either to be a part of this design doc or not) and e.g. interaction with metadata remain not formalized in a design doc, or is it?

satra: >  it is simply the state of a Dandiset which includes its contents and existence being hidden from unauthorized user

i would strike "and existence". it does not matter from an embargo perspective whether dandiset 2024 exists or not. simply knowing that no content about it is exposed is sufficient.

embargo/restricted - only certain authorized individuals/entities have access to that resource (whatever the resource may be).

perhaps a few additional things to note:
- asset ids and blobs are decoupled
- blob ids and blob checksums are decoupled and currently fall under deduplication principles rather than embargo principles. 

thus from an implementation perspective:
- a user should be notified if a blob being uploaded to an embargoed dandiset already exists in public view (could easily be possible for a host of reasons). i believe the CLI doesn't allow 0 length files. i'm not sure the API has any restriction. 
- one needs to consider as @yarikoptic wrote as to what happens when these two worlds collide. same blob uploaded to two different embargoed dandisets (also completely possible). current deduplication rules would not allow a different id. 

both `embargo` (being discussed here) and `restricted` (to be implemented soon) are resource specific concepts. as such i would really bring rebac into the picture. the resources in consideration are dandisets, dandiset metadata, assets, and blobs. whether we allow access control on all of these elements is a policy decision, but implementation wise we should consider them as separate resources. dandisets can contain mixtures of access assets for whatever reason and several users have asked for asset level granularity for access for different use cases. a lot of the considerations in quirks would be covered under rebac, which turns things into policy decisions for access.

waxlamp: Could you expand the acronym "rebac"? (I have seen it stand for different things and I want to disambiguate what you are referring to.)

> dandisets can contain mixtures of access assets for whatever reason and several users have asked for asset level granularity for access for different use cases.

I was really hoping to avoid this type of generality, especially without hard requirements on how this would work. My preference is to work from well-defined concepts such as "NIH embargo" or "restricted access" (the latter is not super well defined but it is at least a domain-level concept) in order to deliver those chunks of features to users.

I understand that users have asked for the fully general idea of mixing access levels of assets, but if we don't understand the actual software requirements here (inlcuding some of the use cases those users are trying to implement), this is a perfect recipe for half-baked implementations plus subsequent rework that we have been desperately trying to avoid.

I am happy to discuss this direction further, but in the current design I would like to pin down how embargo should work in the face of these "quirks" (and some shadow of future capabilities we want to develop). I think we are well on the road to accomplishing this, but I wanted to draw this boundary so that we don't get bogged down in trying to understand three things at once (i.e., embargo, restricted, general asset access).

doc/design/embargo-redesign.md
waxlamp: > We indeed would want interface(s) to make such transitions from EMBARGOED to OPEN to be explicit. I see two "entry points"
> - Explicit "make content of a dandiset (or prefix within it? e.g. a participant)" OPEN -- that goes to "Object Tag Removal Workflow".
> - Make content OPEN via upload.

I think the second option is the right one. It is simpler, and it leaves the idea of "embargoed Dandiset" whole (indeed, the other way to make assets in an embargoed Dandiset OPEN is, quoting from your response, *so, explicit action: UNEMBARGOING of dandiset consists of unembargoing any still embargoed content first*, which also considers "embargo" to be a state on a whole Dandiset).

Everything else you wrote, I believe I understand and agree with. I will rework the PR text to reflect what you've said here and we can do a review round to make sure we agree on how it all should work.

doc/design/embargo-redesign.md
satra: rebac --> resource based access control

regarding use cases:

- ML leaderboards are what has driven most of the user requests on granular access, specifically to create training, validation, and multiple test data sets. the current solution is simply to have multiple dandisets.  
- Embargoed master data with periodic releases of subsets of data associated with papers. again doable from the current setup, not efficient or easy to do from a user perspective.

> so that we don't get bogged down in trying to understand three things at once (i.e., embargo, restricted, general asset access).

i'm fine with not getting bogged down, but if the entire thing will require a complete overhaul again when we implement restricted/general asset access, i would rather have that be discussed in this context.

doc/design/embargo-redesign.md
yarikoptic: I might be misunderstanding you saying "the right one" here. I do not think of these two options as mutually exclusive alternatives - rather they seems to be the two ways we should allow for making content OPEN.  Are you suggesting it ('make content OPEN via upload') to be the only way to make content OPEN?

waxlamp: > Are you suggesting it ('make content OPEN via upload') to be the only way to make content OPEN?

This is indeed what I was suggesting. My reasons for this:
- simplifies the mental model for the user--in particular, the user has less to worry about in terms of managing what is open and what is hidden
- keeps the concept of embargo simpler: a Dandiset is either embargoed or not, and upon performing the "unembargo" action, all contained assets become open

Much of this comes from trying to control complexity, even if in the future we are going to move to that hairier model where any kind of assets can be present in any Dandiset (and the simplicity of a binary embargo state becomes fuzzier). So I may need to compromise here, but the cost is much greater complexity in what we're attempting to build (I just need to be open about this tradeoff).

doc/design/embargo-redesign.md
waxlamp: > i'm fine with not getting bogged down, but if the entire thing will require a complete overhaul again when we implement restricted/general asset access, i would rather have that be discussed in this context.

This is a fair position to take. FWIW, my intuition is that with the ability to hide/open assets already existing, we should be able to build out embargo, and "restricted assets", and the more general mixed-access Dandiset concepts.

For the latter two (which don't exist yet) we still have room to understand requirements and connect them to the existing capabilities of the system--I simply want to make sure we develop that full understanding before anything else (which includes discussions now about those requirements, etc.).

doc/design/embargo-redesign.md
yarikoptic: I don't think that would work since 
- it would pretty much bring back the original motivation for redesign - problem of Inefficiency , with all issues just offloaded to the uploader instead of our automations which now need to move data between buckets
- as a result, would make "UNEMBARGO"ing a very time consuming and unpleasant experience thus demotivating users from making their data OPEN.

I think I could argue also that the first reason you stated would not really be fulfilled, and the 2nd one remains pertinent to the proposed first "entry point" to unembaroing content - done directly on the archive:

> * Explicit "make content of a dandiset (or prefix within it? e.g. a participant)" OPEN -- that goes to "Object Tag Removal Workflow".

So may be we are not fully understanding each other here.

doc/design/embargo-redesign.md
yarikoptic: BTW -- isn't "EMBARGOED" is just a "RESTRICTED" with a dandiset-level rule for when "RESTRICTED" could be automatically be converted to "OPEN"?

satra: > isn't "EMBARGOED" is just a "RESTRICTED" with a dandiset-level rule

yup, just policies are different as to how access is granted. embargoed currently restricted to owners. for restricted, we will need both owners and viewers who have access. hence the consideration of rebac (resource based) + rbac (role based).

yarikoptic: I think model could be simplified if EMBARGOED would just become RESTRICTED with the policy for when it to become OPEN, and otherwise to have the same permission models as RESTRICTED, i.e. to have owners and viewers, where the viewers have no authority to change it, e.g. to make it OPEN through the .  I think it might be one of the possible ways collaborators on EMBARGOED dataset could leak it: through that [Explicit "make content of a dandiset"](https://github.com/dandi/dandi-archive/pull/1802#discussion_r1440043850) entry point I was talking about.  Sure thing they would still be able to make it OPEN through leaking via uploading them as OPEN to another OPEN dataset (unless we forbid such an action as @waxlamp excercised in [this remediation](https://github.com/dandi/dandi-archive/pull/1802/files#diff-f3d72d358571f346f0a3e26e3bf309127d6af6aac1e39f60e291b947d86eeb19R140), but I think it would be potentially causing other issues (pretty much a "race condition" on some possibly trivial content as e.g. a json file with just `[]` or alike)

waxlamp: I really do understand the pull to "simplify" the model so that embargo (which we really need to fix), "restricted", and a more general "mixed access" all appear to become simple extensions thereof, but doing so will actually make it harder to deliver the functionality as envisioned, unless we *write down requirements for each* and then validate that the implementation plan will support all requirements.

I do feel like we're on the verge of figuring out the broad semantics of "restriction" that will support everything, but I am really feeling the exposure to the risk of getting it wrong (and thus triggering another round of discussions and then rework to fix a broken attempted implementation). I'm going to schedule a meeting so we can discuss this synchronously and try to eliminate those risks.

doc/design/embargo-redesign.md
waxlamp: Yes I think we must be miscommunicating a bit. Among other things, I'm not sure how what I proposed causes any performance problems. What I'm suggesting is something like this:
- uploading an asset to an open dandiset would cause it to become unhidden in the bucket
- clicking an "unembargo" button would cause all of an embargoed dandiset's (hidden) assets to become unhidden (and the status of the dandiset becomes `open`)

Both of these "entry points" would work with the redesigned embargo concept (i.e., no performance issues). Allowing users to unhide assets one by one would also be compatible with this way of doing things, now that I think about it (it is equivalent to uploading that asset somewhere else), but I guess I don't see why any user would do so--but perhaps there is a reason (maybe some of the example use cases Satra posted in a different thread).

In any case, maybe we can discuss this face to face.

waxlamp:
@satra I've added some text to reflect the concepts we decided on in our meeting. Please take a look and approve if you like what you see.

doc/design/embargo-redesign.md
```diff
@@ -116,3 +116,107 @@ sequenceDiagram
...
+
+**Quirk.** Dandiset 4 now contains a hidden asset, which contradicts the openness of Dandiset 4 and seems to compromise the secrecy of Dandiset 3.
+
+**Remediation.** There are a couple of ideas here:
+
+1. **Don’t allow embargoed Dandiset assets to be used in open Dandisets.** At upload time, the API would simply refuse to allow hidden assets that are part of any embargoed Dandiset to be uploaded to an open Dandiset.
+
+   This cuts the problem off at the root, but it is subject to a failure of our imaginations: perhaps there is a valid reason a neuroscientist would want to do this. One example: if a standard community tool exists that generates a skeleton Dandiset that includes something boilerplateish like a license file or copyright notice, people would regularly include those in embargoed Dandisets, and that could trigger this error condition somewhat baselessly.
+
+   On the other hand, perhaps scientists would have a reason to want to include the same data in both an embargoed Dandiset and an open one. For instance, perhaps embargoed experiments are using some base but non-secret data within an embargoed Dandiset, and that same data needs to be part of an open Dandiset.
```
satra, line 144: remove this. we agreed on 2.

doc/design/embargo-redesign.md
```diff
@@ -116,3 +116,107 @@ sequenceDiagram
...
+
+2. **Allow this type of sharing but force the user to provide an explicit flag to do so.** In this remediation, the user would receive a warning that they are trying to include embargoed data in an open Dandiset. The API / client would return an error with a prompt to retry with a `--force` type flag, which would then allow the upload to occur. The embargoed assets would remain hidden to the public in the S3 bucket, and data access to the open Dandiset’s asset holding would occur via presigned URLs (the same way as for authorized users to embargoed Dandiset assets).
```
satra, line 146: the embargoes asset would not remain hidden to public in this bucket. we agreed in that call that the moment the force happens the asset becomes open. hence the force.

doc/design/embargo-redesign.md
```diff
@@ -116,3 +116,107 @@ sequenceDiagram
...
+The major questions here are:
+
+- **Are the semantics valid of placing an embargoed asset in an open Dandiset?**
+  *Answer:* Yes. Doing so causes the asset to become unhidden (see "chosen
+	solution" section below).
+
+- **Would a scientist ever want to push data to an embargoed Dandiset, and also
+	to an open Dandiset?** *Answer:* Maybe, but it doesn't matter--if they do
+	this, they'll be warned about their asset within their embargoed Dandiset
+	becoming open, and they can act according to their own wishes.
```
satra, line 157: yup, this is correct and reflects the above. i don't know if we want to capture the full discussion. if yes, then ignore my above comments.

satra (approved):
left some comments, but after reading everything, you can ignore most of them. lgtm.

doc/design/embargo-redesign.md
waxlamp:
```suggestion
2. **Allow this type of sharing but force the user to provide an explicit flag to do so.** In this remediation, the user would receive a warning that they are trying to include embargoed data in an open Dandiset. The API / client would return an error with a prompt to retry with a `--force` type flag, which would then allow the upload to occur. The embargoed assets would become open to the public in the S3 bucket, and data access to the open Dandiset’s asset holding would occur in the usual manner.
```

doc/design/embargo-redesign.md
waxlamp: I'm going to retain it just as a record of the thought process (and alternative ideas) that went into the discussion. It's clear that we're selecting (2) for implementation in the discussion below.

PR #1803, opened 2024-01-02
Retry aggregate_assets_summary_task on version metadata race condition

jjnesbitt:
Fixes https://github.com/dandi/dandi-archive/issues/1734

Related to #1800. Both this PR and #1800 address #1734, although this change should cover _all_ sources of a concurrent metadata change, instead of just the `validate_version_metadata` task.


dandiapi/api/tests/test_version.py
```diff
@@ -359,6 +362,39 @@ def test_version_aggregate_assets_summary(draft_version_factory, draft_asset_fac
...
+    version.metadata['foo'] = 'bar'
+    with pytest.raises(VersionMetadataConcurrentlyModified):
+        version_aggregate_assets_summary(version)
+
+
+@pytest.mark.django_db()
+def test_version_aggregate_assets_summary_task_retries(
+    draft_version_factory, draft_asset_factory, mocker, monkeypatch
+):
+    version = draft_version_factory(status=Version.Status.VALID)
```
danlamanna, line 383: I'm not sure how I feel about this test. It looks it's mostly testing that Celery behaves correctly. If we're worried about configuration changing then I think that could be achieved with a test that just asserts `version_aggregate_assets_summary.autoretry_for == (VersionMetadataConcurrentlyModified,)`, but I would probably leave that out.

danlamanna (changes requested):

dandiapi/api/tests/test_version.py
jjnesbitt: I kind of agree, but it felt wrong having no test, since there's otherwise no way to verify that the task is actually retried :man_shrugging:. If you think it's more of a burden than it's worth, I can remove the test.

danlamanna: I don't think what I'm saying is super authoritative, @mvandenburgh do you have an opinion?

jjnesbitt: I guess I'll also just mention that the spirit of this test is more of a regression test than a behavioral test, if that makes a difference.

mvandenburgh: > I kind of agree, but it felt wrong having no test, since there's otherwise no way to verify that the task is actually retried 🤷‍♂️. 

In principal I would argue that ensuring the task is retried is Celery's responsibility, and it's not something we need to test for ourselves. I think this is somewhat analogous to having a test that examines the RabbitMQ queue after calling `.delay()` to ensure the message is actually placed on the queue. As with other third party libraries, we trust Celery to do this correctly and treat it as a black box that "just works". In a similar manner, the fact that Celery will retry this task seems like a baseline assumption we should make, and the important thing to test is that `VersionMetadataConcurrentlyModified` is being raised correctly (which the other test does).

mvandenburgh: Also, I realize that example is somewhat contrived because there have been real concerns raised about the stability of some of Celery's more advanced features/the guarantees it makes about them, but I think it's a slippery slope to start testing Celery features in our own test suite.

dandiapi/api/services/metadata/__init__.py
```diff
@@ -95,6 +96,7 @@ def version_aggregate_assets_summary(version: Version) -> None:
     )
     if updated_count == 0:
         logger.info('Skipped updating assetsSummary for version %s', version.id)
+        raise VersionMetadataConcurrentlyModified
```
mvandenburgh, line 99: Will this trigger a Sentry error? I imagine we wouldn't want that to happen in this case.

mvandenburgh (approved):

dandiapi/api/services/metadata/__init__.py
jjnesbitt: I don't think it will. I _believe_ what happens is the following:

1. This exception is raised
2. Since the task is decorated with `autoretry_for`, it catches that exception and raises a celery `Retry` exception
3. Sentry (specific the celery integration) sees the `Retry` exception is raised, but ignores it, as it's recognized as part of celery's control flow. [Here](https://github.com/getsentry/sentry-python/blob/2f05ccbae7298978d2f1a8774a07386a018bcce9/sentry_sdk/integrations/celery.py#L346) is the code that controls that.

So we should be good, but I'll do a little more testing to make sure things behave as we expect before merging.

PR #1823, opened 2024-01-23
Boost dandischema to 0.9.* series so we get support for pydantic 2.0 and schema 0.6.5

### LINKED ISSUE #1820: Support dandischema 0.9.0 and schema version 0.6.5
dandischema 0.9.0 has been released; it introduces schema version 0.6.5 and requires Pydantic 2.0 (cf. #1791).  Support for the new schema version is required for dandi/dandi-cli#1381 to pass.

yarikoptic:
Closes #1820 

TODOs
- [x] address fails in e2e tests and investigate if any other changes possibly needed in frontend due to slight changes in the schema (between 0.6.4 and 0.6.6 of https://github.com/dandi/schema/tree/master/releases; documentation of those changes at the top posts of https://github.com/dandi/dandi-schema/pull/203 and https://github.com/dandi/dandi-schema/pull/218). Extra notes from @candleindark :
   - context.json should be the same as before.
   - All other files are generated by Pydantic V2 with a small modification from the default behavior, and they are compliant with JSON Schema Draft 2020-12 (https://docs.pydantic.dev/latest/concepts/json_schema/ for details)
   - The small modification in behavior  is documented in point `#2` in "Additional notes regarding changes entailed by Pydantic V2" in the top post of https://github.com/dandi/dandi-schema/pull/203.
- [x] I expect dandi-cli tests to pass but might fail. 
- [x] test against version in https://github.com/dandi/dandi-cli/pull/1381 in TEMP commit (it was 150297c3f91a08b0748f57824aded1df93228a4e commit, now gone)
- [x] Remove TEMP commit(s)

### DIFF
--- a/dandiapi/api/services/version/metadata.py
+++ b/dandiapi/api/services/version/metadata.py
@@ -43,6 +43,6 @@ def _normalize_version_metadata(
     }
     # Run the version_metadata through the pydantic model to automatically include any boilerplate
     # like the access or repository fields
-    return PydanticDandiset.model_construct(
-        **version_metadata
-    ).model_dump(mode='json',exclude_none=True)
+    return PydanticDandiset.model_construct(**version_metadata).model_dump(
+        mode='json', exclude_none=True
+    )
--- a/dandiapi/api/services/version/metadata.py
+++ b/dandiapi/api/services/version/metadata.py
@@ -43,4 +43,6 @@ def _normalize_version_metadata(
     }
     # Run the version_metadata through the pydantic model to automatically include any boilerplate
     # like the access or repository fields
-    return PydanticDandiset.unvalidated(**version_metadata).json_dict()
+    return PydanticDandiset.model_construct(
+        **version_metadata
+    ).model_dump(mode='json',exclude_none=True)
--- a/dandiapi/api/tests/test_tasks.py
+++ b/dandiapi/api/tests/test_tasks.py
@@ -292,7 +292,8 @@ def test_validate_version_metadata_no_assets(
     assert draft_version.validation_errors == [
         {
             'field': 'assetsSummary',
-            'message': 'A Dandiset containing no files or zero bytes is not publishable',
+            'message': 'Value error, '
+                       'A Dandiset containing no files or zero bytes is not publishable',
         }
     ]
 
--- a/dandiapi/api/tests/test_tasks.py
+++ b/dandiapi/api/tests/test_tasks.py
@@ -153,7 +153,7 @@ def test_validate_asset_metadata_no_digest(draft_asset: Asset):
 
     assert draft_asset.status == Asset.Status.INVALID
     assert draft_asset.validation_errors == [
-        {'field': 'digest', 'message': 'A non-zarr asset must have a sha2_256.'}
+        {'field': 'digest', 'message': 'Value error, A non-zarr asset must have a sha2_256.'}
     ]
 
 
--- a/dandiapi/api/tests/test_tasks.py
+++ b/dandiapi/api/tests/test_tasks.py
@@ -293,7 +293,7 @@ def test_validate_version_metadata_no_assets(
         {
             'field': 'assetsSummary',
             'message': 'Value error, '
-                       'A Dandiset containing no files or zero bytes is not publishable',
+            'A Dandiset containing no files or zero bytes is not publishable',
         }
     ]
 
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
     include_package_data=True,
     install_requires=[
         'celery',
-        'dandischema~=0.8.4',
+        'dandischema~=0.9.0',
         'django~=4.1.0',
         'django-admin-display',
         # Require 0.58.0 as it is the first version to support postgres' native
--- a/web/src/components/Meditor/types.ts
+++ b/web/src/components/Meditor/types.ts
@@ -65,7 +65,7 @@ export const isJSONSchema = (schema: JSONSchemaUnionType): schema is JSONSchema7
 
 export const isBasicSchema = (schema: JSONSchemaUnionType): schema is BasicSchema => (
   isJSONSchema(schema)
-  && isBasicType(schema.type)
+  && (isBasicType(schema.type) || schema.type === undefined)
 );
 
 export const isObjectSchema = (schema: JSONSchemaUnionType): schema is ObjectSchema => (
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -181,7 +181,7 @@ const descriptionMaxLength: ComputedRef<number> = computed(
   () => store.schema.properties.description.maxLength,
 );
 const dandiLicenses: ComputedRef<LicenseType[]> = computed(
-  () => store.schema.definitions.LicenseType.enum,
+  () => store.schema.$defs.LicenseType.enum,
 );
 
 if (!loggedIn()) {


yarikoptic:
there seems to be some changes in behavior with the new dandischema to address , some look like

```
E       AssertionError: assert [{'field': 'a...publishable'}] == [{'field': 'a...publishable'}]
1296
E         At index 0 diff: {'field': 'assetsSummary', 'message': 'Value error, A Dandiset containing no files or zero bytes is not publishable'} != {'field': 'assetsSummary', 'message': 'A Dandiset containing no files or zero bytes is not publishable'}
1297
E         Full diff:
1298
E           [
1299
E            {'field': 'assetsSummary',
1300
E         -   'message': 'A Dandiset containing no files or zero bytes is not publishable'},
1301
E         ?                                                                   ----------- --
1302
E         +   'message': 'Value error, A Dandiset containing no files or zero bytes is not '...
1303
E         
1304
E         ...Full output truncated (3 lines hidden), use '-vv' to show
```

@candleindark does it ring a bell may be right away?

candleindark:
> @candleindark does it ring a bell may be right away?

Yes. Pydantic V2 constructs error objects, including error messages, differently in comparison to Pydantic V1. This particular error is resulting from the failure to pass [the validator for the `assetsSummary` field in the `PublishedDandiset` model](https://github.com/dandi/dandi-schema/blob/dbd704ace5b2abf5fb2718cca427c8fff76dcb62/dandischema/models.py#L1662-L1669).

To adapt to the changes in error generations brought by Pydantic V2, I made changes in the tests in `dandischema`. Specifically, I made a change for the changed error message from the aforementioned error at https://github.com/dandi/dandi-schema/blob/dbd704ace5b2abf5fb2718cca427c8fff76dcb62/dandischema/tests/test_models.py#L422-L423. 

You may have to adjust some of the tests in dandi-archive to adapt to the new error generations in Pydantic V2 if these tests assess  errors  originate from Pydantic.

jwodder:
Some of the dandi-archive code should also be updated to no longer use now-deprecated dandischema methods; see the warnings emitted at the end of the tests.

yarikoptic:
@candleindark could you take over this PR and adjust those exception msgs, and see to @jwodder's comment about deprecated invocations, etc?

candleindark:
> @candleindark could you take over this PR and adjust those exception msgs, and see to @jwodder's comment about deprecated invocations, etc?

OK

candleindark:
@yarikoptic @jwodder I have solved the two failures that are related to the new format of error messages (to be pushed). There are four failures remaining. They are all related. I am suspecting the cause of the remaining failures is that in [POSTing to create a new dandiset with metadata](https://github.com/dandi/dandi-archive/blob/880a1022c590e4ef6606e0f6a00e16a98e90da9b/dandiapi/api/tests/test_dandiset.py#L333-L335), the metadata is not written to the database. I am not familiar with Django and how a Django app accesses and organizes the DB, and I having trouble to locate the tables in the DB for storing dandisets and related metadata. In fact, there were no table in the result of the following SQL script.

```
SELECT table_name
FROM information_schema.tables
WHERE table_schema = 'public'
ORDER BY table_name;
```

candleindark:
> @yarikoptic @jwodder I have solved the two failures that are related to the new format of error messages (to be pushed). There are four failures remaining. They are all related. I am suspecting the cause of the remaining failures is that in [POSTing to create a new dandiset with metadata](https://github.com/dandi/dandi-archive/blob/880a1022c590e4ef6606e0f6a00e16a98e90da9b/dandiapi/api/tests/test_dandiset.py#L333-L335), the metadata is not written to the database. I am not familiar with Django and how a Django app accesses and organizes the DB, and I having trouble to locate the tables in the DB for storing dandisets and related metadata. In fact, there were no table in the result of the following SQL script.
> 
> ```
> SELECT table_name
> FROM information_schema.tables
> WHERE table_schema = 'public'
> ORDER BY table_name;
> ```

I am able to see the tables in running the DB service in "production" mode.

I don't have to execute the steps in https://github.com/dandi/dandi-archive/blob/master/DEVELOPMENT.md#initial-setup for testing, do I?

yarikoptic:
@dandi/dandiarchive could you please guide us here? or may be @kabilar already got hands dirty enough to know this? (otherwise we will need to try to get into chatting to various GPTs to align with my knowledge of testing django apps ;-) )

yarikoptic:
> Do you use SQLite or an equivalent for testing? If you do, where is that DB file?

I think it is relying on a DB since https://github.com/dandi/dandi-archive/blob/master/.github/workflows/backend-ci.yml creates one. DB would leave in the docker (volume) I guess after compose-up ... if no experts follow up, I will dig it up tomorrow.

candleindark:
> > Do you use SQLite or an equivalent for testing? If you do, where is that DB file?
> 
> I think it is relying on a DB since https://github.com/dandi/dandi-archive/blob/master/.github/workflows/backend-ci.yml creates one. DB would leave in the docker (volume) I guess after compose-up ... if no experts follow up, I will dig it up tomorrow.

Yeah, it seems to be a Postgres DB. 

I found out more using the following code in a test.

```
    from django.conf import settings

    databases = settings.DATABASES
    database_engine = settings.DATABASES['default']['ENGINE']
```

It is a database under the name of `test_django` with authentication information being the same as the one for development setup.

Additionally, the access to DB in a test is provided by the `@pytest.mark.django_db()` decorator for the test.

However, for `test_dandiset_rest_create`,  I don't see anything written to the `api_daniset` table.

candleindark:
Because of the transactional nature of `transaction.atomic()` used in the project. To see what's written to the DB through another DB connection in the middle of the test, it seems we need to modify the options to the decorator and use an additional fixture as depicted below.

```py
transaction.atomic()
```

candleindark:
I can now confirm that the metadata was not written correctly to the database in the first place.

The cause of the problem is in

https://github.com/dandi/dandi-archive/blob/880a1022c590e4ef6606e0f6a00e16a98e90da9b/dandiapi/api/services/version/metadata.py#L46

I will continue the play tomorrow.

yarikoptic:
@waxlamp @mvandenburgh - I have adjusted PR description with pointers about json schema changes

yarikoptic:
@jwodder -- I am concerned about our dandi-cli testing here -- all green sounds too good to be true... looked into a sample run log and it says

```
collected 711 items / 443 deselected / 268 selected
...
267 passed, 443 deselected
```
whenever our runs in dandi-cli repo say ` 708 passed, 1 skipped, 2 xfailed in 724.73s (0:12:04) `... so what are will skipping/not testing here and why?

jwodder:
@yarikoptic The CLI integration tests are run with the `--dandi-api` option, which means that only tests that use the Archive Docker image are run.

mvandenburgh:
I pushed two fixes to this branch - one to fix the e2e CI test failures (https://github.com/dandi/dandi-archive/pull/1823/commits/cd5cd040496ff73c3efe9223f6834eaf4419b53c), and the other to fix the Meditor showing a tab for "Schema Key" (https://github.com/dandi/dandi-archive/pull/1823/commits/b3db2310632782ccac935b3a680818fe765ae8e7). I included further explanations in the commit descriptions.

I did some testing locally and things appear to be working after those fixes. I think we should still do extensive testing in staging before cutting a new release.

yarikoptic:
Great, thank you @mvandenburgh !
I would be all for it to merge and test in staging.  But then we must be aware that `master` has this "experimental" feature ATM so we do not rush to tag for release to expedite some fix.

yarikoptic:
And who would be to review & potentially approve?  meanwhile I will request multiple candidates.

yarikoptic:
Thank you @mvandenburgh ! Now we need I guess review and a blessing/merge by @waxlamp ?

yarikoptic:
So, @waxlamp should we merge and test it in staging?

PR #1827, opened 2024-01-24
Replace email.com owned by World Media Group by reserved by W3C for such purposes example.com

yarikoptic:


yarikoptic:
I will just close this in favor of #1828

PR #1828, opened 2024-01-24
DEVELOPMENT.md: set email to the one known to git

yarikoptic:
While following some instructions like this I would prefer to just be able to copy/paste and not follow unnecessary interactive prompts.

In the beginning of the doc we recommend to use the same email as for github which is likely to be the one configured in git. Hence I decided to just query `git`.

One other aspect I am fixing -- I am adding explicitly the email to `createsuperuser` so me/developer does not need to think twice to enter some other email from what I just entered interactively.

Sits on top of #1827

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,10 +6,10 @@ This is the simplest configuration for developers to start with.
 ### Initial Setup
 1. Run `docker-compose run --rm django ./manage.py migrate`
 2. Run `docker-compose run --rm django ./manage.py createcachetable`
-3. Run `docker-compose run --rm django ./manage.py createsuperuser`
+3. Run `docker-compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
    and follow the prompts to create your own user.
-   Set your username to your email to ensure parity with how GitHub logins work.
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -6,10 +6,10 @@ This is the simplest configuration for developers to start with.
 ### Initial Setup
 1. Run `docker-compose run --rm django ./manage.py migrate`
 2. Run `docker-compose run --rm django ./manage.py createcachetable`
-3. Run `docker-compose run --rm django ./manage.py createsuperuser`
+3. Run `docker-compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
    and follow the prompts to create your own user.
-   Set your username to your email to ensure parity with how GitHub logins work.
-4. Run `docker-compose run --rm django ./manage.py create_dev_dandiset --owner your.email@email.com`
+   We set your username to your git email to ensure parity with how GitHub logins work.
```
waxlamp, line 11:
```suggestion
   This sets your username to your git email to ensure parity with how GitHub logins work. You can also replace the command substitution expression with a literal email address, or omit the `--email` option entirely to run the command in interactive mode.
```

waxlamp (approved):
I expanded the discussion of what email address to use a bit to clarify how this works.

PR #1829, opened 2024-01-25
Remove obsolete/not-applicable TODO

yarikoptic:
The comment used to correspond to reality, and was needed for some ad-hoc manual invocation to setTokenHack but it was removed in 984451c822525343a6f54bffad4c8fa39a68ac78 AKA `v0.1.37~27^2~9^2~79^2`
a few years ago.  Now this is the correct specification of DRF


yarikoptic:
ping @waxlamp -- this might be trivial while the memory is fresh on that WWW-Authenticate investigation... please review/merge

waxlamp (approved):
This TODO does seem to be out of date. No one on the engineering team knows what the "setTokenHack login workaround" is, and in fact using DRF's token authentication is the right way to be doing this.

PR #1830, opened 2024-01-25
Add support for "Authorization: Bearer KEY" to follow the RFC 6750

### LINKED ISSUE #1825: REST endpoints send incomplete WWW-Authenticate header on 401
ATM (same example as in #1824 ):

```shell
❯ wget -S 'https://api.dandiarchive.org/api/dandisets/000403/versions/draft/assets/0cb481e5-b979-4eed-8ea4-9cd0d78276a3/download/' 2>&1 | grep Auth
  Www-Authenticate: Bearer realm="api"
```

and for that (Bearer) scheme I see https://datatracker.ietf.org/doc/html/rfc6750#section-2.1 listing

```
Authorization: Bearer mF_9.B5f-4.1JqM
```

whenever we seems to be using `token` not `Bearer`:

```shell
❯ rm -f dataset_description.json; curl -J -O -L -X 'GET' --silent -H "Authorization: token $DANDI_API_KEY" 'https://api.dandiarchive.org/api/dandisets/000403/versions/draft/assets/0cb481e5-b979-4eed-8ea4-9cd0d78276a3/download/' ; ls -ld dataset_description.json
-rw------- 1 yoh yoh 84 Jan 24 13:05 dataset_description.json
```

and if I specify as `Bearer` -- no good:

```shell
❯ rm -f dataset_description.json; curl -J -O -L -X 'GET'  -H "Authorization: Bearer $DANDI_API_KEY" 'https://api.dandiarchive.org/api/dandisets/000403/versions/draft/assets/0cb481e5-b979-4eed-8ea4-9cd0d78276a3/download/' ; ls -ld dataset_description.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Warning: Remote filename has no length
  0    58    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (23) Failure writing output to destination
ls: cannot access 'dataset_description.json': No such file or directory
```

In DataLad we are using `Bearer` keyword there and hence our implementation for this authorization scheme (https://github.com/datalad/datalad/blob/HEAD/datalad/downloaders/http.py#L401) is not working.  It works if I change in our code "Bearer" to "token" but it seems to be invalid and I think the quickest fix would be for dandi-archive to add support for having "Bearer" specified in addition to "token" and slowly (after we have our dandi-cli etc) deprecated "token".

yarikoptic:
ATM non-authenticated request is receiving 401 response with "Bearer" as the auth-scheme:

    ❯ curl --verbose -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/' 2>&1 | grep WW-A
    < WWW-Authenticate: Bearer realm="api"

But according to the
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes and in particular https://datatracker.ietf.org/doc/html/rfc6750 for such request client should provide "Authorization: Bearer KEY" not "Authorization: token KEY".

This commit adds support for both so we could follow the standard and retain support of already implemented client solutions. It would also allow for clients which (try to) follow the standard (e.g. DataLad) to support such requests without implementing some ad-hoc "token handling". Such approach is also taken by GitHub API:
https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28

Verification of functionality:

	❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/'
	{"detail":"Authentication credentials were not provided."}%                                                                                                                                                                                                        ❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/?content_disposition=attachment' -H 'Authorization: Bearer 21a587dff19ec6956364443b97414d8bb4331b09'
	MYDATA
	❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/?content_disposition=attachment' -H 'Authorization: token 21a587dff19ec6956364443b97414d8bb4331b09'
	MYDATA
	❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/?content_disposition=attachment' -H 'Authorization: Token 21a587dff19ec6956364443b97414d8bb4331b09'
	MYDATA
        ❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/?content_disposition=attachment' -H 'Authorization: Bearer 21a587dff19ec6956364443b97414d8bb4331b09'
        MYDATA
	❯ curl -J -L -X 'GET' 'http://localhost:8000/api/assets/ac212d93-525a-4454-afdd-85b90aad6143/download/?content_disposition=attachment' -H 'Authorization: dragon 21a587dff19ec6956364443b97414d8bb4331b09'
	{"detail":"Authentication credentials were not provided."}

Closes #1825

### DIFF
--- a/dandiapi/api/auth.py
+++ b/dandiapi/api/auth.py
@@ -1,0 +1,16 @@
+from __future__ import annotations
+
+from rest_framework.authentication import TokenAuthentication
+
+
+class BearerTokenAuthentication(TokenAuthentication):
+    """
+    To support not only DRF specific but also a standard oauth2 "Bearer" Authorization.
+
+    Supporting both "token" and "Bearer" authorization requests is similar to GitHub behavior:
+    See https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28
+
+    The recipe from https://github.com/encode/django-rest-framework//commit/ffdac0d93619b7ec6039b94ce0e563f0330faeb1
+    """
+
+    keyword = 'Bearer'
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -128,7 +128,8 @@ Then go to `/swagger` and use `GET /auth/token` end-point.
 ### Supplying the Token
 In API endpoint calls, add the `Authorization` HTTP header with a value of
 `Token <token_value>`. For example, for a token `1234`, include the header:
-`Authorization: Token 1234`.
+`Authorization: Bearer 1234` (following OAuth2 RFC 6750) or 
+`Authorization: Token 1234` (following Django rest framework).
 
 ## Scripts
 
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -57,6 +57,7 @@ class DandiMixin(ConfigMixin):
             # TODO: remove TokenAuthentication, it is only here to support
             # the setTokenHack login workaround
             'rest_framework.authentication.TokenAuthentication',
+            'dandiapi.api.auth.BearerTokenAuthentication',
         ]
 
         # Caching


dandiapi/api/auth.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework.authentication import TokenAuthentication
```
yarikoptic, line 1: note  that this is a new file.  If you would like to see it somewhere else -- let me know , but it could not go under dandi.api.views.auth due to then circular import.

waxlamp:
RFC 6750 does not apply to DRF's token authentication scheme, since DRF is not an OAuth 2.0 service. Together with the discussion in #1825, that means we should close this PR.

PR #1840, opened 2024-01-29
Contact owner

### LINKED ISSUE #723: requesting ownership
It would be nice if a dandiset had an option to request ownership.

marySalvi:
Closes #723 

Allows a user to send an e-mail populated with a subject 'Regarding  Dandiset `name` `id`' to either the contact person or owner(if logged in)

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -374,7 +374,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
     )
     # TODO: move these into a viewset
     @action(methods=['GET', 'PUT'], detail=True)
-    def users(self, request, dandiset__pk):  # noqa: C901
+    def users(self, request, dandiset__pk):
         dandiset: Dandiset = self.get_object()
         if request.method == 'PUT':
             if dandiset.unembargo_in_progress:
@@ -423,15 +423,20 @@ class DandisetViewSet(ReadOnlyModelViewSet):
             try:
                 owner_account = SocialAccount.objects.get(user=owner_user)
                 owner_dict = {'username': owner_account.extra_data['login']}
-                if 'name' in owner_account.extra_data:
-                    owner_dict['name'] = owner_account.extra_data['name']
+                owner_dict['name'] = owner_account.extra_data.get('name', None)
+                owner_dict['email'] = (
+                    owner_account.extra_data['email']
+                    if request.user.is_authenticated and 'email' in owner_account.extra_data
+                    else None
+                )
                 owners.append(owner_dict)
             except SocialAccount.DoesNotExist:
                 # Just in case some users aren't using social accounts, have a fallback
-                owners.append(
-                    {
-                        'username': owner_user.username,
-                        'name': f'{owner_user.first_name} {owner_user.last_name}',
-                    }
-                )
+                owner_dict = {
+                    'username': owner_user.username,
+                    'name': f'{owner_user.first_name} {owner_user.last_name}',
+                    'email': owner_user.email if request.user.is_authenticated else None,
+                }
+                owners.append(owner_dict)
+
         return Response(owners, status=status.HTTP_200_OK)
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -432,11 +432,12 @@ class DandisetViewSet(ReadOnlyModelViewSet):
                 owners.append(owner_dict)
             except SocialAccount.DoesNotExist:
                 # Just in case some users aren't using social accounts, have a fallback
-                owner_dict = {
-                    'username': owner_user.username,
-                    'name': f'{owner_user.first_name} {owner_user.last_name}',
-                    'email': owner_user.email if request.user.is_authenticated else None,
-                }
-                owners.append(owner_dict)
+                owners.append(
+                    {
+                        'username': owner_user.username,
+                        'name': f'{owner_user.first_name} {owner_user.last_name}',
+                        'email': owner_user.email if request.user.is_authenticated else None,
+                    }
+                )
 
         return Response(owners, status=status.HTTP_200_OK)
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -426,6 +426,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
                 owner_dict['name'] = owner_account.extra_data.get('name', None)
                 owner_dict['email'] = (
                     owner_account.extra_data['email']
+                    # Only logged-in users can see owners' email addresses
                     if request.user.is_authenticated and 'email' in owner_account.extra_data
                     else None
                 )
--- a/web/src/views/DandisetLandingView/ContactDialog.vue
+++ b/web/src/views/DandisetLandingView/ContactDialog.vue
@@ -33,16 +33,16 @@
       </v-card-title>
       <v-list>
         <v-tooltip
-          :disabled="loggedIn()"
+          :disabled="loggedIn() || !owners"
           open-on-hover
           right
         >
           <template #activator="{ on }">
             <v-list-item
               v-on="on"
-              :class="loggedIn() ? 'black--text' : 'grey--text'"
+              :class="(loggedIn() || !owners) ? 'black--text' : 'grey--text'"
               :selectable="!loggedIn()"
-              :href="`mailto:${owners}?subject=Regarding%20Dandiset%20${currentDandiset?.name}%20${currentDandiset?.metadata?.id} &body=${currentUser?.value?.username}`"
+              :href="makeTemplate(owners)"
             >
             <v-icon
               color="primary"
@@ -54,23 +54,35 @@
               Dandiset Owners
             </v-list-item>
           </template>
-          <span> You must me logged in to contact the owner </span>
+          <span v-if="!loggedIn()"> You must me logged in to contact the owner </span>
+          <span v-else> No owner e-mail available </span>
         </v-tooltip>
       <v-divider />
+        <v-tooltip
+            :disabled="contacts"
+            open-on-hover
+            right
+          >
 
-        <v-list-item
-          :href="`mailto:${contacts}?subject=Regarding%20Dandiset%20${currentDandiset?.name}%20${currentDandiset?.metadata?.id} &body=${currentUser?.value?.username}`"
-        >
-        <v-icon
-          color="primary"
-          left
-          small
-        >
-          mdi-card-account-mail
-        </v-icon>
-          Contact Person
-        </v-list-item>
-
+          <template #activator="{ on }">
+          <v-list-item
+            v-on="on"
+            :class=" !owners ? 'black--text' : 'grey--text'"
+            :selectable="!owners"
+            :href="makeTemplate(contacts)"
+          >
+          <v-icon
+            color="primary"
+            left
+            small
+          >
+            mdi-card-account-mail
+          </v-icon>
+            Contact Person
+          </v-list-item>
+            </template>
+          <span> No contact e-mail available </span>
+        </v-tooltip>
       </v-list>
     </v-card>
   </v-menu>
@@ -78,14 +90,21 @@
 <script setup lang="ts">
 import { computed } from 'vue';
 import { useDandisetStore } from '@/stores/dandiset';
-import { loggedIn, user } from '@/rest';
+import { loggedIn } from '@/rest';
 
 const store = useDandisetStore();
 const owners = computed(() => store.owners?.map((owner) => owner.username).toString());
 const contacts = computed(() => store.dandiset?.metadata?.contributor?.map((contact) => contact.roleName?.includes("dcite:ContactPerson") ? contact.email: '').toString());
-const currentUser = computed(() => user);
 const currentDandiset = computed(() => store.dandiset);
-currentDandiset.value?.metadata?.id
+
+const makeTemplate = (contact: string | undefined) => {
+  if (!contact) {
+    return '';
+  }
+  // Subject is: Regarding [dandiset_name]  [dandiset_id]
+  return `mailto:${contact}?subject=Regarding%20Dandiset%20${currentDandiset?.value?.name}%20${currentDandiset?.value?.metadata?.id}`;
+};
+
 
 </script>
 <style scoped>
--- a/web/src/views/DandisetLandingView/ContactDialog.vue
+++ b/web/src/views/DandisetLandingView/ContactDialog.vue
@@ -7,7 +7,7 @@
       #activator="{ on, attrs }"
     >
       <v-btn
-        id="download"
+        id="contact"
         outlined
         block
         v-bind="attrs"
@@ -91,10 +91,11 @@
 import { computed } from 'vue';
 import { useDandisetStore } from '@/stores/dandiset';
 import { loggedIn } from '@/rest';
+import type { User, Person, Organization} from '@/types';
 
 const store = useDandisetStore();
-const owners = computed(() => store.owners?.map((owner) => owner.email).toString());
-const contacts = computed(() => store.dandiset?.metadata?.contributor?.map((contact) => contact.roleName?.includes("dcite:ContactPerson") ? contact.email: '').toString());
+const owners = computed(() => store.owners?.map((owner: User) => owner.email).toString());
+const contacts = computed(() => store.dandiset?.metadata?.contributor?.map((contact: Person | Organization) => contact.roleName?.includes("dcite:ContactPerson") ? contact.email: '').toString());
 const currentDandiset = computed(() => store.dandiset);
 
 const makeTemplate = (contact: string | undefined) => {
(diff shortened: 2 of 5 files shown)


web/src/views/DandisetLandingView/DownloadDialog.vue
```diff
@@ -1,4 +1,8 @@
 <template>
+  <!--
+    Do we want to set the width here? Default is nearly the same.
+    If we do want to set wouldn't width=500 be sufficient?
+  -->
```
marySalvi, line 5: This comment is in regards to the the following `v-menu`. Just wanted to call attention to it on whether it should be changed or not. No need for it to be done in this PR or at all.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,96 @@
...
+              :href="`mailto:${owners}?subject=Regarding%20Dandiset%20${currentDandiset?.name}%20${currentDandiset?.metadata?.id} &body=${currentUser?.value?.username}`"
+            >
+            <v-icon
+              color="primary"
+              left
+              small
+            >
+              mdi-card-account-mail
+            </v-icon>
+              Dandiset Owner
```
waxlamp, line 53:
```suggestion
              Dandiset Owners
```

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,96 @@
...
+          :disabled="loggedIn()"
+          open-on-hover
+          right
+        >
+          <template #activator="{ on }">
+            <v-list-item
+              v-on="on"
+              :class="loggedIn() ? 'black--text' : 'grey--text'"
+              :selectable="!loggedIn()"
+              :href="`mailto:${owners}?subject=Regarding%20Dandiset%20${currentDandiset?.name}%20${currentDandiset?.metadata?.id} &body=${currentUser?.value?.username}`"
```
waxlamp, line 44: This should be refactored from here and the other place it appears into a computed property of the component. That will furthermore enable you to deal with possible missing data (all the optional chaining operators that appear) in a more robust way (either raising an error or supplying values besides `undefined` to use), as well as improve the readability of this expression.

waxlamp: While you're at it, I think it's odd to have the body filled it with my own email address. I assume that can be left out and you'll just get your email client with a fresh, blank email body?

web/src/views/DandisetLandingView/DownloadDialog.vue
waxlamp: Why would we want to change this? If there's a problem with the width for this other menu, I would file an issue and/or open a different PR to discuss/deal with it.

waxlamp, line 5:
```suggestion
```

waxlamp (changes requested):
Looking good! A few changes requested.

Also, I noticed that the title part of the dialog seems to be perhaps overly restricted by its width? Not sure how to describe it better than that but here's a picture:
overflow

This is happening at a pretty wide browser width, so I wonder if maybe the menu component needs a minimum width or something like that.

web/src/views/DandisetLandingView/DandisetActions.vue
```diff
@@ -37,13 +37,15 @@
...
+      <v-row
+        no-gutters
+      >
         <CiteAsDialog>
           <template
             #activator="{ on }"
           >
             <v-btn
-              id="download"
+              id="cite_as"
```
waxlamp, line 48: What happened here? Was this wrong before?

marySalvi: Yes. I included it in this PR because I am adding to the Dandiset actions.

web/src/views/DandisetLandingView/DownloadDialog.vue
marySalvi: Definitely will be removing the comment, just looking to bring attention to this.

It is odd to have a set width instead of something responsive. It is even odder to have a set width in this manner: 
```
min-width:500px;
max-width:500px;
```

I am bringing attention to this now because the DownloadDialog is the component we chose to influence the design of the ContactDialog

marySalvi: explained below.

web/src/views/DandisetLandingView/ContactDialog.vue
marySalvi: changed in 5f5b0cff9887ebefcd1e5640f1faccf5b9552afc

marySalvi:
> Also, I noticed that the title part of the dialog seems to be perhaps overly restricted by its width? Not sure how to describe it 
> This is happening at a pretty wide browser width, so I wonder if maybe the menu component needs a minimum width or something like that.

fixed in 5f5b0cff9887ebefcd1e5640f1faccf5b9552afc

web/src/views/DandisetLandingView/ContactDialog.vue
marySalvi: changed in d9847970259f7fafb446622621c4ae9ee8b30e6d

web/src/views/DandisetLandingView/DownloadDialog.vue
marySalvi: comment removed in 55db5cd4b43e05d463bc64f895b523248b54b653

dandiapi/api/views/dandiset.py
```diff
@@ -413,19 +408,24 @@ def users(self, request, dandiset__pk):
...
-                    }
-                )
+                owner_dict = {
+                    'username': owner_user.username,
+                    'name': f'{owner_user.first_name} {owner_user.last_name}',
+                }
+                if request.user.is_authenticated:  # Check if user is logged in
+                    owner_dict['email'] = owner_user.email
+                owners.append(owner_dict)
+
```
marySalvi, line 430: adding to the users function throws a linting error that it is now too complex. I could use some guidance on how best to resolve this. @waxlamp

waxlamp: One way to solve this is to always include `email` as a key in the response object, but set it to `None` for a unauthenticated user, and to the email address otherwise. That would do the following:
1. Eliminate the extra control flow complexity that leads to a higher [McCabe score](https://docs.astral.sh/ruff/rules/complex-structure/).
2. Make the response more uniform across the user's login state.

To do this you'd eliminate an if block in favor of a ternary conditional (something like `owner_user.email if request.user.is_authenticated else None`), but I don't know if that counts as a "decision point" to the McCabe algorithm.

Another take: the high complexity score is a clue that we were already doing a lot in this function, and have now tipped over into doing too much. Perhaps a cleaner solution is to add another endpoint that just gets the owner emails. This endpoint could just 403 for unauthenticated callers and unconditionally return email addresses otherwise. The cost would be another function in the codebase and another request at runtime. I'm not sure which approach is better.

A lot of uncertainty in my answer here. We can meet to hash this out -- let me know if you'd like to do so.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+
+const store = useDandisetStore();
+const owners = computed(() => store.owners?.flatMap((owner) => owner.email).toString());
+const contacts = computed(() => store.dandiset?.metadata?.contributor?.flatMap((contact) => contact.roleName?.includes("dcite:ContactPerson") ? contact.email: '').toString());
+const currentDandiset = computed(() => store.dandiset);
+
+const makeTemplate = (contact: string | undefined) => {
+  if (!contact) {
+    return '';
+  }
```
waxlamp, line 103: What is the behavior of the email client when hitting this code path?

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+  // Subject is: Regarding [dandiset_name]  [dandiset_id]
+  return `mailto:${contact}?subject=Regarding%20Dandiset%20${currentDandiset?.value?.name}%20${currentDandiset?.value?.metadata?.id}`;
```
waxlamp, line 105: I'm concerned about how this will break if `currentDandiset` is `null` or `undefined`. But I think in context of this component, that should never happen, which in turn makes me think we should be able to narrow the type of `currentDandiset` to eliminate `null | undefined` by making assertions and throwing an exception if the narrowing fails.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+};
+
+
```
waxlamp, line 108:
```suggestion
```

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+</script>
+<style scoped>
```
waxlamp, line 110:
```suggestion
</script>

<style scoped>
```

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+    </v-card>
+  </v-menu>
+</template>
+<script setup lang="ts">
+import { computed } from 'vue';
+import { useDandisetStore } from '@/stores/dandiset';
+import { loggedIn } from '@/rest';
+
+const store = useDandisetStore();
+const owners = computed(() => store.owners?.flatMap((owner) => owner.email).toString());
```
waxlamp, line 96: The `.toString()` here might be better expressed as an explicit conversion to the string needed for the mailto link. `Array.toString()` happens to do the right thing, but the coincidence indicates that explicit would be better than implicit (to borrow from the Zen of Python).

Also, you might consider leaving this as an array of email addresses, to be converted to a string closer to the point of use (i.e., in the `makeTemplate()` function).

waxlamp (changes requested):

dandiapi/api/views/dandiset.py
```diff
@@ -28,16 +28,12 @@
...
-    DandisetSearchResultListSerializer,
-    UserSerializer,
-    VersionMetadataSerializer,
-)
+from dandiapi.api.views.serializers import (CreateDandisetQueryParameterSerializer,
+                                            DandisetDetailSerializer, DandisetListSerializer,
+                                            DandisetQueryParameterSerializer,
+                                            DandisetSearchQueryParameterSerializer,
+                                            DandisetSearchResultListSerializer, UserSerializer,
+                                            VersionMetadataSerializer)
```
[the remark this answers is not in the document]
marySalvi: Yes this was ruff

dandiapi/api/views/dandiset.py
```diff
@@ -413,19 +408,24 @@ def users(self, request, dandiset__pk):
             send_ownership_change_emails(dandiset, removed_owners, added_owners)
 
         owners = []
+
```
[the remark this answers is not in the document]
marySalvi: yes

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+        <v-tooltip
+          :disabled="!owners ? false : loggedIn()"
+          open-on-hover
+          right
+        >
+          <template #activator="{ on }">
+            <v-list-item
+              v-on="on"
+              :class="(!loggedIn() || !owners) ? 'grey--text': 'black--text'"
+              :selectable="!loggedIn() || !owners"
```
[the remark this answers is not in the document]
marySalvi: using disabled also disables hover which would disable the tooltip. Setting selectable keeps the hover intact without allowing it to be clicked. We could do something else here to be more direct.

web/src/views/DandisetLandingView/ContactDialog.vue
marySalvi: Throw error here

web/src/views/DandisetLandingView/ContactDialog.vue
marySalvi: addressed in c3b06bbcc2361e74fd359ec2c0f80cfba0b02e99

mvandenburgh:
One important thing to point out about this is that we're now going to be exposing user emails in the public API. With this change, anyone can retrieve a DANDI user's email through this endpoint. Previously, user emails were only available to admins (unless of course they manually added it to the dandiset metadata).

I'm not suggesting this is something that should block this PR, but I think it's important to note.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,130 @@
...
+const currentDandiset = computed(() => store.dandiset);
+
+const makeTemplate = (contacts: string[] | undefined) => {
+  if (contacts === undefined) {
+    throw new Error('Contact is undefined.');
+  }
+  if (currentDandiset.value){
+    const subject = encodeURIComponent(`Regarding Dandiset ${currentDandiset.value.name} ${currentDandiset.value.metadata?.id}`);
+    const contact = contacts.join(',');
+    return `mailto:${contact}?subject=${subject}`;
```
waxlamp, line 119: In the deploy preview for this PR, this is working for the "contact person" option, but not for "dandiset owners". Example: in [Dandiset 200560](https://deploy-preview-1840--gui-staging-dandiarchive-org.netlify.app/dandiset/200560), the mailto link includes a blank email address (it reads `mailto:?subject=[...]`).

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,130 @@
...
+    )
+);
+const currentDandiset = computed(() => store.dandiset);
+
+const makeTemplate = (contacts: string[] | undefined) => {
+  if (contacts === undefined) {
+    throw new Error('Contact is undefined.');
+  }
+  if (currentDandiset.value){
+    const subject = encodeURIComponent(`Regarding Dandiset ${currentDandiset.value.name} ${currentDandiset.value.metadata?.id}`);
```
waxlamp, line 117: The subject line ends up rendering as `Regarding Dandiset NeuroConv SDK Testing DANDI:200560/draft` which feels jumbled up. A better template might be:

```suggestion
    const subject = encodeURIComponent(`Regarding Dandiset ${currentDandiset.value.dandiset.identifier} ("${currentDandiset.value.name}")`);
```

This names the Dandiset by its identifier (i.e. its six-digit, zero-padded ID) which is normally how we refer to Dandisets, and puts its name in quotes within parentheses. So more like `Regarding Dandiset 200560 ("NeuroConv SDK Testing")`.

waxlamp (changes requested):
Found some bugs, and I think we can simplify some of the `undefined | null` processing to make the template code clearer.

marySalvi:
> One important thing to point out about this is that we're now going to be exposing user emails in the public API. With this change, anyone can retrieve a DANDI user's email through this endpoint. Previously, user emails were only available to admins (unless of course they manually added it to the dandiset metadata).
> 
> I'm not suggesting this is something that should block this PR, but I think it's important to note.

Agreed that this is an important issue to note. This concern was brought up to @satra  and the decision was that if a person is not logged in it is fine for them to access the contact person's email but they would need to be logged in for the owner's email.

marySalvi:
> I saw that you accepted the spacing suggestion between the <script> and <style> sections. Do you want to apply this one too?

Actually I just resolved it after our conversation.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,138 @@
...
+  }
+};
+
+const isDisabled = (contacts: string[] | undefined): boolean => {
+  if (contacts === undefined || contacts.length === 0 || contacts[0] === '') {
+    return true;
+  }
+  return false;
+};
+
```
marySalvi, line 130: @waxlamp I pulled out most of the disabled logic and I _think_ this addresses most of your concerns. Let me know if this is not what you want

web/src/views/DandisetLandingView/ContactDialog.vue
waxlamp: This is due to the deploy preview going to the production API, which does not have the changes from this PR that are needed to correctly populate the owners list.

web/src/views/DandisetLandingView/ContactDialog.vue
```diff
@@ -0,0 +1,138 @@
...
+    return `mailto:${contact}?subject=${subject}`;
+  }
+};
+
+const isDisabled = (contacts: string[] | undefined): boolean => {
+  if (contacts === undefined || contacts.length === 0 || contacts[0] === '') {
+    return true;
+  }
+  return false;
+};
```
waxlamp, line 129:
```suggestion
const isDisabled = (contacts: string[] | undefined): boolean =>
  contacts === undefined || contacts.length === 0 || contacts[0] === '';
```

mvandenburgh (approved):
I added a couple minor changes around simplifying types and moving some boolean logic out of the template and into the script section.

mvandenburgh:
I assumed the "auto-merge" button would have waited for CI to pass :frowning_face: ... this is only going to staging so worse case I can just revert if it doesn't pass

PR #1845, opened 2024-01-31
Add tooltips to describe FileBrowserView action icons

waxlamp:
This splits out a contribution from #1666 that adds tooltips to the file browser view's action icons.

(attn: @magland)

PR #1846, opened 2024-01-31
Improve file browser view action icons

waxlamp:
This PR improves the presentation of the file browser view icons:
- replace the "eyeball" icon with an "open in app" icon
- replace the triple dot menu with an explicit button-based dropdown menu

The open in app icon makes more sense for what that button does. And the explicit dropdown uses standard language for "open an external program", and makes the presence of those integrations more discoverable.

This was inspired by discussions on #1666.

Depends on #1845.

(attn: @magland)


marySalvi (dismissed):
LGTM

yarikoptic (approved):
Fine with me, thank you @waxlamp

marySalvi (approved):

PR #1847, opened 2024-01-31
Open asset on click

### LINKED ISSUE #761: clicking on NWB file in data viewer does nothing
In the data explorer, the NWB files are clickable, but from what I can tell when you click on them nothing happens:

https://user-images.githubusercontent.com/844306/124321571-f2f36680-db4b-11eb-9106-eade7ab8d066.mov

I'd prefer to have it like before where you can an information panel about that file to the right, but if we aren't going to have that then we shouldn't make the it clickable.

waxlamp:
This splits out a contribution from #1666 that initiates the "view in app" action when the item itself is clicked.

Depends on #1846.

Closes #761.

(attn: @magland)

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -116,7 +116,7 @@
                 v-for="item in items"
                 :key="item.path"
                 color="primary"
-                @click="openItem(item)"
+                @click.self="openItem(item)"
               >
                 <v-icon
                   class="mr-2"
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -116,7 +116,7 @@
                 v-for="item in items"
                 :key="item.path"
                 color="primary"
-                @click="selectPath(item)"
+                @click="openItem(item)"
               >
                 <v-icon
                   class="mr-2"
@@ -445,12 +445,16 @@ function locationSlice(index: number) {
   return `${splitLocation.value.slice(0, index + 1).join('/')}/`;
 }
 
-function selectPath(item: AssetPath) {
+function openItem(item: AssetPath) {
   const { asset, path } = item;
 
-  // Return early if path is a file
-  if (asset) { return; }
-  location.value = path;
+  if (asset) {
+    // If the item is an asset, open it in the browser.
+    window.open(inlineURI(asset.asset_id), "_self");
+  } else {
+    // If it's a directory, move into it.
+    location.value = path;
+  }
 }
 
 function navigateToParent() {
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -158,7 +158,7 @@
                         </v-icon>
                       </v-btn>
                     </template>
-                    <span>Open asset in browser</span>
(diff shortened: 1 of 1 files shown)


web/src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -433,12 +447,16 @@ function locationSlice(index: number) {
...
-function selectPath(item: AssetPath) {
+function openItem(item: AssetPath) {
   const { asset, path } = item;
 
-  // Return early if path is a file
-  if (asset) { return; }
-  location.value = path;
+  if (asset) {
+    // If the item is an asset, open it in a new window
+    window.open(inlineURI(asset.asset_id), '_blank', 'noreferrer');
```
marySalvi, line 455: I think in this scenario I wouldn't expect the asset to open in a new window. I don't have a strong opinion on this though so if this is preferred feel free to dismiss this.

marySalvi (dismissed):
One minor comment to address but otherwise approved.

web/src/views/FileBrowserView/FileBrowser.vue
yarikoptic: I also tend to avoid using explicit "in new window" since it imposes decision on the user, instead of user having control -- there is always a way/shortcut to open in new window/tab if so desired.  But the icon used here (if I got it right)  
image
usually means "new window" for me.   So would somewhat hint on returning the eye

waxlamp: You are both (@marySalvi, @yarikoptic) agreed that clicking on the item should open it in the same window?

@yarikoptic, I'm confused by your comment about the icon--the logic in this function is specifically for what to do when clicking directly on the item. The change of the eye icon to the open-in-new-window icon doesn't change the behavior of that button, it just replaces the eye icon with a more accurate one (see #1846 for where this change was made).

magland: Sorry for the tangential comment, but... if this asset is a nwb file... no browser can handle nwb file type natively. So it results in a download, which is almost never what the user would intend by clicking to open it. If user wants to download, then they should be clicking on an explicit download button, IMO.

waxlamp: Agreed in general, @magland, but there are two things to note:
1. Trying to control what the browser does with its native handling of files is a losing battle, one I don't want to fight. There was some discussion about this in the original PR, along the lines of plugins users might install to extend the browser's native handling capabilities, etc. For that reason, an explicit action to open a file natively is going to do what it does, even if it's not the right thing for, e.g., an NWB file.
2. I filed #1848 to try to improve this situation in the future. That issue says we should simply disallow opening the file if we know enough about its mimetype. It would be even better if we selected a default application for certain mimetypes (like neuroglancer Neurosift for NWB files) that would launch on click.

So for now, we need to stick with the browser-native behavior, even if it's not what we want in the end. We're going to get there!

waxlamp: @marySalvi, @yarikoptic: I modified the [previous PR](#1846) in the chain with 22ca4e2b1977affe79d12c8b4a070c21ca214e8f to open assets in the same window (and use a different icon to reflect that).

I also modified this PR to open on item click in the same window as well.

yarikoptic (approved):
Seems to work as desired. Thanks!

marySalvi (approved):

Issue #1837, opened 2024-01-26
Add endpoint for querying a folder or asset path in a Dandiset


jwodder:
Implementing ~https://github.com/dandi/dandi-webdav~ https://github.com/dandi/dandidav and some other things would be much easier if it were possible to make a single API request that provided a path and received in return either the asset at that path or an indication that the path is a folder, possibly along with the folder's children.  Note that the current `/dandisets/000108/versions/draft/assets/paths/` endpoint is insufficient for this, as it does not return asset properties or metadata, and even if it did, setting `path_prefix` to an asset path returns an empty list, so one cannot make a single request that returns either an asset or the contents of a folder at a path.

Possible endpoint: `GET /dandiset/{dandiset_id}/versions/{version_id}/assets/atpath/?path={path}[&metadata=<0|1>]`

- If `path` is omitted from the request, return a 4xx error.
- If `path` is not an asset path and there are no assets under `{path}/`, return 404.
- If `path` points to an asset, the following is returned:

    ```json
    {
        "type": "asset",
        "asset": {
             // Asset properties (ID, size, created, etc.) as would be returned while paginating over `/dandisets/{dandiset_id}/versions/{version_id}/assets/`
             // If `metadata=1` was included in the query, "asset" also contains a "metadata" field containing the asset metadata
        }
    }
    ```

- If there are assets under `{path}/` (i.e., if `path` is a folder), the following or similar is returned:

    ```json
    {
        "type": "folder",
        "folder": {
             "aggregate_files": INT,
             "aggregate_size": INT
        }
    }
    ```

- Ideally, I'd like it if including a `children=1` query parameter would, if `path` turns out to be a folder, cause the response to include a list of all direct children of `path`, both assets (with properties, and including metadata if `metadata=1`) and paths of subfolders.  However, this would need to be paginated, and having a response that (a) is sometimes paginated and (b) only paginates a subset of the data returned seems awkward.

A parallel endpoint could also be implemented for Zarr entries.

satra:
@jwodder - in the current `paths` api the absence of asset i think indicates it's a folder. doesn't that not satisfy your requirements except for the `metadata` return parameter?

<details>
<summary> current return of `paths` api at the top level of 000108
</summary>

```json

{
  "count": 7,
  "next": null,
  "previous": null,
  "results": [
    {
      "path": "dataset_description.json",
      "version": 500,
      "aggregate_files": 1,
      "aggregate_size": 71,
      "asset": {
[...]

PR #1851, opened 2024-02-01
A discussion on needed `/paths` properties and metadata

yarikoptic:
From https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949

### DIFF
--- a/doc/design/paths-endpoint-webdav.md
+++ b/doc/design/paths-endpoint-webdav.md
@@ -6,16 +6,16 @@ Submitted as a PR to avoid spagetti thread discussion on https://github.com/dand
 @jwodder listed needed properties and metadata fields:
 
 * Asset properties used by dandidav:
-    * `asset_id`
+    * `asset_id` (PRESENT)
     * `blob_id`
     * `zarr_id`
     * `path`
-    * `size`
-    * `created`
-    * `modified`
+    * `size`  (PRESENT)
+    * `created` (GOOD TO HAVE)
+    * `modified` (SUGGESTED to be included)
         * Note: It might be more accurate to use the `blobDateModified` metadata field for this instead; cf. dandi/dandidav#17
 
 * Asset metadata used by dandidav:
-    * `encodingFormat` (blobs only)
-    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs)
+    * `encodingFormat` (blobs only) (STRONGLY DESIRED but not REQUIRED)
+    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs) (INCLUDED)
     * `digest["dandi:dandi-etag"]` (blobs only)
--- a/doc/design/paths-endpoint-webdav.md
+++ b/doc/design/paths-endpoint-webdav.md
@@ -9,7 +9,7 @@ Submitted as a PR to avoid spagetti thread discussion on https://github.com/dand
     * `asset_id` (PRESENT)
     * `blob_id`
     * `zarr_id`
-    * `path`
+    * `path` (PRESENT)
     * `size`  (PRESENT)
     * `created` (GOOD TO HAVE)
     * `modified` (SUGGESTED to be included)
--- a/doc/design/paths-endpoint-webdav.md
+++ b/doc/design/paths-endpoint-webdav.md
@@ -1,6 +1,6 @@
 # Discussion on metadata needed for dandidav
 
-Submitted as a PR to avoid spagetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
+Submitted as a PR to avoid spaghetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
 
 
 @jwodder listed needed properties and metadata fields:
--- a/doc/design/paths-endpoint-webdav.md
+++ b/doc/design/paths-endpoint-webdav.md
@@ -1,0 +1,21 @@
+# Discussion on metadata needed for dandidav
+
+Submitted as a PR to avoid spagetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
+
+
+@jwodder listed needed properties and metadata fields:
+
+* Asset properties used by dandidav:
+    * `asset_id`
+    * `blob_id`
+    * `zarr_id`
+    * `path`
+    * `size`
+    * `created`
+    * `modified`
+        * Note: It might be more accurate to use the `blobDateModified` metadata field for this instead; cf. dandi/dandidav#17
+
+* Asset metadata used by dandidav:
+    * `encodingFormat` (blobs only)
+    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs)
+    * `digest["dandi:dandi-etag"]` (blobs only)


doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
+# Discussion on metadata needed for dandidav
+
+Submitted as a PR to avoid spagetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
+
+
+@jwodder listed needed properties and metadata fields:
+
+* Asset properties used by dandidav:
+    * `asset_id`
```
yarikoptic, line 9:
```suggestion
    * `asset_id` (PRESENT)
```

But I guess only to get metadata, and should generally not be needed if contentURL to redirect to is known, right?

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `blob_id`
+    * `zarr_id`
+    * `path`
```
yarikoptic, line 12:
```suggestion
    * `path` (PRESENT)
```

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `size`
```
yarikoptic, line 13:
```suggestion
    * `size`  (PRESENT)
```

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `created`
+    * `modified`
```
yarikoptic, line 15:
```suggestion
    * `modified` (SUGGESTED to be included)
```

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+
+@jwodder listed needed properties and metadata fields:
+
+* Asset properties used by dandidav:
+    * `asset_id`
+    * `blob_id`
+    * `zarr_id`
+    * `path`
+    * `size`
+    * `created`
```
yarikoptic, line 14: it is not strictly necessary for WebDAV right? so just
```suggestion
    * `created` (GOOD TO HAVE)
```

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `modified`
+        * Note: It might be more accurate to use the `blobDateModified` metadata field for this instead; cf. dandi/dandidav#17
+
+* Asset metadata used by dandidav:
+    * `encodingFormat` (blobs only)
+    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs)
```
yarikoptic, line 20:
```suggestion
    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs) (INCLUDED)
```

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `blob_id`
+    * `zarr_id`
+    * `path`
+    * `size`
+    * `created`
+    * `modified`
+        * Note: It might be more accurate to use the `blobDateModified` metadata field for this instead; cf. dandi/dandidav#17
+
+* Asset metadata used by dandidav:
+    * `encodingFormat` (blobs only)
```
yarikoptic, line 19:
```suggestion
    * `encodingFormat` (blobs only) (STRONGLY DESIRED but not REQUIRED)
```
right?

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `contentUrl` (API download URL for blobs, S3 URL for Zarrs)
+    * `digest["dandi:dandi-etag"]` (blobs only)
```
yarikoptic, line 21: what this one for?

yarikoptic (commented):
@jwodder please review / comment on / accept where my suggestion about state corresponds to what we already have or I suggested in https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921643799

doc/design/paths-endpoint-webdav.md
jwodder: True.

doc/design/paths-endpoint-webdav.md
jwodder: `size`, `created`, `modified`, `blob_id`, `zarr_id`, and metadata are not present in `.../assets/paths/` responses.  `dandidav` currently gets those from just `.../assets/`.

doc/design/paths-endpoint-webdav.md
jwodder: It's needed for both metadata and properties, as the only properties that `../assets/paths/` provides are path and asset ID.

yarikoptic: and `aggregate_size` which is the size for the asset if I got it right

doc/design/paths-endpoint-webdav.md
yarikoptic: `aggregate_size`?

jwodder: `dandidav` currently ignores that and just gets the size from the properties.

yarikoptic: ok, but my point is that the value is available already albeit under obfuscated name somewhat.

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
+# Discussion on metadata needed for dandidav
+
+Submitted as a PR to avoid spagetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
+
+
+@jwodder listed needed properties and metadata fields:
+
+* Asset properties used by dandidav:
+    * `asset_id` (PRESENT)
+    * `blob_id`
```
yarikoptic, line 10: what for do we need blob_id if we have contentUrl?

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
...
+    * `zarr_id`
```
yarikoptic, line 11: I guess `zarr_id` is needed for traversal etc? or could contentUrl version be used? (probably we do not want to rely on parsing it).

So it feels that we might ask for asset record to include `blob_id` (although not sure what for yet exactly) and `zarr_id`, hence

```suggestion
    * `zarr_id`  (DESIRED)
```

doc/design/paths-endpoint-webdav.md
jwodder: The `contentUrl` returned in `.../assets/paths/` responses is a plain S3 URL, and serving it to the user in a web browser would result in the file being downloaded and named with the blob ID.  In order for the downloaded file to be named with the asset's filename instead, we need a signed S3 URL, which is obtained via the API download URL in the `contentUrl` field of the metadata.

doc/design/paths-endpoint-webdav.md
jwodder: `dandidav` does parse S3 `contentUrl`s, as that's necessary in order to discover the name of the S3 bucket without hardcoding it or requiring it to be passed on the command line.

doc/design/paths-endpoint-webdav.md
```diff
@@ -0,0 +1,21 @@
+# Discussion on metadata needed for dandidav
+
+Submitted as a PR to avoid spagetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
```
jwodder, line 3:
```suggestion
Submitted as a PR to avoid spaghetti thread discussion on https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921864949
```

yarikoptic: oh -- codespell didn't pick it up!? ;) thanks @jwodder

doc/design/paths-endpoint-webdav.md
yarikoptic: so in principle we can just parse out bucket, zarr_id and blob_id from contenturl, correct?

jwodder: Yes, but I'd rather dandidav not have to make too many assumptions about how our S3 URLs are structured.

doc/design/paths-endpoint-webdav.md
yarikoptic: ok, so we the actually need `result_type` (Folder, AssetBlob, AssetZarr) as I suggested in https://github.com/dandi/dandi-archive/issues/1837#issuecomment-1921643799 .

doc/design/paths-endpoint-webdav.md
yarikoptic: agree. We would only need to be able to construct zarr "http s3" URLs from manifests so it would need to know that , right?
For blobs -- contentUrl should be good enough.

doc/design/paths-endpoint-webdav.md
yarikoptic: in principle could be any other checksum I guess , right?

note: I thought that besides ETag functionality a commonly known checksum could then be used by receiving end to validate the download if that would be the target purpose for such a listing. 

for versioned zarrs we would need `"dandi:dandi-zarr-checksum"` to use in conjunction with the `zarr_id`.

jwodder: Well, the `dandi:dandi-etag` is the same as the ETag header reported by S3 when downloading a blob asset.  In contrast, there's no single file that has a `dandi:dandi-zarr-checksum` as an ETag.

doc/design/paths-endpoint-webdav.md
jwodder: Note that asset metadata lists two `contentUrl`s, an unsigned S3 URL and an API download URL of the form `https://api.dandiarchive.org/api/assets/{asset_id}/download/`.

* For Zarr assets, `dandidav` needs the S3 URL in order to know how to query the Zarr's entries from S3.

* However, for blob assets, `dandidav` currently uses the API download URL, as it provides a better UX.  Specifically, if `dandidav` were to redirect `GET` requests for a blob to the blob's unsigned S3 URL, then the browser would save the response to a file named after the last component of the S3 URL, which is (always?) the blob ID.  In order to get the downloaded files to be named after the filename portion of the asset's path instead, a signed S3 URL with a Content-Disposition has to be used, and that is acquired by redirecting to the API download URL, which in turn redirects to such a signed S3 URL.

doc/design/paths-endpoint-webdav.md
yarikoptic: I meant that we can use `dandi:dandi-zarr-checksum ` as a value for  ETag on zarr collections (folders).

jwodder: I'm not sure if it makes sense to apply etags to collections, as `GET`ing a collection won't give you a response with that etag.  (Cf. https://github.com/dandi/dandidav/issues/26)

jjnesbitt:
Superceded by https://github.com/dandi/dandi-archive/pull/2155

PR #1854, opened 2024-02-02
Fix lint errors

waxlamp:
I noticed a few linting errors popping up in CI. This PR fixes them, in three bits:
1. Update pyproject.toml to use the correct ruff config section names.
2. Eliminate some explicit dictionary key checks in favor of `dict.get()`.
3. Use `logging.getLogger()` in place of directly instantiating a `Logger` object.

I believe (1) and (2) are fine (though I'd appreciate validation); (3) might be trickier: although the change I made is a best practice (even recommended directly by the [docs for the `logging` module](https://docs.python.org/3/library/logging.html#logger-objects)), it does represent a semantic change in the codebase.

Let me know if I should peel off one or more of these commits into separate PRs--glad to do so if it makes reviewing/merging easier.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,8 @@ exclude = [
 [tool.ruff]
 line-length = 100
 target-version = "py311"
+
+[tool.ruff.lint]
 select = ["ALL"]
 ignore = [
   # Incompatible with formatter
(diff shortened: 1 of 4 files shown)


yarikoptic:
could this be reviewed/merged asap so nobody wastes time chasing those CI fails?

yarikoptic (approved):

pyproject.toml
```diff
@@ -18,6 +18,8 @@ exclude = [
 [tool.ruff]
 line-length = 100
 target-version = "py311"
+
+[tool.ruff.lint]
```
mvandenburgh, line 22: Where does this come from?

mvandenburgh (approved):

PR #1882, opened 2024-03-04
Run all linting commands, even if some of them fail

waxlamp:
Without the `ignore_errors` configuration option set to `True`, Tox exits as soon as any command in the list fails. In the case of the lint tests, we would like to continue running the commands, since they are essentially independent tests.

This was masking formatting errors whenever the `check` command was failing, preventing us from knowing that we should fix those.

(attn: @marySalvi)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,7 @@ envlist =
 
 [testenv:lint]
 package = skip
+ignore_errors = True
 deps =
     codespell~=2.0
     ruff


mvandenburgh (approved):
@brianhelba may also be interested in this, as we move towards integrating Ruff into the girder4/resonant stack. 
[Docs for this option](https://tox.wiki/en/latest/config.html#ignore_errors)

waxlamp:
> @brianhelba may also be interested in this, as we move towards integrating Ruff into the girder4/resonant stack. [Docs for this option](https://tox.wiki/en/latest/config.html#ignore_errors)

Thanks for mentioning the docs, @mvandenburgh; I forgot to mention that `ignore_errors` is a bit of a misnomer since the tox invocation still fails if any of the individual commands do (i.e. it's not really *ignoring* the errors in the end).

Note that some code formatting failures are shown in the CI for this branch; @marySalvi is going to push a fix for that in a separate PR, and I'll go ahead and merge this one now.

brianhelba:
@waxlamp @mvandenburgh Makes sense, especially for this one task, which is a read-only check where the steps aren't dependent on each other.

waxlamp:
@brianhelba, is this enhancement on your radar for "upstreaming" to an eventual ruff-based Resonant cookiecutter?

brianhelba:
I don't think we'll add Codespell as an upstream recommendation. It seems to have too many false positives.

I don't think it's necessary to run the format checks if the linting fails (since formatting shouldn't be fixed manually), but I'm happy to discuss it more in an upstream PR if someone wants to make it.

waxlamp:
> I don't think we'll add Codespell as an upstream recommendation. It seems to have too many false positives.

Oh yes, I didn't mean to include Codespell in this.

> I don't think it's necessary to run the format checks if the linting fails (since formatting shouldn't be fixed manually), but I'm happy to discuss it more in an upstream PR if someone wants to make it.

Are you saying that (auto)formatting should only happen once linting passes? Aren't there cases where fixing the formatting would eliminate some (or maybe even all) linting errors?

yarikoptic:
> I don't think we'll add Codespell as an upstream recommendation. It seems to have too many false positives.

What upstream is the conversation about here?

waxlamp:
> > I don't think we'll add Codespell as an upstream recommendation. It seems to have too many false positives.
> 
> What upstream is the conversation about here?

Resonant (formerly known as Girder 4).

PR #1885, opened 2024-03-05
Use different collation for Asset `path` field

### LINKED ISSUE #1839: `…/assets/?order=path` is horribly broken
Attempting to get a list of assets ordered by path via a request to `/dandisets/:dandiset_id/versions/:version_id/assets/?order=path` does not actually return them in path order.  For example, listing all assets in 000108 with the following script:

```python
import sys

DANDISET = "000108"

if len(sys.argv) > 1:
    path_prefix = sys.argv[1]
else:
    path_prefix = None

with DandiAPIClient.for_dandi_instance("dandi") as client:
    d = client.get_dandiset(DANDISET, "draft")
    if path_prefix is not None:
        for asset in d.get_assets_with_path_prefix(path_prefix, order="path"):
            print(asset.path)
    else:
        for asset in d.get_assets(order="path"):
            print(asset.path)
```

results in the following output:

https://gist.github.com/jwodder/794af3537b95fdc5b0edc8c23edd9815

In particular, lines 170 and 171 are:

```text
sub-mEhm/ses-20220924h14m39s16/micr/sub-mEhm_ses-20220924h14m39s16_sample-ADO6R8_stain-YO_run-1_chunk-3_SPIM.ome.zarr
sub-MITU01h3/ses-20211130h19m15s39/micr/sub-MITU01h3_ses-20211130h19m15s39_sample-20_stain-NN_run-1_chunk-1_SPIM.ome.zarr
```

which are not in order, either by full path or (as I briefly suspected) by final component.  They are in order when sorting case-insensitively, but that doesn't apply to the output as a whole.  In particular, lines 600 and 601 are:

```text
sub-MITU01h3/ses-20220219h16m37s36/micr/sub-MITU01h3_ses-20220219h16m37s36_sample-12_stain-NN_run-1_chunk-1_SPIM_illum_corr.ome.zarr
sub-MITU01h3/ses-20220219h16m37s36/micr/sub-MITU01h3_ses-20220219h16m37s36_sample-12_stain-NN_run-1_chunk-1_SPIM.ome.zarr
```

which are not in order, whether case-sensitively or case-insensitively, as `.` comes before `_` in Unicode.

Most problematic for my purposes are lines 645 and 646:

```text
sub-MITU01h3/ses-20220313h17m18s08/micr/sub-MITU01h3_ses-20220313h17m18s08_sample-25_stain-SST_run-1_chunk-9_SPIM.ome.zarr
sub-MITU01/ses-20210521h17m17s06/micr/sub-MITU01_ses-20210521h17m17s06_sample-178_photo.jpg
```

Why would you ever sort `sub-MITU01h3/…` before `sub-MITU01/…`?

jjnesbitt:
Closes #1839

This fixes an issue where assets were returned seemingly out of order when ordered by `path`. The underlying reason for this is that the default collation used by the database (in my local case and production `en_us.utf-8`) has specific rules when it comes to handling punctuation or special characters. Namely, it seems that they are largely ignored on the first path of sorting, later used to resolve ties if necessary. Aside from that, it seems there are other rules, of which I don't fully understand, but that seem to be in conflict with our requirements for ordering this field.

What this means is that if there are two paths like the following:

* `a/z`
* `aa/z`

Since the default collation will initially ignore the slashes, the comparison it sees is between `az` and `aaz`, so it will sort these in the order `aa/z`, `a/z`. This is obviously in stark contrast to what we want, since these paths are supposed to represent file-system paths, and in that case, you would obviously sort them the other way around.

---
To resolve this, the collation for the Asset `path` field is set to `C.utf8`. The postgres [docs](https://www.postgresql.org/docs/current/collation.html#COLLATION-MANAGING-STANDARD) explain this a bit, but the following quote is fairly concise:

> The C and POSIX collations both specify “traditional C” behavior, in which only the ASCII letters “A” through “Z” are treated as letters, and sorting is done strictly by character code byte values.

As far as I can tell, `C.utf8` is just the C collation with Unicode encoding.


jjnesbitt:
It seems there's some difference between the postgres in CI and locally that I was unaware of. Looking into that now...

jjnesbitt:
@mvandenburgh This is good to go now.

mvandenburgh (approved):
To add to your explanation, I found this post to be informative about this as well - https://dba.stackexchange.com/a/240950

PR #1887, opened 2024-03-06
Always pull postgres image for local docker-compose

jjnesbitt:
In the process of submitting a recent PR, I ran into test failures that only occurred in CI and not locally. This was due to an out of date postgres image locally, which it turns out was over a year old.

This PR adds `pull_policy: always` to the postgres docker-compose service, to prevent this from occuring.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ version: '3'
 services:
   postgres:
     image: postgres:latest
+    pull_policy: always
     command: postgres -c log_lock_waits=on -c log_min_duration_statement=100
     environment:
       POSTGRES_DB: django


yarikoptic:
not sure then if it shouldn't be a general pattern since similar gotchas might happen with any container image.  IIRC we had similar situation in dandi-cli and thus added `DANDI_TESTS_PULL_DOCKER_COMPOSE` env var so we `docker-compose pull` by default before running `docker-compose run`.

```python
            if os.environ.get("DANDI_TESTS_PULL_DOCKER_COMPOSE", "1") not in ("", "0"):
                run(["docker-compose", "pull"], cwd=str(LOCAL_DOCKER_DIR), check=True)
```
so may be a better solution should be adding more of "Run `docker-compose pull`" to https://github.com/dandi/dandi-archive/blob/HEAD/DEVELOPMENT.md and adhering to such practice?

PR #1892, opened 2024-03-18
Design doc for generating Zarr Manifest Files

jwodder:
Unresolved issues:

- [ ] This doc currently specifies that only one manifest file per Zarr (the one reflecting the Zarr's current state) is kept around.  This is in contrast to @yarikoptic's prior art at <https://github.com/dandi/zarr-manifests>, which stores multiple manifest files for Zarrs which have changed over time.  Do we want the Archive to store separate manifest files for each version of each Zarr?

- [ ] The design of the `fields` key and entry arrays (copied from YOH's prior art) is odd.  I do not foresee any circumstances in which `fields` would have anything other than the recommended value (and `fields` values other than the recommended would be tricky to support in `dandidav`), so we might want to change the manifest file format in one of the following ways:

    - Eliminate the `fields` key and define the elements of entry arrays to
      always be version ID, last modified, size, and ETag in that order

        - Con: Manifest files become harder to read for people not sufficiently
          familiar with the format definition

    - Eliminate the `fields` key and change the entry arrays to objects with
      `versionId`, `lastModified`, `size`, and `ETag` fields

        - Con: Much larger manifest files

CC @yarikoptic @dandi/dandiarchive


yarikoptic:
> * Do we want the Archive to store separate manifest files for each version of each Zarr?

Yes. This is one of the main motivations behind this feature. So we could "cheaply" provide support for versioning of zarrs and this way then allow for their releases: ATM nobody can publish/release a versioned dandiset if there is zarr in it.  If we have specific version manifest for zarr -- we can get to that manifest / version no problem and thus can release zarr and access that specific version of zarr.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+Archive Behavior
+----------------
+
+Whenever Dandi Archive calculates the checksum for a Zarr in the Archive, it
+shall additionally produce a *manifest file* listing various information about
+the Zarr and its entries in the format described in the next section.  This
+manifest file shall be stored in the Archive's S3 bucket at the path
+`zarr-manifest/{zarr_id}.json`, where `{zarr_id}` is replaced by the ID of the
```
yarikoptic, line 22: here we indeed should introduce versioning similar (if not identical) to the one currently present in https://github.com/dandi/zarr-manifests/tree/master/zarr-manifests-v2-sorted .

jwodder:
@yarikoptic How should manifests for different versions of the same Zarr be organized in S3?  Naming the files after the Zarr checksum like in your repository isn't guaranteed to be stable, as a user could create a Zarr, then change one entry, then change that entry back to its original contents, resulting in an S3 object with three versions while there are only two different checksums for the Zarr across its history.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,150 @@
...
+Zarr.  The manifest file shall be world-readable, unless the Zarr is embargoed
+or belongs to an embargoed Dandiset, in which case appropriate steps shall be
+taken to limit read access to the file.
+
+If a manifest file is generated for a Zarr for which an earlier manifest file
+was already generated, the newer file shall replace the older.
+
+Manifest files shall also be generated for all Zarrs already in the Archive
+when this feature is first implemented.
```
yarikoptic, line 31: After versions support added above, here should follow description on what is to happen for publishing dandisets which have zarrs in them. 

We need to review/analyze what should now happen for zarr records or assets so we capture version (checksum) information for a zarr whenever it becomes part of the released dandiset. In case of blobs it is easy since blobs are not mutable. But with zarrs, since zarr could have multiple versions - we would need to make sure that published asset has versionId for zarr which would not be changed, whenever that asset zarr would be modified in draft version.

yarikoptic:
> then change that entry back to its original contents, resulting in an S3 object with three versions while there are only two different checksums for the Zarr across its history.

I think that should be fine --  since we would just carry about a manifest with a specific checksum of the content. So pretty much to have similarly to blobs a "context addressable storage" (of zarrs). Then given zarr id + checksum (as stored in DB, or in assets dumps which would have zarr id and its checksum) we can get needed manifest and provide access to that particular version, possibly the same one across different points in the version history.

doc/design/zarr-manifests.md
jwodder: Done: 40ebed86

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,148 @@
...
+where noted, the manifest file format defined herein matches the format used by
+the proof of concept.
+
+
+Archive Behavior
+----------------
+
+Whenever Dandi Archive calculates the checksum for a Zarr in the Archive, it
+shall additionally produce a *manifest file* listing various information about
+the Zarr and its entries in the format described in the next section.  This
```
yarikoptic, line 20: Add an explicit note that it would need to use an S3 API call to list keys versioned which might be more expensive.
May be state that optimizations could potentially be achieved to do only partial (e.g. subtree) reads for what is being changed by Zarr upload etc, but not part of this design document to be worked out.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,148 @@
...
+manifest file shall be stored in the Archive's S3 bucket at the path
+`zarr-manifest/{zarr_id}/{checksum}.json`, where `{zarr_id}` is replaced by the
+ID of the Zarr and `{checksum}` is replaced by the Dandi Zarr checksum of the
+Zarr at that point in time.  The manifest file shall be world-readable, unless
+the Zarr is embargoed or belongs to an embargoed Dandiset, in which case
+appropriate steps shall be taken to limit read access to the file.
+
+Manifest files shall also be generated for all Zarrs already in the Archive
+when this feature is first implemented.
+
```
yarikoptic, line 30: here needs to be analyzed and expanded on what would happen to 
- DANDI API calls:  where and how we should extend returned records with a specific version of zarr (since zarr id would not be enough)
- metadata records , i.e. `contentsUrl`, e.g. replacing direct links to `https://dandiarchive.s3.amazonaws.com/zarr/...` with `https://webdav.dandiarchive.org/zarr/...` URLs?
- review what happens upon publishing of a dandiset in the view of the above. E.g. that we do not need to do anything to zarr assets or should anything particular be done?

yarikoptic (changes requested):
That is a great start, but now needs to come a hard part - analysis of current API/behavior and proposal on how to augment it so we have clear association between dandiset asset and version of zarr for an asset, and what would happen on publish, and how people (and web UI) would provide access to published versions of dandisets with corresponding versions of zarrs

doc/design/zarr-manifests.md
jwodder: Done in f4bfc3e7.

doc/design/zarr-manifests.md
jwodder: @yarikoptic 

> - metadata records , i.e. `contentsUrl`, e.g. replacing direct links to `https://dandiarchive.s3.amazonaws.com/zarr/...` with `https://webdav.dandiarchive.org/zarr/...` URLs?

Why would we want to do this?  This would actually break `dandidav`.  Do you mean replacing the `https://api.dandiarchive.org/api/assets/{asset_id}/download/` `contentUrl`s of Zarr assets?

jwodder: @yarikoptic 

> DANDI API calls: where and how we should extend returned records with a specific version of zarr

Are you envisioning giving each Zarr a "version ID" field of some sort?  Seeing as the per-Zarr manifests are going to be distinguished solely based on checksum, couldn't the checksum just do double duty as the version ID?

yarikoptic: yes -- checksum (similarly somewhat to checksum of blobs) serves as "version ID" for zarrs.  But the difference from blobs is that blob's checksum would never change. But in current design we have mutable zarrs, so we need to associate zarr_id with a version in the asset, and likely mint a new asset when zarr gets changed -- is that what happening or not ATM in dandi-archive?

> Why would we want to do this? This would actually break `dandidav`. Do you mean replacing the `https://api.dandiarchive.org/api/assets/{asset_id}/download/` `contentUrl`s of Zarr assets?

this one we do not have to change since based on the information in the `{asset_id}` should be able to forward (that is what it does) to correct version of zarr on `webdav`. But we need to change the other URL which is currently points to `https://dandiarchive.s3.amazonaws.com/zarr` which would be "draft" (mutating) version of zarr.

jwodder: @yarikoptic 

> and likely mint a new asset when zarr gets changed -- is that what happening or not ATM in dandi-archive?

I don't believe the Archive does anything to the asset when a Zarr changes.  However, when dandi-cli uploads to a pre-existing Zarr, it does a `PUT /dandisets/{dandiset_id}/versions/{version_id}/assets/{asset_id}/` to set the asset's metadata and `zarr` fields in case they changed; I don't know whether the Archive would mint a new asset if the fields' values are unchanged.

> should be able to forward (that is what it does) to correct version of zarr on `webdav`

API download URLs for Zarrs do not currently redirect to webdav; they just return an error.  See, e.g., [this URL](https://api.dandiarchive.org/api/assets/9e92e32b-93e1-48d2-a46d-42165dacc4cd/download/) for a Zarr in 000108.

yarikoptic: > However, when dandi-cli uploads to a pre-existing Zarr, it does a `PUT /dandisets/{dandiset_id}/versions/{version_id}/assets/{asset_id}/` to set the asset's metadata and `zarr` fields in case they changed; I don't know whether the Archive would mint a new asset if the fields' values are unchanged.

how do you see archive behavior/API could be changed so we can assure consistent versioning of zarrs?



> API download URLs for Zarrs do not currently redirect to webdav; they just return an error.

please add to this design document the needed change in behavior that it would redirect to webdav for that version.



> > But we need to change the other URL which is currently points to https://dandiarchive.s3.amazonaws.com/zarr which would be "draft" (mutating) version of zarr.
> 
> The only way I could see S3 URLs for published Zarrs pointing to fixed content working (without changing too many things) would be if each Zarr in a published version was copied to a dedicated location (that is afterwards never mutated again) separate from the draft Zarrs upon publication, and then the content URL could point to the copy.

the whole idea with zarr manifests and webdav instance was to **avoid** needing for such a copy: if we operate on a specific version (checksum) of zarr through its manifest/webdav we have guaranteed access to that version even if zarr on s3 would mutate.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,199 @@
...
+* Add `zarr_version` field to …/assets/path/ results
+
+* Zarr `contentUrl`s:
+    - Make API download URLs for Zarrs redirect to dandidav
+    - Replace S3 URLs with webdav.{archive_domain}/zarr/ URLs?
+        - Document needed changes to dandidav?
+            - The bucket for the Archive instance will now be given on the command line (only required if a custom/non-default API URL is given)
+            - The bucket's region will have to be looked up & stored before starting the webserver
+            - Zarrs under `/dandisets/` will no longer determine their S3 location via `contentUrl`; instead, they will combine the Archive's bucket & region with the Zarr ID in the asset properties (templated into "zarr/{zarr_id}/")
```
yarikoptic, line 188: note to ourselves: requests to modify a Zarr from previous non draft version might be "hard to impossible". we should keep that in mind while developing life cycle here.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,200 @@
...
+    - The current `/zarr/{zarr_id}/…` endpoints operate on the most recent version of the Zarr
+    - `GET /zarr/{zarr_id}/versions/` (paginated)
+    - `GET /zarr/{zarr_id}/versions/{version_id}/` ?
+    - `GET /zarr/{zarr_id}/versions/{version_id}/files/[?prefix=...]` (paginated)
+        - The Zarr entry objects returned in `…/files/` responses (with & without `versions/{version_id}/`) will need to gain a `VersionId` field containing the S3 object version ID
+    - Nothing under /zarr/versions/ is writable over the API
+
+* Publishing Zarrs: Just ensure that the `zarr_version` in Zarr assets is frozen and that no entries/S3 object versions from the referenced version are ever deleted ?
+
+* Does garbage collection of old Zarr versions need to be discussed?
```
yarikoptic, line 200: I think it should indeed be touched on. And we should touch on ["trailing delete"](https://github.com/dandi/dandi-archive/blob/0a3524d774baf2d0df4c820fb0509af7b68bafcc/doc/design/s3-trailing-delete.md?plain=1#L57) to be disabled for `/zarr/` prefix on s3.

doc/design/zarr-manifests.md
jwodder: @yarikoptic 

> requests to modify a Zarr from a previous non draft version

What are you talking about?  A "non draft version" is a published version, and published versions and their contents can't be modified.

> we need association of zarr to a dandiset

Each Zarr is already associated with a Dandiset.  You can see a Zarr's Dandiset by requesting `/zarr/{zarr_id}/`.

doc/design/zarr-manifests.md
yarikoptic: related to this discussion on being unable to `/download/` zarr: filed

- https://github.com/dandi/dandi-archive/issues/1993

since with already deployed dandidav we can easily have it implemented even without waiting for the final version of the design.

doc/design/zarr-manifests.md
```diff
@@ -0,0 +1,220 @@
...
+Actual data is served from the Archive's S3 bucket, but the WebDAV server uses the manifest files to determine the structure of the Zarrs and the versions of the Zarrs' entries.
+Two "end-points" within that namespace are provided:
+
+- [webdav.dandiarchive.org/zarrs](https://webdav.dandiarchive.org/zarrs) -- all Zarrs across all dandisets, possibly with multiple versions. E.g. see [zarrs/057/f84/057f84d5-a88b-490a-bedf-06f3f50e9e62](https://webdav.dandiarchive.org/zarrs/057/f84/057f84d5-a88b-490a-bedf-06f3f50e9e62) which ATM has 3 versions.
+- [webdav.dandiarchive.org/dandisets/](https://webdav.dandiarchive.org/dandisets/)`{dandiset_id}/{version}/{path}/`. E.g. for aforementioned Zarr - https://webdav.dandiarchive.org/dandisets/000026/draft/sub-I48/ses-SPIM/micr/sub-I48_ses-SPIM_sample-BrocaAreaS09_stain-Somatostatin_SPIM.ome.zarr/ -- a specific version (the latest, currently [6efea0a8e95e67ecb5af7aa028dece14-18147--30560865836](https://webdav.dandiarchive.org/zarrs/057/f84/057f84d5-a88b-490a-bedf-06f3f50e9e62/6efea0a8e95e67ecb5af7aa028dece14-18147--30560865836.zarr/)).
```
jwodder, line 28: @yarikoptic Why is this mentioned here?  The `/dandisets/` hierarchy gets its information directly from S3; it does not use the Zarr manifest files.

yarikoptic: ah! good to know since I assumed that this one gets it also from the manifests!  So we need smth like

```suggestion
- [webdav.dandiarchive.org/dandisets/](https://webdav.dandiarchive.org/dandisets/)`{dandiset_id}/{version}/{path}/` although providing access to Zarr files, it **does not** uses manifest files but gets listing directly from S3, so points only to the most recent version (possibly not even finalized yet during upload).
```
but with that "not finalized"  aspect, we might (not yet sure) want to change that webdav behavior and also use manifest file here to ensure access to the legit version of zarr (not finalized might be partial etc).

doc/design/zarr-manifests.md
yarikoptic: I will mark this resolved for now

satra:
perhaps we can look into tensorstore (https://google.github.io/tensorstore/kvstore/s3/index.html#json-kvstore/s3) and which backend could work for simply providing the index. also this may work directly with neuroglancer which now has support for both icechunk and middleauth (https://github.com/google/neuroglancer/tree/master/src/kvstore).

PR #1911, opened 2024-03-27
Only include total count in the first page of list views

jjnesbitt:
Currently regardless of which page from a paginated endpoint is fetched, we return the count of the _entire_ queryset, which requires a full table scan. This is a problem for endpoints where optimization is key, since regardless of how efficient you structure the database queries, you'll always scan the _entire_ table to get the total number of objects that match query.

This can be wildly inefficient for queries with a large result set. If, for example, you were to list the assets from a dandiset containing 50k assets and page through them 100 at a time, you'd be making 500 requests, and for every request would scan the _entire_ table to count all of these assets, even though you're only fetching 100.

To fix this behavior, this PR introduces a few new pagination related classes. Since the existing implementation of DRF's  [`PageNumberPagination`](https://github.com/encode/django-rest-framework/blob/085b7e166ba80aa973645e5249b441f2dbdc0c96/rest_framework/pagination.py#L165) class (and the classes it invokes by default) involves many calls to the queryset `count` method from one path or another, specific classes and methods are overridden in order to provide pagination without invoking `count`, unless explicitly done so.

At the moment this new pagination is applied to all paginated views in the archive. If this is undesirable it could easily be applied to only the asset list endpoint, but I'm not a fan of that approach. It's more confusing to have two different behaviors across the archive, and the performance improvement will now benefit all of those endpoints.

The consequences of this change at the API surface are the following:
1. `count` will only be returned if `page` is unspecified or equals `1`, and `null` otherwise
2. `last` can no longer be specified as a page number. I don't think anyone was really using this to begin with so that should have no effect.
3. The page number controls will no longer be shown on the stock DRF page for paginated views, as that requires knowing the full count of the queryset on _every_ page.

As far as I can tell the dandi-cli won't be affected by this change, as it only checks for the checks for the count on the [first page](https://github.com/dandi/dandi-cli/blob/d5e2a76a3adc9105c77bf7eb37ed0bbae410a3a3/dandi/dandiapi.py#L326).

### DIFF
--- a/dandiapi/api/views/pagination.py
+++ b/dandiapi/api/views/pagination.py
@@ -1,0 +1,93 @@
+from __future__ import annotations
+
+from collections import OrderedDict
+
+from django.core.paginator import Page, Paginator
+from django.utils.functional import cached_property
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
+
+
+class LazyPage(Page):
+    """
+    A page class that doesn't call .count() on the queryset it's paging from.
+
+    This class should be used with `LazyPaginator` unless you know what you're doing.
+    """
+
+    # Override to store the real object list under self._object_last
+    def __init__(self, object_list, number, paginator):
+        self._object_list = list(object_list)
+        self.number = number
+        self.paginator = paginator
+
+    # Override to prevent returning the extra record
+    @cached_property
+    def object_list(self):
+        return self._object_list[: self.paginator.per_page]
+
+    # Because we fetch one extra object to check for more rows, we know that if the number of
+    # objects returned is the page size or less, we have no more pages.
+    def has_next(self) -> bool:
+        return len(self._object_list) > self.paginator.per_page
+
+    # Override to prevent calling `.count` on the queryset. To my knowledge we don't use this.
+    def end_index(self) -> int:
+        raise NotImplementedError
+
+
+class LazyPaginator(Paginator):
+    """A Paginator that doesn't call .count() on the queryset."""
+
+    # Set this to infinity so that inherited code doesn't assume we're done paginating
+    num_pages = float('inf')
+
+    def page(self, number):
+        """Return a page with one extra row, used to determine if there are more pages."""
+        number = self.validate_number(number)
+        bottom = (number - 1) * self.per_page
+
+        # Intentionally fetch one extra to see if there are any more pages left
+        top = bottom + self.per_page + 1
+
+        return self._get_page(self.object_list[bottom:top], number, self)
+
+    def _get_page(self, *args, **kwargs):
+        return LazyPage(*args, **kwargs)
+
+
+class LazyPagination(PageNumberPagination):
+    page_size = 100
+    max_page_size = 1000
+    page_size_query_param = 'page_size'
+    django_paginator_class = LazyPaginator
+
+    # Set to always false since we only know the full number of pages on page 1
+    display_page_controls = False
+
+    # Define as empty to prevent `get_page_number` from calling `count`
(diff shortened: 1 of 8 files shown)


mvandenburgh:
> `count` will only be returned if `page` is unspecified or equals `1``, and `null` otherwise

Is it possible to just exclude `count` from the response?

waxlamp:
> 2\. `last` can no longer be specified as a page number. I don't think anyone was really using this to begin with so that should have no effect.

It could be included in the same instances when `count` is returned, right?

jjnesbitt:
> Is it possible to just exclude `count` from the response?

Good question, I forgot to mention why this implementation is necessary. I initially tried to simply exclude `count` from the response, but the underlying pagination class calls `count` on the queryset in various places. This happens when calling `paginate_queryset`,  and by the time the response is generated, it's simply using the cached value of `count` on that page instance.

jjnesbitt:
> > 2. `last` can no longer be specified as a page number. I don't think anyone was really using this to begin with so that should have no effect.
> 
> It could be included in the same instances when `count` is returned, right?

The use case is someone specifying `page=last` to return the last page, and doesn't really apply to normal page queries. I suppose it's actually not strictly necessary to disallow this, since it won't apply to the pages where we're excluding `count`, so I could keep it.

jjnesbitt:
FYI, there are some flaky tests which should be fixed by #1910. I noticed a failure in this PR's CI due to that and re-ran the test, which this time succeeded. This may continue to happen if I update this branch.

dandiapi/api/views/pagination.py
```diff
@@ -0,0 +1,97 @@
...
+                ('count', None),
+                ('next', self.get_next_link()),
+                ('previous', self.get_previous_link()),
+                ('results', data),
+            ]
+        )
+
+        # Only get the full count if it's the first page
+        if self.page.number == 1:
+            page_dict['count'] = self.page.paginator.count
```
waxlamp, line 91: Can this just be "transplanted" into line 82 as a conditional expression? That is, something like `('count', self.page.paginator.count if self.page.number == 1 else None)`?

waxlamp (commented):
Looks good. Relying on @mvandenburgh for a substantive review.

dandiapi/api/views/pagination.py
jjnesbitt: Good idea, I've applied this change.

yarikoptic:
NICE digging!  Could someone though run this by https://github.com/encode/django-rest-framework/ , e.g. inquire on 
- either observation on the overhead is expected 
- whether DRF has some means to remove that overhead without providing an alternative paginator
- would they be interested to receive contribution to adopt proposed here paginator?

my points are 
- would be great to run by this discovery/observations with the authors of the framework used here
- see if they might come up with some less intrusive solution
- long term -- offload maintenance of a needed solution to DRF instead of maintaining custom solution here.

dandiapi/api/views/pagination.py
```diff
@@ -0,0 +1,93 @@
+from __future__ import annotations
```
mvandenburgh, line 1: It might be worth having a short comment blurb at the start of this file just explaining why this new `pagination` module exists. It's clear from the docstrings on each overridden class that they don't call `.count()`, but I think a top-level explanation on _why_ it's important that they don't call `.count()` would make this more readable.

mvandenburgh (approved):
LGTM, just left one suggestion to add a clarifying comment explaining the motivation for this reworked pagination as a whole. It's unfortunate that we have to override so much to accomplish this, but it all makes sense to me.

PR #1919, opened 2024-04-12
Update our instructions for installation to state newer versions of python and dandi-cli

yarikoptic:
@dandi/archive-maintainers is there an easy way to avoid hardcoding dandi-cli version here and take it from `cli-minimal-version` of https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/api/views/info.py#L65 ?

Change is "inspired" by a user who pointed me to that line which didn't work on his mac with OSX having python 3.7 systemwide

### DIFF
--- a/web/src/views/DandisetLandingView/DownloadDialog.vue
+++ b/web/src/views/DandisetLandingView/DownloadDialog.vue
@@ -102,10 +102,10 @@
               <v-list>
                 <v-list-item>
                   Install the Python client (DANDI CLI)
-                  in a Python 3.7+ environment using command:
+                  in a Python 3.8+ environment using command:
                 </v-list-item>
                 <v-list-item>
-                  <kbd>pip install "dandi>=0.13.0"</kbd>
+                  <kbd>pip install "dandi>=0.60.0"</kbd>
                 </v-list-item>
               </v-list>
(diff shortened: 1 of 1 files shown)


waxlamp:
> @dandi/archive-maintainers is there an easy way to avoid hardcoding dandi-cli version here and take it from `cli-minimal-version` of https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/api/views/info.py#L65 ?
> 
> Change is "inspired" by a user who pointed me to that line which didn't work on his mac with OSX having python 3.7 systemwide

One way would be to hit the `/server-info` endpoint and grab the version number from there. @mvandenburgh, @jjnesbitt, @marySalvi, can you think of other ways that don't involve asking the server? Can we build this value into both the server and client by using a common build-time resource somehow?

jjnesbitt:
> One way would be to hit the `/server-info` endpoint and grab the version number from there. @mvandenburgh, @jjnesbitt, @marySalvi, can you think of other ways that don't involve asking the server? Can we build this value into both the server and client by using a common build-time resource somehow?

Calling `/server-info` would be the easiest way imo, and isn't really a compromise since the client relies on server up-time anyway.

PR #1937, opened 2024-05-09
Add retries to sha256 checksum calculation task

### LINKED ISSUE #471: Monitor for AssetBlobs with undefined sha256
The sha256 checksum of newly uploaded blobs is calculated by an asynchronous task, which may fail. We should monitor the DB for older AssetBlobs which do not have a sh256 field, and have a way to conveniently recompute those checksums as necessary.

#254 is an instance of this happening.

jjnesbitt:
Closes #471

### DIFF
--- a/dandiapi/api/tasks/__init__.py
+++ b/dandiapi/api/tasks/__init__.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from celery import shared_task
+from celery.exceptions import SoftTimeLimitExceeded
 from celery.utils.log import get_task_logger
 
 from dandiapi.api.doi import delete_doi
@@ -15,6 +18,9 @@ from dandiapi.api.manifests import (
 from dandiapi.api.models import Asset, AssetBlob, Version
 from dandiapi.api.models.dandiset import Dandiset
 
+if TYPE_CHECKING:
+    from uuid import UUID
+
 logger = get_task_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


dandiapi/api/tasks/scheduled.py
```diff
@@ -63,6 +64,19 @@ def aggregate_assets_summary_task(version_id: int):
     version_aggregate_assets_summary(version)
 
 
+@shared_task(soft_time_limit=30)
+def recalculate_missing_sha256_checksums():
+    blobs_missing = AssetBlob.objects.filter(sha256__isnull=True).values_list('blob_id', flat=True)
```
yarikoptic, line 69: there should probably be some condition that blob creation time is 24 h ago or more since otherwise it might trigger on fresh blobs on which such a process already running or about to run (race condition)?

yarikoptic:
I think this would close https://github.com/dandi/dandi-archive/issues/471

dandiapi/api/tasks/scheduled.py
```diff
@@ -63,6 +64,24 @@ def aggregate_assets_summary_task(version_id: int):
...
+        sha256__isnull=True, created__lte=date_threshold
+    ).values_list('blob_id', flat=True)
+
+    num_missing = blobs_missing.count()
+    if num_missing == 0:
+        logger.info('No blobs with missing checksums found')
+        return
+
+    logger.info('Found %s blobs with missing checksums', num_missing)
+    for blob_id in blobs_missing:
```
mvandenburgh, line 81:
```suggestion
    for blob_id in blobs_missing.iterator():
```

It most likely won't ever happen that this queryset is so large that it becomes a problem, but since we're inside a celery task it costs nothing to use `.iterator()` here so :shrug:

mvandenburgh (changes requested):

jjnesbitt:
I've reworked this PR to instead simply add retries to the sha256 checksum calculation task. This largely overlaps functionally, and will be easier to incrementally determine exactly why this is happening (a cron job would simply mask and/or extend the problem indefinitely).

As a follow up PR, I'd like to investigate if adding a `sha256_status` field would be helpful in determining at exactly what step the checksum calculation workflow breaks down, if any are still missing,

mvandenburgh (approved):

Issue #1578, opened 2023-04-24
assets can still be missing sha256


yarikoptic:
prior instances of this issue could be found in [search](https://github.com/dandi/dandi-archive/search?q=assets+can+still+be+missing+sha256&type=issues) so the issue remains that there is no automated monitoring/fixing for the cases where some jobs failed for some reason.

While dataladifying -- got following error

```
2023-04-24T02:00:05-0400 [ERROR   ] backups2datalad: Dandiset 000251: sub-271/sub-271_ses-Ca-VS-VR-CLOOP-SOLID-1.nwb: Asset created more than a day ago but SHA256 digest has not yet been computed                                                                
2023-04-24T02:00:05-0400 [ERROR   ] backups2datalad: Dandiset 000251: sub-3044/sub-3044_ses-SU-VR-ALL-TP-17.nwb: Asset created more than a day ago but SHA256 digest has not yet been computed                                                                     
2023-04-24T02:00:06-0400 [ERROR   ] backups2datalad: Job failed on input <Dandiset 000251/draft>:                                                                                                                                                                  
Traceback (most recent call last):                                                                                                                                                                                                                                 
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/aioutil.py", line 173, in dowork                                                                                                                                                                         
    outp = await func(inp)                                                                                                                                                                                                                                         
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/datasetter.py", line 149, in update_dandiset                                                                                                                                                             
    changed = await self.sync_dataset(                                                                                                                                                                                                                             
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/datasetter.py", line 186, in sync_dataset                                                                                                                                                                
    await syncer.sync_assets(error_on_change)                                                                                                                                                                                                                      
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/syncer.py", line 46, in sync_assets                                                                                                                                                                      
    self.report.check()                                                                                                                                                                                                                                            
  File "/mnt/backup/dandi/dandisets/tools/backups2datalad/asyncer.py", line 105, in check                                                                                                                                                                          
    raise RuntimeError(                                                                                                                                                                                                                                            
RuntimeError: Errors occurred while downloading: 2 assets on server had no SHA256 hash despite advanced age     
```
[...]

PR #1938, opened 2024-05-10
Add calculate_sha256 management command to trigger (re)computation for a blob

yarikoptic:
Added ability to specify either blob id (where it applies) or asset id since that is the id we typically know for which a blob lacks sha256.

Refs:
- https://github.com/dandi/dandi-archive/issues/471
- https://github.com/dandi/dandi-archive/issues/1578

- #1937 

has a chance to properly address the underlying issue but IMHO having an interface to trigger such a task manually could still be valuable e.g. in case of that automation failing to perform for some reason etc.


yarikoptic:
In 4 days this non-draft PR will celebrate its 3rd month without feedback. @waxlamp please attend to decide on its destiny one way or another.

yarikoptic:
ping @waxlamp @jjnesbitt @mvandenburgh .  I felt like this is a trivial PR which could get either blessed or rejected speedily.

yarikoptic:
so, overall code is good and we could proceed?

waxlamp (approved):

Issue #1940, opened 2024-05-10
backend tests started to fail a week or so ago


yarikoptic:
Noticed while looking at the status of the PR and seeing that has failures in tests which are unrelated. Checked the actions page and saw that

![image](https://github.com/dandi/dandi-archive/assets/39889/9ddd707c-5fa6-4a8c-b3a5-69f96e48afe1)

dandibot:
:rocket: Issue was released in [`v0.3.85`](https://github.com/dandi/dandi-archive/releases/tag/v0.3.85) :rocket:

PR #1945, opened 2024-05-24
add handbook to welcome email

bendichter:


yarikoptic:
Makes total sense and is a no brainer. Tests are broken for awhile now
- https://github.com/dandi/dandi-archive/issues/1940
.  As such, I would just merge it .  Thank you @bendichter

PR #1946, opened 2024-05-24
In 1.14.3 it became client_config and .config was announced deprecated

### LINKED ISSUE #1940: backend tests started to fail a week or so ago

yarikoptic:
and caused .config to become None

and I believe that caused our tests to start failing etc

If tests succeed, Closes #1940

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
         # Production-only
         'django-composed-configuration[prod]>=0.23.0',
         'django-s3-file-field[s3]>=1.0.0',
-        'django-storages[s3]>=1.14.2',
+        'django-storages[s3]>=1.14.3',
         'gunicorn',
         # Development-only, but required
         'django-minio-storage',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -67,7 +67,7 @@
         # Production-only
         'django-composed-configuration[prod]>=0.23.0',
         'django-s3-file-field[s3]>=1.0.0',
-        'django-storages[s3]>=1.14.2',
+        'django-storages[s3]>=1.14.3',
```
mvandenburgh, line 70: Just to note for future reference - I think in this case, a better approach would have been to pin this to the older version, i.e.
`'django-storages[s3]==1.14.2',`. In cases where we're unsure about a dependancy upgrade, we usually do that and add a TODO comment to look into it in the future.

mvandenburgh (commented):
LGTM based on the changelog for `django-storages` :heavy_check_mark:

setup.py
yarikoptic: yeah, i had it in mind too and agree it would have been safer, but given that we have already lots of things to be developed, 
```shell
❯ git grep TODO | wc -l
69
```
and looking at the django-storages code decided that it was safe enough.

PR #1948, opened 2024-05-28
Draft: Reference count from cached object_list for lazy paginator

aaronkanzer:
Follow-up from https://github.com/dandi/dandi-archive/pull/1911

Unfortunately, returning `count` as `null` violates the requirement for `v-paginator` to operate as intended -- rather than extending logic in the UI, this PR references `cached_property` of `self.object_list` as a method of retrieving an accurate count.

### DIFF
--- a/admin_dev_startup.sh
+++ b/admin_dev_startup.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# This script can be used to launch the front-end and back-end applications locally.
+# Happy developing!
+
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 <image_name> <your.email@email.com>"
+  exit 1
+fi
+
+image_name=$1
+email=$2
+
+cd web/
+
+# Build Docker image (include the path to the Dockerfile's context)
+docker build -t $image_name -f Dockerfile.dev .
+
+# Run the Docker container for frontend in background
+docker run -d -v "$(pwd):/usr/src/app" -v /usr/src/app/node_modules -p 8085:8085 -e CHOKIDAR_USEPOLLING=true $image_name
+
+cd ..
+
+# Run Docker Compose commands for backend
+docker compose run --rm django ./manage.py migrate
+docker compose run --rm django ./manage.py createcachetable
+docker compose run --rm django ./manage.py createsuperuser
+docker compose run --rm django ./manage.py create_dev_dandiset --owner $email
+
+# Bring backend application to life!
+docker compose up
--- a/dandiapi/api/management/commands/create_dev_dandiset.py
+++ b/dandiapi/api/management/commands/create_dev_dandiset.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import hashlib
 
 from uuid import uuid4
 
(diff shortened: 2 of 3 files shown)


jjnesbitt:
This will become moot once #1947 is merged as this is only really supposed to apply to listing endpoints that _don't_ need to make use of the `count` field. Furthermore, I believe what this change actually does is return the length of the current page in the `count` field on each page. The `count` field is supposed to refer to the count of the entire dataset from which a single page is retrieved.

aaronkanzer:
> The `count` field is supposed to refer to the count of the entire dataset from which a single page is retrieved.

With the change in this PR, I'm seeing the appropriate behavior via a localhost API with this retrieved `count` value being the total count (cached) w/ the v-paginator not reverting back to a length of 1.

However, if this will become moot post-that fix, I can hold off -- I'm still a bit stumped though for any case where `count: null` is returned, since it breaks the underlying logic of the `v-paginator` UI component

Let me know -- thanks again for taking a look. If this change makes sense, I'll push some unit tests alongside.

jjnesbitt:
> With the change in this PR, I'm seeing the appropriate behavior via a localhost API & UI with this retrieved `count` value being the total count (cached) w/ the v-paginator not reverting back to a length of 1.

While it's true that [this](https://github.com/dandi/dandi-archive/blob/288b8545f63b6f8ce452d4490a2487d53ed3834d/dandiapi/api/views/pagination.py#L34-L36) function has a `cached_property` decorator on it, it's not caching its result for future requests to use it. It's caching the contents of the fetched page, so that if it were to be called multiple times within the same request, it will only actually run once. 

Accessing that cached property would be done with `self.page.object_list`, however, your change accesses `self.page.paginator.object_list`, which is a separate entity. That will return the `QuerySet` that was passed into the paginator (before pagination), and so when `len` is called on it, what actually occurs is that a list of _all_ assets in that version is fetched from the database and coerced into a list, and then the length measured and returned. You can observe this for yourself with swagger and the Django debug toolbar, there will be an extra query to the database at the end that looks like this

```sql
SELECT "api_asset"."id"
  FROM "api_asset"
 INNER JOIN "api_asset_versions"
    ON ("api_asset"."id" = "api_asset_versions"."asset_id")
 WHERE "api_asset_versions"."version_id" = 416
 ORDER BY "api_asset"."created" ASC
```

So this change is effectively just re-introducing the old behavior in a roundabout way.

> I'm still a bit stumped though for any case where count: null is returned, since it breaks the underlying logic of the v-paginator UI component

This was only really intended to be used by the asset list endpoint, which has no access in the UI. The asset paths endpoint _is_ accessed by the UI, and is what it used for the file browser in the web browser. That's located at `/assets/paths`, and in my above mentioned PR, is switched back to normal pagination (along with all other list endpoints). I had originally thought that not many places in the UI relied on that `count` field being present, but obviously I was mistaken.

jjnesbitt:
Anyway, if it's okay with you, I think we can close this PR. The change in #1947 will remove the use of this custom pagination class from all UI facing endpoints, and restrict it to just the asset list endpoint, which has no front-end utility.

aaronkanzer:
> Anyway, if it's okay with you, I think we can close this PR. The change in #1947 will remove the use of this custom pagination class from all UI facing endpoints, and restrict it to just the asset list endpoint, which has no front-end utility.

Sounds good -- thanks Jake for the reply here and all the explanation -- makes sense! Will close this PR in anticipation for #1947

PR #1952, opened 2024-06-03
Only include APPROVED users for stats on homepage

aaronkanzer:
Cc @kabilar -- per our fix in LINC

The current query on the landing page that shows stats does not filter by user type (e.g. it takes in account REJECTED uers)

This PR includes the proper query.


kabilar:
Thank you, @aaronkanzer.

Hi @waxlamp, we noticed this behavior on lincbrain.org where the `users` count on the homepage included those users with a `STATUS` of `APPROVED`, `PENDING`, `INCOMPLETE` and `REJECTED`.  Presumably the desired behavior is only to count the users who are `APPROVED`.

dandiapi/api/tests/test_stats.py
```diff
@@ -1,15 +1,18 @@
...
+
 
 @pytest.mark.django_db()
 def test_stats_baseline(api_client):
     assert api_client.get('/api/stats/').data == {
         'dandiset_count': 0,
         'published_dandiset_count': 0,
-        # django-guardian automatically creates an AnonymousUser
-        'user_count': 1,
+        # django-guardian automatically creates an AnonymousUser, but user is not approved
```
waxlamp, line 14:
```suggestion
```

I don't think the comment is needed any more.

dandiapi/api/tests/test_stats.py
```diff
@@ -33,10 +36,23 @@ def test_stats_published(api_client, published_version_factory):
...
+        UserMetadata.objects.create(user=user, status=status_value)
+        if status_value == UserMetadata.Status.APPROVED:
+            approved_user_count += 1
+
     stats = api_client.get('/api/stats/').data
 
-    # django-guardian automatically creates an AnonymousUser
-    assert stats['user_count'] == 2
+    # Assert that the user count only includes users with APPROVED status
+    assert stats['user_count'] == 1
```
waxlamp, line 54:
```suggestion
```

This assertion is no longer useful, nor even correct (in general; see suggestion in the block above that creates users).

dandiapi/api/tests/test_stats.py
```diff
@@ -33,10 +36,23 @@ def test_stats_published(api_client, published_version_factory):
...
+    User.objects.all().delete()
+
+    # Create users with different statuses
+    approved_user_count = 0
+    for status in UserMetadata.Status.choices:
+        status_value = status[0]
+        user = User.objects.create(username=f'{status_value.lower()}_user')
+        UserMetadata.objects.create(user=user, status=status_value)
+        if status_value == UserMetadata.Status.APPROVED:
+            approved_user_count += 1
```
waxlamp, line 49: I think this would be improved if it created multiple users of each type (keeping track of how many are `APPROVED`), rather than just one user of each status type.

(This is connected to my comment in the line below, where the `== 1` assertion was retained.)

dandiapi/api/models/asset.py
```diff
@@ -218,10 +218,7 @@ def is_different_from(
         if self.path != path:
             return True
 
-        if self.metadata != metadata:
-            return True
-
-        return False
+        return self.metadata != metadata
```
waxlamp, line 221: Please revert this change, as it is not connected to the PR's purpose.

(I believe the code was written this way for stylistic consistency with the previous `if` blocks, each of which tests a single condition.)

waxlamp (changes requested):

dandiapi/api/models/asset.py
waxlamp: I see now that this was done to avoid linting/formatting failures. But this was resolved in a different way, so merging master into this branch will resolve this problem.

yarikoptic:
@aaronkanzer what is your plans for this PR?

kabilar:
> @aaronkanzer what is your plans for this PR?

Sorry, @yarikoptic, not sure what you mean.  After discussion with @aaronkanzer, he will be coming back to these PRs over the next few weeks.

dandiapi/api/tests/test_stats.py
aaronkanzer: Updated

dandiapi/api/tests/test_stats.py
aaronkanzer: Updated

dandiapi/api/tests/test_stats.py
aaronkanzer: Updated

aaronkanzer:
@waxlamp thanks for the review -- I've updated the code accordingly to your review.

Cc @kabilar @yarikoptic

dandiapi/api/tests/test_stats.py
```diff
@@ -32,11 +34,37 @@ def test_stats_published(api_client, published_version_factory):
...
+                    'status': status_value,
+                    'questionnaire_form': None,
+                    'rejection_reason': ''
+                }
+            )
+            if not created:
+                pass
+            if status_value == UserMetadata.Status.APPROVED:
+                approved_user_count += 1
+            user_index += 1
```
mvandenburgh, line 62: I think this test could be rewritten to be more concise and readable. Maybe something like

```suggestion
    
    # Create multiple users with different statuses
    users_per_status = 3
    
    for status in UserMetadata.Status.choices:
        users = [user_factory(username=username, metadata__status=status[0]) for _ in range(users_per_status)]
    
    stats = api_client.get('/api/stats/').data

    # Assert that the user count only includes users with APPROVED status
    assert stats['user_count'] == approved_user_count
```

Some reasoning behind this:
- The nested loops and multiple counter variables make it difficult to discern what is going on in the test. 
- I think the test is more dynamic than it needs to be. It seems that the number of users created could vary depending on certain conditions; these types of tests should be deterministic when possible, IMO.
- `user_factory` should be preferred over directly calling `UserMetadata.objects.get_or_create` where possible

mvandenburgh (changes requested):

waxlamp (approved):

kabilar:
Thanks team.  Based on these changes it looks like the number of users listed on the homepage (https://dandiarchive.org/) went from 1908 to 1412.

PR #1954, opened 2024-06-06
Restrict updates to metadata `access` field

### LINKED ISSUE #1787: dandi:EmbargoedAccess metadata can be silently inconsistent with the actual embargo status

### LINKED ISSUE #1831: unembargo appears to not update access information
see this unembargoed dataset for example: https://dandiarchive.org/dandiset/000871

when unembargoing, the access information field of the metadata should be set to open access.

jjnesbitt:
Closes #1787
Closes #1831

It seems both these issues were due to us not prohibiting the `access` field from being updated. A user should never update the `access` field (as it is derived from the embargo status of the dandiset), nor is there a natural way for that to happen using the meditor. However, it was not actually disallowed in anyway, so the following situations were possible:

1. After the dandiset is embargoed, if the meditor uses old metadata in some way, it can overwrite this field with the old embargoed access field. All without the user being aware, since that field isn't available in the meditor
2. If someone were to simply use the API outside of the meditor, they could set that field to whatever they want, intentional or not.

~This PR updates that field to treat it as `computed` and thus not able to be changed from a simple metadata update.~ (See below)

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 import logging
 
+from dandischema.models import AccessType
 from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.validators import RegexValidator
@@ -167,6 +168,28 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
         ]
         return {key: metadata[key] for key in metadata if key not in computed_fields}
 
+    def _populate_access_metadata(self):
+        default_access = [{}]
+        access = self.metadata.get('access', default_access)
+
+        # Ensure access is a non-empty list
+        if not (isinstance(access, list) and access):
+            access = default_access
+
+        # Ensure that every item in access is a dict
+        access = [x for x in access if isinstance(x, dict)] or default_access
+
+        # Set first access item
+        access[0] = {
+            **access[0],
+            'schemaKey': 'AccessRequirements',
+            'status': AccessType.EmbargoedAccess.value
+            if self.dandiset.embargoed
+            else AccessType.OpenAccess.value,
+        }
+
+        return access
+
     def _populate_metadata(self):
         from dandiapi.api.manifests import manifest_location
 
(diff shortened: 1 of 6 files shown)


waxlamp (dismissed):
I had two questions, but if the answers are what I think, they don't block approval and merge of this PR:
1. The new test that checks for non-mutability of the access field relies solely on the addition of `access` to the `computed_fields` list in `Version.strip_metadata()`, right?
2. There is no added test to see whether the correct access field value pops out of an embargoed Dandiset--is that already covered by existing tests (i.e., those tests unaffected by adding `access` as an immutable field)?

As I said, aside from those questions... LGTM.

jjnesbitt:
This PR has changed slightly, as I realized that there are sometimes other fields in `access`, that may indeed require updating (`contactPoint` is one such example). As such, I've updated this change to instead restrict updates to `access` to be in the correct format, while still allowing extra fields to be present. So `access` is not really being treated as "computed", rather is just handled in a special way.

waxlamp (approved):

dandiapi/api/models/version.py
```diff
@@ -165,7 +166,41 @@ def strip_metadata(cls, metadata):
             'publishedBy',
             'manifestLocation',
         ]
-        return {key: metadata[key] for key in metadata if key not in computed_fields}
+        stripped = {key: metadata[key] for key in metadata if key not in computed_fields}
+
+        # Ensure specific fields not modified
```
mvandenburgh, line 171: I think this is a "what-not-why" comment (to borrow a saying from @danlamanna :slightly_smiling_face:). It's clear from reading the code that it is programatically removing certain fields, but I think the comment would be more useful if it explained _why_ we're excluding those fields from being modified, as imo it's not clear at first glance that it's at all related to embargo.

mvandenburgh (commented):
Just one improvement that I think can be made to a comment, otherwise this makes sense and lgtm

PR #1957, opened 2024-06-14
Lock dandisets during un-embargo

jjnesbitt:
This change will help prevent race conditions that could arise if modifications are allowed while un-embargo is ongoing.

### DIFF
--- a/dandiapi/api/mail.py
+++ b/dandiapi/api/mail.py
@@ -194,14 +194,14 @@ def build_dandisets_to_unembargo_message(dandisets: Iterable[Dandiset]):
     ]
     render_context = {**BASE_RENDER_CONTEXT, 'dandisets': dandiset_context}
     return build_message(
-        subject='DANDI: New Dandisets to un-embargo',
+        subject='DANDI: New Dandisets to unembargo',
         message=render_to_string('api/mail/dandisets_to_unembargo.txt', render_context),
         to=[settings.DANDI_DEV_EMAIL],
     )
 
 
 def send_dandisets_to_unembargo_message(dandisets: Iterable[Dandiset]):
-    logger.info('Sending dandisets to un-embargo message to devs at %s', settings.DANDI_DEV_EMAIL)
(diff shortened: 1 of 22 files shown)


waxlamp (approved):
This all looks good.

One note: I didn't realize that we prohibit publishing of embargoed Dandisets. I wonder if we should consider lifting that ban--I can see why it would be useful to be able to do so, though the wording we're using makes it confusing ("it's embargoed... how can I be publishing this?") when really "publish" mostly means to create an immutable snapshot.

jjnesbitt:
> One note: I didn't realize that we prohibit publishing of embargoed Dandisets. I wonder if we should consider lifting that ban--I can see why it would be useful to be able to do so, though the wording we're using makes it confusing ("it's embargoed... how can I be publishing this?") when really "publish" mostly means to create an immutable snapshot.

I thought it was a hard requirement that published dandisets be open access, from more of a NIH policy standpoint.

satra:
> I thought it was a hard requirement that published dandisets be open access, from more of a NIH policy standpoint.

yes, embargoed dandisets should not receive a published doi. publishing is only for non-embargoed dandisets.

in the future we may publish restricted dandisets, but we haven't implemented that yet. 

to speak about roni's point of freezing versions on embargoed data, that's a different feature which we don't have. we should discuss whether that feature should even exist.

jjnesbitt:
@waxlamp I had forgotten to include uploads as a part of the locking, as well as prohibit un-embargo in the first place if there were active uploads. Commits ef7c530 and b789645 cover that.

waxlamp (changes requested):
This looks good to me, with one requested change: within the exception names (and elsewhere as appropriate), please change the "UnEmbargo" capitalization to "Unembargo" (the word is spelled "unembargo", as "one word" and no hyphens, etc.).

waxlamp (changes requested):
The `UnEmbargo` spelling within the exception classes should be shifted to `Unembargo`.

waxlamp (approved):

PR #1965, opened 2024-07-02
Automate dandiset unembargo

jjnesbitt:
This automates the new unembargo workflow introduced in #1890. Now, when a dandiset unembargo is kicked off from the API, the unembargo task will immediately be put on the worker queue. If that task fails, both the devs and dandiset owners will be notified through an email, as well as through sentry. There's no retrying implemented for the dandiset unembargo task, as my thinking is that retrying an unknown error would simply cause the error to fail further. Instead, I think we should simply investigate and directly resolve any errors we might find.

~~The only qualm I have with this at the moment is that I wasn't able to find a good way to test that the desired email is sent when the dandiset unembargo task fails. I can test this myself locally when running the application in dev mode, but doing so in a testing context has proved more difficult, as it seems that either the `on_failure` hook isn't called, or the custom Task class isn't used at all. @mvandenburgh @danlamanna if you have any ideas here I'd be interested.~~ This has been addressed.


dandiapi/api/tasks/__init__.py
```diff
@@ -12,10 +13,18 @@
     write_dandiset_yaml,
 )
 from dandiapi.api.models import Asset, AssetBlob, Version
+from dandiapi.api.models.dandiset import Dandiset
 
 logger = get_task_logger(__name__)
 
 
+class UnembargoTask(Task):
+    def on_failure(self, *task_args, **task_kwargs):
```
danlamanna, line 22: I would probably just do this logic inline rather than as a base task, since I don't see this class being reused for other tasks.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -76,10 +120,8 @@ def unembargo_dandiset(*, user: User, dandiset: Dandiset):
...
-    dandiset.embargo_status = Dandiset.EmbargoStatus.UNEMBARGOING
-    dandiset.save()
-
-    # Send initial email to ensure it's seen in a timely manner
-    send_dandisets_to_unembargo_message(dandisets=[dandiset])
+    with transaction.atomic():
+        Dandiset.objects.filter(pk=dandiset.pk).update(
+            embargo_status=Dandiset.EmbargoStatus.UNEMBARGOING
+        )
+        transaction.on_commit(lambda: unembargo_dandiset_task.delay(dandiset.pk))
```
danlamanna, line 127: I haven't looked closely at the PR, but I'm wondering if the lack of [django_capture_on_commit_callbacks](https://pytest-django.readthedocs.io/en/latest/helpers.html#django-capture-on-commit-callbacks) is the cause of problems?

jjnesbitt: That wasn't actually the problem, this part was working correctly (all I needed was `transaction=True` in the test django_db mark). The problem was due to my overall implementation which you pointed out was unnecessary in your other comment.

dandiapi/api/tasks/__init__.py
jjnesbitt: Good point, not sure why I didn't think of that. This makes everything much simpler and easy to test!

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +39,66 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
+    logger.info('Updated %s asset blobs', updated)
+
+    # Set status to OPEN
+    Dandiset.objects.filter(pk=ds.pk).update(embargo_status=Dandiset.EmbargoStatus.OPEN)
+    logger.info('Dandiset embargo status updated')
+
+    # Fetch version to ensure changed embargo_status is included
+    # Save version to update metadata through populate_metadata
+    v = Version.objects.get(dandiset=ds, version='draft')
+    v.save()
```
mvandenburgh, line 94:
```suggestion
    v = Version.objects.select_for_update().get(dandiset=ds, version='draft')
    v.save()
```

`v` is potentially stale by the time `.save()` is called on it. If that's the case, `.save()` would potentially clobber any recent changes to `v`. Since we're already inside a transaction, I think using `select_for_update` is an easy way to prevent this.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +39,66 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
+        )
+    )
 
     client = get_boto_client(config=Config(max_pool_connections=100))
     with ThreadPoolExecutor(max_workers=100) as e:
-        for blob in embargoed_asset_blobs:
+        futures = [
             e.submit(_delete_asset_blob_tags, client=client, blob=blob)
+            for blob in embargoed_asset_blobs
+        ]
```
mvandenburgh, line 53: Should we considering batching these using `.iterator()` and `ichunked`? The Heroku worker dyno has 512MB of memory, so I wonder if this would use too much memory given enough assets

mvandenburgh (changes requested):

dandiapi/api/services/embargo/__init__.py
mvandenburgh: Also, since the `ThreadPoolExecutor` has a max_workers of 100, it seems there would be no downside to batching things provided that the batch size is >= 100

jjnesbitt: I think it's unlikely since we just return the `blob` value, but nevertheless it's better to err on the side of caution, to reduce memory consumption if possible.

waxlamp: I had a similar thought about the possible number of embargoed assets in a given Dandiset.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +38,65 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
+
+    # Check if any failed and raise exception if so
+    failed = [
+        blob for i, blob in enumerate(embargoed_asset_blobs) if futures[i].exception() is not None
+    ]
```
waxlamp, line 58: According to the [`Future` docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.exception) the `.exception()` method will wait forever for the job to complete without a `timeout` parameter. If something goes wrong with `_delete_asset_blob_tags`, what do we do to recover?

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +38,65 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
-def _unembargo_dandiset(dandiset: Dandiset):
-    # NOTE: Before proceeding, all asset blobs must have their embargoed tags removed in s3
+def _unembargo_dandiset(ds: Dandiset):
+    logger.info('Unembargoing Dandiset %s', ds.identifier)
+    logger.info('\t%s assets', ds.draft_version.assets.count())
+
+    if ds.embargo_status != Dandiset.EmbargoStatus.UNEMBARGOING:
+        raise DandiError(
+            message=f'Expected dandiset state {Dandiset.EmbargoStatus.UNEMBARGOING}, found {ds.embargo_status}',  # noqa: E501
+            http_status_code=500,
```
waxlamp, line 72: Is this really a 5xx error? What if a user hits the unembargo endpoint for an ineligible Dandiset? That wouldn't be a server error.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +38,65 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
+        )
+    if ds.uploads.all().exists():
+        raise DandisetActiveUploadsError(http_status_code=500)
```
waxlamp, line 75: Same thing here--a Dandiset with active uploads can be requested for unembargo through user error.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -33,28 +38,65 @@ def _delete_asset_blob_tags(client: S3Client, blob: str):
...
+
+    # Update embargoed flag on asset blobs
+    updated = AssetBlob.objects.filter(embargoed=True, assets__versions__dandiset=ds).update(
+        embargoed=False
+    )
+    logger.info('Updated %s asset blobs', updated)
+
+    # Set status to OPEN
+    Dandiset.objects.filter(pk=ds.pk).update(embargo_status=Dandiset.EmbargoStatus.OPEN)
+    logger.info('Dandiset embargo status updated')
```
waxlamp, line 89: Do these operations need to be part of a transaction?

dandiapi/api/models/dandiset.py
```diff
@@ -62,6 +62,10 @@ def identifier(self) -> str:
     def embargoed(self) -> bool:
         return self.embargo_status == self.EmbargoStatus.EMBARGOED
 
+    @property
+    def unembargo_in_progress(self) -> bool:
```
waxlamp, line 66: This might be better named as `unembargoing` but I don't have very strong feelings about it.

dandiapi/api/mail.py
```diff
@@ -43,8 +43,17 @@ def user_greeting_name(user: User, socialaccount: SocialAccount = None) -> str:
...
-def build_message(subject: str, message: str, to: list[str], html_message: str | None = None):
-    email_message = mail.EmailMultiAlternatives(subject=subject, body=message, to=to)
+def build_message(  # noqa: PLR0913
+    to: list[str],
+    subject: str,
+    message: str,
+    html_message: str | None = None,
+    cc: list[str] | None = None,
+    bcc: list[str] | None = None,
+):
```
waxlamp, line 53: I don't know if this PR is the place for it, but since you've altered this function already: I wonder if this function would be better off with keyword-only arguments. That would make the use of the function elsewhere much clearer at a glance.

dandiapi/api/templates/api/mail/dandiset_unembargo_failed.html
```diff
@@ -0,0 +1,6 @@
+There was an error during the unembargo of dandiset
+<a href="{{ dandiset.ui_link }}">{{dandiset.identifier }}</a>. This has been reported to the developers, and will be
+investigated as soon as possible. We hope to have this resolved soon! Please avoid making any major changes to this
+dandiset in the meantime, as that may hinder or delay the process of resolving this issue.
+<br><br><br>
+<em>You are receiving this email because you are an owner of this dandiset.</em>
```
waxlamp, line 6: This is good copy. We should add a note saying to reply to this email (assuming it is sent by info@dandiarchive.org) if you haven't heard from us in such-and-such amount of time.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: The `ThreadPoolExecutor` context manager only exits once all futures have completed, so by the time this line is reached, we can be sure all futures have completed, successfully or otherwise. `.exception()` will then return (not raise) an exception if any were raised in `_delete_asset_blob_tags`. Otherwise, it returns `None`.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: It'a a 5xx error here because this is an internal function, and is only intended to be called on dandisets for which the embargo status is `UNEMBARGOING`. The workflow is:

1. User calls `unembargo` endpoint, which calls `kickoff_dandiset_unembargo`
2. `kickoff_dandiset_unembargo` updates that dandiset's status to `UNEMBARGOING` in a transaction, which once completed, kicks off `unembargo_dandiset_task`
3. `unembargo_dandiset_task`, once picked up by a celery worker, calls this function, `unembargo_dandiset`

So at step 3, if the dandiset were to _not_ be in the `UNEMBARGOING` state, that would be an internal error.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: Same answer as above, this is also handled in `kickoff_dandiset_unembargo`

dandiapi/api/services/embargo/__init__.py
jjnesbitt: This function (`unembargo_dandiset`) is wrapped in the `transaction.atomic()` decorator.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: [720018b](https://github.com/dandi/dandi-archive/pull/1965/commits/720018b09785d49f760380c85e187243c7b61fd8)

dandiapi/api/services/embargo/__init__.py
jjnesbitt: [566295a](https://github.com/dandi/dandi-archive/pull/1965/commits/566295a9c2461cc2bdecc1cbc4224c46c834651e). I ended up using `chunked` instead of `ichunked`, since the chunk is accessed more than once.

dandiapi/api/models/dandiset.py
jjnesbitt: Having a property `unembargoing` felt a bit vague, and this felt more clear.

dandiapi/api/mail.py
jjnesbitt: This method is honestly pretty superfluous, as it simply wraps the construction of a `mail.EmailMultiAlternatives` class instance. I think it'd be better to remove it entirely.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: We discussed offline and decided that this specific problem isn't much of an issue, but that more care should be taken around timeouts. This will be fully addressed in follow-on work.

dandiapi/api/templates/api/mail/dandiset_unembargo_failed.html
jjnesbitt: [93d9392](https://github.com/dandi/dandi-archive/pull/1965/commits/93d93927d86fd10858ae6f51f5304ff9f848b77b)

dandiapi/api/mail.py
jjnesbitt: I'll file an issue for this.

mvandenburgh (approved):

PR #1973, opened 2024-07-10
Suppress lint error (SIM103)

jjnesbitt:
This linting rule goes against the intention of the function it's raised in, and enforcing it would make further modification of that function more difficult.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -218,7 +218,7 @@ class Asset(PublishableMetadataMixin, TimeStampedModel):
         if self.path != path:
             return True
 
-        if self.metadata != metadata:
+        if self.metadata != metadata:  # noqa: SIM103
             return True
 
         return False


mvandenburgh (approved):
I was struggling to figure out how to resolve this too when I saw the nightly CI started failing. I don't have any better ideas so this lgtm

jjnesbitt:
The pytest failures we're seeing here are unrelated to this change, and are present on other PRs as well.

waxlamp:
You *could* follow the style recommendation here with a `return self.metadata != metadata`, right? It's just that doing that would confuse human readers attempting to understand this function?

jjnesbitt:
We could, but I argue that is the wrong approach. This method is one by one checking specific cases that we know differentiate two assets, returning `True` when those cases are reached. So in that sense, we're systematically eliminating cases where the assets are different from each other, to reach the base case of them being equal, where we return `False`. Through the happenstance of the check that linting rule is flagging being the final one, it thinks we're doing a useless statement, but that's only because it's the _final_ check before we reach the base case. I think following that rule here simply makes the function harder to read and refactor.

waxlamp:
Yeah, that's a good way of explaining it and is in line with my comment. Thanks.

PR #1978, opened 2024-07-12
Use md5 hasher in tests to speed them up

mvandenburgh:
This speeds up tests by using a simpler md5 password hasher instead of the more performance intensive bcrypt hasher.

### DIFF
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -169,7 +169,9 @@ class TestingConfiguration(DandiMixin, TestingBaseConfiguration):
 
     @staticmethod
     def mutate_configuration(configuration: ComposedConfiguration):
-        # use md5 in testing for quicker user creation
+        # use md5 in testing for quicker user creation.
+        # TODO: remove this when it is implemented upstream
+        # Issue: https://github.com/kitware-resonant/django-composed-configuration/issues/211
         configuration.PASSWORD_HASHERS.insert(0, 'django.contrib.auth.hashers.MD5PasswordHasher')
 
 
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):

danlamanna (approved):

yarikoptic:
I am curious on how much of improvement that would result in, any guestimates?

mvandenburgh:
> I am curious on how much of improvement that would result in, any guestimates?

It seems not by much. In fact, it's actually a few seconds slower than some of the recent test runs on `master` :laughing: 

https://github.com/dandi/dandi-archive/actions/runs/9966017574/job/27537367397
https://github.com/dandi/dandi-archive/actions/runs/9923912981/job/27414622824
https://github.com/dandi/dandi-archive/actions/runs/9915723096/job/27397036453

Still, it is recommended by Django for speeding up tests https://docs.djangoproject.com/en/5.0/topics/testing/overview/#password-hashing.

yarikoptic:
". If you’re authenticating many users in your tests," which is I guess not the case here, thus my 1c : not sure if worth the hassle/later wondering on why there is that code.

mvandenburgh:
> ". If you’re authenticating many users in your tests," which is I guess not the case here, thus my 1c : not sure if worth the hassle/later wondering on why there is that code.

We do authenticate a decent amount of users in tests, so I'm surprised this doesn't have any impact on runtime. I guess not at the scale they are suggesting this for :shrug:. 

Anyways, I agree it's not worth the additional code.

PR #1979, opened 2024-07-16
Add dashboard link to view to emit user list as Mailchimp-compatible CSV

waxlamp:
This is needed to automate a somewhat tricky process when done by hand. The resulting CSV can be used to import the users to a Mailchimp account in order to keep the Mailchimp audience accurate when sending (infrequent) mass emails.

### DIFF
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -88,7 +88,7 @@ def mailchimp_csv_view(request: HttpRequest) -> HttpResponse:
     # In production, there's a placeholder user with a blank email that we want
     # to avoid.
     users = User.objects.filter(is_active=True).exclude(email='')
-    data = users.values_list('email', 'first_name', 'last_name')
+    data = users.values_list('email', 'first_name', 'last_name').iterator()
 
     response = HttpResponse(
         content_type='text/csv',
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -88,7 +88,9 @@ def mailchimp_csv_view(request: HttpRequest) -> StreamingHttpResponse:
     # In production, there's a placeholder user with a blank email that we want
     # to avoid.
     users = User.objects.filter(is_active=True).exclude(email='')
-    data = users.values_list('email', 'first_name', 'last_name').iterator()
+
+    fieldnames = ['email', 'first_name', 'last_name']
+    data = users.values(*fieldnames).iterator()
 
     def streaming_output():
         """A generator that "streams" CSV rows using a pseudo-buffer."""
@@ -100,10 +102,10 @@ def mailchimp_csv_view(request: HttpRequest) -> StreamingHttpResponse:
                 return value
 
         # Yield back the rows of the CSV file.
-        writer = csv.writer(Echo())
-        yield writer.writerow(('Email Address', 'First Name', 'Last Name'))
+        writer = csv.DictWriter(Echo(), fieldnames=fieldnames)
+        yield writer.writeheader()
         for row in data:
-            yield writer.writerow(data)
+            yield writer.writerow(row)
 
     response = StreamingHttpResponse(
         streaming_output(),
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -92,7 +92,7 @@ def mailchimp_csv_view(request: HttpRequest) -> StreamingHttpResponse:
     fieldnames = ['email', 'first_name', 'last_name']
     data = users.values(*fieldnames).iterator()
 
-    def streaming_output():
+    def streaming_output() -> Iterator[str]:
         """A generator that "streams" CSV rows using a pseudo-buffer."""
 
         # This class implements a filelike's write() interface to provide a way
--- a/dandiapi/api/views/dashboard.py
+++ b/dandiapi/api/views/dashboard.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import csv
 from typing import TYPE_CHECKING
 
 from django.conf import settings
@@ -7,7 +8,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.db.models import Exists, OuterRef
-from django.http import HttpRequest, HttpResponseRedirect
+from django.http import HttpRequest, HttpResponse, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
 from django.views.decorators.http import require_http_methods
 from django.views.generic.base import TemplateView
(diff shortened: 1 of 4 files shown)


kabilar:
Thank you @waxlamp.

mvandenburgh:
We already have this feature as a Django admin action (code is [here](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/admin.py#L66-L72)), which IMO is more appropriate than a management command for this particular action. I'm not aware of what the driving use case is behind this, though. Any thoughts @waxlamp (or anyone else)?

waxlamp:
> We already have this feature as a Django admin action (code is [here](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/admin.py#L66-L72)), which IMO is more appropriate than a management command for this particular action. I'm not aware of what the driving use case is behind this, though. Any thoughts @waxlamp (or anyone else)?

Thanks Mike, I hadn't realized that (obviously).

The existing functionality doesn't cover the new use case, which is to export a CSV file consisting of just the active users' email addresses, first, and last names. I'm sure we could extend the existing code to cover this as well. I'll reach out to you to discuss further.

dandiapi/api/management/commands/mailchimp_csv.py
```diff
@@ -0,0 +1,24 @@
...
+    """Generate a Mailchimp-compatible CSV file of all (active) users."""
+    users = User.objects.all()
+    if not all_users:
+        users = users.filter(is_active=True)
+
+    data = users.values_list('email', 'first_name', 'last_name')
+
+    writer = csv.writer(sys.stdout)
+    writer.writerow(('Email Address', 'First Name', 'Last Name'))
+    writer.writerows(d for d in data if d[0])
```
asmacdo, line 24:
```suggestion
   #  Do not include the empty user ('', '', '')
    writer.writerows(d for d in data if d[0])
```

waxlamp:
@mvandenburgh, attempt number two -- in the dashboard view, I added a "navbar" link to a new view that generates the necessary CSV. Please take a look and let me know if there's a better way to accomplish this. In particular, perhaps I don't need to manufacture a URL to the new view and can somehow use the view function directly, rather than by reversing the named view? Or really any other improvements.

dandiapi/api/views/dashboard.py
```diff
@@ -82,6 +83,26 @@ def _users(self):
...
 
 
+def mailchimp_csv_view(request: HttpRequest) -> HttpResponse:
+    """Generate a Mailchimp-compatible CSV file of all active users."""
+    # In production, there's a placeholder user with a blank email that we want
+    # to avoid.
+    users = User.objects.filter(is_active=True).exclude(email='')
+    data = users.values_list('email', 'first_name', 'last_name')
+
+    response = HttpResponse(
```
danlamanna, line 93: I would use [StreamingHttpResponse](https://docs.djangoproject.com/en/5.0/ref/request-response/#streaminghttpresponse-objects) if possible.

dandiapi/api/views/dashboard.py
```diff
@@ -82,6 +83,26 @@ def _users(self):
         )
 
 
+def mailchimp_csv_view(request: HttpRequest) -> HttpResponse:
+    """Generate a Mailchimp-compatible CSV file of all active users."""
+    # In production, there's a placeholder user with a blank email that we want
+    # to avoid.
+    users = User.objects.filter(is_active=True).exclude(email='')
+    data = users.values_list('email', 'first_name', 'last_name')
```
danlamanna, line 91: Maybe use [.iterator](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#iterator).

dandiapi/api/views/dashboard.py
```diff
@@ -82,6 +83,26 @@ def _users(self):
...
+
+    response = HttpResponse(
+        content_type='text/csv',
+        headers={
+            'Content-Disposition': 'attachment; filename="dandi_users_mailchimp.csv"',
+        },
+    )
+    writer = csv.writer(response)
```
danlamanna, line 99: Now I'm being pedantic, but I find [csv.DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter) to be harder to misuse.

dandiapi/api/views/dashboard.py
waxlamp: Done in 88d9798884bedd0652f4b9a9ce605b8b4a1de893.

dandiapi/api/views/dashboard.py
waxlamp: Done in e2808d15baf4643352924eaecffea9643a26d9af.

dandiapi/api/views/dashboard.py
```diff
@@ -82,6 +83,39 @@ def _users(self):
         )
 
 
+def mailchimp_csv_view(request: HttpRequest) -> StreamingHttpResponse:
+    """Generate a Mailchimp-compatible CSV file of all active users."""
+    # In production, there's a placeholder user with a blank email that we want
+    # to avoid.
+    users = User.objects.filter(is_active=True).exclude(email='')
```
mvandenburgh, line 90:
```suggestion
    users = User.objects.filter(metadata__status=UserMetadata.Status.APPROVED).exclude(email='')
```

The built-in `is_active` field is used to determine if the user can login or not. A PENDING user (i.e. a user waiting for approval) is still allowed to login, they just can't do most operations, so they are actually have `is_active=True`.

dandiapi/api/views/dashboard.py
```diff
@@ -82,6 +83,39 @@ def _users(self):
...
+    data = users.values_list('email', 'first_name', 'last_name').iterator()
+
+    def streaming_output():
```
mvandenburgh, line 93:
```suggestion
    def streaming_output() -> Iterator[tuple[str, str, str]]:
```

dandiapi/urls.py
```diff
@@ -105,6 +106,7 @@ def to_url(self, value):
     path('admin/', admin.site.urls),
     path('dashboard/', DashboardView.as_view(), name='dashboard-index'),
     path('dashboard/user/<str:username>/', user_approval_view, name='user-approval'),
+    path('dashboard/mailchimp', mailchimp_csv_view, name='mailchimp-csv'),
```
mvandenburgh, line 109:
```suggestion
    path('dashboard/mailchimp/', mailchimp_csv_view, name='mailchimp-csv'),
```

Django will automatically redirect, but just to be consistent with the rest of the routes.

mvandenburgh (changes requested):

dandiapi/api/views/dashboard.py
waxlamp: IIUC, this generator yields strings. The reason we're using that `Echo` pseudo-buffer thing is to cause the  `writeheader()` and `writerow()` methods to return the string that they write. (Apparently, those methods just return the result of calling `.write()` on the underlying filelike object. By contrast, if we did this with a real filelike, whose `.write()` method returns the number of bytes written, we'd end up sending those byte counts to the streaming response.) It's sort of a mess 🫤 but it works.

If my analysis is correct, I'll update the suggestion to use `Iterator[str]` and commit it. Let me know what you think.

mvandenburgh: Ah you're right, I verified it locally and it is indeed a `Iterator[str]`.

mvandenburgh (approved):

PR #1986, opened 2024-07-23
Fix N query problem with VersionStatusFilter

jjnesbitt:
The queryset pulled from `VersionAdmin` when used in this way was causing Django to generate a `SELECT DISTINCT(status, version)` instead of just a `SELECT DISTINCT(status)`. It seems there's about 234 unique `version` values in the database as of now, which was causing 234 `SELECT COUNT(*)` queries being made when `qs.filter(status=status).count()` was called. This was causing this page to crash.

This PR replaces that queryset with one that does the same thing in one query.

### DIFF
--- a/dandiapi/api/admin.py
+++ b/dandiapi/api/admin.py
@@ -135,12 +135,17 @@ class VersionStatusFilter(admin.SimpleListFilter):
     title = 'status'
     parameter_name = 'status'
 
-    def lookups(self, request, model_admin):
-        qs = model_admin.get_queryset(request)
-        for status in qs.values_list('status', flat=True).distinct():
-            count = qs.filter(status=status).count()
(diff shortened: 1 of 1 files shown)


dandiapi/api/admin.py
```diff
@@ -135,12 +135,17 @@ class VersionStatusFilter(admin.SimpleListFilter):
     title = 'status'
     parameter_name = 'status'
 
-    def lookups(self, request, model_admin):
-        qs = model_admin.get_queryset(request)
-        for status in qs.values_list('status', flat=True).distinct():
-            count = qs.filter(status=status).count()
```
mvandenburgh, line 141: 😮

mvandenburgh (approved):
Nice find!

PR #1989, opened 2024-07-29
Re-validate version metadata during unembargo

jjnesbitt:
This addresses what I believe to be the cause of https://github.com/dandi/helpdesk/issues/158. Previously, when a dandiset was unembargo, we re-populated the version metadata but never validated it. This is in contrast to the normal flow of updating a version, which sets its status to `PENDING` and is eventually validated.

Now, this is performed inline as a part of unembargo.

@danlamanna Curious about your thoughts on my changes and additional test for `validate_version_metadata`. Previously, we assumed that versions with a status besides `PENDING` were passed into the function correctly, and were concurrently modified, changing the status. However, the function could easily be called mistakenly with a different version status, and so I think the log message should be updated to reflect that.


kabilar:
Thank you, Jake.

mvandenburgh (approved):

PR #1991, opened 2024-08-04
Separate core model logic from top-level asset service layer functions

jjnesbitt:
Depends on #1886 

As a part of the audit backend PR, I noticed that we do something peculiar in the `change_asset` service layer function, which required an undesirable workaround for audit. The `change_asset` function at its core just calls `remove_asset_from_version` and `add_asset_to_version` sequentially. This achieves the desired outcome, but with the following downsides:

1. Each service layer function (including `change_asset`) contains its own set of checks and filters to ensure the asset being created/modified/destroyed is not malformed. This is largely similar between the services, but they each differ slightly. So when `remove_asset_from_version` and `add_asset_to_version` are called from `change_asset`, they are each making checks that were already made in `change_asset`, making the overall operation inefficient.
2. This general architecture of calling one service function from another is not ideal, as it means that those two top level service layer functions have the responsibility of both their intended service layer purpose, as well as needing to conform to being called by _other_ service layer functions, which could potentially diverge in use case.
3. The core model logic is intertwined with the service layer functions, making it impossible to elegantly perform the low level model operations without the service layer side effects. This lead to the `do_audit` flag being introduced into `add_asset_to_version` and `remove_asset_from_version`, to workaround this limitation.

This PR separates out the core model logic from those two respective functions, allowing all of them, including `change_asset`, to utilize them. This removes the need for the `do_audit` flag, and generally cleans up the asset service layer a bit.

Once #1886 is merged, this can be rebased off master and fully evaluated.

### DIFF
--- a/dandiapi/api/services/asset/__init__.py
+++ b/dandiapi/api/services/asset/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from django.db import transaction
+from django.utils import timezone
 
 from dandiapi.api.asset_paths import add_asset_paths, delete_asset_paths, get_conflicting_paths
 from dandiapi.api.models.asset import Asset, AssetBlob
@@ -59,9 +60,9 @@ def _add_asset_to_version(
     add_asset_paths(asset, version)
 
     # Trigger a version metadata validation, as saving the version might change the metadata
-    version.status = Version.Status.PENDING
-    # Save the version so that the modified field is updated
-    version.save()
+    Version.objects.filter(id=version.id).update(
+        status=Version.Status.PENDING, modified=timezone.now()
+    )
 
     return asset
 
@@ -72,9 +73,9 @@ def _remove_asset_from_version(*, asset: Asset, version: Version):
     version.assets.remove(asset)
 
     # Trigger a version metadata validation, as saving the version might change the metadata
-    version.status = Version.Status.PENDING
-    # Save the version so that the modified field is updated
-    version.save()
+    Version.objects.filter(id=version.id).update(
+        status=Version.Status.PENDING, modified=timezone.now()
+    )
 
 
 def change_asset(  # noqa: PLR0913
--- a/dandiapi/api/services/asset/__init__.py
+++ b/dandiapi/api/services/asset/__init__.py
@@ -44,6 +44,39 @@ def _create_asset(
     return asset
 
 
+def _add_asset_to_version(
+    *,
+    version: Version,
+    asset_blob: AssetBlob | None,
+    zarr_archive: ZarrArchive | None,
+    metadata: dict,
+) -> Asset:
+    path = metadata['path']
+    asset = _create_asset(
+        path=path, asset_blob=asset_blob, zarr_archive=zarr_archive, metadata=metadata
+    )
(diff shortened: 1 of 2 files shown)


dandiapi/api/services/asset/__init__.py
```diff
@@ -140,59 +170,39 @@ def add_asset_to_version(  # noqa: PLR0913, C901
...
+        if (
+            asset_blob is not None
+            and asset_blob.embargoed
+            and version.dandiset.embargo_status == Dandiset.EmbargoStatus.OPEN
+        ):
             asset_blob.embargoed = False
             asset_blob.save()
+            transaction.on_commit(
+                lambda: remove_asset_blob_embargoed_tag_task.delay(blob_id=asset_blob.blob_id)
+            )
```
mvandenburgh, line 185:
```suggestion
            remove_asset_blob_embargoed_tag_task.delay_on_commit(blob_id=asset_blob.blob_id)
```

dandiapi/api/services/asset/__init__.py
```diff
@@ -44,6 +44,39 @@ def _create_asset(
...
+    asset = _create_asset(
+        path=path, asset_blob=asset_blob, zarr_archive=zarr_archive, metadata=metadata
+    )
+    version.assets.add(asset)
+    add_asset_paths(asset, version)
+
+    # Trigger a version metadata validation, as saving the version might change the metadata
+    version.status = Version.Status.PENDING
+    # Save the version so that the modified field is updated
+    version.save()
```
mvandenburgh, line 64: Since the version isn't necessarily locked here, it might be unsafe to call .save() on it like this. Couldn't this just be

```suggestion
    Version.objects.filter(id=version.id).update(status=Version.Status.PENDING, modified=timezone.now())
```

dandiapi/api/services/asset/__init__.py
```diff
@@ -44,6 +44,39 @@ def _create_asset(
...
+
+def _remove_asset_from_version(*, asset: Asset, version: Version):
+    # Remove asset paths and asset itself from version
+    delete_asset_paths(asset, version)
+    version.assets.remove(asset)
+
+    # Trigger a version metadata validation, as saving the version might change the metadata
+    version.status = Version.Status.PENDING
+    # Save the version so that the modified field is updated
+    version.save()
```
mvandenburgh, line 77: Same idea as above

```suggestion
    Version.objects.filter(id=version.id).update(status=Version.Status.PENDING, modified=timezone.now())
```

mvandenburgh (changes requested):

dandiapi/api/services/asset/__init__.py
jjnesbitt: Since this requires ensuring we're on celery `>=5.4`, and we have several other places in the codebase where this could be leveraged, I'd rather do this in a separate PR.

jjnesbitt: https://github.com/dandi/dandi-archive/issues/2006

jjnesbitt:
It seems the current linting errors are from a rule change in ruff, affecting our pytest marks. I'd like to defer that fix to a separate PR.

mvandenburgh (approved):

PR #2008, opened 2024-08-20
Remove `Asset.previous` field

mvandenburgh:
Now that audit is implemented, we can remove the `previous` field from the `Asset` model. This field was used to store the previous version of an asset, but it is no longer needed since we can track changes using the audit log.

This also unblocks garbage collection, which will follow this PR.

(GC docs - 
https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-1.md
https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-uploads-asset-blobs.md
)

Keeping this as draft until it's ready to merge, since we'll want to perform a DB backup before doing so.


mvandenburgh:
Keeping this as draft until it's ready to merge, since we'll want to perform a DB backup before doing so.

dandiapi/api/fixtures/playwright.json.xz
jjnesbitt, line 1: Just curious, why does this file need to be changed as a part of this PR? Since I don't believe the API surface is changing at all.

waxlamp:
> Keeping this as draft until it's ready to merge, since we'll want to perform a DB backup before doing so.

I think we talked about it already, but can you remind me why we need a DB backup before rolling this out? We have the DB snapshot just before audit rolled out, and since that time everything we want to track should be in the audit table. We could take another DB backup out of an abundance of caution, but it's not strictly necessary, is that right?

(The tradeoff here is just a downtime if we want to take the backup for safety purposes.)

satra:
@waxlamp - was the previous link captured in the audit log? i.e. did audit backfill old entries based on the database?

waxlamp:
> was the previous link captured in the audit log? i.e. did audit backfill old entries based on the database?

No, instead of backfilling the audit records table with that information, we captured a database backup that preserves whatever historical information exists, giving us the option of performing the backfill at any time. The major problem will be that most of the information that we currently capture in audit records is not recoverable from the previous links (see the [audit design doc](https://github.com/dandi/dandi-archive/blob/master/doc/design/audit-trail-mvp.md#asset-chain-analysis) for details).

If we decide we want to do this backfill, we should open an issue and discuss it there. Let me know what you think.

dandiapi/api/fixtures/playwright.json.xz
mvandenburgh: That file is a bunch of assets serialized to JSON. It includes the `previous` field, so it needed to be updated to remove it from each asset.

dandiapi/api/services/asset/__init__.py
```diff
@@ -127,10 +127,6 @@ def change_asset(  # noqa: PLR0913
             metadata=new_metadata,
         )
 
-        # Set previous asset and save
-        new_asset.previous = asset
-        new_asset.save()
-
         audit.update_asset(dandiset=version.dandiset, user=user, asset=new_asset)
```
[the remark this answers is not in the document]
waxlamp: Yes I believe that's the case. You reconstruct the history of an asset by looking at the audit records relating to it. Currently that's through the path; in the future, we may be able to stop manufacturing new `Asset` objects, enabling us to track changes by keying on the asset ID.

waxlamp (approved):

PR #2010, opened 2024-08-20
Derive asset `access` field from asset blob

### LINKED ISSUE #1997: Asset's `access/status` is `dandi:OpenAccess` even though it is uploaded to an embargoed Dandiset
Hi team, I recently created an embargoed Dandiset (001089) and uploaded some assets.  When I look at the `Asset metadata`, the `access/status` key is set to `dandi:OpenAccess`.   This is not a blocker, but I wanted to make sure that this is the desired behavior.

Below is the entire metadata for a simple asset (`dataset_description.json`) that is apart of this embargoed Dandiset.
```
HTTP 200 OK
Allow: GET, PUT, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "access": [
        {
            "status": "dandi:OpenAccess",
            "schemaKey": "AccessRequirements"
        }
    ],
    "schemaKey": "Asset",
    "dateModified": "2024-08-07T16:36:55.757493-05:00",
    "schemaVersion": "0.6.8",
    "encodingFormat": "application/json",
    "wasGeneratedBy": [
        {
            "id": "urn:uuid:5e169c27-abeb-4697-b78f-0b126421af38",
            "name": "Metadata generation",
            "endDate": "2024-08-07T16:36:55.757472-05:00",
            "schemaKey": "Activity",
            "startDate": "2024-08-07T16:36:55.757472-05:00",
            "description": "Metadata generated by DANDI cli",
            "wasAssociatedWith": [
                {
                    "url": "https://github.com/dandi/dandi-cli",
                    "name": "DANDI Command Line Interface",
                    "version": "0.62.4+17.g9ae6128",
                    "schemaKey": "Software",
                    "identifier": "RRID:SCR_019009"
                }
            ]
        }
    ],
    "blobDateModified": "2024-08-07T16:35:07.344006-05:00",
    "id": "dandiasset:c39a9d1d-efe2-4574-8bed-1564f44abf5b",
    "path": "dataset_description.json",
    "identifier": "c39a9d1d-efe2-4574-8bed-1564f44abf5b",
    "contentUrl": [
        "https://api.dandiarchive.org/api/assets/c39a9d1d-efe2-4574-8bed-1564f44abf5b/download/",
        "https://dandiarchive.s3.amazonaws.com/blobs/3fd/8e1/3fd8e186-73fc-494e-9da8-17f229af2f38"
    ],
    "contentSize": 332,
    "digest": {
        "dandi:dandi-etag": "8f4fdbbf7e606861d31e638f10e4caef-1",
        "dandi:sha2-256": "8db7c6e61fb3a24d4c168e091f66e4d8b40961b45a635d48c2cac4d8087f80d1"
    },
    "@context": "https://raw.githubusercontent.com/dandi/schema/master/releases/0.6.8/context.json"
}
```

Thank you.

jjnesbitt:
Closes #1997

Previously, this field was left unaltered, which meant it assumed whatever value was provided during creation through the API. This meant a default of `dandi:OpenAccess` most of the time, but it could have actually had any value. This has led to almost all of the currently embargoed assets to mistakenly display an `access` value of `dandi:OpenAccess`.

This PR changes this behavior by instead using the asset blob `embargoed` field to determine the value of 'access'.

TODO
- [ ] migration


dandiapi/api/migrations/0011_asset_access_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+from django.db.models.expressions import RawSQL
+
+
+def remove_access_fields(apps, _):
+    Asset = apps.get_model('api.Asset')
+
+    # Use the postgres jsonb '-' operator to delete the 'access' field from metadata
+    Asset.objects.filter(metadata__access__isnull=False).update(
+        metadata=RawSQL("metadata - 'access'", [])
+    )
```
jjnesbitt, line 14: Just a note, I tested this locally and it works as expected. Here is the generated SQL:

```sql
UPDATE "api_asset"
   SET "metadata" = (metadata - 'access')
 WHERE ("api_asset"."metadata" -> 'access') IS NOT NULL
```

mvandenburgh:
The `AuditRecord` model also stores a copy of `Asset.metadata`, which means that any asset audit records created up until this PR is merged will have an `access` field, while future ones won't.

This also made me realize we're just storing the `metadata`, not `full_metadata` in the asset audit record; i.e. all the "computed" fields are being left out of the audit records. I wonder if we should do the latter (@waxlamp).

jjnesbitt:
> The `AuditRecord` model also stores a copy of `Asset.metadata`, which means that any asset audit records created up until this PR is merged will have an `access` field, while future ones won't.

Since the `access` field for assets has always been superfluous, I think it would be appropriate to also purge it from the existing audit records. What do you think @waxlamp?

waxlamp:
> This also made me realize we're just storing the `metadata`, not `full_metadata` in the asset audit record; i.e. all the "computed" fields are being left out of the audit records. I wonder if we should do the latter (@waxlamp).

I'm open to this. If the full metadata can be derived from the "small" metadata, I think leaving it as is would be ok too. But let's discuss next week.

waxlamp:
> > The `AuditRecord` model also stores a copy of `Asset.metadata`, which means that any asset audit records created up until this PR is merged will have an `access` field, while future ones won't.
> 
> Since the `access` field for assets has always been superfluous, I think it would be appropriate to also purge it from the existing audit records. What do you think @waxlamp?

I think that would be ok.

mvandenburgh (commented):
LGTM, just need the migration for the audit record table :+1:

dandiapi/api/migrations/0011_asset_access_metadata.py
waxlamp: Why not use Django ORM style for this? i.e., replace `Asset.metadata` with the `access`-less copy, followed by `.save()`?

jjnesbitt: We could do that, but that means we'd be doing as many `.save` calls are there are assets, which seems unnecessarily burdensome, as well as the potential for a race condition (if someone were to update the asset at the same time, either their update may be lost, or this one could be). It would also probably take much longer.

mvandenburgh (approved):

yarikoptic:
PR gained approval 3 weeks ago. @jjnesbitt is there anything else to be done here or could be just merged?

waxlamp:
> PR gained approval 3 weeks ago. @jjnesbitt is there anything else to be done here or could be just merged?

This will require a migration; as such I'll move it back to draft status.

yarikoptic:
I have added to the PR description the TODO "section" with a checkbox for migration.  Would be great if similar approach is adopted for other PRs to make it clear(er) on what outstanding TODOs left.

Issue #1932, opened 2024-04-30
Unpublished Dandiset 724 shows a DOI


kabilar:
1. [Dandiset 724](https://dandiarchive.org/dandiset/000724) has not been published but the `Cite As` button displays a DOI - https://doi.org/10.80507/dandi.123456/0.123456.1234 - that appears to be placeholder numbers.
2. Inspection of a few other random unpublished Dandisets do not have this issue.

Thank you

rly:
These dandisets also have the DOI placeholder:
https://dandiarchive.org/dandiset/000482
https://dandiarchive.org/dandiset/000541
https://dandiarchive.org/dandiset/000565
https://dandiarchive.org/dandiset/000776

Could be others as well. I did not look thoroughly. Could be that all draft dandisets between ID 400 some and 800 some have the issue. The earlier ones and later ones don't have this DOI.

yarikoptic:
Related issue here I think is 
- #1709

Issue #1709, opened 2023-10-16
Stop injecting "fake" DOIs into draft dandisets


mvandenburgh:
Our current publish/DOI creation workflow needs improvement. Currently, when publishing a dandiset, we inject a "fake" DOI into the dandiset metadata in order to allow it to be validated against the `PublishedDandiset` schema; this is because we cannot validate a dandiset without putting a valid DOI in its metadata, but we don't want to hit the DOI server for a real DOI without first validating our dandiset. 

Instead of using a "fake" DOI , we can create a "draft DOI" for all draft dandisets. Then, we can promote those to a "Findable" DOI when a publish actually happens.

waxlamp:
There's a confluence here with #1710. Perhaps they can be folded into a single issue.

waxlamp:
The other idea here is to split the concepts up, so that our current validation process is meant as a clearinghouse for publishing, and a separate validation process would be for draft dandisets specifically. The latter might be similar to the former, but for certain values being optional, etc. (including, crucially, DOIs).

satra:
> Instead of using a "fake" DOI , we can create a "draft DOI" for all draft dandisets. Then, we can promote those to a "Findable" DOI when a publish actually happens.

this seems very reasonable to me and would also help users who want to insert a doi in their publication for review without it being set in stone if reviewers want changes to the dandiset. and yes it aligns with thoughts in #1710. we can also garbage collect drafts as needed if we see a dandiset abandoned.

CodyCBakerPhD:
> this seems very reasonable to me and would also help users who want to insert a doi in their publication for review without it being set in stone if reviewers want changes to the dandiset.

+1 to 'draft DOI' idea; we get this question/request from users very frequently

yarikoptic:
I think we could do even better. Similarly to how Zenodo does it, we can have a "dataset DOI" which would first point to draft and then most recent released version. For that reason we do not even need to make it mutable since our DLP shows most released one IIRC.  But the only concern is -- metadata which would be absent upon creation and then improved.  So we could create DOI with minimal/fake metadata and then keep updating it upon every metadata editing. I bet there is type of mutable DOI we could use here, couldn't we?

rly:
Just pinging this as I found a fake DOI in the wild in https://arxiv.org/pdf/2406.19492

```
29. Mazzamuto, G. et al. Human brain cell census for ba 44/45 (version draft). DANDI archive https://doi.org/10.80507/dandi.
[...]

Issue #275, opened 2020-06-03
Add endpoint to return the number of items in a dandiset with 0 files


AlmightyYakob:

AlmightyYakob:
I don't recall the initial reason for this issue, and there's been no action taken, so we can close this. If this crops up again we can address it with more context.

PR #2012, opened 2024-08-22
Initial draft of a design document for the Zenodo like DOI per dandiset

yarikoptic:
A design doc composed with @djarecka to avoid dummy DOIs for dandisets

refs: 
- #1709
- #1932

TODOs
- [ ] complete initial pass
- [ ] seek review

but could already be checked out by @dandi/archive-maintainers folks since overall idea is formulated already and some early concerns/questions could already be asked/answered

### DIFF
--- a/doc/design/doi-generation-2.md
+++ b/doc/design/doi-generation-2.md
@@ -1,0 +1,66 @@
+# DOI for Draft Dandisets
+
+Author: Yaroslav O. Halchenko & Dorota Jarecka
+
+The current approach:
+
+- [initial design doc](./doi-generation-1.md)
+- overall:
+   - inject fake DOI upon dandiset creation
+   - mint proper DOI only upon dandiset publication
+
+## Issues with the Existing Approach
+
+- [Stop injecting "fake" DOIs into draft dandisets](https://github.com/dandi/dandi-archive/issues/1709)
+- [Unpublished Dandisets display a DOI under `Cite As`](https://github.com/dandi/dandi-archive/issues/1932)
+
+## Proposed Solution
+
+Initially proposed/discussed in
+
+- [Create and maintain a "Findable" DOI for the Dandiset as a whole](https://github.com/dandi/dandi-archive/issues/1319)
+
+and boils down to the adoption of approach of Zenodo of having a DOI which always points to the latest version of the record.
+
+DataCite allows for three types of DOIs ([DataCite](https://support.datacite.org/docs/what-does-the-state-of-the-doi-mean)):
+
+- `Draft`. We do not use those.
+  *Can be deleted, and they require only the identifier itself in order to be created or saved. They can be updated to either Registered or Findable DOIs. Registered and Findable DOIs may not be returned to the Draft state, which means that changing the state of a Draft is final.*
+- `Registered`. Like `Findable` but not indexed for search, so we do not use them.
+- `Findable`. Is the type we use for published dandisets.
+  Requires to be valid (pass validation to fit the datacite schema) to be created.
+
+We propose to:
+
+- Instead of a fake DOI, upon creation of dandiset, mint and use a legit `Draft` DOI `10.48324/dandi.{dandiset.id}` with
+ - metadata entered during creation request (title, description, license)
+ - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+ - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
+- Upon changes to dandiset metadata record, for public (non-embargoed dandisets), try to update datacite metadata record while keeping the same target URL.
+  - For Draft DOI (dandiset was not published yet), there is no validation.
+    - **Question to clear up**: what happens to Draft DOI if metadata record is invalid? Does it fail to update altogether? does it update only the fields it knows about?
+  - For Findable DOI (dandiset was published), metadata record must pass validation, so we might fail to update.
+    But that should be ok.
+- Upon unembargoing dandiset: update (Draft) DOI metadata record with current metadata.
+- Upon publication of the dandiset:
+  - (already done currently) mint a proper `Findable` version DOI `10.48324/dandi.{dandiset.id}/{version}`
+  - update Dandiset-wide DOI `10.48324/dandi.{dandiset.id}` with metadata provided for version DOI, while keeping URL pointing to DLP instead of the released version.
+    - if Dandiset-wide DOI was in Draft state, it would be updated to Findable state (should work since we know metadata record passed validation).
+
+## Concerns to keep in mind/address
+
+- Draft dandiset might not have sufficient metadata to mint a proper DOI, or metadata might not be "proper" (fail validation) thus causing issues with minting a DOI
+  - Solution: start with Draft (not findable) DOI, and then upon publication mint a "findable" DOI
+  - **Follow up concern**: after dandiset and DOI publish, metadata of the Draft version of the dandiset could still be changed.
+    This potentially making changed record again "invalid".then when people change metadata
+- test site of datacite had different result of validation that the primary one
+
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - We might want a dedicated 404 page for deleted dandisets, or at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+  - Then we adjust DOI record to point to that page.
+
+- Should we do anything at dandischema level?
+
+- Should we do anything at DLP level?
+
+- Should we somehow reflect interactions with DataCite in Audit log?


doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,66 @@
...
+
+We propose to:
+
+- Instead of a fake DOI, upon creation of dandiset, mint and use a legit `Draft` DOI `10.48324/dandi.{dandiset.id}` with
+ - metadata entered during creation request (title, description, license)
+ - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+ - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
+- Upon changes to dandiset metadata record, for public (non-embargoed dandisets), try to update datacite metadata record while keeping the same target URL.
+  - For Draft DOI (dandiset was not published yet), there is no validation.
+    - **Question to clear up**: what happens to Draft DOI if metadata record is invalid? Does it fail to update altogether? does it update only the fields it knows about?
```
djarecka, line 41: @yarikoptic  - was testing, and sometimes it fails completely.
we should check probably specific fields that we are interested in changing.

yarikoptic: well, it could be any field.  We could establish minimal set of cause... at least we could try, and it should "definitely" work for whenever we would be updating for a release DOI

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,66 @@
...
+
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - We might want a dedicated 404 page for deleted dandisets, or at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+  - Then we adjust DOI record to point to that page.
+
+- Should we do anything at dandischema level?
+
+- Should we do anything at DLP level?
+
+- Should we somehow reflect interactions with DataCite in Audit log?
```
yarikoptic, line 66:
```suggestion
- Should we somehow reflect interactions with DataCite in Audit log?

# Targets

- [ ] develop a script, which tests on test fabric of datacite changes as introduced to all dandisets in the archive by 
     - for each dandiset
       - generate a record for overall *dandiset DOI* corresponding to metadata of the first release if any exists, otherwise corresponding to metadata of the draft version
       - for each release: mint a new *version DOI* for that release + possibly update *dandiset DOI* to correspond to potential changes in metadata
       - update *dandiset DOI*  to metadata of draft version
- implementation in dandi-archive
```

doc/design/doi-generation-2.md
```diff
@@ -33,10 +33,10 @@ DataCite allows for three types of DOIs ([DataCite](https://support.datacite.org
 We propose to:
 
 - Instead of a fake DOI, upon creation of a **public** dandiset, mint and use a legit `Draft DOI` `10.48324/dandi.{dandiset.id}` with
- - *minimal metadata* entered during creation request (title, description, license)
- - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
- - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
- - If minting a DOI fails, we need to raise exception to inform developers about the issue but proceed with the creation of the dandiset.
+  - *minimal metadata* entered during creation request (title, description, license)
+  - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+  - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
```
satra, line 38: a thought: perhaps the doi generation process can be implemented only when dataset is unembargoed. i.e. embargoed datasets cannot be pointed to by doi (even if we get reviewer view only access). an owner would have to unembargo it. 

for implementation, doi generation happens at creation for public and unembargoing for embargoed dandisets.

yarikoptic: We could. But one of the goals here is to get away from using "fake DOIs" (#1709) which this would "prevent".

In current design (might change) we would make overall dandiset DOI Findable only upon initial publication. As embargoed dataset would never be published, its DOI would remain `Draft` thus not available to users, and thus IMHO there is no harm.  I think we could even keep updating it with metadata etc. That IMHO would simplify the logic and make "embargoed" less special (thus easier to code/troubleshoot etc).

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,83 @@
...
+  - If minting a DOI fails, we need to raise exception to inform developers about the issue but proceed with the creation of the dandiset.
+- Upon changes to dandiset metadata record (so, of a draft version of dandiset), for public (non-embargoed dandisets):
+  - For `Draft DOI` (dandiset was not published yet), there is no validation, try to update datacite metadata record while keeping the same target URL
+    - **Question to clear up**: what happens to Draft DOI if metadata record is invalid? Does it fail to update altogether? does it update only the fields it knows about?
+  - For `Findable DOI` (dandiset was published at least once), we do not update anything since DLP points to that published version.
+  - **TODO: figure out how to annotate Draft version, so it always says that it is a draft version and thus potentially not used for citation if that could be avoided**
```
yarikoptic, line 44: To facilitate citation of dandisets which are not yet published, and immediate current use case -- dandisets with zarrs which we cannot publish yet since we cannot guarantee their versions, we could try migrating `Draft` DOI to Findable upon every modification of metadata.  If fails -- keep prior state (Draft).  Then if Findable already and fails --  doomed to keep prior one until edits bring it to "good state".

```suggestion
  - For `Draft DOI` (dandiset was not published yet): try to update/make it `Findable`.
    - If fails - keep Draft since there is no validation, try to update datacite metadata record while keeping the same target URL.
    - **Question to clear up**: what happens to Draft DOI if metadata record is invalid? It seems to create one with no metadata, but does it update only the fields it knows about?
  - For `Findable DOI` 
    - if it is still a draft version but which had legit metadata, we try to update metadata. If fails, we either ignore or just add a comment somewhere that "record might not reflect the most recent changes to draft version".
      - I think we need to add to validation procedures, validation against datacite metadata record, and reporting errors to the user so that users address them before trying to publish. May be we should validate only if no other errors (our schema validation) were detected to reduce noise, or just give a summary that "Metadata is not satisfying datacite model, fix known metadata errors first."
    - if dandiset was published at least once (has version) -- we do not update anything since DLP points to that published version.
  - **TODO: figure out how to annotate Draft version, so it always says that it is a draft version and thus potentially not used for citation if that could be avoided**
```

djarecka: regarding validation, I think we aimed to create `PublishedDandised` class that has already all the fields that datacite requires, so if we are able to create `PublishedDandiset`, we should be able to create dandiset.
However, we so have at the end of `to_datacite` validation_datacite` that checks against datacite schema (or at least one of the versions...)

So I believe once we have published dandiset and findable doi, we should only update with new publishe version and if our schema is right, we should not have problem with updating doi.

of course, datacite can change the schema (or at least the validation function), and we could have issues.

djarecka:
I created some test to simulate the workflow in #275

doc/design/doi-generation-2.md
yarikoptic: > In current design (might change) we would make overall dandiset DOI Findable only upon initial publication.

Given demand from zarr users to get DOIs for dandisets with zarrs, we better make them Findable as soon as possible in the life cycle of those dandisets. ref: https://github.com/dandi/helpdesk/discussions/165#discussioncomment-11824068 .  So I think we should proceed that way -- make them Findable as soon as datacite validation passes. Also inform user about datacite model issues as part of the validation.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,83 @@
...
+    This potentially making changed record again "invalid".
+    Should be Ok'ish
+- Test site of datacite had different result of validation that the primary one
+
+- `Draft` DOI is not visible/usable by users. We might want to switch it to `Findable` as soon ASAP (when datacite validates record ok).
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - We might want a dedicated 404 page for deleted dandisets, or at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+  - Then we adjust DOI record to point to that page.
+
+- Should we do anything at dandischema level?
```
yarikoptic, line 68: @djarecka we are to answer those I think before we could call this "done" ;-)

doc/design/doi-generation-2.md
yarikoptic: @djarecka comments are inline with my thinking, I will merge my suggestion and we will go through all of it again to make sure we are in alignemnt

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,87 @@
...
+  - *minimal metadata* entered during creation request (title, description, license)
+  - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+  - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
+  - If minting a DOI fails, we need to raise exception to inform developers about the issue but proceed with the creation of the dandiset.
+- Upon changes to dandiset metadata record (so, of a draft version of dandiset), for public (non-embargoed dandisets):
+  - For `Draft DOI` (dandiset was not published yet): try to update/make it `Findable`.
+    - If fails - keep Draft since there is no validation, try to update datacite metadata record while keeping the same target URL.
+    - **Question to clear up**: what happens to Draft DOI if metadata record is invalid? It seems to create one with no metadata, but does it update only the fields it knows about?
+  - For `Findable DOI` 
+    - if it is still a draft version but which had legit metadata, we try to update metadata. If fails, we either ignore or just add a comment somewhere that "record might not reflect the most recent changes to draft version".
```
yarikoptic, line 45:
```suggestion
    - if it is still a draft version of dandiset but which had legit metadata, we try to update metadata. If fails, we either ignore or just add a comment somewhere that "record might not reflect the most recent changes to draft version".
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,87 @@
...
+- [Stop injecting "fake" DOIs into draft dandisets](https://github.com/dandi/dandi-archive/issues/1709)
+- [Unpublished Dandisets display a DOI under `Cite As`](https://github.com/dandi/dandi-archive/issues/1932)
+
+## Proposed Solution
+
+Initially proposed/discussed in
+
+- [Create and maintain a "Findable" DOI for the Dandiset as a whole](https://github.com/dandi/dandi-archive/issues/1319)
+
+and boils down to the adoption of approach of Zenodo of having a DOI which always points to the latest version of the record.
```
yarikoptic, line 23:
```suggestion
and boils down to the adoption of approach of Zenodo of having a "Concept DOI" which always points to the latest version of the record.
```

pre @asmacdo finding on the terminology of Zenodo. In our case it is more of a "Dandiset DOI" vs "Dandiset Published Version DOI"

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,87 @@
...
+- Upon changes to dandiset metadata record, for embargoed dandisets don't do anything.
+- Upon unembargoing dandiset: update `Draft DOI` metadata record with current metadata **after** unembargoing.
+- Upon publication of the dandiset:
+  - (already done currently) mint a proper `Findable` version DOI `10.48324/dandi.{dandiset.id}/{version}`
+  - update Dandiset-wide DOI (`Draft` or `Findable`) `10.48324/dandi.{dandiset.id}` with metadata provided for the version DOI, while keeping URL pointing to DLP instead of the released version.
+    - if Dandiset-wide DOI was in `Draft` state, it would be updated to `Findable` state (should work since we know metadata record passed validation).
+       - **Question to clear up**:how to do that in API
+    - **Question to clear up**: behavior on what happens if metadata record is invalid?
+
+## Concerns to keep in mind/address
```
yarikoptic, line 58: TODO: add sequence diagram. Here is the one claude produced ATM for this document


```mermaid
sequenceDiagram
    participant User
    participant DandiArchive as Dandi Archive
    participant DataCite
    
    Note over User,DataCite: Public Dandiset Creation
    User->>DandiArchive: Create public dandiset
    DandiArchive->>DataCite: Request Draft DOI (10.48324/dandi.{id})
    DataCite-->>DandiArchive: Return Draft DOI
    Note right of DataCite: Minimal metadata + DLP URL
    DandiArchive-->>User: Return dandiset with Draft DOI
    
    Note over User,DataCite: Metadata Updates (Public Dandiset)
    User->>DandiArchive: Update dandiset metadata
    alt Draft DOI (not published yet)
        DandiArchive->>DataCite: Try to update & make DOI Findable
        alt Validation successful
            DataCite-->>DandiArchive: Update to Findable DOI
        else Validation fails
            DataCite-->>DandiArchive: Keep as Draft DOI
            DandiArchive->>DataCite: Update metadata record only
        end
    else Findable DOI (previous draft with valid metadata)
        DandiArchive->>DataCite: Try to update metadata
        alt Validation successful
            DataCite-->>DandiArchive: Update metadata
        else Validation fails
            DandiArchive-->>DandiArchive: Log error, continue
        end
    else Published dandiset (has version)
        DandiArchive-->>DandiArchive: No DOI update (points to published version)
    end
    DandiArchive-->>User: Return updated dandiset
    
    Note over User,DataCite: Embargoed Dandiset Handling
    User->>DandiArchive: Create embargoed dandiset
    DandiArchive->>DataCite: Request Draft DOI (minimal info)
    Note right of DataCite: Only DLP URL, no metadata
    DataCite-->>DandiArchive: Return Draft DOI
    DandiArchive-->>User: Return dandiset with Draft DOI
    
    User->>DandiArchive: Unembargo dandiset
    DandiArchive->>DataCite: Update Draft DOI with current metadata
    DataCite-->>DandiArchive: Update Draft DOI
    DandiArchive-->>User: Return unembaroed dandiset
    
    Note over User,DataCite: Dandiset Publication
    User->>DandiArchive: Publish dandiset
    DandiArchive->>DataCite: Mint version DOI (10.48324/dandi.{id}/{version})
    DataCite-->>DandiArchive: Return Findable version DOI
    
    DandiArchive->>DataCite: Update dandiset-wide DOI with version metadata
    alt Draft DOI
        DataCite-->>DandiArchive: Update to Findable state
    else Already Findable
        DataCite-->>DandiArchive: Update metadata only
    end
    
    DandiArchive-->>User: Return published dandiset with both DOIs
    
    Note over User,DataCite: Dandiset Deletion
    User->>DandiArchive: Delete dandiset
    DandiArchive->>DataCite: Update DOI to point to deletion page
    DataCite-->>DandiArchive: Update DOI target URL
    DandiArchive-->>User: Confirm deletion
```

yarikoptic: @asmacdo , had a chance to review this auto generated one?

asmacdo: needed some changes, included in latest push

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+      - **Question to clear up** If fails, we either ignore or just add a comment somewhere that "record might not reflect the most recent changes to draft version".
+      - **Question to clear up** If we add to validation procedures to dandiset updates, (validation against datacite metadata record), we can report errors to the user so they can be addressed prior to attempted publication. May be we should validate only if no other errors (our schema validation) were detected to reduce noise, or just give a summary that "Metadata is not satisfying datacite model, fix known metadata errors first."
+    - if dandiset was published at least once (has version) -- we do not update anything since DLP points to that published version.
+  - **TODO: figure out how to annotate Draft version, so it always says that it is a draft version and thus potentially not used for citation if that could be avoided**
+- Upon changes to embargoed dandiset metadata record, don't do anything.
+- Upon unembargoing dandiset: update `Draft DOI` metadata record with current metadata and attempt to promote to a `Findable DOI` **after** unembargoing.
```
asmacdo, line 71: There is a small difference here with the diagram. Can we safely assume that an embargoed dataset has any DOI at all? This may boil down to the question "Can a dandiset be embargoed at creation time"

Assuming that creating an embargoed dandiset does NOT get a DOI, but creating a dandiset first, then embargoing it are both possibilities, at unembargo time we just check to see if theres a DOI or not, and proceed accordingly.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+  - **Follow up concern**: after dandiset and DOI publish, metadata of the Draft version of the dandiset could still be changed.
+    This potentially making changed record again "invalid".
+    Should be Ok'ish
+- Test site of datacite had different result of validation that the primary one
+
+- `Draft` DOI is not visible/usable by users. We might want to switch it to `Findable` as soon ASAP (when datacite validates record ok).
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - Option 1: We might want a dedicated 404 page for deleted dandisets, then we adjust DOI record to point to that page.
+  - Option 2: at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+
```
asmacdo, line 188:
```suggestion
  - Option 3: We can "hide" the DOI by changing it from Findable to Registered
```

doc/design/doi-generation-2.md
asmacdo: An alternative workflow would be to create a draft DOI for embargoed datasets, and attempt to promote it on each update, but instead of promoting to Findable, we promote to Registered (which is hidden from public).

We can "publish" Registered DOIs which become Findable when unembargoed.

yarikoptic: > This may boil down to the question "Can a dandiset be embargoed at creation time"

This is the only way to get dandiset embargoed -- at creation time. And as a part of this design we want to avoid adding any fake DOIs as it is done now.

> An alternative workflow would be to create a draft DOI for embargoed datasets, and attempt to promote it on each update, but instead of promoting to Findable, we promote to Registered (which is hidden from public).

It is a sound alternative but ATM we are not exposing metadata of embargoed dandisets at all. If we expose it in Registered, it would be a "side effect"... so let's keep it just not populated at all.

doc/design/doi-generation-2.md
yarikoptic: **Good idea!** but is that a possibility (also ping @djarecka, did you try may be?)?
I thought that once Findable , it would not be allowed go to Registered, although on a 2nd thought I do not see why it should be forbidden!

asmacdo: On the test Datacite api I created a Findable DOI and converted it to Registered. (The action is called "hide", so I think this is the intended use more than promoting a draft to registered)

djarecka: correct, it is possible, never thought about it...

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+  - **Follow up concern**: after dandiset and DOI publish, metadata of the Draft version of the dandiset could still be changed.
+    This potentially making changed record again "invalid".
+    Should be Ok'ish
+- Test site of datacite had different result of validation that the primary one
+
+- `Draft` DOI is not visible/usable by users. We might want to switch it to `Findable` as soon ASAP (when datacite validates record ok).
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - Option 1: We might want a dedicated 404 page for deleted dandisets, then we adjust DOI record to point to that page.
+  - Option 2: at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+  - Option 3: We can "hide" the DOI by changing it from Findable to Registered
```
yarikoptic, line 188: Now that we have Option 3, let's finalize into:

```suggestion
- Dandisets can be deleted. In such cases, upon deletion of a dandiset:
  - if DOI was a `Draft` DOI - just delete it as well.
  - if DOI was a `Findable` DOI - convert to `Registered` DOI
    - Also at the level of the DANDI archive itself we should provide tombstone page so URL is still "working" (#3211)
```

doc/design/doi-generation-2.md
yarikoptic: And otherwise you didn't spot any other divergence in the diagram? (I am yet to look too)

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+
+We propose to:
+
+- Upon creation of a **public** dandiset, mint a `Dandiset DOI` (a DataCite `Draft DOI`) `10.48324/dandi.{dandiset.id}` with
+  - *minimal metadata* entered during creation request (title, description, license)
+  - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+  - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
+  - If minting a DOI fails, we need to raise exception to inform developers about the issue but proceed with the creation of the dandiset.
+- Upon changes to a non-embargoed, draft dandiset metadata record:
+  - If `Draft DOI`, attempt to "promote" it to `Findable`.
```
yarikoptic, line 61: @asmacdo @djarecka how easy/reliable to check if DOI is `Draft` or `Findable`?  

On one hand, ideally we should avoid needing to check the state on datacite since it would require extra interaction which could go bad.  On the other, this way we would not assume knowing what is the state of the DOI, and whether it was successfully produced/promoted before.  It brings possibility for a race condition, but I think the danger and possible side-effects are minimal (unless I see some big problem).

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+    - If validation fails - keep `Draft DOI` (very limited validation), attempt to update datacite metadata record while keeping the same target URL.
+    - **Question to clear up**: what happens to `Draft DOI` if metadata record is invalid? It seems to create one with no metadata, but does it update only the fields it knows about?
```
yarikoptic, line 63: @djarecka I feel like you clarified on this but we did not put it "in writing" here.  What do you remember on this aspect?

doc/design/doi-generation-2.md
asmacdo: Easy, a GET to the DOI, response["data"]["attributes"]["state"]

doc/design/doi-generation-2.md
asmacdo: I need to have another careful look-- after "first contact with reality", ie playing with the code, this does need to change a bit.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+    end
+    DandiArchive->>DataCite: Mint Version DOI (Findable) (10.48324/dandi.{id}/{version})
+    DataCite-->>DandiArchive: Return Version DOI (Findable)
+    DandiArchive->>DataCite: Update Dandiset DOI with version metadata
+    Note right of DandiArchive: Keep URL pointing to DLP
+    DandiArchive-->>User: Return published dandiset with both DOIs
+
+    Note over User,DataCite: Dandiset Deletion (Optional)
+    User->>DandiArchive: Delete dandiset
+    DandiArchive->>DataCite: Update DOI to point to deletion page
```
asmacdo, line 164:
```suggestion
    alt Dandiset DOI is Draft
          DandiArchive->>DataCite: Delete Draft DOI
    else Dandiset DOI is  Findable
          DandiArchive->>DataCite: "hide" DOI (Convert to "Registered") and point DOI to tombstone page
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
+# DOI for Draft Dandisets
+
+Authors: Yaroslav O. Halchenko, Dorota Jarecka, Austin Macdonald
+
+## Overview
+
+This document describes an updated strategy for DOI management within the Dandi Archive.
+Upon creation, every public Dandiset will receive a **Dandiset DOI** that will represent the current draft and all future versions.
+Every public published version of a Dandiset will recieve a **Version DOI**.
```
asmacdo, line 9:
```suggestion
Every public published version of a Dandiset will receive a **Version DOI**.
```

yarikoptic: oh no -- you found a typo which codespell does not know!  

Starting to collect ideas for bot commands etc, codename `@myuniquebot` , so I could 

```
@myuniquebot please submit a PR with a new typo fix for codespell: "recieve" should be "receive"
```

with some shortcut like

```
@myuniquebot new typo: recieve -> receive
```

and it would prepare a PR on my behalf (first in my personal fork to ensure tests pass etc, and then preparing PR to be approved/submitted "as by me")

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+
+For example:
+Dandiset DOI: `https://doi.org/10.48324/dandi.000027/`
+Version DOI: `https://doi.org/10.48324/dandi.000027/0.210831.2033`
+
+Prior to publication, the Dandiset DOI will refer to the latest draft version.
+Following publication, the Dandiset DOI will refer to the latest published version.
```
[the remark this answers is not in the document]
yarikoptic: I thought we discussed that and for some reason thought to make it more explicit... but may be indeed, let's just do this

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+- `Registered`. Like `Findable` but not indexed for search, so we do not use them.
+- `Findable`. Is the type we use for published dandisets.
+  Requires to be valid (pass validation to fit the datacite schema) to be created.
+
+We propose to:
+
+- Upon creation of a **public** dandiset, mint a `Dandiset DOI` (a DataCite `Draft DOI`) `10.48324/dandi.{dandiset.id}` with
+  - *minimal metadata* entered during creation request (title, description, license)
+  - DLP URL `https://dandiarchive.org/dandiset/{dandiset.id}`
+  - For embargoed dandiset, **do not** specify any metadata besides the DLP URL.
```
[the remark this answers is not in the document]
yarikoptic: I think we still want to mint it since we do not want a fake one, so we better reserve a real one right away. Could be draft.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+### Migration
+
+A django-admin script should be created and executed to create a `Dandiset DOI` for all existing dandisets.
+
+**Question to address**: Will adding a `Dandiset DOI` in addition to `Version DOI` require a db migration?
+
+## Concerns to keep in mind/address
+
+- Draft dandiset might not have sufficient (or valid) metadata to promote to a `Findable` DOI, thus causing issues with minting a DOI
+  - **Solution**: start with Draft (not findable) DOI, and then upon publication promote to a "findable" DOI.
```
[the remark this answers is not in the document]
yarikoptic: I don't remember if we discussed it, but I think we should better keep it as originally worded:  since ATM we allow for easy deletion of dandisets without published versions.  So, here then we would still be able to delete Draft DOI easily.  So let's keep original wording/logic

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+
+- `Draft` DOI is not visible/usable by users. We might want to switch it to `Findable` as soon ASAP (when datacite validates record ok).
+- `Findable` DOI cannot be deleted, but in principle we allow for deletion of dandisets.
+  - Option 1: We might want a dedicated 404 page for deleted dandisets, then we adjust DOI record to point to that page.
+  - Option 2: at least a message that the dandiset was deleted, and ideally describe the reason why it was deleted ("Upon request of maintainer", "Due to violation of terms of service", etc.)
+  - Option 3: We can "hide" the DOI by changing it from Findable to Registered
+- Should we do anything at dandischema level?
+  - yes
+    - Needs to be able to mint `Draft DOI`
+    - Needs to be able to promote `Draft DOI` to `Findable DOI`
```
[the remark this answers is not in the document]
yarikoptic: why we would no longer need to promote `Draft` DOI into `Findable`?

doc/design/doi-generation-2.md
asmacdo: No codespell alerted me to this :)

asmacdo:
For dandi-schema, we may also want to pull the validation out of `to_datacite`. This validation does not use the datacite API, rather validation occurs against the schema which has been pre-fetched and committed into the dandi-schema repo.

Currently `to_datacite` accepts an optional arg `validate`, which defaults to `False`--  **[the only use of  `to_datacite` in dandi-archive](https://github.com/dandi/dandi-archive/blob/24f68130266c2e280e147ef97e5ca2c51bbf287d/dandiapi/api/doi.py#L38) DOES NOT enable validation!**. I'm curious what problems this could cause, and if those problems have actually occurred-- maybe there are more Dandisets with DOIs stored in the data model without a DOI minted? Or maybe they were created but set to Draft due to validation errors? We would have logged exceptions if the datacite API did not accept a new DOI.

From this design doc only, I gather that we believe Draft DOIs have less stringent validation than Findable, but I have not found any upstream documentation that confirms this. If it is the case, I suggest we pull the validation out of the `to_datacite` function, which would only be responsible for constructing the API payload. Then, in dandi-archive, we perform a validation. If valid we publish a Findable DOI, and if invalid we fallback to Draft DOI if it fails validation. 

Either way, we need to consider using some kind of validation. @djarecka @yarikoptic I suggest this as  topic for our discussion tomorrow.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+**Question to address**: Will adding a `Dandiset DOI` in addition to `Version DOI` require a db migration?
+
+## Concerns to keep in mind/address
+
+- Draft dandiset might not have sufficient (or valid) metadata to promote to a `Findable` DOI, thus causing issues with minting a DOI
+  - **Solution**: start with Draft (not findable) DOI, and then upon publication promote to a "findable" DOI.
+  - **Follow up concern**: after dandiset and DOI publish, metadata of the Draft version of the dandiset could still be changed.
+    This potentially making changed record again "invalid".
+    Should be Ok'ish
+- Test site of datacite had different result of validation that the primary one
```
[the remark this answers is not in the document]
yarikoptic: It might have been a bad memory, @djarecka do you have any information on this or should we just remove this?
```suggestion
```

doc/design/doi-generation-2.md
yarikoptic: note: if we do not use validated models (e.g. pydantic), it would require try/except or conversion to chain of `.get` to ensure robust operation and appropriate logic

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
asmacdo: we would still do that, its just that the only difference is in the payload generated by dandischema, which already has this capability

doc/design/doi-generation-2.md
asmacdo: The validation is jsonschema based https://github.com/dandi/dandi-archive/pull/2012#issuecomment-2828205136

The only thing I think we need to worry about is if it exists or not. Whether draft or findable, I think we can do the same payload with "event=publish", pretty sure it would promote if draft or keep as findable as long as the data is valid. 

see also
https://github.com/dandi/dandi-schema/blob/master/dandischema/datacite/__init__.py#L265-L269

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
djarecka: let's remove it. It was based on my memory of some old issue, but wasn't able to reproduce.

djarecka:
> For dandi-schema, we may also want to pull the validation out of `to_datacite`. This validation does not use the datacite API, rather validation occurs against the schema which has been pre-fetched and committed into the dandi-schema repo.

correct!

> 
> Currently `to_datacite` accepts an optional arg `validate`, which defaults to `False`-- **[the only use of `to_datacite` in dandi-archive](https://github.com/dandi/dandi-archive/blob/24f68130266c2e280e147ef97e5ca2c51bbf287d/dandiapi/api/doi.py#L38) DOES NOT enable validation!**. I'm curious what problems this could cause, and if those problems have actually occurred-- maybe there are more Dandisets with DOIs stored in the data model without a DOI minted? Or maybe they were created but set to Draft due to validation errors? We would have logged exceptions if the datacite API did not accept a new DOI.

That's probably not good, if the validation is not used before publishing. Was not aware.

> 
> From this design doc only, I gather that we believe Draft DOIs have less stringent validation than Findable, but I have not found any upstream documentation that confirms this. If it is the case, I suggest we pull the validation out of the `to_datacite` function, which would only be responsible for constructing the API payload. Then, in dandi-archive, we perform a validation. If valid we publish a Findable DOI, and if invalid we fallback to Draft DOI if it fails validation.

Yes, no documentation, but that's the case...

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
asmacdo: I don't think theres a need to "reserve" a DOI. given that we have our own prefix, each dandiset_id suffix is guaranteed to be available.

The bigger problem is the metadata-- for an embargoed dataset, we probably dont want to send any potentially non-public metadata to a 3rd party. We could possibly use fake data, but I dont see any value.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,214 @@
...
+- Upon changes to a non-embargoed, draft dandiset metadata record:
+  - If `Draft DOI`, attempt to "promote" it to `Findable`.
+    - If validation fails - keep `Draft DOI` (very limited validation), attempt to update datacite metadata record while keeping the same target URL.
+    - **Question to clear up**: what happens to `Draft DOI` if metadata record is invalid? It seems to create one with no metadata, but does it update only the fields it knows about?
+  - If `Findable DOI`
+    - if draft version with no prior publications, but which had legit metadata, we try to update metadata.
+      - **Question to clear up** If fails, we either ignore or just add a comment somewhere that "record might not reflect the most recent changes to draft version".
+      - **Question to clear up** If we add to validation procedures to dandiset updates, (validation against datacite metadata record), we can report errors to the user so they can be addressed prior to attempted publication. May be we should validate only if no other errors (our schema validation) were detected to reduce noise, or just give a summary that "Metadata is not satisfying datacite model, fix known metadata errors first."
```
asmacdo, line 66: @djarecka and I discussed this and we aren't convinced this is really necessary, especially since it leads to that awkward scenario of a Findable DOI with  incorrect metadata

"Adding a comment somewhere" -- Since we cant use the data on the dandiset, (it would be invalid) we could possibly retrieve the DOI, and modify that data, but this feels unclean to me.

IMO the cleaner path would be to simply leave Draft Dandisets as Draft DOIs until their first publish. At that point mint the Dandiset DOI. And only update the Dandiset DOI after further successful publishes.

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
yarikoptic: The value is that ATM AFAIK code requires having some "fake DOI" to start with.  Sure thing we could avoid "reserving" it at datacite level, but I think it might be easier to just have it so that 

- we always have a DOI for dandiset, and it is not a "fake DOI" in case of our main instance
- DOI is "real" as it also known to datacite (draft or not)

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,212 @@
...
+    DandiArchive->>DataCite: Update DOI to point to deletion page
+    DataCite-->>DandiArchive: Update DOI target URL
+    DandiArchive-->>User: Confirm deletion
+```
+
+### Migration
+
+A django-admin script should be created and executed to create a `Dandiset DOI` for all existing dandisets.
+
+**Question to address**: Will adding a `Dandiset DOI` in addition to `Version DOI` require a db migration?
```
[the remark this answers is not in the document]
yarikoptic: isn't there also a "draft" Version with DOI (and that's where I guess we inject a fake one), i.e. could we avoid changing DB model?

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
yarikoptic: so we are to list only what to implement (i.e. what is not there yet) here, makes sense, but then we might want to clarify, or make it into checklist, e.g. 

```suggestion
    Need to be able to
    - [x] promote `Draft DOI` to `Findable DOI` (can be done via ...)
    - [ ] "hide"  `Findable DOI` to `Registered DOI` for deletion
    - [ ] mint Dandiset DOI as well as Version DOI. 
```

not immediately clear  on what exactly payload there  to use since the only options are the bools we have . https://github.com/dandi/dandi-schema/pull/297 seems an attempt to make it more flexible but not yet there, so remaining a TODO?

doc/design/doi-generation-2.md
yarikoptic: I do not think here it is concerned with Draft-vs-Findable, and rather to validate metadata against datacite for draft version. May be this comment is mis-placed?

doc/design/doi-generation-2.md
asmacdo: (@yarikoptic this is my previously misplaced comment)
@djarecka and I discussed this and we aren't convinced it is really necessary to promote unpublished Dandisets to Findable, especially since it leads to that awkward scenario of a Findable DOI with incorrect metadata

"Adding a comment somewhere" -- Since we cant use the data on the dandiset, (it would be invalid) we could possibly retrieve the DOI, and modify that data, but this feels unclean to me.

IMO the cleaner path would be to simply leave Draft Dandisets as Draft DOIs until their first publish. At that point mint the Dandiset DOI. And only update the Dandiset DOI after further successful publishes.

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
asmacdo: What metadata should we send to datacite? Are the title, description, authors potentially secret? We could provide fake placeholder data?

We could also just determine what the DOI url will be, and save that to the db without sending anything to datacite.

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
asmacdo: I have to admit that I "just dont like that" since the Dandiset DOI semantically belongs to the Dandiset. I cant think of a reason this wouldnt work, but just feels messy. 

To retrieve the Dandiset DOI via Django, `someversion.dandiset.draft_version.doi` IMO violates the "principle of least surprise".

Prior to publication the Dandiset DOI will point to the draft version (via the DLP) **but after publication** the dandiset DOI will point to the latest publication, so that would also be surprising

doc/design/doi-generation-2.md
yarikoptic: ping @djarecka

doc/design/doi-generation-2.md
yarikoptic: agree, let's keep Draft until publish, and after that keep updating only upon subsequent Publishes

doc/design/doi-generation-2.md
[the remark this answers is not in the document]
asmacdo: We decided not to mint dois for embargoed dandisets

doc/design/doi-generation-2.md
djarecka: yes, sorry, i just gave "thumbs up", since you were right, `Draft` DOI doesn't have to have metadata, it can only have url, and if has more, it will be added

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Proposed Solution
+
+- For **Public dandisets**:
+    - Upon creation:
+        - mint a `Dandiset DOI` (a DataCite `Draft DOI`) `10.48324/dandi.{dandiset.id}` with *minimal metadata* entered during creation request (title, description, license)
+        - URL should point be DLP `https://dandiarchive.org/dandiset/{dandiset.id}`
+        - If minting a DOI fails, we need to raise exception to inform developers about the issue but proceed with the creation of the dandiset.
+    - Upon updates to a draft dandiset metadata **prior to first publication**:
+        - Update the datacite metadata of the `Draft DOI`, (leave as draft)
```
yarikoptic, line 60:
```suggestion
        - Update the datacite metadata and if still a `Draft` DOI, try to make it a `Findable` DOI
           - if `Findable` fails, try to update metadata as a `Draft` DOI
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+- Dandisets can be deleted. In such cases, upon deletion of a dandiset:
+  - if DOI was a `Draft` DOI - just delete it as well.
+  - if DOI was a `Findable` DOI - convert to `Registered` DOI (follows [datacite best practices](https://support.datacite.org/docs/tombstone-pages))
+  - Also at the level of the DANDI archive itself we should provide tombstone page so URL is still "working" (#3211)
+  - If no tombstone page support added, just adjusted URL in datacite record to point to https://www.datacite.org/invalid.html
+
+## Concerns to keep in mind/address
+
+- **Question to clear up**: what happens to `Draft DOI` if metadata record is invalid?
+    - It seems to create one with no metadata, but does it update only the fields it knows about?
```
yarikoptic, line 216: @djarecka did you check in your experiments?

doc/design/doi-generation-2.md
yarikoptic: this was to provide citeable DOIs for Draft Dandisets with zarrs... for now let's postpone this and if needed, will implement/enable later.

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+    - Upon deletion of a draft dandiset metadata **after the first publication**:
+        - "hide" the `Dandiset DOI` (Findable) to `Registered DOI`
+    - Upon **subsequent publications** of a dandiset:
+        - Mint a new `Version DOI`
+        - Update `Dandiset DOI` metadata to match published version
+- For **embargoed dandiset**:
+    - Upon creation, no DOI is created.
+    - Upon changes to embargoed dandiset metadata record, don't do anything.
+    - Upon deletion of an embargoed dandiset:
+        - Delete the `Dandiset DOI` (Draft) from Datacite
```
yarikoptic, line 79:
```suggestion
```

yarikoptic, line 79:
```suggestion
    - Upon deletion of an embargoed dandiset: don't do anything.
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+    - Upon unembargoing dandiset:
+        - If there are published versions:
+            - Mint `Dandiset DOI` (Findable) with latest published version of metadata,
+            - Mint `Version DOI` for each published version.
+        - If there are no published versions:
+            - Mint `Dandiset DOI` (Draft) with latest metadata,
```
yarikoptic, line 85:
```suggestion
        - Mint `Dandiset DOI` (Draft) with latest metadata,
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Alternatives Explored
+
+### Creating DOIs for Embargoed Dandisets
+
+We opted not to create DOIs for embargoed Dandisets because:
+ - We own the prefix, and so there is no need to "reserve"
+ - We should avoid sending any potentially secret metadata to a 3rd party, even if it is not publicly searchable.
+ - If we were to create a DOI with fake metadata that probably would not have any value at all.
+ - What the DOIs will eventually be upon publication is semantically determined, so the value can be used even prior to being "real".
```
yarikoptic, line 197:
```suggestion
 - What the DOIs will eventually be upon publication is semantically determined, so the value can be used even prior to being "real".
 
 We might reconsider, if decision would be made to expose metadata of Embargoed Dandisets for the purpose of discovery.
```

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+    - Upon deletion of a draft dandiset metadata **prior to first publication**:
+        - Delete the `Dandiset DOI` (Draft) from Datacite
+    - Upon **first publication** of a dandiset:
+        - Mint a new `Version DOI` (Findable) (already done currently), ie `10.48324/dandi.{dandiset.id}/{version}`
+        - Update `Dandiset DOI` metadata to match published version
+        - promote `Dandiset DOI` (Draft) to `Findable DOI`
+    - Upon updates to draft dandiset metadata **after the first publication**"
+        - no-op. The `Dandiset DOI` metadata will match the most recent publication.
+    - Upon deletion of a draft dandiset metadata **after the first publication**:
+        - "hide" the `Dandiset DOI` (Findable) to `Registered DOI`
```
yarikoptic, line 71:
```suggestion
    - Upon deletion of a published dandiset version (`VersionViewSet.destroy`) :
        - "hide" the `Version DOI` (Findable) to `Registered DOI`
    - Upon deletion of a dandiset (`DandisetViewSet.destroy`):
        - "hide"  the `Dandiset DOI`  if `Findable` and delete if `Draft`
```

note also that currently `delete_doi` which is invoked from within `VersionViewSet.destroy` seems to be useless since can delete only Draft ones, but this is for published version, so cannot do that

doc/design/doi-generation-2.md
```diff
@@ -0,0 +1,229 @@
...
+    end
+    DandiArchive-->>User: Confirm deletion
+```
+
+### Migration
+
+A django-admin script should be created and executed to create a `Dandiset DOI` for all existing dandisets.
+
+No new field will be added for `Dandiset DOI`.
+Instead, the `Draft Dandiset` DOI field will be where the `Dandiset DOI` is stored.
```
yarikoptic, line 173:
```suggestion
No DB migration will be needed, as no new field will be added to `Dandiset` model, and
instead, the `Dandiset DOI` will be stored in the "draft" `Version`.
```

yarikoptic:
@asmacdo then please approve this PR "officially" -- I can't since I am the author.  I have enabled auto-merge though ;)

PR #2015, opened 2024-08-27
Disable GUI "Unembargo" button if there are active uploads

jjnesbitt:
Currently, there's a quirk of the GUI where you're allowed to hit the "Unembargo" button, but which produces an error every time, without any progression. This is due to there being active uploads in the dandiset, which prohibits unembargo.

This PR prevents taking such an action in the GUI, disabling the button and giving the reason. I'd like to follow up this PR with a more robust improvement to uploads, adding the ability to "clear" any in progress uploads. Currently this isn't possible, leading to a very awkward situation.


satra:
thanks @jjnesbitt

dandiapi/api/models/version.py
```diff
@@ -71,6 +71,13 @@ def size(self):
             or 0
         )
 
+    @property
+    def active_uploads(self):
+        if self.version != 'draft':
+            return 0
+
+        return self.dandiset.uploads.count()
```
waxlamp, line 79: Just a minor style thing, but Python's "ternary operator" would be great here.

mvandenburgh (approved):

waxlamp:
> I'd like to follow up this PR with a more robust improvement to uploads, adding the ability to "clear" any in progress uploads.

That would address the newly filed #2021.

PR #2017, opened 2024-08-28
admin view: Also show (list) zarr for Assets view

yarikoptic:
ATM shows only blobs, so to get to zarr requires 2 clicks. As long as we have zarr as valuable as blob, we better list both.

Having said that, I do not feel strongly about it yet, so feel free to quickly close if you feel we better keep it just to blobs for now.


yarikoptic:
ping @waxlamp on this trivial PR

PR #2022, opened 2024-09-05
Add text from the Handbook to the rejected user email

kabilar:


waxlamp (changes requested):
I think the new text makes more sense _after_ the rejection reason block. Then it's "you were denied; here's why; here's some helpful info as a call to action".

kabilar:
> I think the new text makes more sense _after_ the rejection reason block. Then it's "you were denied; here's why; here's some helpful info as a call to action".

Thanks @waxlamp.  I was actually hoping that we don't use the rejection reason block often.  This would limit the amount of manual work that Yarik has to do, and provide a standard response.  The reason for denial would be that an academic email is not associated with the GitHub account.  And if an admin wanted to provide further explanation, then the rejection reason block would be available.

kabilar:
> > I think the new text makes more sense _after_ the rejection reason block. Then it's "you were denied; here's why; here's some helpful info as a call to action".
> 
> Thanks @waxlamp. I was actually hoping that we don't use the rejection reason block often. This would limit the amount of manual work that Yarik has to do, and provide a standard response. The reason for denial would be that an academic email is not associated with the GitHub account. And if an admin wanted to provide further explanation, then the rejection reason block would be available.

Agree that if we do use the rejection reason block, it makes more sense to be above the template text.  So I have moved the block.

dandiapi/api/templates/api/mail/rejected_user_message.txt
```diff
@@ -3,11 +3,17 @@ Dear {{ greeting_name }},
...
+    Detailed reason:
     {{ rejection_reason }}
 {% endif %}
 
-If you would like to appeal this decision, please contact the DANDI admins at help@dandiarchive.org.
+DANDI is freely accessible to the neuroscience research community. Membership is usually granted automatically to GitHub accounts with a .edu or similar academic email. If your registration is denied:
+
+- With an academic email not linked to your GitHub, please contact help@dandiarchive.org for assistance using this academic email address to verify your academic affiliation.
+
+- Without an academic email, account approval is still possible under specific circumstances. Appeal decisions at help@dandiarchive.org.
```
waxlamp, line 14:
```suggestion
- Without an academic email, account approval is still possible under specific circumstances. You can appeal the decision by contacting help@dandiarchive.org.
```

waxlamp (approved):
I left one suggestion to soften and clarify what to do without an academic email address, but otherwise this looks good.

yarikoptic (approved):

kabilar:
Hi @waxlamp, would you be able to deploy these changes this week?  Thank you.

waxlamp:
> Hi @waxlamp, would you be able to deploy these changes this week? Thank you.

Done!

Issue #1900, opened 2024-03-20
Should process_s3_log_file_task be removed/turned off?


yarikoptic:
It was added, improved in #1605 .
But AFAIK we do never cared about that collected global download count statistic and it is not displayed/used anywhere. Or is it? (I even failed to figure out in admin django interface where it is )

But I see it quite active.
But if not used, why bother? (especially since we are trying to figure out what makes dandi archive slow ATM).

relevant issues:
- https://github.com/dandi/dandi-archive/issues/1424
- https://github.com/dandi/dandiarchive-logs/issues/5

and ATM we are just harvesting full S3 logs on drogon server with @CodyCBakerPhD doing initial analytics/visualization so we pin point target needed collection to actually implement/collect long term.

CodyCBakerPhD:
See more in-depth comments in #2392, but yes  I would recommend turning this off for the time being

dandibot:
:rocket: Issue was released in [`v0.11.5`](https://github.com/dandi/dandi-archive/releases/tag/v0.11.5) :rocket:

PR #2025, opened 2024-09-09
Use bulk_update for updating blob download counts

danlamanna:
This removes a linear scaling set of update queries in favor of `bulk_update`. This should prevent us from hitting our preconfigured time limits for processing S3 log files.

### DIFF
--- a/dandiapi/analytics/tasks/__init__.py
+++ b/dandiapi/analytics/tasks/__init__.py
@@ -10,6 +10,7 @@ from django.db import transaction
 from django.db.models.aggregates import Max
 from django.db.models.expressions import F
 from django.db.utils import IntegrityError
+from more_itertools import batched
 from s3logparse import s3logparse
 
 from dandiapi.analytics.models import ProcessedS3Log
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):

dandiapi/analytics/tasks/__init__.py
```diff
@@ -80,10 +81,20 @@ def process_s3_log_file_task(s3_log_key: str) -> None:
                 logger.info('Already processed log file %s', s3_log_key)
             return
 
+        # we need to store all of the fully hydrated blob objects in memory in order to use
+        # bulk_update, but this turns out to not be very costly. 1,000 blobs use about ~8kb
+        # of memory.
+        # breakpoint()
```
jjnesbitt, line 87:
```suggestion
```

danlamanna: :upside_down_face:

dandiapi/analytics/tasks/__init__.py
```diff
@@ -80,10 +81,19 @@ def process_s3_log_file_task(s3_log_key: str) -> None:
...
+        # bulk_update, but this turns out to not be very costly. 1,000 blobs use about ~8kb
+        # of memory.
+        asset_blobs = []
+
+        # batch the blob queries to avoid a large WHERE IN clause
+        for batch in batched(download_counts, 1_000):
+            asset_blobs += AssetBlob.objects.filter(blob__in=batch)
+
+        for asset_blob in asset_blobs:
+            asset_blob.download_count = F('download_count') + download_counts[asset_blob.blob]
```
jjnesbitt, line 94: What does using `F('download_count')` do here? As opposed to just writing:

```python
asset_blob.download_count += download_counts[asset_blob.blob]
```

danlamanna:
```sql
-- assuming download_count is already 20 and we're adding 10

-- using F('download_count')
UPDATE asset_blob SET download_count = download_count + 10;

-- without
UPDATE asset_blob SET download_count = 30;
```

The latter is susceptible to the lost update problem.

yarikoptic:
PLEASE review/chime in on 
- https://github.com/dandi/dandi-archive/issues/1900

We will also meet with @CodyCBakerPhD  this Wed 10-11am to go through what he has done in https://github.com/dandi/dandiarchive-logs/issues/10#issuecomment-2339068431 etc

danlamanna:
> PLEASE review/chime in on
> 
>     * [Should process_s3_log_file_task be removed/turned off? #1900](https://github.com/dandi/dandi-archive/issues/1900)

I'll leave a comment there, but I think this is still worth deploying regardless.

PR #2031, opened 2024-09-20
Provide workaround to specify "correct" content_type for JSON files in "inline" mode

### LINKED ISSUE #2027: "Open in browser" on dataset_description.json results in Download request
image

on https://dandiarchive.org/dandiset/000108/draft/files?location= clicking on the .json file which could be easily visualized by browser leads to download.

```
Chromium	128.0.6613.113 (Official Build) built on Debian GNU/Linux trixie/sid (64-bit) 
Revision	9597ae93a15d4d03089b4e9997b1072228baa9ad-refs/branch-heads/6613@{#1429}
OS	Linux
JavaScript	V8 12.8.374.24
User Agent	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36
Command Line	/usr/lib/chromium/chromium --show-component-extension-options --enable-gpu-rasterization --no-default-browser-check --disable-pings --media-router=0 --enable-remote-extensions --load-extension --flag-switches-begin --disable-quic --flag-switches-end
Executable Path	/usr/lib/chromium/chromium
Profile Path	/home/yoh/.config/chromium/Default
Variations Seed Type	Null
```

References:
- https://github.com/dandi/dandi-archive/issues/742

yarikoptic:
Closes #2027

### DIFF
--- a/dandiapi/api/views/asset.py
+++ b/dandiapi/api/views/asset.py
@@ -164,6 +164,11 @@ class AssetViewSet(DetailSerializerMixin, GenericViewSet):
                 storage.generate_presigned_download_url(asset_blob.blob.name, asset_basename)
             )
         if content_disposition == 'inline':
+            # Client might have not correctly identified/provided the content type which
+            # now forbids correct handling of the file in the browser.
+            if asset_basename.endswith('.json') and content_type == 'application/octet-stream':
+                content_type = 'application/json'
+
             url = storage.generate_presigned_inline_url(
                 asset_blob.blob.name,
                 asset_basename,


waxlamp (commented):
I know this fixes the problem in the short term, but it causes another wider problem: this kind of thing means that we are starting not to trust the data's own self-reports of its content type. Is there something we can do to fix the problem at the root?

waxlamp:
> I know this fixes the problem in the short term, but it causes another wider problem: this kind of thing means that we are starting not to trust the data's own self-reports of its content type. Is there something we can do to fix the problem at the root?

Ok I see you already raised this issue over in https://github.com/dandi/dandi-archive/issues/2027#issuecomment-2364043643. Let's continue discussion there.

waxlamp:
This PR is no longer needs, as the JSON assets in the system have been updated with the correct `encodingFormat`.

PR #2038, opened 2024-10-05
Update new user registration questions

kabilar:
This pull request will let users know at this step of the registration flow that accounts are not needed to search/view/download.

### DIFF
--- a/dandiapi/api/views/auth.py
+++ b/dandiapi/api/views/auth.py
@@ -49,8 +49,9 @@ def auth_token_view(request: Request) -> HttpResponseBase:
 QUESTIONS = [
     {'question': 'First Name', 'max_length': 100},
     {'question': 'Last Name', 'max_length': 100},
-    {'question': 'What do you plan to use DANDI for?', 'max_length': 1000},
-    {'question': 'Please list any affiliations you have.', 'max_length': 1000},
+    {'question': 'Affiliation(s)', 'max_length': 1000},
+    {'question': 'Lab/project website', 'max_length': 1000},
+    {'question': 'Please describe how your research project will utilize DANDI resources. (Please note that users can freely search, view, and download public datasets without an account.)', 'max_length': 1000},
 ]
 
 # questions for new users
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
1. I would like @mvandenburgh to provide a technical review, to make sure changing the list of questions in this way doesn't somehow break the form or its processing.
2. I note that the "lab/project website" question presents yet another hurdle to registering an account. Do we need to have this information? If this is part of making account review easier, I think Ben's proposal to verify .edu email addresses is a better solution.
3. The note about not needing an account strikes me as too little too late: by this point in the process, the account is already created and put in the INCOMPLETE state, pending completion of this form and an approve/reject decision. I imagine most users will simply power through at this point in accordance with a sort of sunk cost fallacy (applied to their own effort, and without any reason to not simply apply for the account anyway, lest they wish to upload data tomorrow, even if they don't want to now). I think a better approach would be to have an interstitial screen after clicking the "log in with github" button that presents this information, along with a pair of buttons, "go back" and "continue to create account" or something like that.

yarikoptic:
I feel we had discussion re `3.` in the past, either on 

- https://github.com/dandi/dandi-archive/pull/496

or elsewhere.  That exemplifies that we have this outstanding problem and hopes for its resolution.  I *roger* the fact that it is "too late" and account is created, but if user leaves it in INCOMPLETE, we could potentially "garbage collect" it after some time if it does not get ACCEPTed or REJECTed, and completely avoid interaction requiring review/reject/accept which I would greatly appreciate.  Freshly filed related

- https://github.com/dandi/dandi-archive/issues/2040

Adding information about lab/project also would be useful to discover more about person.  I usually go to github page and/or google to often end up at such a page.

So, overall, I support the intention behind these changes.

waxlamp:
Ok, so as soon as Mike can verify that this will be ok, and we fix the linting error, let's merge.

Mike's already got the review on his TODOs, and I'm talking with Kabi about resolving the linting problem.

mvandenburgh:
This works as expected, but the styling is bit off due to the longer text in one of the new form fields -
Screenshot from 2024-10-16 17-12-15

Issue #2039, opened 2024-10-08
Add `Visualize in Neurosift` on Dandiset Landing Page


kabilar:
## Current state
Neurosift is available for NWB files on DANDI and when in the file browser can be accessed for a single asset by selecting the `External Services` menu.  However, it takes many clicks to get to a Neurosift page.

## Feature request
I would suggest that we also add a `Visualize in Neurosift` button on the landing page of Dandisets that contain NWB files.  This will help surface this tool so that more users see it.  Each Dandiset on Neurosift has a corresponding "summary" page of all the assets which can be linked to from the DLP (e.g. https://neurosift.app/?p=/dandiset&dandisetId=001065&dandisetVersion=draft).

Perhaps we can add this button under the `Dandiset Actions` section:
<img height="256" alt="image" src="https://github.com/user-attachments/assets/c30340fd-85cc-4a6d-babf-48dd09ce8237">

cc @magland @bendichter

magland:
@kabilar I'm in favor. :)

yarikoptic:
me too!  I just want us to ensure that we implement such specification in a generic "fashion".

<details>
<summary>Notes on a potential design</summary> 

 ATM we decide on allowing for neurosift simply given the filename `.nwb` extension.  But at the dandiset level we do not have in our summary file extensions, we have "Data standards" an in particular

```
  "assetsSummary": {
...
    "dataStandard": [
      {
        "name": "Neurodata Without Borders (NWB)",
        "schemaKey": "StandardsType",
        "identifier": "RRID:SCR_015242"
      }
    ],
```

so IMHO the decision should be based on that. But we might like to add to dandischema mapping to IDs like we did in unfinished

- https://github.com/dandi/dandi-cli/pull/958/files
 
We might later also add apps which work on BIDS datasets etc. Hence let's extend our specification/list of external services we coded in https://github.com/dandi/dandi-archive/blob/HEAD/web/src/views/FileBrowserView/FileBrowser.vue#L319 to 

1. specify the `target` ("asset" vs "dataset"; default to "asset"); 
2. rename `regex` into `path_regex` unless we want to add regexing based on dandiset id (makes little sense besides possible blacklisting, but then better be explicit)
  - we can even apply limitation by size at "dataset" level but not sure if there is a use case to limit??
3. in dandi-schema add shorter `id` to standards, i.e. `NWB` or `BIDS` instead of extended "name". Create a mapping from those ids to identifiers
[...]

PR #1810, opened 2024-01-11
For staging environment, include dynamic text for `dandi download` command

aaronkanzer:
When using `gui-staging.dandiarchive.org`, a minor inconsistency was observed where the CLI command was incorrect.

This PR makes a simple fix to render the appropriate text depending on if the user is using staging or production.

Closes #1553 

Cc @kabilar

PR #2041, opened 2024-10-14
Add neurosift external service for dandisets

magland:
Addresses #2039 

This doesn't yet include any of the general framework proposed by @yarikoptic in #2039. I do agree with what you are describing there, so we can view this as step 1 toward that. Whether it gets merged as is first or whether we wait until it is generalized that of course is up to you guys. I can work on it futher, but it will take me time to become more familiar with how data are passed around in the application.

### DIFF
--- a/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
+++ b/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
@@ -1,0 +1,94 @@
+<template>
+  <v-menu
+    offset-y
+    left
+  >
+    <template
+      #activator="{ on, attrs }"
+    >
+      <v-btn
+        id="external-dandiset-services"
+        outlined
+        block
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          color="primary"
+          left
+        >
+          mdi-web
+        </v-icon>
+        <span>External services</span>
+        <v-spacer />
+        <v-icon right>
+          mdi-chevron-down
+        </v-icon>
+      </v-btn>
+    </template>
+    <v-card
+    >
+      <v-card-title class="pb-0" style="min-width: fit-content;">
+        Select an external service:
+      </v-card-title>
+      <v-list>
+        <v-tooltip
+          open-on-hover
+          left
+        >
+          <template #activator="{ on }">
+            <div
+            v-on="on"
+            >
+              <v-list-item
+                :href="openInNeurosift()"
+              >
+                <v-icon
+                  color="primary"
+                  left
+                  small
+                >
+                  mdi-web
+                </v-icon>
+                Neurosift
+              </v-list-item>
+            </div>
+          </template>
+          <span>Open the Dandiset in Neurosift</span>
+        </v-tooltip>
+      </v-list>
+    </v-card>
+  </v-menu>
+</template>
+<script setup lang="ts">
+import { computed } from 'vue';
+import { useDandisetStore } from '@/stores/dandiset';
+
+const store = useDandisetStore();
+
+const currentDandiset = computed(() => store.dandiset);
+
+const openInNeurosift = () => {
+  const fullId = currentDandiset.value?.metadata?.id;
+  // e.g., DANDI:000776/0.241009.1509
+  if (!fullId) {
+    alert('Unexpected: No id field found in the metadata');
+    return;
+  }
+  const dandisetId = fullId.split('/')[0].split(':')[1];
+  const dandisetVersion = fullId.split('/')[1];
+  if (!dandisetId || !dandisetVersion) {
+    alert(`Unexpected: Invalid id field found in the metadata: ${fullId}`);
+    return;
+  }
+  return `https://neurosift.app/?p=/dandiset&dandisetId=${dandisetId}&dandisetVersion=${dandisetVersion}`;
+};
+
+</script>
+<style scoped>
+.v-btn--outlined {
+  border: thin solid #E0E0E0;
+  color: #424242;
+  font-weight: 400;
+}
+</style>
--- a/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
+++ b/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
@@ -28,9 +28,6 @@
     </template>
     <v-card
     >
-      <v-card-title class="pb-0" style="min-width: fit-content;">
-        External services:
-      </v-card-title>
       <v-list>
         <v-tooltip
           open-on-hover
--- a/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
+++ b/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
@@ -76,18 +76,11 @@ const openInNeurosift = () => {
   }
 
   const metadata = currentDandiset.value.metadata;
-
-  const dandisetUrl = metadata.url;
-  const staging = dandisetUrl.startsWith('https://gui-staging.dandiarchive.org/');
-
   const dandisetId = metadata.identifier.split(':')[1];
   const dandisetVersion = metadata.version;
+  const stagingParam = metadata.url.startsWith('https://gui-staging.dandiarchive.org/') ? '&staging=1' : '';
 
-  let href = `https://neurosift.app/?p=/dandiset&dandisetId=${dandisetId}&dandisetVersion=${dandisetVersion}`;
-  if (staging) {
-    href += '&staging=1';
-  }
-  return href;
+  return `https://neurosift.app/?p=/dandiset&dandisetId=${dandisetId}&dandisetVersion=${dandisetVersion}${stagingParam}`;
 };
 
 </script>
--- a/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
+++ b/web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
@@ -79,14 +79,10 @@ const openInNeurosift = () => {
 
   const dandisetUrl = dandiset.metadata.url;
   const staging = dandisetUrl.startsWith('https://gui-staging.dandiarchive.org/');
-  const fullId = dandiset.metadata.id;
-  // e.g., DANDI:000776/0.241009.1509
-  const dandisetId = fullId.split('/')[0].split(':')[1];
-  const dandisetVersion = fullId.split('/')[1];
-  if (!dandisetId || !dandisetVersion) {
-    alert(`Unexpected: Invalid id field found in the metadata: ${fullId}`);
-    return;
-  }
+
+  const dandisetId = dandiset.metadata.identifier.split(':')[1];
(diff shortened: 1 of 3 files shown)


yarikoptic:
I don't mind to start with an ad-hoc solution first, but we might need to figure out how to make styling better since it doesn't fit nicely ATM:

image

I think it might make sense it aligned with how it looks in Files browser where we have "Open with"

image


Also, while trying on that CI-driven against staging, I tried to go to 
https://neurosift.app/?p=/dandiset&dandisetId=215297&dandisetVersion=draft
and it is simply stuck on "Loading dataset..." instead of stating something more specific -- (that there is no such dataset ID on the main instance of the DANDI archive).

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,94 @@
...
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          color="primary"
+          left
+        >
+          mdi-web
+        </v-icon>
+        <span>External services</span>
```
yarikoptic, line 22:
```suggestion
        <span>Open with</span>
```

magland:
> Also, while trying on that CI-driven against staging, I tried to go to https://neurosift.app/?p=/dandiset&dandisetId=215297&dandisetVersion=draft and it is simply stuck on "Loading dataset..." instead of stating something more specific -- (that there is no such dataset ID on the main instance of the DANDI archive).

I updated Neurosift so it gives a more specific error message now.

And to solve this, query parameter staging=1 needs to be included in the URL when this is the staging site.

I accepted your commits. Do you think the styling is okay now?

magland:
@yarikoptic I have add handling for staging site, although haven't tested that.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,104 @@
...
+
+const currentDandiset = computed(() => store.dandiset);
+
+const openInNeurosift = () => {
+  const dandisetUrl = currentDandiset.value?.metadata?.url;
+  if (!dandisetUrl) {
+    alert('Unexpected: No url field found in the metadata of the dandiset');
+    return;
+  }
+  const staging = dandisetUrl.startsWith('https://gui-staging.dandiarchive.org/');
```
yarikoptic, line 77: Is there a better way to decide on the "instance name" used to tell apart at the Vue client level not only from staging but also from LINC or temporary testing one?  attn @kabilar @aaronkanzer @jjnesbitt . If so we better use that one and overall enable this only for stock DANDI main and staging but not for anything else.

kabilar: I don't think that we have yet abstracted the instance name.  At present, it is hard-coded in the LINC fork.

kabilar: #1810 does abstract the DANDI main and staging deployments.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,104 @@
...
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          color="primary"
+          left
+        >
+          mdi-web
+        </v-icon>
+        <span>Open with</span>
```
kabilar, line 22:
```suggestion
        <span>Visualize</span>
```

I like how NEMAR and OpenNeuro use verbs that describe what can be done by clicking the button (e.g. compute, analyze).  As other external services are added, we can add other general categories.

NEMAR:
image

OpenNeuro:
image

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,104 @@
...
+                  left
+                  small
+                >
+                  mdi-web
+                </v-icon>
+                Neurosift
+              </v-list-item>
+            </div>
+          </template>
+          <span>Open the Dandiset in Neurosift</span>
```
kabilar, line 57:
```suggestion
          <span>Visualize the Dandiset in Neurosift</span>
```

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,104 @@
...
+        <v-icon right>
+          mdi-chevron-down
+        </v-icon>
+      </v-btn>
+    </template>
+    <v-card
+    >
+      <v-card-title class="pb-0" style="min-width: fit-content;">
+        External services:
+      </v-card-title>
```
kabilar, line 33: To simplify, I would suggest removing this text.

```suggestion
```

kabilar (commented):
Thank you, @magland.  This is great.  I left a few minor comments above on the phrasing of the text.  I will leave the review of the javascript to the Kitware team.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
magland: I think in this case "Open with" is more accurate than "Visualize" because in Neurosift you can browse, explore and (with Dendro) analyze. Visualize is more relevant when opening an individual NWB file. Really, for the Dandiset view, Neurosift is an alternate way of viewing a subset of the information on the DANDI site, with the ability to do Neurosift-specific exploration and analysis.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
magland: See above.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
yarikoptic: And "Open with" is what I suggested based on what we have in file browser -- making it consistent is good. 
I do see a good UX in clear verbs, but indeed here it is not applicable IMHO, **but** it brings an idea to annotate those external services with "actions" or alike, which would be `['visualize', 'analyze']` or `['download']` which we could then somehow depict (icons?) along with the entries within `Open with` drop down.

yarikoptic:
> I updated Neurosift so it gives a more specific error message now.

thanks, confirming it does ;)

> And to solve this, query parameter staging=1 needs to be included in the URL when this is the staging site.

works well. FWIW, attempt to find similar doesn't work for staging

> > Similar dandisets Problem loading embeddings: Embedding not found for this dandiset 

> I accepted your commits. Do you think the styling is okay now?

yes, looks nice as to my non-artistic eye ;-)

> @yarikoptic I have add handling for staging site, although haven't tested that.

FWIW, if you missed, in the status comment from CIs there is
image
so it provides URL to the preview  at https://deploy-preview-2041--gui-staging-dandiarchive-org.netlify.app/ . It does look consistent/nice as to me, e.g

image

magland:
Wondering, is anything blocking this? Thx

yarikoptic:
@waxlamp please review 

@magland please remove changes to change log - they are automatically populated from the PR title when release comes

magland:
changelog entry removed.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
kabilar: That sounds good, thanks

kabilar:
Hi @magland, perhaps the Neurosift link should open in a new tab similar to the behavior in the file browser.

kabilar:
It looks like the option to open with Neurosift is also available for Dandisets that don't contain NWB files (e.g. [211293](https://deploy-preview-2041--gui-staging-dandiarchive-org.netlify.app/dandiset/211293?pos=6)).  On the [Neurosift page](https://neurosift.app/?p=/dandiset&dandisetId=211293&dandisetVersion=draft&staging=1) for this Dandiset it then says that zero assets are loaded.  Perhaps we can catch this case and disable the Neurosift link when there are no assets to be visualized with Neurosift?

magland:
@kabilar 

> Hi @magland, perhaps the Neurosift link should open in a new tab, which would be similar to the behavior in the file browser.

That sounds good to me... although I don't know how to add the target=_blank attribute to that link. Here's the relevant snippet

```
<v-list-item
    :href="openInNeurosift()"
>
```

Does anyone know how to add that?

> Perhaps we can catch this case and disable the Neurosift link when there are no assets to be visualized with Neurosift?

That also sounds good, but I'm not sure how to detect that in this file. My suspicion is that it would be a challenge.

aaronkanzer:
> @kabilar
> 
> > Hi @magland, perhaps the Neurosift link should open in a new tab, which would be similar to the behavior in the file browser.
> 
> That sounds good to me... although I don't know how to add the target=_blank attribute to that link. Here's the relevant snippet
> 
> ```
> <v-list-item
>     :href="openInNeurosift()"
> >
> ```
> 
> Does anyone know how to add that?
> 
> > Perhaps we can catch this case and disable the Neurosift link when there are no assets to be visualized with Neurosift?
> 
> That also sounds good, but I'm not sure how to detect that in this file. My suspicion is that it would be a challenge.

Hi @magland -- you should be able to bind a `target` HTML attribute to the `v-list-item`:

```
<v-list-item
    :href="openInNeurosift()"
    target="_blank"
>
```

magland:
Thanks @aaronkanzer that seems to have worked, and I made the change.

yarikoptic:
FWIW, I would have (and probably in the past did) argue to avoid breeding new tabs -- it is IMHO up to the user to open in a new tab/window or just to continue in the current one.  But we do have it to open in new tab for per-file browsing, so for consistency of UX it is indeed good to have it here in a new tab too.

waxlamp:
This looks pretty good so far. I am working on a few changes to the Javascript part to bring it into line with our style in other frontend components.

web/src/views/DandisetLandingView/ExternalDandisetServicesDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+  if (!currentDandiset.value.metadata) {
+    throw new Error('Dandiset metadata is undefined');
+  }
+
+  if (!currentDandiset.value.metadata.url) {
+    throw new Error('Dandiset metadata.url is undefined');
+  }
+
+  const metadata = currentDandiset.value.metadata;
+  const dandisetId = metadata.identifier.split(':')[1];
```
mvandenburgh, line 83: You can retrieve the dandiset identifier directly from the `dandiset` object instead of doing string manipulation:
```suggestion
  const dandisetId = currentDandiset.value.dandiset.identifier;
```

mvandenburgh (changes requested):
Looks good to me, just one minor code improvement

mvandenburgh (approved):

waxlamp (approved):

PR #2044, opened 2024-10-14
Auto-allow people with @alleninstitute.org email addresses

yarikoptic:
To quickly further lessen the burden on yours truly

### DIFF
--- a/dandiapi/api/views/auth.py
+++ b/dandiapi/api/views/auth.py
@@ -124,13 +124,13 @@ def user_questionnaire_form_view(request: HttpRequest) -> HttpResponse:
             not questionnaire_already_filled_out
             and user_metadata.status == UserMetadata.Status.INCOMPLETE
         ):
-            is_edu_email: bool = user.email.endswith('.edu') or user.email.endswith(
+            should_auto_approve: bool = user.email.endswith('.edu') or user.email.endswith(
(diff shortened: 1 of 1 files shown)


kabilar (approved):

dandiapi/api/views/auth.py
```diff
@@ -124,7 +124,9 @@ def user_questionnaire_form_view(request: HttpRequest) -> HttpResponse:
             not questionnaire_already_filled_out
             and user_metadata.status == UserMetadata.Status.INCOMPLETE
         ):
-            is_edu_email: bool = user.email.endswith('.edu')
+            is_edu_email: bool = user.email.endswith('.edu') or user.email.endswith(
+                '@alleninstitute.org'
+            )
```
mvandenburgh, line 129: I would just suggest changing this variable name to reflect it's new meaning -

```suggestion
            should_auto_approve: bool = user.email.endswith('.edu') or user.email.endswith(
                '@alleninstitute.org'
            )
```

mvandenburgh (changes requested):

dandiapi/api/views/auth.py
yarikoptic: done

mvandenburgh (approved):

PR #2053, opened 2024-10-18
Pin ruff version to avoid spurious test failures

waxlamp:
Since ruff is fast moving and freely publishes [new versions with breaking changes](https://docs.astral.sh/ruff/versioning/), it behooves us to pin a ruff version so that local dev and CI don't go out of sync with each other, complicating the review process for PRs.

This PR pins ruff to the 0.6.x minor series and adds a version report to each test (to make version numbers easier to determine in the future).

(See commit messages for details.)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ deps =
     codespell~=2.0
     ruff~=0.6.2
 commands =
+    ruff -V
     ruff check
     ruff format --check
     codespell .
@@ -23,6 +24,7 @@ package = skip
 deps =
     ruff~=0.6.2
 commands =
+    ruff -V
     ruff check --fix-only
     ruff format
 
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ package = skip
 ignore_errors = True
 deps =
     codespell~=2.0
-    ruff
+    ruff~=0.6.2
 commands =
     ruff check
     ruff format --check
@@ -21,7 +21,7 @@ commands =
 [testenv:format]
 package = skip
 deps =
-    ruff
+    ruff~=0.6.2
 commands =
     ruff check --fix-only
     ruff format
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
> see if this actually works

So, it works, but in a way that is kind of useless: the `ruff-latest` test fails inside its run (as expected) but the job shows a green checkmark in the results panel. Not sure if there's a better thing here (something that shows the failure but doesn't fail the run).

waxlamp:
> > see if this actually works
> 
> So, it works, but in a way that is kind of useless: the `ruff-latest` test fails inside its run (as expected) but the job shows a green checkmark in the results panel. Not sure if there's a better thing here (something that shows the failure but doesn't fail the run).

@danlamanna, @jjnesbitt, @mvandenburgh: would this be better as a nightly test? That would require duplicating the GH Action spec (which is a negative) but we could allow it to fail freely without compromising the ability to merge PRs, etc.

tox.ini
```diff
@@ -12,17 +12,19 @@ package = skip
 ignore_errors = True
 deps =
     codespell~=2.0
-    ruff
+    ruff~=0.6.2
 commands =
+    ruff -V
```
mvandenburgh, line 17: I usually go with the longer form CLI arguments if they're being invoked programatically.
```suggestion
    ruff --version
```

tox.ini
```diff
@@ -12,17 +12,19 @@ package = skip
...
     ruff format --check
     codespell .
 
 [testenv:format]
 package = skip
 deps =
-    ruff
+    ruff~=0.6.2
 commands =
+    ruff -V
```
mvandenburgh, line 27:
```suggestion
    ruff --version
```

mvandenburgh (approved):
Minor suggestion about invoking `ruff`, either way this looks good to me.

tox.ini
waxlamp: Oh yes this is better!

mvandenburgh (approved):

PR #2054, opened 2024-10-21
Update new user registration questions

kabilar:
Supersedes #2038

### DIFF
--- a/dandiapi/api/templates/api/account/questionnaire_form.html
+++ b/dandiapi/api/templates/api/account/questionnaire_form.html
@@ -17,6 +17,13 @@
         {% for name, value in query_params.items %}
           <input type="hidden" name="{{ name }}" value="{{ value }}">
         {% endfor %}
+
+        <div>
+          <b><i>
+              Please note that users can freely search, view, and download public datasets without an account!
+          </b></i>
+        </div>
+
         <button id="btn" class="btn waves-effect waves-light blue" type="submit" disabled>
           <span>Submit</span>
           <i class="material-icons right">send</i>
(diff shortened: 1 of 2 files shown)


waxlamp:
@kabilar, @mvandenburgh: I moved the parenthesis from the question itself down into a bolded, italicized div right above the submit button. Take a look and lmk what you think. Particularly:
- @kabilar: is it ok to have the admonition there instead of in the question?
- @mvandenburgh: there is still trouble with the styling for those form fields (crunch your screen width down to see what I mean) but at least it's less likely with the shorter question -- is that ok for now at least?
    - Related question: should we bother to "fix" this? It doesn't seem worth the effort. Thoughts?

Note also that the dandi-cli integratin test is failing due to the fork not having access to the repo secrets. In this case I don't think it matters because this is purely a cosmetic change in the user signup flow (which doesn't affect the CLI). But in the future, there are apparently ways to share the appropriate secrets with forks and we can look into that.

waxlamp (approved):
This looks good to me (pending @kabilar and @mvandenburgh verbally approving my change).

mvandenburgh:
Looks good to me :+1:
Screenshot from 2024-10-22 12-40-45

kabilar:
> Looks good to me 👍 Screenshot from 2024-10-22 12-40-45

Hi @mvandenburgh, this looks like the current form.

kabilar:
> * @kabilar: is it ok to have the admonition there instead of in the question?

Hi @waxlamp Yes, thank you.  I just noticed the [note](https://github.com/kabilar/dandi-archive/blob/5f7805f63e361597629a9ee6241206175583d1ed/dandiapi/api/templates/api/account/questionnaire_form.html#L5) at the top of the form.  Perhaps we should remove this note or replace it with the phrasing that I suggested.

mvandenburgh:
Sorry for the confusion, here's the correct form 
Screenshot from 2024-10-22 13-28-33

PR #2055, opened 2024-10-21
Add a CI job that runs linting and formatting against latest ruff

waxlamp:
This is an experimental idea based off of the work in #2053, which pins Ruff to a specific minor version so that dev and CI don't go out of sync. The idea is to give CI the ability to run against _latest_ Ruff and let the failure of that run indicate new rules, etc., that we may want to adopt by fixing those errors and re-pinning the version used in the "real" runs.

### DIFF
--- a/.github/workflows/backend-ci.yml
+++ b/.github/workflows/backend-ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tox-env: [lint, type, test, check-migrations]
+        tox-env: [lint, type, test, check-migrations, ruff-latest]
     services:
       postgres:
         image: postgres:latest
@@ -48,6 +48,7 @@ jobs:
           pip install --upgrade pip
           pip install tox
       - name: Run tests
+        continue-on-error: ${{ matrix.tox-env == 'ruff-latest' }}
         run: |
           tox -e ${{ matrix.tox-env }}
         env:
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,16 @@ commands =
     ruff check --fix-only
     ruff format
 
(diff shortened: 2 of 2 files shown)


waxlamp:
Since this PR was spun out from #2053, I'm copying over the comments I left regarding this experiment:

---

> > see if this actually works
> 
> So, it works, but in a way that is kind of useless: the `ruff-latest` test fails inside its run (as expected) but the job shows a green checkmark in the results panel. Not sure if there's a better thing here (something that shows the failure but doesn't fail the run).

@danlamanna, @jjnesbitt, @mvandenburgh: would this be better as a nightly test? That would require duplicating the GH Action spec (which is a negative) but we could allow it to fail freely without compromising the ability to merge PRs, etc.

waxlamp:
> > > see if this actually works
> > 
> > 
> > So, it works, but in a way that is kind of useless: the `ruff-latest` test fails inside its run (as expected) but the job shows a green checkmark in the results panel. Not sure if there's a better thing here (something that shows the failure but doesn't fail the run).
> 
> @danlamanna, @jjnesbitt, @mvandenburgh: would this be better as a nightly test? That would require duplicating the GH Action spec (which is a negative) but we could allow it to fail freely without compromising the ability to merge PRs, etc.

Dan pointed out that even with a nightly run, alarm fatigue will set in. A better approach would be for us to just pay attention and run against latest Ruff when new versions come out.

I will try converting this to use the ["workflow dispatch" mode](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch) which would avoid alarm fatigue, but make it a bit easier to actually run against the latest version.

PR #2057, opened 2024-10-23
Use pre-commit hooks in tox configuration

waxlamp:
This extends #2045 to use the same pre-commit hooks to perform linting and formatting via tox.ini. In particular, this "refactors" the specification of what versions of codespell and ruff to use to just one place: the pre-commit config file.

The variants of the pre-commit hooks that are needed to replace the tox.ini commands are listed in the pre-commit config using the `manual` stage specifier with appropriate aliases, meaning they will not be invoked as a git hook ever, and enabling them to be invoked by name in the tox config.

This PR also modifies the dev Dockerfile to build it to run as the user from the host system. That is, instead of running as root inside the container, it will instead run as the same (non-root) user who is doing development work (including building this container, etc.). That enables the use of `pre-commit` both inside and outside the container, since `pre-commit` requires examining Git status, etc. Without this modification, you get errors about "dubious ownership" of the Git repository.

I am totally willing to peel off the Dockerfile changes to a separate PR if we require further discussion, especially if some of these ideas are suitable for general use with Docker containers. Let me know what you think, @mvandenburgh.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -5,7 +5,7 @@ This is the simplest configuration for developers to start with.
 
 ### Initial Setup
 1. Run `docker compose pull` to ensure you have the latest versions of the service container images.
-2. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
+2. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u) --build-arg GROUPNAME=$(id -n -g)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
 3. Run `docker compose run --rm django ./manage.py migrate`
 4. Run `docker compose run --rm django ./manage.py createcachetable`
 5. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,13 +4,14 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
-2. Run `docker compose run --rm django ./manage.py migrate`
-3. Run `docker compose run --rm django ./manage.py createcachetable`
-4. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
+1. Run `docker compose pull` to ensure you have the latest versions of the service container images.
+2. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
+3. Run `docker compose run --rm django ./manage.py migrate`
+4. Run `docker compose run --rm django ./manage.py createcachetable`
+5. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
    and follow the prompts to create your own user.
    This sets your username to your git email to ensure parity with how GitHub logins work. You can also replace the command substitution expression with a literal email address, or omit the `--email` option entirely to run the command in interactive mode.
-5. Run `docker compose run --rm django ./manage.py create_dev_dandiset --owner $(git config user.email)`
+6. Run `docker compose run --rm django ./manage.py create_dev_dandiset --owner $(git config user.email)`
    to create a dummy dandiset to start working with.
 
 ### Run Application
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,12 +4,13 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker compose run --rm django ./manage.py migrate`
-2. Run `docker compose run --rm django ./manage.py createcachetable`
-3. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
+1. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
+2. Run `docker compose run --rm django ./manage.py migrate`
+3. Run `docker compose run --rm django ./manage.py createcachetable`
+4. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
    and follow the prompts to create your own user.
    This sets your username to your git email to ensure parity with how GitHub logins work. You can also replace the command substitution expression with a literal email address, or omit the `--email` option entirely to run the command in interactive mode.
-4. Run `docker compose run --rm django ./manage.py create_dev_dandiset --owner $(git config user.email)`
+5. Run `docker compose run --rm django ./manage.py create_dev_dandiset --owner $(git config user.email)`
    to create a dummy dandiset to start working with.
 
 ### Run Application
@@ -21,7 +22,7 @@ This is the simplest configuration for developers to start with.
 Occasionally, new package dependencies or schema changes will necessitate
 maintenance. To non-destructively update your development stack at any time:
 1. Run `docker compose pull`
-2. Run `docker compose build --pull --no-cache`
+2. Run `docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` (omitting the `--build-arg` arguments if you did so in Step 1 of *Initial Setup* above).
 3. Run `docker compose run --rm django ./manage.py migrate`
 
 ## Develop Natively (advanced)
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,7 +23,7 @@ This is the simplest configuration for developers to start with.
 Occasionally, new package dependencies or schema changes will necessitate
 maintenance. To non-destructively update your development stack at any time:
 1. Run `docker compose pull`
-2. Run `docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` (omitting the `--build-arg` arguments if you did so in Step 1 of *Initial Setup* above).
+2. Run `docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u) --build-arg GROUPNAME=$(id -n -g)` (omitting the `--build-arg` arguments if you did so in Step 1 of *Initial Setup* above).
 3. Run `docker compose run --rm django ./manage.py migrate`
 
 ## Develop Natively (advanced)
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -18,7 +18,7 @@ ARG USERID=0
 ARG LOGIN=root
 ARG GROUPID=0
 ARG GROUPNAME=root
-RUN getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}
+RUN getent group ${GROUPID} || addgroup --allow-all-names --gid ${GROUPID} ${GROUPNAME}
 RUN getent passwd ${USERID} || adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN
 
 # Create the project folder and make the user its owner.
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -19,7 +19,7 @@ ARG LOGIN=root
 ARG GROUPID=0
 ARG GROUPNAME=root
 RUN getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}
-RUN if [ "${USERID}" != 0 ]; then adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi
+RUN getent passwd ${USERID} || adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN
 
 # Create the project folder and make the user its owner.
 RUN mkdir -p /opt/django-project
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -1,15 +1,15 @@
 FROM python:3.11-slim
 
-# Set some behaviors for Python.
-ENV PYTHONDONTWRITEBYTECODE 1
-ENV PYTHONUNBUFFERED 1
-
 # Install system librarires for Python packages:
 # * psycopg2
 RUN apt-get update && \
     apt-get install --no-install-recommends --yes \
     libpq-dev gcc libc6-dev git && \
     rm -rf /var/lib/apt/lists/*
+
+# Set some behaviors for Python.
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
 
 # Install pre-commit. Use `pip` instead of `apt` in order to get a recent
 # version instead of whatever is in the image's package lists.
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -19,7 +19,7 @@ ARG LOGIN=root
 ARG GROUPID=0
 ARG GROUPNAME=root
 RUN getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}
-RUN if [ "${USERID}" != 0 ]; then adduser --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi
+RUN if [ "${USERID}" != 0 ]; then adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi
 
 # Create the project folder and make the user its owner.
 RUN mkdir -p /opt/django-project
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -11,10 +11,6 @@ RUN apt-get update && \
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
-# Install pre-commit. Use `pip` instead of `apt` in order to get a recent
-# version instead of whatever is in the image's package lists.
-RUN pip install pre-commit
-
 # Create a normal user (that can be made to match the dev's user stats), falling
 # back to root; if the root user is specified, don't actually run the `adduser`
 # command.
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -15,8 +15,10 @@ ENV PYTHONUNBUFFERED 1
 # back to root; if the root user is specified, don't actually run the `adduser`
 # command.
 ARG USERID=0
-ARG GROUPID=0
 ARG LOGIN=root
+ARG GROUPID=0
+ARG GROUPNAME=root
+RUN getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}
 RUN if [ "${USERID}" != 0 ]; then adduser --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi
 
 # Create the project folder and make the user its owner.
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -1,4 +1,9 @@
 FROM python:3.11-slim
+
+# Set some behaviors for Python.
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
 # Install system librarires for Python packages:
 # * psycopg2
 RUN apt-get update && \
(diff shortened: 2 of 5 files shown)


yarikoptic:
> There are some quirks w.r.t. running this via Docker:
> 
> 1. Because the permissions for the `.git` directory are not those of `root`,

IMHO nobody should run dev tools as root -- should be mapped to the real UID via `--user` option etc. (in https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd#L198C1-L199C19 we passed UID and GID env vars)  -- what container we are talking about to try on?

mvandenburgh:
Should the difficulties around running linting inside docker be documented in `DEVELOPMENT.md`?

yarikoptic:
when do `git` and `tox` need to run **inside** docker environment?

waxlamp:
> IMHO nobody should run dev tools as root -- should be mapped to the real UID via --user option etc. (in https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd#L198C1-L199C19 we passed UID and GID env vars) -- what container we are talking about to try on?

Nobody is running dev tools as root 🙂. Docker containers, by default, run as root inside the container (even when launched by a non-root user on the host system). Your example is using Singularity, which was designed in part to work with non-root users by default.

> when do git and tox need to run inside docker environment?

Tox needs to run inside the container precisely when the dev wants it to. For example, I often run the command `docker compose run --rm django tox -e lint` etc. to perform linting in my dev environment. Git does not need to run inside the container, though `pre-commit` uses git to determine how to run its hooks, which is where I was running into trouble.

Resonant applications provides this container as a convenience. Adopting pre-commit in dandi-archive caused some of the "rev matching" problems to surface, so I'm looking into how to fix it to provide a generally better experience across Resonant apps.

In any case, I am working on a way to build the container so that it runs as the same user as the dev who built it. If that works, then it will provide parity between host OS and Docker container. Stay tuned for updates.

yarikoptic:
> > There are some quirks w.r.t. running this via Docker:
> > 
> > 1. Because the permissions for the `.git` directory are not those of `root`,
> 
> IMHO nobody should run dev tools as root -- should be mapped to the real UID via `--user` option etc. (in https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd#L198C1-L199C19 we passed UID and GID env vars) -- what container we are talking about to try on?

FWIW -- checked that it works:

```shell
❯ docker compose run --rm --user "$(id -u)" django touch 123
WARN[0000] /home/yoh/proj/dandi/dandi-archive/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
WARN[0000] /home/yoh/proj/dandi/dandi-archive/docker-compose.override.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
[+] Creating 3/0
 ✔ Container dandi-archive-minio-1     Running                                                                                                                                 0.0s 
 ✔ Container dandi-archive-rabbitmq-1  Running                                                                                                                                 0.0s 
 ✔ Container dandi-archive-postgres-1  Running                                                                                                                                 0.0s 

❯ ls -ld 123
-rw-r--r-- 1 yoh root 0 Oct 28 12:18 123
```

so just add ` --user "$(id -u)"` to your invocation of tox within container.  Depending on scripts, you might also benefit from passing env vars as I pointed our so more complete would be ` --user "$(id -u)" -e "UID=$(id -u)" -e "GID=$(id -g)"` (potentially improved to your liking)

yarikoptic:
note: might need to clean up "root owned" artifacts from prior runs of tox first, and then it seems to work just fine

```
❯ sudo rm -rf .tox/lint/log .ruff_cache
❯ docker compose run --rm --user "$(id -u)" -e "UID=$(id -u)" -e "GID=$(id -g)" django tox -e lint
WARN[0000] /home/yoh/proj/dandi/dandi-archive/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
WARN[0000] /home/yoh/proj/dandi/dandi-archive/docker-compose.override.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
[+] Creating 3/0
 ✔ Container dandi-archive-postgres-1  Running                                                                                                                                 0.0s 
 ✔ Container dandi-archive-minio-1     Running                                                                                                                                 0.0s 
 ✔ Container dandi-archive-rabbitmq-1  Running                                                                                                                                 0.0s 
lint: commands[0]> ruff --version
ruff 0.6.9
lint: commands[1]> ruff check
All checks passed!
lint: commands[2]> ruff format --check
155 files already formatted
lint: commands[3]> codespell .
  lint: OK (0.34=setup[0.06]+cmd[0.01,0.02,0.02,0.23] seconds)
  congratulations :) (0.44 seconds)

```

waxlamp:
> so just add  --user "$(id -u)" to your invocation of tox within container

This mostly works--depending on how the container image was *built* you can still run into problems. Plus, having to add the `--user` flag on every invocation is annoying and error-prone (you might forget to do it, etc.)

> note: might need to clean up "root owned" artifacts from prior runs of tox first, and then it seems to work just fine

This is really the core of the issue--having to remember the `--user` flag, plus the fact that your root user might be "polluting" your dev directory makes for an unergonomic dev experience.

But I'm experimenting with moving these things to the build phase so that the regular commands "just work". Stand by, I have an incoming fix for all this (I hope).

yarikoptic:
> > note: might need to clean up "root owned" artifacts from prior runs of tox first, and then it seems to work just fine
> 
> This is really the core of the issue--having to remember the `--user` flag, plus the fact that your root user might be "polluting" your dev directory makes for an unergonomic dev experience.

The core issue is :

> > IMHO nobody should run dev tools as root -- should be mapped to the real UID via --user option etc. (in https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd#L198C1-L199C19 we passed UID and GID env vars) -- what container we are talking about to try on?
> 
> Nobody is running dev tools as root 🙂. Docker containers, by default, run as root inside the container (even when launched by a non-root user on the host system).

and the fact that "Docker containers" are not animate, they do not run themselves. They are executed by people who run them and as a result running dev tools as root.  It is just a fact.

> Tox needs to run inside the container precisely when the dev wants it to.

It is something to remember about and to not execute `docker` with any dev tools/scripts without precautions being taken -- ideally to avoid that generally.  Same advice for `sudo` - too many people often did `sudo pip install` causing all kinds of pains later on -- all of that is IMHO just a bad practice. 

In the longer run it might be better to look to switch to podman and having fake root inside the container space so touched bind-mounted files would still be fine.

I have expressed my opinion ("never run dev tools ran as root" - whether through `docker` or `sudo` or ...), and will quiet down on this particular issue.

waxlamp:
> > > note: might need to clean up "root owned" artifacts from prior runs of tox first, and then it seems to work just fine
> > 
> > 
> > This is really the core of the issue--having to remember the `--user` flag, plus the fact that your root user might be "polluting" your dev directory makes for an unergonomic dev experience.
> 
> The core issue is :
> 
> > > IMHO nobody should run dev tools as root -- should be mapped to the real UID via --user option etc. (in https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd#L198C1-L199C19 we passed UID and GID env vars) -- what container we are talking about to try on?
> > 
> > 
> > Nobody is running dev tools as root 🙂. Docker containers, by default, run as root inside the container (even when launched by a non-root user on the host system).
> 
> and the fact that "Docker containers" are not animate, they do not run themselves. They are executed by people who run them and as a result running dev tools as root. It is just a fact.
> 
> > Tox needs to run inside the container precisely when the dev wants it to.
> 
> It is something to remember about and to not execute `docker` with any dev tools/scripts without precautions being taken -- ideally to avoid that generally. Same advice for `sudo` - too many people often did `sudo pip install` causing all kinds of pains later on -- all of that is IMHO just a bad practice.
> 
> In the longer run it might be better to look to switch to podman and having fake root inside the container space so touched bind-mounted files would still be fine.
> 
> I have expressed my opinion ("never run dev tools as root" - whether through `docker` or `sudo` or ...), and will quiet down on this particular issue.

I did not mean to be dismissive--I was just pointing out a reality of how most people run Docker containers. As I said, I'm working on a solution that addresses this issue head on (it is similar to what you suggested, but even better).

DEVELOPMENT.md
```diff
@@ -4,12 +4,13 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker compose run --rm django ./manage.py migrate`
-2. Run `docker compose run --rm django ./manage.py createcachetable`
-3. Run `docker compose run --rm django ./manage.py createsuperuser --email $(git config user.email)`
+1. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
```
yarikoptic, line 7: In my case following instructions failed on the server (`smaug`, note: used `docker-compose` since my docker is old there -- 20.x.x) where I had prior copy of the `minio/minio:latest` already available and docker compose did not aggressively pull them before running.


```suggestion
0. Run `docker compose pull` to ensure that possibly present local `:latest` docker images are up to date.
1. Run `docker compose build --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` to build the development container image. This builds the image to work with your (non-root) development user so that the linting and formatting commands work inside and outside of the container. If you prefer to build the container image so that it runs as `root`, you can omit the `--build-arg` arguments (but you will likely run into trouble running those commands).
```

DEVELOPMENT.md
```diff
@@ -4,12 +4,13 @@
...
+2. Run `docker compose run --rm django ./manage.py migrate`
```
yarikoptic, line 8: note, it might be worth just to number them all as `1.` to avoid causing the diff if only indexing to change. Then numbers would be rendered by the markdown renderer

waxlamp: It is a nice feature of Markdown but I'm going to avoid it because for the most part, people are reading the text version of this file in their dev environment, rather than the rendered version that appears on GitHub (meaning, numbering the steps explicitly is necessary).

DEVELOPMENT.md
waxlamp: Good catch, thank you. I took your suggestion, rewording a bit: 6acf8163534fc1dc5adf6ec74487804d70d8e5de

waxlamp: I suppose we can specify a minimal version, but as you discovered, even your "ancient" version worked (you just had to supply a different command). What do you recommend here?

dev/django.Dockerfile
```diff
@@ -1,13 +1,31 @@
...
 RUN apt-get update && \
     apt-get install --no-install-recommends --yes \
     libpq-dev gcc libc6-dev git && \
     rm -rf /var/lib/apt/lists/*
 
-ENV PYTHONDONTWRITEBYTECODE 1
-ENV PYTHONUNBUFFERED 1
+# Install pre-commit. Use `pip` instead of `apt` in order to get a recent
+# version instead of whatever is in the image's package lists.
+RUN pip install pre-commit
```
mvandenburgh, line 16: Should we just add `pre-commit` to the `dev` extra in `setup.py`? That way, it will be installed in the docker image by the existing `pip install` step without needing a separate one for pre-commit. Also, it will be automatically installed for native users when they do a `pip install -e .[dev]`. We already include `tox` there, so it seems like adding `pre-commit` makes sense.

dev/django.Dockerfile
```diff
@@ -1,13 +1,31 @@
 FROM python:3.11-slim
+
+# Set some behaviors for Python.
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
```
mvandenburgh, line 6: Why move these? It seems unrelated to this PR.

mvandenburgh (commented):
I get this when I run the `docker compose build` command. I think the `adduser` command doesn't support the `.` character in usernames (my username on my ubuntu system is `michael.vandenburgh`).  

```
[+] Building 1.2s (11/33)                                                                                                                                          docker:default
 => [celery internal] load build definition from django.Dockerfile                                                                                                           0.0s
 => => transferring dockerfile: 2.63kB                                                                                                                                       0.0s
 => [django internal] load build definition from django.Dockerfile                                                                                                           0.0s
 => => transferring dockerfile: 2.63kB                                                                                                                                       0.0s
 => [celery internal] load metadata for docker.io/library/python:3.11-slim                                                                                                   0.6s
 => [celery internal] load .dockerignore                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                              0.0s
 => [django internal] load .dockerignore                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                              0.0s
 => [celery internal] load build context                                                                                                                                     0.4s
 => => transferring context: 657.14kB                                                                                                                                        0.4s
 => [django  1/16] FROM docker.io/library/python:3.11-slim@sha256:5148c0e4bbb64271bca1d3322360ebf4bfb7564507ae32dd639322e4952a6b16                                           0.0s
 => CACHED [django  2/16] RUN apt-get update &&     apt-get install --no-install-recommends --yes     libpq-dev gcc libc6-dev git &&     rm -rf /var/lib/apt/lists/*         0.0s
 => CACHED [django  3/16] RUN pip install pre-commit                                                                                                                         0.0s
 => ERROR [django  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh; fi                  0.5s
 => [django internal] load build context                                                                                                                                     0.4s
 => => transferring context: 657.14kB                                                                                                                                        0.4s
------
 > [django  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh; fi:
0.489 adduser: Please enter a username matching the regular expression
0.489             configured via the NAME_REGEX configuration variable.  Use the
0.489             `--allow-bad-names' option to relax this check or reconfigure
0.489             NAME_REGEX in configuration.
------
failed to solve: process "/bin/sh -c if [ \"${USERID}\" != 0 ]; then adduser --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi" did not complete successfully: exit code: 1
```

The man page for `adduser` says you can use `--allow-bad-names` to override this, but then I get this error:
```
 > [celery  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh --allow-bad-names; fi:
0.361 adduser: The GID 692584961 does not exist.
```

dev/django.Dockerfile
waxlamp: Good idea. See 22c8fdf57a5264093b252760cd0ef518ea20c67f.

waxlamp:
> I get this when I run the `docker compose build` command. I think the `adduser` command doesn't support the `.` character in usernames (my username on my ubuntu system is `michael.vandenburgh`).
> 
> ```
> [+] Building 1.2s (11/33)                                                                                                                                          docker:default
>  => [celery internal] load build definition from django.Dockerfile                                                                                                           0.0s
>  => => transferring dockerfile: 2.63kB                                                                                                                                       0.0s
>  => [django internal] load build definition from django.Dockerfile                                                                                                           0.0s
>  => => transferring dockerfile: 2.63kB                                                                                                                                       0.0s
>  => [celery internal] load metadata for docker.io/library/python:3.11-slim                                                                                                   0.6s
>  => [celery internal] load .dockerignore                                                                                                                                     0.0s
>  => => transferring context: 2B                                                                                                                                              0.0s
>  => [django internal] load .dockerignore                                                                                                                                     0.0s
>  => => transferring context: 2B                                                                                                                                              0.0s
>  => [celery internal] load build context                                                                                                                                     0.4s
>  => => transferring context: 657.14kB                                                                                                                                        0.4s
>  => [django  1/16] FROM docker.io/library/python:3.11-slim@sha256:5148c0e4bbb64271bca1d3322360ebf4bfb7564507ae32dd639322e4952a6b16                                           0.0s
>  => CACHED [django  2/16] RUN apt-get update &&     apt-get install --no-install-recommends --yes     libpq-dev gcc libc6-dev git &&     rm -rf /var/lib/apt/lists/*         0.0s
>  => CACHED [django  3/16] RUN pip install pre-commit                                                                                                                         0.0s
>  => ERROR [django  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh; fi                  0.5s
>  => [django internal] load build context                                                                                                                                     0.4s
>  => => transferring context: 657.14kB                                                                                                                                        0.4s
> ------
>  > [django  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh; fi:
> 0.489 adduser: Please enter a username matching the regular expression
> 0.489             configured via the NAME_REGEX configuration variable.  Use the
> 0.489             `--allow-bad-names' option to relax this check or reconfigure
> 0.489             NAME_REGEX in configuration.
> ------
> failed to solve: process "/bin/sh -c if [ \"${USERID}\" != 0 ]; then adduser --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN; fi" did not complete successfully: exit code: 1
> ```
> 
> The man page for `adduser` says you can use `--allow-bad-names` to override this, but then I get this error:
> 
> ```
>  > [celery  4/16] RUN if [ "692589655" != 0 ]; then adduser --uid 692589655 --gid 692584961 --home /home/michael.vandenburgh michael.vandenburgh --allow-bad-names; fi:
> 0.361 adduser: The GID 692584961 does not exist.
> ```

Thanks for the edge cases (regarding username nonconformance and group existence). Commits d7026cc83eb0e1f50062a2c976bcbb1cec7c27f0 and 2ada467972bb246e1a3a7faa6f3a21353d1e851f should help.

DEVELOPMENT.md
```diff
@@ -21,7 +23,7 @@ This is the simplest configuration for developers to start with.
 Occasionally, new package dependencies or schema changes will necessitate
 maintenance. To non-destructively update your development stack at any time:
 1. Run `docker compose pull`
-2. Run `docker compose build --pull --no-cache`
+2. Run `docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)` (omitting the `--build-arg` arguments if you did so in Step 1 of *Initial Setup* above).
```
asmacdo, line 26: This didnt work for me 

<details><summary>Fail copypaste</summary>

```
austin@fancy ~/devel/dandi-archive (branch detox) docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u)
WARN[0000] /home/austin/devel/dandi-archive/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
WARN[0000] /home/austin/devel/dandi-archive/docker-compose.override.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
[+] Building 25.8s (10/33)                            docker:default
 => [django internal] load build definition from django.Docker  0.0s
 => => transferring dockerfile: 2.62kB                          0.0s
 => [celery internal] load build definition from django.Docker  0.0s
 => => transferring dockerfile: 2.62kB                          0.0s
 => [celery internal] load metadata for docker.io/library/pyth  0.6s
 => [celery internal] load .dockerignore                        0.0s
 => => transferring context: 2B                                 0.0s
 => [django internal] load .dockerignore                        0.0s
 => => transferring context: 2B                                 0.0s
 => CACHED [celery  1/16] FROM docker.io/library/python:3.11-s  0.0s
 => [django internal] load build context                        0.0s
 => => transferring context: 5.23kB                             0.0s
 => [django  2/16] RUN apt-get update &&     apt-get install   24.7s
 => [celery internal] load build context                        0.0s
 => => transferring context: 5.23kB                             0.0s
 => ERROR [celery  3/16] RUN getent group 1000 || addgroup --g  0.4s
------                                                               
 > [celery  3/16] RUN getent group 1000 || addgroup --gid 1000 root: 
0.289 addgroup: The group `root' already exists.                     
------                                                               
failed to solve: process "/bin/sh -c getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}" did not complete successfully: exit code: 1
```
</details>

Pretty sure we just need to add GROUPNAME, since it was trying to use `root`. I tried the suggestion below, and the build progressed past this point.

```suggestion
2. Run `docker compose build --pull --no-cache --build-arg USERID=$(id -u) --build-arg GROUPID=$(id -g) --build-arg LOGIN=$(id -n -u) --build-arg GROUPNAME=$(id -gn)` (omitting the `--build-arg` arguments if you did so in Step 1 of *Initial Setup* above).
```

waxlamp: There is 2ada467 in the commit stream. Can you pull this branch and try again?

asmacdo: @waxlamp yeah this line needs to match the line you changed in that commit

waxlamp: Ah got it, see 026dc13.

dev/django.Dockerfile
```diff
@@ -1,14 +1,30 @@
...
 
+# Create a normal user (that can be made to match the dev's user stats), falling
+# back to root; if the root user is specified, don't actually run the `adduser`
+# command.
+ARG USERID=0
+ARG LOGIN=root
+ARG GROUPID=0
+ARG GROUPNAME=root
+RUN getent group ${GROUPID} || addgroup --gid ${GROUPID} ${GROUPNAME}
+RUN getent passwd ${USERID} || adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN
```
mvandenburgh, line 22: Now, I get 

```
 => ERROR [django  3/16] RUN getent group 692584961 || addgroup --gid 692584961 domain^users                                                0.4s
------                                                                                                                                           
 > [django  3/16] RUN getent group 692584961 || addgroup --gid 692584961 domain^users:                                                           
0.386 addgroup: To avoid problems, the username should consist only of                                                                           
0.386             letters, digits, underscores, periods, at signs and dashes, and                                                                
0.386             not start with a dash (as defined by IEEE Std 1003.1-2001). For                                                                
0.386             compatibility with Samba machine accounts, $ is also supported                                                                 
0.386             at the end of the username.  (Use the `--allow-all-names' option
0.386             to bypass this restriction.)
```

It seems like we also need to allow "bad" group names. Adding `--allow-all-names` as the error message suggests, seems to fix it.

```suggestion
RUN getent group ${GROUPID} || addgroup --allow-all-names --gid ${GROUPID} ${GROUPNAME}
RUN getent passwd ${USERID} || adduser --allow-bad-names --uid ${USERID} --gid ${GROUPID} --home /home/${LOGIN} $LOGIN
```

dev/django.Dockerfile
```diff
@@ -17,13 +33,28 @@ COPY ./pyproject.toml /opt/django-project/pyproject.toml
...
+USER $LOGIN
+
 # Use a directory name which will never be an import name, as isort considers this as first-party.
 WORKDIR /opt/django-project
+
+# Run the pre-commit hooks (without --all-files, so that it won't actually run
+# on any files) to create a reusable package cache.
+RUN pre-commit run ruff-fix-only --hook-stage manual
+RUN pre-commit run ruff-format-check --hook-stage manual
+RUN pre-commit run codespell
```
mvandenburgh, line 60: I think you can just do

```suggestion
# Run the pre-commit hooks (without --all-files, so that it won't actually run
# on any files) to create a reusable package cache.
RUN pre-commit
```
here, right? Then, we don't need to remember to update the Dockerfile when adding new pre-commit hooks.

mvandenburgh (commented):
This works for me now. Just left one suggestion

dev/django.Dockerfile
waxlamp: Very cool!

mvandenburgh (approved):

PR #2062, opened 2024-10-30
Fix validation error when only Zarr assets are uploaded

### LINKED ISSUE #1814: Asset Summary validation fails to evaluate all "assets" size during validation of Version metadata
During Version validation, `version.assets` is queried to understand the byte size and file count for a given `Version` object associated with a Dandiset.

For the `version.assets`, there are three potential "assets" ForeignKey models that could be referenced: "blob", "embargoed_blob" and "zarr" -- see here for the [model definition
](https://github.com/dandi/dandi-archive/blob/1be249aaff28586e8cd0fadbe2c719a3fb83ef45/dandiapi/api/models/asset.py#L137C1-L145C6)

Downstream, during validation, only the `blob` asset FK is queried: [see here](https://github.com/dandi/dandi-archive/blob/1be249aaff28586e8cd0fadbe2c719a3fb83ef45/dandiapi/api/services/metadata/__init__.py#L120) -- `'numberOfBytes': 1 if version.assets.filter(blob__size__gt=0).exists() else 0`

This discrepancy was noticed during zarr-only uploads, where the validation error of `assetsSummary: A Dandiset containing no files or zero bytes is not publishable` was noticed in the UI after an initial upload.

@aaronkanzer to continue to provide more context and an eventual fix here -- just wanted to make this Issue to record what I've found thus far.

Cc @kabilar @waxlamp @yarikoptic @satra

aaronkanzer:
## Description of error

As documented in #1814, we receive the following validation error on the web application when uploading only Zarr asset(s) to a Dandiset (and not blobs):

```
assetsSummary: A Dandiset containing no files or zero bytes is not publishable
```

The error is raised from `dandi-schema` when a validation check occurs on the `assetSummary` field in the `dandiset.yaml`.  See the [models module](https://github.com/dandi/dandi-schema/blob/bdb7dd681435f1092814041eb01eb4fabe87bf57/dandischema/models.py#L1495-L1501):

```python
@validator("assetsSummary")
def check_filesbytes(cls, values: AssetsSummary) -> AssetsSummary:
    if values.numberOfBytes == 0 or values.numberOfFiles == 0:
        raise ValueError(
            "A Dandiset containing no files or zero bytes is not publishable"
        )
    return values
```

And it arises because in the query below only the `blob` asset is evaluated:

https://github.com/dandi/dandi-archive/blob/0a4bf7057fe176097b09c9ce9c6541dfa966fc17/dandiapi/api/services/metadata/__init__.py#L127

## Description of proposed fix

The proposed changes additionally evaluate the size of Zarr assets that are in the `COMPLETE` state to generate the `assetSummary:numberOfBytes` metadata field.

cc @kabilar

Closes #1814

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -7,6 +7,7 @@ import dandischema.exceptions
 from dandischema.metadata import aggregate_assets_summary, validate
 from django.conf import settings
 from django.db import transaction
+from django.db.models.query_utils import Q
 from django.utils import timezone
 
 from dandiapi.api.models import Asset, Version
@@ -124,7 +125,9 @@ def validate_version_metadata(*, version: Version) -> None:
         metadata_for_validation['doi'] = '10.80507/dandi.123456/0.123456.1234'
         metadata_for_validation['assetsSummary'] = {
             'schemaKey': 'AssetsSummary',
-            'numberOfBytes': 1 if version.assets.filter(blob__size__gt=0).exists() else 0,
+            'numberOfBytes': 1
+            if version.assets.filter(Q(blob__size__gt=0) | Q(zarr__size__gt=0)).exists()
+            else 0,
             'numberOfFiles': 1 if version.assets.exists() else 0,
         }
         return metadata_for_validation
--- a/dandiapi/api/tests/test_tasks.py
+++ b/dandiapi/api/tests/test_tasks.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime
 import hashlib
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from django.conf import settings
@@ -10,9 +11,13 @@ from django.utils import timezone
 from guardian.shortcuts import assign_perm
 import pytest
 from rest_framework.test import APIClient
+from zarr_checksum import compute_zarr_checksum
+from zarr_checksum.checksum import EMPTY_CHECKSUM
+from zarr_checksum.generators import ZarrArchiveFile
 
 from dandiapi.api import tasks
 from dandiapi.api.models import Asset, AssetBlob, Version
+from dandiapi.zarr.models import ZarrArchiveStatus
 
 from .fuzzy import URN_RE, UTC_ISO_TIMESTAMP_RE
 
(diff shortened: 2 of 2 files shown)


dandiapi/api/services/metadata/__init__.py
```diff
@@ -90,7 +92,9 @@ def version_aggregate_assets_summary(version: Version) -> None:
 
     assets_summary = aggregate_assets_summary(
         asset.full_metadata
-        for asset in version.assets.filter(status=Asset.Status.VALID)
+        for asset in version.assets.filter(
+            Q(status=Asset.Status.VALID) | Q(zarr__status=ZarrArchiveStatus.UPLOADED)
```
waxlamp, line 96: What was the problem that led to this particular change? We need @jjnesbitt to confirm, but I don't think we should be counting Zarr archives in the `UPLOADED` state (those have not yet been "finalized" in the terminology of the Zarr API; once they are, I believe they move into the `COMPLETE` state).

waxlamp (commented):
I think we need a bit more explanation of the actual bug, and why this fixes it. We need to be careful with Zarr statuses because those have a complex relationship with Assets that results in (perhaps overly) complex semantics.

aaronkanzer:
> I think we need a bit more explanation of the actual bug, and why this fixes it. We need to be careful with Zarr statuses because those have a complex relationship with Assets that results in (perhaps overly) complex semantics.

In the case where there is no `blob` but just `zarr` FKs associated with the `Asset`, validation fails since the query of:

```
'numberOfBytes': 1 if version.assets.filter(blob__size__gt=0).exists() else 0,
```

only evaluates the `blob` Foreign Key. I stumbled upon this bug when uploading a dandiset of pure `zarr`

dandiapi/api/services/metadata/__init__.py
aaronkanzer: Good catch -- updated to `COMPLETE` here and in the unit test.

The expansion of the `filter` query here is to match similar logic in the `validate_pending_asset_metadata` Celery task -- [specific code reference here](https://github.com/dandi/dandi-archive/blob/0d4f6038b1a208617716eaf68859ea6b157a4eb3/dandiapi/api/tasks/scheduled.py#L72-L74)

kabilar:
Hi @aaronkanzer, thank you for the original fix and pushing these changes upstream to DANDI.

Hi @waxlamp @jjnesbitt, I made some minor changes to the original comment above to further describe the issue and fix.  Please take a look at it when you have a chance.  Thank you.

dandiapi/api/tests/test_version.py
```diff
@@ -356,16 +358,39 @@ def test_version_publish_version(draft_version, asset):
...
 
-    assert version.metadata['assetsSummary']['numberOfBytes'] == asset.blob.size
+    assert version.assets.count() == 1
+    asset = version.assets.first()
+
+    if asset_type == 'blob':
+        assert version.metadata['assetsSummary']['numberOfBytes'] == asset.blob.size
+
+    elif asset_type == 'zarr':
+        assert version.metadata['assetsSummary']['numberOfBytes'] == asset.size
```
jjnesbitt, line 392: I believe `asset.size` can be used for both here, since it abstracts over blob vs zarr.
```suggestion
    assert version.metadata['assetsSummary']['numberOfBytes'] == asset.size
```

aaronkanzer: Good point @jjnesbitt -- updated!

jjnesbitt:
The original implementation of this PR was modifying both the `version_aggregate_assets_summary` and `validate_version_metadata` functions, testing only `version_aggregate_assets_summary`. However, only `validate_version_metadata` required any change to address issue this PR was created for, and the changes to `version_aggregate_assets_summary` would have no impact. I've reworked this PR to reflect that.

@aaronkanzer I believe the `test_validate_version_metadata_only_zarr_assets` test reflects the situation in the PR description. If that's not the case, please elaborate.

mvandenburgh (approved):
LGTM

aaronkanzer:
> The original implementation of this PR was modifying both the `version_aggregate_assets_summary` and `validate_version_metadata` functions, testing only `version_aggregate_assets_summary`. However, only `validate_version_metadata` required any change to address issue this PR was created for, and the changes to `version_aggregate_assets_summary` would have no impact. I've reworked this PR to reflect that.
> 
> @aaronkanzer I believe the `test_validate_version_metadata_only_zarr_assets` test reflects the situation in the PR description. If that's not the case, please elaborate.

@jjnesbitt Just reviewed the tests you added -- thanks so much for this -- yes, this makes sense, and looks good to me

PR #2069, opened 2024-11-04
Add API support for Embargoed Zarrs

jjnesbitt:
This PR adds API support for embargoed zarrs, with the following notable changes:

1. `ZarrArchive`, `EmbargoedZarrArchive`, and `BaseZarrArchive` have all been merged into one `ZarrArchive` model.
2. The `Asset` model now has an `is_embargoed` property, which will account for both asset blob and zarr embargo status.
3. The `unembargo_dandiset` function has been extended to include zarrs.
4. The zarr list and retrieve endpoints have been modified to account for embargoed zarrs. If the requested zarr is embargoed, and the user is either unauthenticated or not an owner on the respective dandiset, a `404 Not Found` is returned.

This does not address any changes in the GUI, as I'm unsure what exactly that entails. These changes are necessary for embargoed zarr support in any case, and so can be merged before any GUI re-work.


dandiapi/api/models/asset.py
```diff
@@ -162,6 +162,13 @@ def is_blob(self):
     def is_zarr(self):
         return self.zarr is not None
 
+    @property
+    def is_embargoed(self):
```
mvandenburgh, line 166: Pylance shows the return type here as `(bool | Any)` for me, I assume due to it not being able to infer the type of `self.zarr.embargoed`.
```suggestion
    def is_embargoed(self) -> bool:
```

dandiapi/api/services/asset/__init__.py
```diff
@@ -168,15 +168,16 @@ def add_asset_to_version(
...
     with transaction.atomic():
-        # Creating an asset in an OPEN dandiset that points to an embargoed blob results in that
-        # blob being unembargoed
+        # Creating an asset in an OPEN dandiset that points to an
+        # embargoed blob results in that blob being unembargoed.
+        # NOTE: This only applies to asset blobs, as zarrs cannot belong to
+        # multiple dandisets at once.
         if (
-            asset_blob is not None
+            asset_blob
```
mvandenburgh, line 176: What is the reason for removing the `is not None` condition here?

dandiapi/zarr/migrations/0004_zarrarchive_embargoed_delete_embargoedzarrarchive.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='zarrarchive',
+            name='embargoed',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.DeleteModel(
+            name='EmbargoedZarrArchive',
+        ),
+    ]
```
mvandenburgh, line 21: Don't we need to migrate the existing `EmbargoedZarrArchive` objects to the `ZarrArchive` table before deleting the `EmbargoedZarrArchive` table?

dandiapi/api/services/audit/__init__.py
```diff
@@ -106,6 +106,7 @@ def create_zarr(*, dandiset: Dandiset, user: User, zarr_archive: ZarrArchive) ->
     details = {
         'zarr_id': str(zarr_archive.zarr_id),
         'name': zarr_archive.name,
+        'embargoed': dandiset.embargoed,
```
jjnesbitt, line 109: @waxlamp does this make sense? I figured it's good to store this since I don't know if it's readily available on the zarr otherwise.

dandiapi/api/services/asset/__init__.py
jjnesbitt: Good point, i'll revert that.

dandiapi/zarr/migrations/0004_zarrarchive_embargoed_delete_embargoedzarrarchive.py
jjnesbitt: There are no existing `EmbargoedZarrArchive`s, and aren't supported in any real way even if they did exist. So I think this achieves what we want.

dandiapi/api/services/audit/__init__.py
jjnesbitt: Alright, I've removed that commit.

mvandenburgh (approved):

PR #2076, opened 2024-11-20
ENH: run pytest  against dandi-cli with -v to see which tests ran

yarikoptic:
and which FAILed or PASSed.  Just now realized that we do not run in verbose mode so cannot `grep` for specific test in the logs collected by tinuous for the past


yarikoptic:
This one is trivial.  If there would be no objections, I would merge in few days.

mvandenburgh (approved):

PR #2077, opened 2024-11-20
Use a dedicated logger. not top level logging. module

yarikoptic:
Most of the other submodules already used a dedicated logger. I have just identifier where root  logging.  was still used and similarly established loggers based on the files __name__ which should include the path starting with dandiapi. happen someone needs tighter control over logging of the application


mvandenburgh (approved):

yarikoptic:
Thank you @mvandenburgh for the rapid review/approval -- let's proceed then!

PR #2078, opened 2024-11-20
Make logging level of the django app configurable

yarikoptic:
ATM we do not have a single DEBUG level log statement, but I hope that logging would be enriched. Adding such control would also allow to make log less talkative if so desired by setting it to WARNING level. But also would simplify addition and use of new log statements.  E.g. I would right away add it into dandi-cli docker-compose setup so to help in the future troubleshooting issues like

- https://github.com/dandi/dandi-cli/issues/1488


dandiapi/api/apps.py
```diff
@@ -50,13 +50,15 @@ def ready(self):
         import dandiapi.api.checks  # noqa: F401, RUF100
         import dandiapi.api.signals  # noqa: F401
 
+        logging.getLogger(__name__).setLevel(settings.DANDI_LOG_LEVEL)
```
waxlamp, line 53: This was originally `logging.getLogger(__name__.split('.')[0])`, but looking through the codebase and at the logging docs, it seems `logging.getLogger(__name__)` is a standard practice. Was there a reason you had it the other way? (If necessary, we can revert 6addfacce0d6fac0bf11c0da77e038be412ce344, which makes the change.)

waxlamp (commented):
I added a few commits to simplify the code. If you're ok with those, let me know and I will merge.

dandiapi/api/apps.py
```diff
@@ -50,13 +50,15 @@ def ready(self):
...
+
         if hasattr(settings, 'SENTRY_DSN'):
             sentry_sdk.init(
                 dsn=settings.SENTRY_DSN,
                 environment=settings.SENTRY_ENVIRONMENT,
                 release=settings.SENTRY_RELEASE,
                 integrations=[
-                    LoggingIntegration(level=logging.INFO, event_level=logging.WARNING),
+                    LoggingIntegration(level=settings.DANDI_LOG_LEVEL, event_level=logging.WARNING),
```
waxlamp: This should actually remain unchanged: we need for Sentry to receive `INFO` level events, as it uses those to create the breadcrumbs that make up its frame-level reporting for error events. See 74697eaf862d122153a1f29357c400cfb58ade1f.

dandiapi/api/apps.py
waxlamp: Actually, this does not have the desired effect (of changing the logging level for all loggers). Instead, we need to omit the `__name__` argument so as to affect the root logger. See b07bfe126db504ab7c879b37eefbad647c2baac3.

waxlamp (commented):
A couple more changes to eliminate some bugs.

dandiapi/api/apps.py
yarikoptic: roger that, makes sense, thanks!

dandiapi/api/apps.py
```diff
@@ -50,6 +50,8 @@ def ready(self):
         import dandiapi.api.checks  # noqa: F401, RUF100
         import dandiapi.api.signals  # noqa: F401
 
+        logging.getLogger().setLevel(settings.DANDI_LOG_LEVEL)
```
yarikoptic, line 53: Thanks @waxlamp for catching the initial limitation! but I do not think we should go this wild, what about affecting only `dandi` ? i.e. smth like

```suggestion
        logging.getLogger(__name__.split('.', 1)[0]).setLevel(settings.DANDI_LOG_LEVEL)
```
?  otherwise, if to debug our code we turn to DEBUG level globally -- could lead to excessive amount of noise from all kinds of libraries we use.  WDYT?

waxlamp: Ah I see what you were going for now. I've added this back (along with explanatory comment) in dc3549d.

waxlamp (approved):

PR #2082, opened 2024-11-21
Add incomplete upload dialog to DLP when unembargo is blocked

jjnesbitt:
Recently we've had several instances of a dandiset unembargo being blocked by incomplete/active uploads. Most of the time, these uploads are junk and the owners simply wish to clear them, which requires admin intervention. This PR adds the API and Vue client support for users to list and delete these lingering uploads themselves.

Here is a video of the new UI.

https://github.com/user-attachments/assets/6033cd6b-9b10-4c8d-96e1-f85ea3df2598

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -43,6 +43,7 @@ from dandiapi.api.views.serializers import (
     DandisetSearchQueryParameterSerializer,
     DandisetSearchResultListSerializer,
     DandisetUploadSerializer,
+    PaginationQuerySerializer,
     UserSerializer,
     VersionMetadataSerializer,
 )
@@ -467,6 +468,7 @@ class DandisetViewSet(ReadOnlyModelViewSet):
     @swagger_auto_schema(
         methods=['GET'],
         manual_parameters=[DANDISET_PK_PARAM],
+        query_serializer=PaginationQuerySerializer,
         request_body=no_body,
         operation_summary='List active/incomplete uploads in this dandiset.',
     )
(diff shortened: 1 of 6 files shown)


dandiapi/api/views/dandiset.py
```diff
@@ -453,3 +463,33 @@ def users(self, request, dandiset__pk):  # noqa: C901
...
+    @action(methods=['GET'], detail=True)
+    def uploads(self, request, dandiset__pk):
+        dandiset: Dandiset = self.get_object()
+
+        # Special case where a "safe" method is access restricted, due to the nature of uploads
+        self.require_owner_perm(dandiset)
+
+        uploads: QuerySet[Upload] = dandiset.uploads.all()
+        serializer = DandisetUploadSerializer(instance=uploads, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
mvandenburgh, line 482: I think we should paginate these. Someone could theoretically DoS the server by creating thousands of uploads (which may be a completely legitimate action) and then hitting this endpoint.

mvandenburgh (changes requested):
LGTM except one note about pagination

dandiapi/api/views/dandiset.py
jjnesbitt: Just pushed that change. It's a bit awkward since we can't use the default pagination behavior, due to this being a "detailed" endpoint that still needs pagination. That was originally the reason I didn't bother, since it seemed unlikely to be an issue anyway, but better safe than sorry. I've also changed the client to match that, and tested it locally, confirming the overall behavior is unchanged.

mvandenburgh (approved):

PR #2084, opened 2024-11-25
Include robots.txt in UI and API for handling of web crawlers

aaronkanzer:
@yarikoptic -- per https://dandiarchive.slack.com/archives/GMRLT5RQ8/p1732553458007359 -- includes `robots.txt` in both UI and backend, per @jwodder suggestion

Cc @jjnesbitt @mvandenburgh

### DIFF
--- a/dandiapi/api/views/robots.py
+++ b/dandiapi/api/views/robots.py
@@ -1,0 +1,10 @@
+from django.views.decorators.cache import cache_page
+from django.http import HttpResponse
+
+@cache_page(60 * 60 * 24)  # 24 hour cache
+def robots_txt_view(request):
+    content = """
+    User-agent: *
+    Disallow: /
+    """
+    return HttpResponse(content, content_type="text/plain")
--- a/dandiapi/urls.py
+++ b/dandiapi/urls.py
@@ -19,6 +19,7 @@ from dandiapi.api.views import (
     blob_read_view,
     info_view,
     mailchimp_csv_view,
+    robots_txt_view,
     root_content_view,
     stats_view,
     upload_complete_view,
@@ -80,8 +81,10 @@ class DandisetIDConverter:
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('', root_content_view),
+    path("robots.txt", robots_txt_view),
     path('api/', include(router.urls)),
     path('api/auth/token/', auth_token_view, name='auth-token'),
+    path('api/stats/', stats_view),
     path('api/stats/', stats_view),
     path('api/info/', info_view),
     path('api/blobs/digest/', blob_read_view, name='blob-read'),
--- a/.gitignore
+++ b/.gitignore
@@ -125,5 +125,5 @@ dmypy.json
 
 # Editor settings
 .vscode
-
+.idea/
 .DS_Store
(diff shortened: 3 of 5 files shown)


aaronkanzer:
Adding unit tests currently for backend -- will move out of draft after that is done

aaronkanzer:
Unite tests added -- moving out of draft

dandiapi/urls.py
```diff
@@ -80,9 +81,11 @@ def to_url(self, value):
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('', root_content_view),
+    path('robots.txt', robots_txt_view),
     path('api/', include(router.urls)),
     path('api/auth/token/', auth_token_view, name='auth-token'),
     path('api/stats/', stats_view),
+    path('api/stats/', stats_view),
```
jjnesbitt, line 88:
```suggestion
```

aaronkanzer: whoops, forgot to remove -- should be removed now

dandiapi/api/views/robots.py
```diff
@@ -0,0 +1,10 @@
+from django.views.decorators.cache import cache_page
+from django.http import HttpResponse
+
+@cache_page(60 * 60 * 24)  # 24 hour cache
+def robots_txt_view(request):
+    content = """
+    User-agent: *
+    Disallow: /
+    """
+    return HttpResponse(content, content_type="text/plain")
```
jjnesbitt, line 10: This suggestion includes a couple things.

First, the removal of the `cache_page` decorator. Since this is returning a simple string, and making no database calls, introducing caching here has no payoff. In fact, it's probably _worse_ to use caching here, as we use the database as a cache backend, so it's actually making a database call where it usually wouldn't.

Second, since triple quote strings preserve whitespace, the webpage content was indented by 4 spaces. The suggestion removes that, putting it in line with a normal `robots.txt`. I also included a matching suggestion for the test.
```suggestion
from __future__ import annotations

from django.http import HttpResponse


def robots_txt_view(request):
    content = 'User-agent: *\nDisallow: /'
    return HttpResponse(content, content_type='text/plain')
```

dandiapi/api/tests/test_robots.py
```diff
@@ -0,0 +1,13 @@
...
+
+@pytest.mark.django_db
+def test_robots_txt(api_client):
+    response = api_client.get(reverse('robots_txt'))
+    assert response.status_code == 200
+    assert response['Content-Type'] == 'text/plain'
+    expected_content = """
+    User-agent: *
+    Disallow: /
+    """
```
jjnesbitt, line 12:
```suggestion
    expected_content = 'User-agent: *\nDisallow: /'
```

dandiapi/urls.py
```diff
@@ -80,6 +81,7 @@ def to_url(self, value):
 register_converter(DandisetIDConverter, 'dandiset_id')
 urlpatterns = [
     path('', root_content_view),
+    path('robots.txt', robots_txt_view),
```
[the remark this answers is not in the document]
aaronkanzer: Updated!

jjnesbitt (changes requested):
Once these suggestions are applied along with formatting to fix the linting errors, I think this looks good.

jjnesbitt (approved):

PR #2087, opened 2024-12-02
`Upload` and `AssetBlob` garbage collection implementation

mvandenburgh:
Implements the design outlined in #2068.


dandiapi/api/models/garbage_collection.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.db import models
+
+
+class GarbageCollectionEvent(models.Model):
+    created = models.DateTimeField(auto_now_add=True)
+    type = models.CharField(
+        max_length=255, help_text='The model name of the records that were garbage collected.'
+    )
+    records = models.JSONField(
```
jjnesbitt, line 11: Is there a particular reason you chose to store a collection of records in one "Event" object, rather than one row per object? I.e. `records` would become `record`, and `garbage_collection_event_id` could be a foreign key to a separate table that stores the events.

One thing that it seems would be easier in that model is searching for a particular garbage collected upload or asset blob.

dandiapi/api/services/garbage_collection/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+)  # TODO: pick this up from env var set by Terraform to ensure consistency?
+
+
+def _garbage_collect_uploads() -> int:
+    qs = Upload.objects.filter(
+        created__lt=timezone.now() - UPLOAD_EXPIRATION_TIME,
+    )
+
+    # Chunk the queryset to avoid creating a single
+    # GarbageCollectionEvent with a huge JSONField
```
jjnesbitt, line 37: This may also be something that's easier with an `Event <- Record` model, as I mentioned in my other comment. You could have many records pointing to one event, which would allow for just one event record to be stored per "actual" event. In the current case, there could be 10 `GarbageCollectionEvent` rows for what's really one "event', i.e. one invocation of `_garbage_collect_uploads` or `_garbage_collect_asset_blobs`.

jjnesbitt: This would also allow you to point both asset blobs and uploads back to the same single event record, rather than two separate records.

jjnesbitt (changes requested):

dandiapi/api/models/garbage_collection.py
jjnesbitt: (I realize this is how it's outlined in the design doc, but this question is just coming to me now.)

dandiapi/api/models/garbage_collection.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+from django.db import models
+
+
+class GarbageCollectionEvent(models.Model):
+    created = models.DateTimeField(auto_now_add=True)
```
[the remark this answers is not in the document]
mvandenburgh: Yeah I agree `timestamp` is clearer.

dandiapi/api/services/garbage_collection/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                )
+                for u in uploads_chunk
+            )
+
+            # Delete the blobs from S3
+            futures.append(
+                executor.submit(
+                    lambda chunk: [u.blob.delete(save=False) for u in chunk],
+                    uploads_chunk,
+                )
```
jjnesbitt, line 61: Should we use `transaction.on_commit` here, in  case an error occurs and this transaction gets rolled back? I'm not sure if that plays nicely with waiting on those futures, you might have to define those two context managers separately.

dandiapi/api/services/garbage_collection/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+            deleted_records += AssetBlob.objects.filter(
+                pk__in=[a.pk for a in asset_blobs_chunk],
+            ).delete()[0]
+
+        wait(futures)
+
+    return deleted_records
```
jjnesbitt, line 108: This code block is essentially the same as the one in `_garbage_collect_uploads`, with a few different variables. Should we parameterize this within a function?

jjnesbitt (changes requested):

dandiapi/api/services/garbage_collection/__init__.py
mvandenburgh: My thinking was that we would want the DB delete to get rolled back if an S3 delete fails. If that were to happen, we could potentially have asset blobs in the DB that have no equivalent S3 blob, which isn't a big deal and the DB asset blobs would just get cleaned up on the next garbage collection run. 
But if we delete the DB asset blobs, commit the transaction, and _then_ delete the S3 blobs, any S3 blobs that fail to delete would be "dangling" blobs with no DB record, and wouldn't get cleaned up.

dandiapi/api/services/garbage_collection/__init__.py
mvandenburgh: I chose not to because the future `_garbage_collect_assets` function will also be similar, but won't be as easy to parameterize. Also, while we could parameterize just the upload and asset blob functions, I think structurally it might be misleading. Asset blobs and uploads are two different concepts in the system, and while they both happen to have a `blob` field that points to an S3 file, that's more of a coincidence in naming than an intentional design decision.

jjnesbitt: For others, we discussed offline, and agreed to leave these code blocks separate for now.

jjnesbitt (approved):

mvandenburgh:
Because this is such a sensitive operation, we'll need to careful coordinate merging/deploying of this PR. I'll mark it as draft for now so it doesn't get merged by accident.

mvandenburgh:
We discussed this offline, and decided to temporarily disable the garbage collection cron and merge this. Once we've run this manually a few times, we'll enable the automatic cron again.

PR #2102, opened 2024-12-09
DLP: Have "Cite as" only for OPEN (non embargoed) dandisets

### LINKED ISSUE #2094: Embargoed dandisets must not show `Cite As` in the sidebar
Because they cannot and should not be citeable.

In part addresses
- https://github.com/dandi/dandi-archive/issues/1932

yarikoptic:
Closes #2094

A trivial PR. Seems to work locally I see

image

and then for OPEN

image

NB error is there even on master, I didn't dig in.


jjnesbitt (approved):
LGTM, deploy preview seems to behave as expected.

Issue #865, opened 2022-01-31
Make NIH awards clickable links


yarikoptic:
From my toying around with https://api.reporter.nih.gov/#/Search/post_v2_projects_search on an example of our ReproNim project which is 1P41EB019936-01A1 . 

```
$> curl --silent -X POST "https://api.reporter.nih.gov/v2/projects/search" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"criteria\":{\"project_nums\":[\"1P41EB019936-01A1\"]},\"include_fields\":[\"ApplId\",\"SubprojectId\",\"FiscalYear\",\"Organization\",\"ProjectNum\",\"ProjectNumSplit\",\"ContactPiName\",\"AllText\",\"FullStudySection\",\"ProjectStartDate\",\"ProjectEndDate\"],\"offset\":0,\"limit\":25,\"sort_field\":\"project_start_date\",\"sort_order\":\"desc\"}" >| /tmp/out.json

$> jq '.results[].appl_id' < /tmp/out.json
8999833
8999834
8999841
8999840
8999838
8999835
```
didn't check but I believe they all go for the same award but different years, we could may be choose the latest and then URL is e.g. https://reporter.nih.gov/project-details/8999835 .

API seems to respond quite fast so in principle could be built in into grant number form entry to list matching grant records (PI, title) for the user to choose from

yarikoptic:
I was thinking of making it "dynamic", i.e. it could be some on-click action which would perform that query (with caching) and redirect to that URL.  This way we could adjust action whenever target URL was to change etc, without needing to go and change the metadata. Persistent information here is the award number, not URL really.

satra:
i think we could also help the process of adding the info by searching through awards via reporter in the meditor.

waxlamp:
> i think we could also help the process of adding the info by searching through awards via reporter in the meditor.

This is an interesting idea. Essentially, a little embedded search engine for award numbers in the meditor.

waxlamp:
> I was thinking of making it "dynamic", i.e. it could be some on-click action which would perform that query (with caching) and redirect to that URL. This way we could adjust action whenever target URL was to change etc, without needing to go and change the metadata. Persistent information here is the award number, not URL really.

I don't think this is a good idea. If we have the award number then presumably we can construct the URL statically from it--no need for a live query on click. On the other hand, we could also simply link to the search results page for that award number (if indeed it's not possible to programmatically construct the right URL).

PR #2103, opened 2024-12-09
DLP: Add "Protocols" card

### LINKED ISSUE #2071: "protocol" is not displayed on DLP
E.g. https://dandiarchive.org/dandiset/001084 does not visualize anywhere on DLP but present in  metadata

image

NB That URL is not clickable, and not even selectable -- unclear how user is supposed to visit that page?!

yarikoptic:
Closes #2071

There seems to be a little design gotcha -- a space from prior section which I do not know framework well enough to see where to add anything to make it not indent itself.

image

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -1,5 +1,7 @@
 <template>
   <div>
+    <v-row>
+      <v-col>
     <v-card
       v-if="contributors && contributors.length"
       outlined
@@ -42,7 +44,8 @@
         </v-chip>
       </div>
     </v-card>
-
+      </v-col>
+    </v-row>
     <MetadataCard
       :items="fundingInformation"
       name="Funding information"
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -78,18 +78,13 @@
     >
       <template #content="slotProps">
         <div class="text-caption grey--text text--darken-1">
-          <span class="pl-2">
-            - {{ slotProps.item }}
-               <v-btn
-                      v-if="isURL(slotProps.item)"
-                      icon
-                      :href="slotProps.item"
-                      target="_blank"
-                      rel="noopener"
-               >
-                   <v-icon>mdi-link</v-icon>
-               </v-btn>
-          </span>
+            <a
+                :href="slotProps.item"
+                target="_blank"
+                rel="noopener"
+            >
+                {{ slotProps.item }}
+            </a>
         </div>
       </template>
     </MetadataCard>
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -135,6 +135,8 @@
       </template>
     </MetadataCard>
 
+    <v-row>
+      <v-col>
     <v-card
       v-if="assetSummary"
       outlined
@@ -230,6 +232,8 @@
         </div>
       </v-list>
     </v-card>
+      </v-col>
+    </v-row>
 
     <MetadataCard
       v-if="associatedProjects && associatedProjects.length"
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -137,101 +137,101 @@
 
     <v-row>
       <v-col>
-    <v-card
-      v-if="assetSummary"
-      outlined
-    >
-      <v-card-title class="font-weight-regular">
-        <v-progress-circular
-          v-if="assetSummaryBeingComputed"
-          class="mr-3"
-          size="20"
-          indeterminate
-        />
-        <v-icon
-          v-else
-          class="mr-3 grey--text text--lighten-1"
+        <v-card
+          v-if="assetSummary"
+          outlined
         >
-          mdi-clipboard-list
-        </v-icon>
-        Assets Summary
-      </v-card-title>
-      <v-list
-        :style="`column-count: ${assetSummaryColumnCount};`"
-        class="px-3 ml-2"
-      >
-        <div
-          v-if="assetSummaryBeingComputed"
-          class="text-subtitle-2"
-        >
-          The assets summary is being computed, please wait.
-        </div>
-        <div
-          v-else-if="!assetSummary || !Object.keys(assetSummary).length"
-          class="font-italic font-weight-bold"
-        >
-          This Dandiset does not contain any assets.
-        </div>
-        <div
-          v-for="([type, items], i) in Object.entries(assetSummary)"
-          v-else
-          :key="i"
-        >
-          <div
-            v-if="items && items.length"
-            class="d-inline-block"
-            style="width: 100%;"
+          <v-card-title class="font-weight-regular">
+            <v-progress-circular
+              v-if="assetSummaryBeingComputed"
+              class="mr-3"
+              size="20"
+              indeterminate
+            />
+            <v-icon
+              v-else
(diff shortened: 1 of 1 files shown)


web/src/components/DLP/OverviewTab.vue
```diff
@@ -67,6 +67,30 @@
...
+                      v-if="isURL(slotProps.item)"
+                      icon
+                      :href="slotProps.item"
+                      target="_blank"
+                      rel="noopener"
+               >
+                   <v-icon>mdi-link</v-icon>
+               </v-btn>
+          </span>
+        </div>
```
waxlamp, line 90: It seems like we can put together better content for this section.

The only restriction on the protocol metadata field is that the elements be URIs, so it seems the only thing we can do with these is turn them into links. But displaying the link text with a markdown-style list identifier within a "quotation" style block, then with a link element to its side seems strange.

```suggestion
        <div class="text-caption grey--text text--darken-1">
            <a
                :href="slotProps.item"
                target="_blank"
                rel="noopener"
            >
                {{ slotProps.item }}
            </a>
        </div>
```
This looks better IMO, unless there is more, specific information about the protocols we can take advantage of.

web/src/components/DLP/OverviewTab.vue
```diff
@@ -67,6 +67,30 @@
       </template>
     </MetadataCard>
 
+    <MetadataCard
+      :items="protocols"
+      v-if="protocols && protocols.length"
+      name="Protocols"
+      icon="mdi-protocol"
```
waxlamp, line 74:
```suggestion
      icon="mdi-rodent"
```

waxlamp (commented):
My suggested changes are visible at the [deploy preview](https://deploy-preview-2109--gui-staging-dandiarchive-org.netlify.app/dandiset/200026?search=200026&pos=1) for a dummy PR I opened against this branch.

As for the padding issue around the protocols card, I looked into it briefly but could not come up with anything immediately. We will need @mvandenburgh or @jjnesbitt's expertise.

web/src/components/DLP/OverviewTab.vue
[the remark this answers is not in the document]
waxlamp:
```suggestion
      icon="mdi-rodent"
```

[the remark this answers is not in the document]
yarikoptic: indeed `://` is quite cryptic... but I am not sure if a mouse (i.e.)
would be not  misleading -- since it might be not even about mice research and/or could be mis-treated as "species" or some other aspect.  May be [file-document-check](https://pictogrammers.com/library/mdi/icon/file-document-check/) as "protocol" being a checklist to follow?

image

```suggestion
      icon="mdi-file-document-check"
```

?

web/src/components/DLP/OverviewTab.vue
yarikoptic: In general -- I do not mind and like hyperlinks to be "more explicit", but they stand out too much IMHO:

- we do not have hyperlinks (as blue underlined text) anywhere else on the DLP (besides footer)
- even where it looks like a URL we have it with that extra link icon

image

so, I was just trying to stay consistent with what we have for other fields.  I wonder if someone with good UX eye could have a look here to decide between the two?

web/src/components/DLP/OverviewTab.vue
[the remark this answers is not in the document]
waxlamp: I like it.

web/src/components/DLP/OverviewTab.vue
waxlamp: FWIW, I don't think we should even be listing the DOI URL in that example since there is more pertinent information to display already (the title), and the URL information is available in a very standard way by hovering over the link icon.

> I wonder if someone with good UX eye could have a look here to decide between the two?

I had the same thought. I'll report back here after I talk to someone.

mvandenburgh:
> As for the padding issue around the protocols card, I looked into it briefly but could not come up with anything immediately. We will need @mvandenburgh or @jjnesbitt's expertise.

It seems that the issue isn't that the protocols card has this padding, but that the Contributors card does not. i.e. every card seems to have this padding except the Contributors one. See #2112 for the fix.

mvandenburgh:
Ah, the "Assets Summary" card has the same issue as "Contributors". I'll update my PR.

web/src/components/DLP/OverviewTab.vue
waxlamp: > I had the same thought. I'll report back here after I talk to someone.

I spoke with Jared, and he said that if we're constrained to just having URLs, showing them as hyperlinks would be the preference (though he also dislikes the edge case of a potentially long URL standing out too much as a glowing blue link).

He also had an interesting thought: that both options here are bad, and much better would be if the Protocol metadata entry had not only a URL, but also a title property. Then, we could use the same approach as other entries of this type. If that were the case, we could even think about a strategy where the user could enter the protocol URL, and we could try to retrieve metadata to autofill the title field, etc. (but of course that is something to consider in the future).

Since we are stuck with two suboptimal choices, I think we should just pick one for now while considering this sort of upgrade in the future. I am ok with either choice in the end (though I still lean towards hyperlinks). Let me know what you think, @yarikoptic.

yarikoptic: ### Re "title" property

FTR, ATM we have https://github.com/dandi/dandi-schema/blob/master/dandischema/models.py#L1534C5-L1540C6 `Protocols` only as a list of http URLs, so there is no metadata support for providing some title etc, as pretty much for any citeable thing.   Not yet sure if adding `title` alone would be "sufficient" since it is IMHO as any other "citation" and ideally also should include other pertinent information: e.g. author(s), and location (AKA publisher) of protocol.

In principle in case of DOI URL being used, we could at the level of UI here (somewhat relating to a similar desire #865) query/cache citation information for DOI and thus provide an enhanced record more in style with the other records. ... E.g. for a random one on protocols https://www.protocols.io/view/mri-pilot-81wgbe7ogpko/v2 -- it has [dx.doi.org/10.17504/protocols.io.w8hfht6](https://dx.doi.org/10.17504/protocols.io.w8hfht6) for which there is 

```bash
❯ doiref 10.17504/protocols.io.w8hfht6
 @article{Snoek_2019, title={MRI pilot v2}, url={http://dx.doi.org/10.17504/protocols.io.w8hfht6}, DOI={10.17504/protocols.io.w8hfht6}, publisher={ZappyLab, Inc.}, author={Snoek, Lukas and Beemsterboer, Tinka}, year={2019}, month=jan }
```
or even rendered one like

```bash
❯ curl --silent https://api.citeas.org/product/https://doi.org/10.17504/protocols.io.w8hfht6 | jq .citations[0]
{
  "citation": "Snoek, L., & Beemsterboer, T. (2019, January 17). MRI pilot v2. <i>[]</i>. ZappyLab, Inc. http://doi.org/10.17504/protocols.io.w8hfht6",
  "style_fullname": "American Psychological Association 6th edition",
  "style_shortname": "apa"
}
```

Not sure if worth filing an issue ATM - we have enough ;)

### Re matter at hands

I am ok with either one.  I can live with hyperlinks -- please accept your suggestions if you make a decision to go with them. (although I think we should adjust style for them but that is a separate issue/PR to do ;-) )

waxlamp: > (although I think we should adjust style for them but that is a separate issue/PR to do ;-) )

Which style do you think needs updating? Please do go ahead and file an issue if needed.

waxlamp (approved):

PR #2111, opened 2024-12-16
Set version status to PENDING in ingest_zarr_archive

### LINKED ISSUE #2091: Finalizing a zarr should put dandiset in `PENDING` state
Currently, the interaction between draft versions, assets, and asset summaries is as follows:

1. Draft version is in `Valid` or `Invalid` state.
2. Asset is added to or updated on dandiset, which sets the draft version status to `PENDING`.
3. The recurring [`validate_draft_version_metadata`](https://github.com/dandi/dandi-archive/blob/dc454f44b9e814b8a09781e9b798662e6cb8528f/dandiapi/api/tasks/scheduled.py#L87) task picks up this `PENDING` version, and runs both version validation and asset summary aggregation. 

This ensures that if an asset is created or modified, the version metadata appropriately follows suit. However, zarr assets have an additional layer of functionality, and behave slightly differently, which allows for a broken chain of events.

1. Draft version is in `Valid` or `Invalid` state.
2. Zarr asset is added to or updated on dandiset, which sets the draft version status to `PENDING`. Let's assume this zarr is not in a finalized state, and so its `size` shows as zero.
3. `validate_draft_version_metadata` is once again run on this version, but since the zarr is not finalized, is not picked up. Let's assume this zarr is the only asset in this dandiset, and so the draft version is marked as `INVALID`, with the message of: `assetsSummary: Value error, A Dandiset containing no files or zero bytes is not publishable`.
4. User notices this and finalizes the zarr with the `/zarr/{zarr_id}/finalize` endpoint. This kicks off the `ingest_zarr_archive` task, which completes and sets the appropriate size, checksum, etc. values on the zarr archive. However, this task **never places the draft version back in a `PENDING` state**. This results in the dandiset draft version remaining in an `INVALID` state until something else kicks it.

To address this, the [`ingest_zarr_archive`](https://github.com/dandi/dandi-archive/blob/dc454f44b9e814b8a09781e9b798662e6cb8528f/dandiapi/zarr/tasks/__init__.py#L17) task should finish by setting the draft verison back to `PENDING`, as is done in [`_add_asset_to_version`](https://github.com/dandi/dandi-archive/blob/dc454f44b9e814b8a09781e9b798662e6cb8528f/dandiapi/api/services/asset/__init__.py#L62-L65).

jjnesbitt:
Closes #2091

@mvandenburgh The last commit fixes some zarr tests that didn't have a draft version associated with their dandiset, due the chain of `SubFactory` definitions. I originally tried to fix this on the `DandisetFactory` class itself, so that all dandisets generated in this way would always have a draft version, but that proved troublesome. For now, I've settled for just fixing the zarr tests.

### DIFF
--- a/dandiapi/zarr/tasks/__init__.py
+++ b/dandiapi/zarr/tasks/__init__.py
@@ -3,10 +3,12 @@ from __future__ import annotations
 from celery import shared_task
 from celery.utils.log import get_task_logger
 from django.db import transaction
+from django.utils import timezone
 from zarr_checksum import compute_zarr_checksum
 from zarr_checksum.generators import S3ClientOptions, yield_files_s3
 
 from dandiapi.api.asset_paths import add_zarr_paths, delete_zarr_paths
+from dandiapi.api.models.version import Version
 from dandiapi.api.storage import get_storage_params
 from dandiapi.zarr.models import ZarrArchive, ZarrArchiveStatus
 
@@ -29,8 +31,10 @@ def ingest_zarr_archive(zarr_id: str, *, force: bool = False):
 
     # Zarr is in correct state, lock until ingestion finishes
     with transaction.atomic():
-        zarr = ZarrArchive.objects.select_for_update().get(
-            zarr_id=zarr_id, status=ZarrArchiveStatus.INGESTING
+        zarr = (
+            ZarrArchive.objects.select_related('dandiset')
+            .select_for_update()
+            .get(zarr_id=zarr_id, status=ZarrArchiveStatus.INGESTING)
         )
 
         # Remove all asset paths associated with this zarr before ingest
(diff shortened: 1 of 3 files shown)


dandiapi/zarr/tasks/__init__.py
```diff
@@ -62,6 +64,11 @@ def ingest_zarr_archive(zarr_id: str, *, force: bool = False):
         # Add asset paths after ingest is finished
         add_zarr_paths(zarr)
 
+        # Set version status back to PENDING, and update modified.
+        Version.objects.filter(id=zarr.dandiset.draft_version.id).update(
+            status=Version.Status.PENDING, modified=timezone.now()
+        )
```
mvandenburgh, line 70: To avoid an additional query -
```suggestion
        # Set version status back to PENDING, and update modified.
        Version.objects.filter(id=zarr.dandiset.draft_version_id).update(
            status=Version.Status.PENDING, modified=timezone.now()
        )
```

GitHub won't let me comment on code that hasn't been modified, but we should also have a `.select_related('dandiset')` on the query that gets this `zarr` object.

mvandenburgh (changes requested):
Just some query optimizations.

dandiapi/zarr/tasks/__init__.py
jjnesbitt: Since `.draft_version` is just a property that calls `self.versions.filter(version='draft').get()`, `draft_version_id` can't be used here. The `select_related` call will cut down on a query, but I think otherwise we're stuck with an extra query to grab the draft version.

mvandenburgh (approved):

PR #2112, opened 2024-12-17
Fix inconsistent padding on Contributors card in DLP

mvandenburgh:
The `MetadataCard` component wraps its contents in a `v-row` and `v-col`, which adds some additional padding around the card. The "Contributors" card, however, is the only DLP card that doesn't use the `MetadataCard` component. Prior to this change, the "Contributors" card didn't use `v-row` and `v-col`, so it didn't have the same padding as the other cards.

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -1,5 +1,7 @@
 <template>
   <div>
+    <v-row>
+      <v-col>
     <v-card
       v-if="contributors && contributors.length"
       outlined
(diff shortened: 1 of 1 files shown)


mvandenburgh:
See https://github.com/dandi/dandi-archive/pull/2112/commits/705704729a40383b569d6b6098cffd5d8305e5c9 for the actual change this PR makes. The other commit indents the html template to fix the formatting, leading to a large diff with no actual changes in code.

yarikoptic:
note:we do not have netlify previews here likely because merge is into another PR branch for https://github.com/dandi/dandi-archive/pull/2103  . I will "rebase and merge" .

PR #2115, opened 2024-12-18
Create permissions service layer

jjnesbitt:
The goal of this PR is to abstract most of the use of `django-guardian` into the permissions service layer, instead exposing functions that achieve the same effect. This has the following benefits:

1. Consolidates permission logic in the application, most notably used in endpoints.
2. Allows for easier modification or extension of our permissions logic.

One notable use case for this is forks of the codebase that may need to modify how permissions are done (e.g. LINC). This refactor and consolidation makes that much easier, as everything can be easily found in one file. cc @aaronkanzer.

### DIFF
--- a/dandiapi/api/models/dandiset.py
+++ b/dandiapi/api/models/dandiset.py
@@ -3,40 +3,11 @@ from __future__ import annotations
 from django.db import models
 from django_extensions.db.models import TimeStampedModel
 from guardian.models import GroupObjectPermissionBase, UserObjectPermissionBase
-from guardian.shortcuts import assign_perm, get_objects_for_user, get_users_with_perms, remove_perm
-
-
-class DandisetManager(models.Manager):
-    def visible_to(self, user) -> models.QuerySet[Dandiset]:
-        """
-        Return a queryset containing all dandisets visible to the given user.
-
-        This is basically all dandisets except embargoed dandisets not owned by the given user.
-        """
-        # We would like to do something like:
-        # Dandiset.filter(embargo_status=OPEN | permission__owned)
-        # but this is not possible with django-guardian; the `get_objects_for_user` shortcut must
-        # be used instead.
-        # We would like to do something like:
-        # queryset = Dandiset.objects.filter(embargo_status=OPEN).union(get_objects_for_user(...))
-        # but you cannot filter or perform many other common queryset operations after using
-        # union().
-        # Therefore, we resort to fetching the list of all primary keys for dandisets owned by the
-        # current user with `get_objects_for_user(...)`, then filter for either those dandisets or
-        # OPEN dandisets. There aren't very many dandisets and most users won't own very many of
-        # them, so there shouldn't be too many dandisets in the list.
-        owned_dandiset_pks = get_objects_for_user(user, 'owner', Dandiset).values('pk').all()
-        return self.filter(
-            models.Q(embargo_status=Dandiset.EmbargoStatus.OPEN)
-            | models.Q(pk__in=owned_dandiset_pks)
-        ).order_by('created')
 
 
 class Dandiset(TimeStampedModel):
     # Don't add beginning and end markers, so this can be embedded in larger regexes
     IDENTIFIER_REGEX = r'\d{6}'
-
-    objects: DandisetManager = DandisetManager()
 
     class EmbargoStatus(models.TextChoices):
         EMBARGOED = 'EMBARGOED', 'Embargoed'
@@ -76,28 +47,9 @@ class Dandiset(TimeStampedModel):
 
     @property
     def owners(self):
-        return get_users_with_perms(self, only_with_perms_in=['owner']).order_by('date_joined')
+        from dandiapi.api.services.permissions.dandiset import get_dandiset_owners
 
-    def set_owners(self, new_owners):
-        old_owners = get_users_with_perms(self, only_with_perms_in=['owner'])
-
-        removed_owners = []
-        added_owners = []
-
-        # Remove old owners
-        for old_owner in old_owners:
-            if old_owner not in new_owners:
-                remove_perm('owner', old_owner, self)
-                removed_owners.append(old_owner)
-
-        # Add new owners
-        for new_owner in new_owners:
-            if new_owner not in old_owners:
-                assign_perm('owner', new_owner, self)
-                added_owners.append(new_owner)
-
-        # Return the owners added/removed so they can be emailed
-        return removed_owners, added_owners
+        return get_dandiset_owners(self)
 
     @classmethod
     def published_count(cls):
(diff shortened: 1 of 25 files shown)


dandiapi/api/views/asset.py
```diff
@@ -94,19 +99,15 @@ def raise_if_unauthorized(self):
         # Clients must be authenticated to access it
         if not self.request.user.is_authenticated:
             raise NotAuthenticated
+        self.request.user = typing.cast(User, self.request.user)
 
         # Admins are allowed to access any embargoed asset blob
         if self.request.user.is_superuser:
             return
 
         # User must be an owner on any of the dandisets this asset belongs to
```
mvandenburgh, line 108: The logic changed so that we're not checking the user's dandisets here to determine asset ownership, so this comment is outdated now
```suggestion
```

dandiapi/api/services/permissions/dandiset.py
```diff
@@ -0,0 +1,118 @@
...
+    # union().
+    # Therefore, we resort to fetching the list of all primary keys for dandisets owned by the
+    # current user with `get_objects_for_user(...)`, then filter for either those dandisets or
+    # OPEN dandisets. There aren't very many dandisets and most users won't own very many of
+    # them, so there shouldn't be too many dandisets in the list.
+
+    owned_dandiset_pks = get_owned_dandisets(user).values('pk').all()
+    return Dandiset.objects.filter(
+        Q(embargo_status=Dandiset.EmbargoStatus.OPEN) | Q(pk__in=owned_dandiset_pks)
+    ).order_by('created')
```
mvandenburgh, line 105: This comment is incorrect, this actually can be done with one query -
```suggestion
def get_visible_dandisets(user: AbstractBaseUser | AnonymousUser) -> QuerySet[Dandiset]:
    """
    Return a queryset containing all dandisets visible to the given user.

    This is basically all dandisets except embargoed dandisets not owned by the given user.
    """

    # Get dandisets owned by this user, potentially including embargoed dandisets
    owned_dandisets = Dandiset.objects.filter(
        pk__in=(
            DandisetUserObjectPermission.objects.filter(
                user=user, permission__codename='owner'
            ).values('pk'),
        )
    )

    open_dandisets = Dandiset.objects.filter(embargo_status=Dandiset.EmbargoStatus.OPEN)

    # The union of these two sets is the set of dandisets visible to the user
    return owned_dandisets.union(open_dandisets)
```

Although I see now that this was copied from another file and introduced in this PR, so we can also defer it to a follow up if you prefer.

mvandenburgh (changes requested):

dandiapi/api/models/dandiset.py
```diff
@@ -76,28 +47,9 @@ def draft_version(self):
...
-        # Add new owners
-        for new_owner in new_owners:
-            if new_owner not in old_owners:
-                assign_perm('owner', new_owner, self)
-                added_owners.append(new_owner)
+        from dandiapi.api.services.permissions.dandiset import get_dandiset_owners
 
-        # Return the owners added/removed so they can be emailed
-        return removed_owners, added_owners
+        return get_dandiset_owners(self)
```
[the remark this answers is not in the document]
jjnesbitt: That sounds great to me. I didn't do that in an attempt to limit the scope of this PR, but if you think it's warranted, I'm happy to include that change.

dandiapi/api/services/permissions/dandiset.py
```diff
@@ -0,0 +1,118 @@
...
+    ).delete()
+
+    # Set owners to new list
+    for user in users:
+        add_dandiset_owner(dandiset, user)
+
+    # Return the owners added/removed so they can be emailed
+    removed_owners = existing_owner_set - new_owner_set
+    added_owners = new_owner_set - existing_owner_set
+    return removed_owners, added_owners
```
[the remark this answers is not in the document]
jjnesbitt: Good idea :+1:

dandiapi/api/services/permissions/dandiset.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def is_dandiset_owner(dandiset: Dandiset, user: AbstractBaseUser | AnonymousUser) -> bool:
+    if isinstance(user, AnonymousUser):
+        return False
+
+    user = typing.cast(User, user)
+    return user.has_perm('owner', dandiset)
```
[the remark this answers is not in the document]
jjnesbitt: This is mostly a typing issue. The user type that is thrown around before any checks are done on it is `AbstractBaseUser | AnonymousUser`. So if we use the type `User` instead of `AbstractBaseUser`, type errors will be thrown in all the places where we invoke this function, and we'd either have to ignore that type error, or cast the type before calling this function, which is very messy.

I figure it's easiest to accept the most generic user type possible, and centralize the filtering in this function, rather than all the places it's invoked. I originally had the user type as `User`, but ran into the typing issues I mentioned, and so went with this.

dandiapi/api/views/asset.py
jjnesbitt: How has the logic changed? `is_owned_asset` is called right below this line, which is what this comment is referring to.

dandiapi/api/services/permissions/dandiset.py
jjnesbitt: It seems like implementing this causes some tests and endpoints to fail, due to the union operation preventing much else from happening to a queryset. It actually seems like that was captured in the comment:

```
    # We would like to do something like:
    # queryset = Dandiset.objects.filter(embargo_status=OPEN).union(get_objects_for_user(...))
    # but you cannot filter or perform many other common queryset operations after using
    # union().
```

Although I imagine you're referring to the first part of that comment that mentions django-guardian. If we can clean this up, I'd prefer to do it in a follow-on PR. Although I'd like to continue to stay away from `union` if we can.

mvandenburgh: Sounds good. I think we could avoid `union`, but we can defer further discussion to a future PR

dandiapi/api/views/asset.py
mvandenburgh: "Logic changed" was probably the wrong wording, I meant that we're not explicitly checking dandiset ownership in the code under this comment anymore, but instead just calling the opaque `is_owned_asset` function.

jjnesbitt: I still think the comment is valid, as it describes both why we're calling `is_owned_asset`, as well as the fact that this check returns true if the user is an owner of _any_ dandiset this asset belongs to.

mvandenburgh: Fair enough, I don't have any objections to keeping it

mvandenburgh (approved):

dandiapi/api/services/permissions/dandiset.py
waxlamp: @jjnesbitt or @mvandenburgh, could one of you file an issue to capture this?

Issue #2116, opened 2024-12-18
Embargo end date should not be a user populated field


jjnesbitt:
Currently, the `embargoedUntil` field in the asset metadata is required to be set if the dandiset is embargoed. However, this is often _not_ set, and as a result, people are left with validation errors. [Here](https://github.com/dandi/dandi-schema/blob/395bc5b38579d29701e9629ce17daa34290a5891/dandischema/models.py#L1116-L1124) is where that error is generated in dandi-schema.

IMO this should not be a user filled field, and should instead be automatically populated by the archive. When a user first creates an embargoed dandiset, we could require that the user set an embargo end date, with a maximum of 1 year from the current date (as per NIH requirements). This would then automatically populate that field in the schema, and when that time comes, we would automatically unembargo that dandiset. If the user unembargos before that previously set unembargo end date, we can update it to when it was actually unembargoed.

Thoughts @satra @yarikoptic?

satra:
@jjnesbitt - the nih requirement is more complicated: end of award or publication of paper whichever is earlier. we could ask the user for end of award date when they create an embargoed dataset. and use that to set any asset. i agree that this shouldn't be a user filled element, and should be updated when unembargoed.

jjnesbitt:
> @jjnesbitt - the nih requirement is more complicated: end of award or publication of paper whichever is earlier. we could ask the user for end of award date when they create an embargoed dataset. and use that to set any asset. i agree that this shouldn't be a user filled element, and should be updated when unembargoed.

Thanks for the clarification. Is there any concern about verifying this end of award date? For example, a user could set an "end of award" date 10 years in the future, just because they don't feel like dealing with it.

satra:
no award is typically more than 5 years. so we should put in that sanity check. we can perhaps verify and update as an admin component.  when that happens, i suspect some command may need to be run to update everything. for nih, awards we should really use the reporter.nih.gov API to pull in award name and project end date. however, all awards will not be nih. hence this may require a bit of work to restructure the registration (people have also complained that we only say nih on that dandiset embargo registration page).

kabilar:
Thanks @jjnesbitt.

> ...we could require that the user set an embargo end date, with a maximum of 1 year from the current date (as per NIH requirements).

I might be missing some context but why can the embargo end date only be a maximum of 1 year from the current date?

> ...when that time comes, we would automatically unembargo that dandiset.

[...]

PR #2114, opened 2024-12-17
Capitalize `Archive` when it follows `DANDI`

kabilar:
The changes in this pull request are based on the [style guidelines](https://github.com/dandi/handbook?tab=readme-ov-file#dandi-style-guidelines).

cc @yarikoptic

PR #2117, opened 2024-12-18
Set assets to pending during unembargo

jjnesbitt:
This addresses the effects of #2116. Currently, without this change, every asset of a previously unembargoed dandiset can be left with the validation error `An embargo end date is required for NIH awards to be in compliance with NIH resource sharing policy.`. Now, when a dandiset is unembargoed, all assets will be set to `PENDING` status, so that their metadata can be re-validated.

### DIFF
--- a/dandiapi/api/services/embargo/__init__.py
+++ b/dandiapi/api/services/embargo/__init__.py
@@ -7,6 +7,7 @@ from django.db import transaction
 
 from dandiapi.api.mail import send_dandiset_unembargoed_message
 from dandiapi.api.models import AssetBlob, Dandiset, Version
+from dandiapi.api.models.asset import Asset
 from dandiapi.api.services import audit
 from dandiapi.api.services.asset.exceptions import DandisetOwnerRequiredError
 from dandiapi.api.services.embargo.utils import _delete_object_tags, remove_dandiset_embargo_tags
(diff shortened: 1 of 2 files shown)


jjnesbitt:
@yarikoptic @jwodder It seems the integration tests are failing, although I can't really discern why, as there's so much log output. It's just the `test_update_dandiset_from_doi` test that's failing, but in all parameterizations. Maybe the tests need to be updated? I'm not really sure why this change would affect those tests.

jwodder:
@jjnesbitt It looks like the tests are failing because they expect Dandiset "citation" fields to say "DANDI archive" instead of "DANDI Archive".

jjnesbitt:
> @jjnesbitt It looks like the tests are failing because they expect Dandiset "citation" fields to say "DANDI archive" instead of "DANDI Archive".

Hmm okay, seems like it's due to https://github.com/dandi/dandi-archive/pull/2114 being merged then. Could those tests be updated then?

jwodder:
@jjnesbitt https://github.com/dandi/dandi-cli/pull/1555

jjnesbitt:
> @jjnesbitt [dandi/dandi-cli#1555](https://github.com/dandi/dandi-cli/pull/1555)

Much appreciated!

mvandenburgh (approved):

Issue #2107, opened 2024-12-11
CI can't read GitHub Actions secrets on fork-based PRs


mvandenburgh:
Example: https://github.com/dandi/dandi-archive/pull/2105

Because that PR was based on a fork of `dandi-archive`, and not a branch internal to the `dandi/dandi-archive` repo, the `DOCKER_LOGIN` and  `DOCKER_TOKEN` GitHub Actions secrets could not be read due to permission issues.

PR #2118, opened 2024-12-19
Login to docker hub only if credentials were provided

yarikoptic:
I think this could be the only solution (may be besides also adding retries in addition) to allow testing against dandi-cli to be done also in PRs from forks of the project, where there is no credentials.  We could also advise our developers to provide such (own) credentials in their forks.

FTR: login originally was added in 6b815dc6aabf33aebf846aa56b67fb952e13bc77

- I think we can state that it closes https://github.com/dandi/dandi-archive/issues/2107 . If it fails to pull from docker -- it would be a differently titled issue ;-)
- note: this PR submitted from a personal fork, so would test ability to pass for similar ones

### DIFF
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -22,7 +22,7 @@ jobs:
           if [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ]; then
               docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
           else
-              echo "Not login-in to docker since no credentials were provided, might hit limits below"
+              echo "Not logging in to docker since no credentials were provided, might hit limits below"
           fi
           docker build \
             -t dandiarchive/dandiarchive-api \
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -19,7 +19,9 @@ jobs:
 
       - name: Build Docker image
         run: |
-          docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
+          [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ] \
+          && docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN" \
+          || echo "Not login-in to docker since no credentials were provided, might hit limits below"
           docker build \
             -t dandiarchive/dandiarchive-api \
             -f dev/django-public.Dockerfile \
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -19,9 +19,11 @@ jobs:
 
       - name: Build Docker image
         run: |
-          [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ] \
-          && docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN" \
-          || echo "Not login-in to docker since no credentials were provided, might hit limits below"
+          if [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ]; then
+              docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
+          else
+              echo "Not login-in to docker since no credentials were provided, might hit limits below"
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-integration.yml
```diff
@@ -19,7 +19,9 @@ jobs:
 
       - name: Build Docker image
         run: |
-          docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
+          [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ] \
+          && docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN" \
+          || echo "Not login-in to docker since no credentials were provided, might hit limits below"
```
jwodder, line 24: This echo will run if `docker login` fails.  It would be better to use an `if` block.

jwodder (changes requested):

.github/workflows/cli-integration.yml
yarikoptic: indeed! thanks... alternative is to revert condition here to `-z` and doing echo on `&&`... but indeed if/then/else is more explicit, will do

yarikoptic:
success as we did build image without credentials:
```
Not login-in to docker since no credentials were provided, might hit limits below
14
#0 building with "default" instance using docker driver
15

16
#1 [internal] load build definition from django-public.Dockerfile
17
#1 transferring dockerfile: 373B done
18
#1 DONE 0.0s
```

the fails are actually due to that
- #2114 was not "complete" and thus broke it and started to fail in master, e.g. https://github.com/dandi/dandi-archive/actions/runs/12415512591

I will be addressing @jwodder comment and will see if could fix it while at this

yarikoptic:
scrape above "comment" - sure thing it broke testing of dandi-cli since we need now to fix it up there too! ;-)

.github/workflows/cli-integration.yml
```diff
@@ -19,7 +19,11 @@ jobs:
 
       - name: Build Docker image
         run: |
-          docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
+          if [ -n "$DOCKER_LOGIN$DOCKER_TOKEN" ]; then
+              docker login -u "$DOCKER_LOGIN" --password-stdin <<<"$DOCKER_TOKEN"
+          else
+              echo "Not login-in to docker since no credentials were provided, might hit limits below"
```
jwodder, line 25:
```suggestion
              echo "Not logging in to docker since no credentials were provided, might hit limits below"
```

jwodder (changes requested):

.github/workflows/cli-integration.yml
yarikoptic: <details>
<summary>that is a word on itself which AFAIK has nothing to do with the process to login ;-)</summary> 

```shell
❯ dict logging
3 definitions found

From The Collaborative International Dictionary of English v.0.48 [gcide]:

  Log \Log\, v. t. [imp. & p. p. {Logged}; p. pr. & vb. n.
     {Logging}.]
     1. (Naut.), To enter in a ship's log book; as, to log the
        miles run. --J. F. Cooper.
        [1913 Webster]
  
     2. To record any event in a logbook, especially an event
        relating to the operation of a machine or device.
        [PJC]

From The Collaborative International Dictionary of English v.0.48 [gcide]:

  Logging \Log"ging\, n.
     The business of felling trees, cutting them into logs, and
     transporting the logs to sawmills or to market.
     [1913 Webster]

From WordNet (r) 3.0 (2006) [wn]:

  logging
      n 1: the work of cutting down trees for timber

```
</details>

ok, what about 

```suggestion
              echo "No credentials for docker were provided, no 'docker login' used."
```

?

yarikoptic:
FTR -- build image is fine again!

.github/workflows/cli-integration.yml
jwodder: "Logging in" is the progressive form of "to log in": https://www.merriam-webster.com/dictionary/log%20in

yarikoptic: thanks for digging, let's then go with

```suggestion
              echo "Not logging in to docker since no credentials were provided, might hit limits below"
```

PR #2119, opened 2024-12-19
Enable User search by GitHub username

waxlamp:
This adds the `github_username` field to the list of searchable fields in the `User` admin console.


jjnesbitt (approved):
The integration tests will likely fail for an unrelated reason.

PR #2122, opened 2024-12-21
Display validation errors on DLP for embargoed dandisets

### LINKED ISSUE #2089: Embargoed dandisets assets are not showing up in "Assets Summary"
I have an issue with two embargoed dandisets ([here](https://dandiarchive.org/dandiset/001264/draft) and [here](https://dandiarchive.org/dandiset/001265/draft)), both have multiple assets uploaded but the "Assets Summary" is empty.

I tried to use `dandi service-scripts reextract-metadata https://dandiarchive.org/dandiset/001264/draft` to update the asset metadata, but it still shows an empty "Assets Summary".

Screenshot 2024-12-05 at 15 56 34 (2)

Is there another way to force updating the metadata for the dandisets?

mvandenburgh:
Fixes #2089.

We currently don't display validation errors on the DLP for embargoed dandisets. This seems to be an oversight due to much of the UI being duplicated between the `DandisetPublish` and `DandisetUnembargo` components. I've taken an initial stab at refactoring some of the duplicate templates into a reusable component in this PR, but we could probably clean it up a bit more. 

The text displayed when the Assets Summary is missing (`This Dandiset does not contain any assets.`) is also slightly misleading. A dandiset that contains _only invalid assets_ will also not have an Assets Summary; I've updated that text to reflect that.

### DIFF
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -171,8 +171,8 @@
             <div
               v-else-if="!assetSummary || !Object.keys(assetSummary).length"
               class="font-italic font-weight-bold"
+              v-text="`This Dandiset does not contain any valid assets.${currentDandiset?.asset_validation_errors.length ? ' Please check the asset validation errors on the right panel.' : ''}`"
             >
-              This Dandiset does not contain any valid assets.
             </div>
             <div
               v-for="([type, items], i) in Object.entries(assetSummary)"
--- a/web/src/components/DLP/OverviewTab.vue
+++ b/web/src/components/DLP/OverviewTab.vue
@@ -172,7 +172,7 @@
               v-else-if="!assetSummary || !Object.keys(assetSummary).length"
               class="font-italic font-weight-bold"
             >
-              This Dandiset does not contain any assets.
+              This Dandiset does not contain any valid assets.
             </div>
             <div
               v-for="([type, items], i) in Object.entries(assetSummary)"
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -153,124 +153,7 @@
       </v-tooltip>
     </v-row>
 
-    <v-row
-      v-if="currentDandiset.status === 'Pending'"
-      class="my-2 px-1"
-      no-gutters
-    >
-      <v-menu
-        :nudge-width="200"
-      >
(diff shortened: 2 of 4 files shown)


mvandenburgh:
See https://deploy-preview-2122--gui-staging-dandiarchive-org.netlify.app/dandiset/215957 for working example

web/src/components/DLP/OverviewTab.vue
```diff
@@ -165,7 +165,7 @@
               v-else-if="!assetSummary || !Object.keys(assetSummary).length"
               class="font-italic font-weight-bold"
             >
-              This Dandiset does not contain any assets.
+              This Dandiset does not contain any valid assets.
```
jjnesbitt, line 168: I know this is technically the most concise way to phrase this, but I wonder if people will take both wordings to mean the same thing. They might say "hold on, I just uploaded a bunch of assets that I _know_ are valid", even though "valid" might mean different things to us vs them. Maybe we should reword it to something like:

`This dandiset does not contain any valid assets. Please check the validation errors to address any issue with asset metadata.`, or something like that :man_shrugging: .

jjnesbitt (changes requested):
LGTM, just had one suggestion.

web/src/components/DLP/OverviewTab.vue
mvandenburgh: I added a message about the validation errors that only appears when there are asset validation errors - https://github.com/dandi/dandi-archive/pull/2122/commits/9fcf1dd8053a0a96ec4dd2c06f980b30ce8bb01b. What do you think?

jjnesbitt: I think the logic in that commit is inverted. This can be seen in [this](https://deploy-preview-2122--gui-staging-dandiarchive-org.netlify.app/dandiset/101233) example. It should show the extended message, but just shows the short one. While the [other](https://deploy-preview-2122--gui-staging-dandiarchive-org.netlify.app/dandiset/215957) example shows the extended message, even though it has zero assets.

jjnesbitt: Besides that I think the extended message is perfect.

mvandenburgh: Oh I inverted the logic to test it, and forgot to change it back :man_facepalming: . Thanks for catching that, I fixed it

jjnesbitt (approved):

PR #2123, opened 2024-12-26
Add Dandiset star functionality with UI components

### LINKED ISSUE #1119: GitHub stars on dandisets
I think having a star system similar to GitHub's would be helpful for promoting dandisets.
image

You are all familiar with this feature, but let me just lay out all of the aspects of its implementation:

1. Dandiset landing pages have a star icon. If a user is signed in they can star the dandiset. Users can also unstar the dandiset.
2. Next to the star is a number that indicates the number of times any user has starred it.
3. Users have a tab for "Starred dandisets" which displays all of the dandisets they have starred.
4. On the list page, the number of stars is visible for each dandiset.
5. Users can sort dandisets by most starred.

This would have several benefits:
* This creates a prestige that users can aspire to. I know programmers pay attention when someone's repo has a lot of stars, and I think this will have a similar effect.
* This is a much better measure of influence than number of downloads, which may be inflated for any number of reasons.
* This would help a new user quickly identify democratically selected well-established dandisets.

bendichter:
fix #1119 

Feature developed with Cursor

- Introduced DandisetStar model and admin interface for managing starred Dandisets.
- Implemented star/unstar actions in the DandisetViewSet and corresponding API endpoints.
- Added star count and starred status to DandisetSerializer.
- Created StarButton component for toggling star status in the UI.
- Developed StarredDandisetsView to display Dandisets starred by the user.
- Updated DandisetList and DandisetLandingView to include star functionality.
- Enhanced DandisetFilterBackend to support ordering by star count.
- Update existing tests and add new tests
- Add package-lock.json to ignore list for spell check

I haven't been able to log in on my dev server so I can't test that behavior, but the unlogged in web UI and the admin interface appear to work as expected


waxlamp (commented):
Very interesting!

I'm a dinosaur who's not ready to trust AI-generated code just yet, so I'll want to get careful reviews from @jjnesbitt and @mvandenburgh (both to identify problems with the code, and so we have team members with a mental model of how it works) before merging. Furthermore, I think this will benefit from some enhancements (like using a service layer to manage stars), as well as some minor stuff that I can push fixes for myself. Finally, I would like for @jtomeck to help out with the frontend design here as well.

Thanks @bendichter for kicking this feature off!

web/package-lock.json
```diff
@@ -0,0 +1,6546 @@
+{
```
yarikoptic, line 1: @bendichter why that package-lock.json is getting introduced here at all, or why is it becoming needed for adding this feature?

web/yarn.lock
```diff
@@ -4,17 +4,17 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
```
yarikoptic, line 7: doesn't feel like needed changes either for the feature ("migration from yarnpkg.com to npmjs.org")?

web/package-lock.json
bendichter: isn't this a file that gets automatically generated? Maybe it should be in .gitignore?

web/yarn.lock
bendichter: This also seems like it might belong in .gitingore

bendichter:
@waxlamp yes, I mentioned Cursor because I agree that this will need careful eyes on it before integrating. But looking at the changes the majority of them seem straightforward

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-exclude: (^\.codespellrc|yarn.lock)
+exclude: (^\.codespellrc|yarn.lock|package-lock.json)
```
bendichter, line 3:
```suggestion
exclude: (^\.codespellrc|yarn.lock)
```

web/package-lock.json
waxlamp: This file is generated by `npm` whereas we have always used `yarn` to build the frontend. I'm planning to revert the addition of this file, along with the changes in yarn.lock, as I don't believe they are needed for this PR. (You can do this yourself, @bendichter, if you prefer, but it's something I can easily do when I work on this.)

web/yarn.lock
bendichter: oh, I see, this was created when I accidentally ran nmp install instead of yarn install

waxlamp: We can't put this it .gitignore because there are times when indeed we need to track/commit changes to this file.

web/package-lock.json
bendichter: oh, I see, this was created when I accidentally ran nmp install instead of yarn install.

web/yarn.lock
bendichter: ah ok

bendichter:
ok yeah the npm-related issues are my fault. I am used to installing with npm and having large automatically generated files as a result of that. I'll make sure to use yarn in the future.

bendichter:
I don't know what to do about the remaining CI check failures

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -74,3 +74,45 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
...
+        The new star count for the Dandiset.
+    """
+    if not user.is_authenticated:
+        return dandiset.star_count
+
+    with transaction.atomic():
+        DandisetStar.objects.get_or_create(user=user, dandiset=dandiset)
+        # Refresh star count from database
+        dandiset.refresh_from_db()
+        return dandiset.star_count
```
jjnesbitt, line 97: There's not really a reason to use `transaction.atomic` here, since there's only one write action being performed. The `refresh_from_db` call is also superfluous, since `star_count` makes a query to the DB anyway.

```suggestion
    DandisetStar.objects.get_or_create(user=user, dandiset=dandiset)
    return dandiset.star_count
```

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -74,3 +74,45 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
...
+        The new star count for the Dandiset.
+    """
+    if not user.is_authenticated:
+        return dandiset.star_count
+
+    with transaction.atomic():
+        DandisetStar.objects.filter(user=user, dandiset=dandiset).delete()
+        # Refresh star count from database
+        dandiset.refresh_from_db()
+        return dandiset.star_count
```
jjnesbitt, line 118: Same complaint as above
```suggestion
    DandisetStar.objects.filter(user=user, dandiset=dandiset).delete()
    return dandiset.star_count
```

dandiapi/api/views/dandiset.py
```diff
@@ -101,6 +106,9 @@ def filter_queryset(self, request, queryset, view):
                     )
                 )
                 return queryset.order_by(ordering)
+            if ordering.endswith('stars'):
+                queryset = queryset.annotate(stars_count=Count('stars'))
+                return queryset.order_by(f"{'-' if ordering.startswith('-') else ''}stars_count")
```
jjnesbitt, line 111: I find the original a bit hard to read
```suggestion
                queryset = queryset.annotate(stars_count=Count('stars'))
                prefix = '-' if ordering.startswith('-') else ''
                return queryset.order_by(f"{prefix}stars_count")
```

dandiapi/api/views/dandiset.py
```diff
@@ -502,3 +512,59 @@ def clear_uploads(self, request, dandiset__pk):
...
+        request_body=no_body,
+        responses={
+            200: 'Dandiset starred successfully',
+            401: 'Authentication required',
+        },
+        operation_summary='Star a dandiset.',
+        operation_description='Star a dandiset. User must be authenticated.',
+    )
+    @action(methods=['POST'], detail=True)
+    def star(self, request, dandiset__pk) -> Response:
```
jjnesbitt, line 528: I think this PR should be merged after #2115, and one of the reasons is that this PR could take advantage of the new permission service layer. For example, a decorator that requires authentication could be used, similar to the `require_dandiset_owner_or_403` decorator introduced in that PR.

dandiapi/api/views/serializers.py
```diff
@@ -62,6 +66,15 @@ def get_contact_person(self, dandiset: Dandiset):
...
         return extract_contact_person(latest_version)
 
+    def get_star_count(self, dandiset):
+        return dandiset.star_count
+
+    def get_is_starred(self, dandiset):
+        request = self.context.get('request')
+        if not request or not request.user.is_authenticated:
+            return False
+        return dandiset.is_starred_by(request.user)
```
jjnesbitt, line 76: This is inefficient as it leads to a SQL query for every dandiset that's listed (and I can confirm this from local testing). I think something more complex needs to be done, for example, retrieving all of the stars that are associated with the dandisets being listed for one page, and add them to the serialized result.

web/yarn.lock
jjnesbitt, line 1: I think this has already been settled but just in case, this file should not be modified, as nothing in `package.json` has been modified.

web/src/components/StarButton.vue
jjnesbitt, line 1: It seems there's some bugs with this component, or how it's invoked by other components. I've included two videos that showcase the following:

1. Starred dandisets don't show up as "filled in" yellow stars on the list page, just outlined (although the number is correct).
2. Clicking on the star button on the listing page brings you to that dandiset, even though it does still star the dandiset (the click is being propagated down to the component underneath). This is unexpected behavior from a UX perspective, and this action should only star the dandiset.
3. On the DLP, if a dandiset is already starred, it once again does not show as filled in. Furthermore, if you click on the button, it will increase the displayed star count, even though you've already starred it. On the backend, no new star is being added, but still, the behavior is undesirable.

https://github.com/user-attachments/assets/149dd196-93c4-40e6-811c-d644f3f41787

https://github.com/user-attachments/assets/828c9e4b-cc75-46a1-9abc-d39ef4939204

web/src/router.ts
```diff
@@ -56,6 +57,14 @@ const routes: RouteConfig[] = [
...
     component: SearchView,
   },
+  {
+    path: '/dandiset/starred',
+    name: 'starredDandisets',
+    component: StarredDandisetsView,
+    meta: {
+      requiresAuth: true,
+    },
+  },
```
jjnesbitt, line 67: This page is unviewable at the moment (at least for me), as it's lower in priority than the `'/dandiset/:identifier/:version?` URL pattern above. So when `/dandiset/starred` is navigated to, it attempts to load the dandiset with the ID of `starred`, which obviously doesn't exist. This needs to be moved above that pattern, so that it's matched first.

web/src/views/StarredDandisetsView/StarredDandisetsView.vue
jjnesbitt, line 1: It seems something is wrong with the filtering logic of this component. Even if I fix the url matching issue that I mentioned elsewhere, it simply shows all dandisets, instead of just dandisets that I've starred.

jjnesbitt, line 1: I think we need to add an entry to https://github.com/dandi/dandi-archive/blob/stars/web/src/components/AppBar/AppBar.vue#L166, which navigates to the starred dandisets page. As it is now, there's no way to get there aside from changing the URL.

jjnesbitt (changes requested):
Thanks for the work, @bendichter. I've done a first pass of this PR and identified some necessary changes, but nonetheless it's great to have the outline of this feature out of the way. 

I'd be happy to address some of these changes myself, if that makes more sense. I'll still leave the comments either way.

bendichter:
@jjnesbitt thanks for the prompt and thorough review. I can confirm that I can replicate those UI bugs. I might have time to address them myself but feel free to take this over to any extent you have the bandwidth to do so.

dandiapi/api/views/serializers.py
bendichter: @waxlamp mentioned an "N+1 query problem", which sounds like it's the same or similar to what you are bringing up here. This would also be the case for the contact person of each dandiset, right?

jjnesbitt: Indeed this is an N+1 problem. This doesn't happen with contact person as the `get_contact_person` method that _would_ cause this (originally defined in `DandisetSerializer`) is [overwritten ](https://github.com/dandi/dandi-archive/blob/stars/dandiapi/api/views/serializers.py#L160-L173) in `DandisetListSerializer`, to operate without making extra queries. (Side note, I think this is a bit unnecessary. I don't see a real reason we'd need to display the contact person at the dandiset list level).

Anyway, I think that code snippet is a good example of how we would update this implementation to solve the N+1 query problem, using that `self.context` field.

jjnesbitt:
> @jjnesbitt thanks for the prompt and thorough review. I can confirm that I can replicate those UI bugs. I might have time to address them myself but feel free to take this over if you have the bandwidth to do so.

I think I have the bandwidth, so today I can take a pass at changing someone of the things I mentioned.

dandiapi/api/views/serializers.py
jjnesbitt: Addressed in [b27390a](https://github.com/dandi/dandi-archive/pull/2123/commits/b27390a07d9952ac06d2343c90aa2954cc9d3f89).

dandiapi/api/views/version.py
```diff
@@ -79,7 +79,9 @@ def retrieve(self, request, **kwargs):
     def info(self, request, **kwargs):
         """Django serialization of a version."""
         version = self.get_object()
-        serializer = VersionDetailSerializer(instance=version)
+        serializer = VersionDetailSerializer(
+            instance=version, context=self.get_serializer_context()
+        )
```
jjnesbitt, line 84: @mvandenburgh @waxlamp I don't love this solution, but it works. I was curious how the other serializers work without this (e.g. the Dandiset retrieve endpoint), and it turns out the DRFViewSet class we inherit from (specifically `ReadOnlyModelViewSet` -> `GenericViewSet` -> `generics.GenericAPIView`) automatically populate this context when the `get_serializer` method is called. This means that on all untouched methods we inherit and use, it's populated, while on any methods we've overwritten, it's missing (unless we explicitly specify it, like it done here).

To me this suggests that we should do an overarching refactor of our view and serialization code, since I think we have a lot of technical debt there. But for now, this is what seems to be the easiest change that's in scope.

mvandenburgh (changes requested):
We'll also need to add audit events for starring/unstarring a dandiset.

web/src/components/StarButton.vue
jjnesbitt: I believe I've addressed all of these issues.

web/package-lock.json
jjnesbitt: This change has been reverted in [7b4441d](https://github.com/dandi/dandi-archive/pull/2123/commits/7b4441dc4dba108d02d83dcea13b780365b8d267).

jjnesbitt:
> We'll also need to add audit events for starring/unstarring a dandiset.

For posterity: We met offline and agreed to exclude dandiset stars from audit events.

dandiapi/api/views/dandiset.py
jjnesbitt: Yes, good point. I'll see what can be done there.

jjnesbitt: I'd actually like to do this as a follow up. To functionally matter it would require a decent refactor of the `DandisetViewSet` class, and i'd say that's out of scope for this PR.

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -74,3 +78,39 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
...
+        user: The user unstarring the Dandiset.
+        dandiset: The Dandiset to unstar.
+
+    Returns:
+        The new star count for the Dandiset.
+    """
+    if not user.is_authenticated:
+        raise NotAuthenticatedError
+
+    DandisetStar.objects.filter(user=user, dandiset=dandiset).delete()
```
mvandenburgh, line 115: Similar to the previous suggestion about `star_dandiset`. This would mainly be catching the case where the function gets called on a dandiset that does not have a star from this user (i.e. when `objects_deleted` is `0`)
```suggestion
    objects_deleted = DandisetStar.objects.filter(user=user, dandiset=dandiset).delete()
    if objects_deleted != 1:
        # replace with a custom exception class that propogates a 4xx error up to the view response
        raise Exception(f'{objects_deleted} objects deleted by unstar_dandiset!!')
```

jjnesbitt:
> @jjnesbitt just one point about the failure conditions for starring/unstarring. I think you could make an argument for either way (i.e., starring/unstarring is idempotent and always returns a 2xx as it is now, vs. verifying validity of the request and possibly rejecting it with a 4xx), any thoughts?

IMO the star/unstar operation should be idempotent. Raising an error in this case doesn't really provide us with much benefit.

It can actually lead to an undesirable outcome in the following situation:

1. User has a dandiset open in two tabs for some reason.
2. In the first tab they star the dandiset, showing it reflected as being starred.
3. Later, in the second tab, they see it isn't starred for some reason, and hit star again. Since the front-end is out of sync with the back-end, that would call `/star` again, and raise an exception. 

This is unnecessarily confusing to the user with no real upside. It's not like the user can do anything, and we couldn't really do anything either, except for try to prevent that situation from occurring in the first place.

mvandenburgh (approved):

dandiapi/api/tests/test_dandiset.py
```diff
@@ -1230,3 +1248,86 @@ def test_dandiset_rest_clear_active_uploads(
...
+
+@pytest.mark.django_db
+def test_dandiset_star_count(api_client, user_factory, dandiset):
+    users = [user_factory() for _ in range(3)]
+    for user in users:
+        api_client.force_authenticate(user=user)
+        api_client.post(f'/api/dandisets/{dandiset.identifier}/star/')
+
+    response = api_client.get(f'/api/dandisets/{dandiset.identifier}/')
+    assert response.data['star_count'] == 3
```
waxlamp, line 1291:
```suggestion
    assert response.data['star_count'] == len(users)
```

dandiapi/api/views/dandiset.py
```diff
@@ -538,3 +598,37 @@ def clear_uploads(self, request, dandiset__pk):
...
+            401: 'Authentication required',
+        },
+        operation_summary='Unstar a dandiset.',
+        operation_description='Unstar a dandiset. User must be authenticated.',
+    )
+    @action(methods=['POST'], detail=True)
+    def unstar(self, request, dandiset__pk) -> Response:
+        dandiset = self.get_object()
+        star_count = unstar_dandiset(user=request.user, dandiset=dandiset)
+        return Response({'count': star_count}, status=status.HTTP_200_OK)
```
waxlamp, line 634: We could unify these two functions into one if we had them as
- `POST /dandiset/{pk}/star`
- `DELETE /dandiset/{pk}/star`
(you'd just have a conditional on the method to select between `star_dandiset` and `unstar_dandiset` and the rest of the function is identical between the two) and it would be more idiomatically "RESTful".

But this existing design is (RESTfully) valid as well. @mvandenburgh, @jjnesbitt, what do you think?

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -74,3 +78,39 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
...
+        user: The user starring the Dandiset.
+        dandiset: The Dandiset to star.
+
+    Returns:
+        The new star count for the Dandiset.
+    """
+    if not user.is_authenticated:
+        raise NotAuthenticatedError
+
+    DandisetStar.objects.get_or_create(user=user, dandiset=dandiset)
```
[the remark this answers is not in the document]
waxlamp: I see you resolved this, @mvandenburgh, but there's no explanation as to your concern being resolved. I was curious about the argument here.

dandiapi/api/models/dandiset.py
```diff
@@ -58,10 +59,33 @@ def published_count(cls):
...
+    dandiset = models.ForeignKey(Dandiset, on_delete=models.CASCADE, related_name='stars')
+    created = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(name='unique-user-dandiset-star', fields=['user', 'dandiset'])
+        ]
+
+    def __str__(self) -> str:
+        return f'Star {self.user.username} ★ {self.dandiset.identifier}'
```
waxlamp, line 91: I am curious why we don't express this as a direct many to many relationship between `Dandiset` and `User`. I see that expressing a "through table" model like this is exactly equivalent, but I'm still curious about the different approaches. Thoughts @mvandenburgh and @jjnesbitt?

waxlamp (approved):
This looks great, and I don't wish to hold up merge-and-deploy any longer, but I did have some questions and some minor suggestions in case they're useful.

dandiapi/api/services/dandiset/__init__.py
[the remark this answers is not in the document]
mvandenburgh: It was resolved by this commen https://github.com/dandi/dandi-archive/pull/2123#issuecomment-2625262813
Sorry, should have linked this to close out this thread

dandiapi/api/services/dandiset/__init__.py
mvandenburgh: Resolution - https://github.com/dandi/dandi-archive/pull/2123#issuecomment-2625262813

dandiapi/api/views/dandiset.py
mvandenburgh: I actually like the POST/DELETE better, that seems more inline with the RESTful philosophy of POST==create and DELETE==destroy

dandiapi/api/models/dandiset.py
mvandenburgh: I agree the `DandisetStar` table is essentially acting as a through table here, in the same way as if we had a `starred_dandisets = ManyToManyField(Dandiset)` on the `User` model. Practically, it is difficult to add new fields to the built-in `User` model though, see https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#substituting-a-custom-user-model

dandiapi/api/views/dandiset.py
jjnesbitt: I agree, good idea. I'll make that change.

dandiapi/api/models/dandiset.py
jjnesbitt: This was part of the Cursor generated code, but I don't see anything wrong with it. While it _may_ be more concise to have it as a `ManyToManyField` on the `Dandiset` model, I don't see a particularly good reason to do so. Django's `ManytoManyField` is doing a good deal of magic, with not really much of a benefit. With a separate model like this, it's functionally equivalent, with the ability to add more fields to the model (like the `created` field), as well as query flexibility (e.g. ability to do `DandisetStar.objects.filter(dandiset=000123)`).

dandiapi/api/services/dandiset/__init__.py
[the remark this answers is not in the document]
waxlamp: Ah, got it, thanks!

dandiapi/api/models/dandiset.py
waxlamp: > Practically, it is difficult to add new fields to the built-in User model though

We could just as easily place the `ManyToManyField` in `Dandiset.stars`, right? Would that still cause a problem with `User` on the other end of the connection?

> While it may be more concise to have it as a ManyToManyField on the Dandiset model, I don't see a particularly good reason to do so. Django's ManytoManyField is doing a good deal of magic, with not really much of a benefit.

I believe the main benefit is to directly express to developers that there's a many-to-many relationship happening here. And I think the magic is minimal: Django creates the through table and exposes the related models.

> With a separate model like this, it's functionally equivalent, with the ability to add more fields to the model (like the created field), as well as query flexibility (e.g. ability to do DandisetStar.objects.filter(dandiset=000123))

See [the `through` option](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ManyToManyField.through) for how to specify your own through table model in a many-to-many relationship. This delivers the added benefits of additional fields and the ability to query the table directly.

If this all actually works in our case, I'd like to shift over to using it.

mvandenburgh: > We could just as easily place the ManyToManyField in Dandiset.stars, right? Would that still cause a problem with User on the other end of the connection?

No, you're correct. I wasn't thinking about the relation in that that way, but it is completely valid to put the `ManyToManyField` on the `Dandiset` model. 

I agree we should keep the `DandisetStar` model, but use it with the `through` kwarg of the `ManyToManyField`.

dandiapi/api/views/dandiset.py
jjnesbitt: Changed in [afe79a3](https://github.com/dandi/dandi-archive/pull/2123/commits/afe79a35ef57a2c59cea5ec804529d389b358076)

dandiapi/api/models/dandiset.py
jjnesbitt: Changed in [4a02dbe](https://github.com/dandi/dandi-archive/pull/2123/commits/4a02dbe53f188e4073097dd0379cb5d8a91f3b20). Worth noting is that the `stars` relation still exists on `Dandiset`, and is used in the same way as before. The new explicit relationship to users is named `starred_users`.

dandiapi/api/views/dandiset.py
```diff
@@ -538,3 +598,36 @@ def clear_uploads(self, request, dandiset__pk):
...
+        manual_parameters=[DANDISET_PK_PARAM],
+        request_body=no_body,
+        responses={
+            200: 'Dandiset unstarred successfully',
+            401: 'Authentication required',
+        },
+        operation_summary='Unstar a dandiset.',
+        operation_description='Unstar a dandiset. User must be authenticated.',
+    )
+    @star.mapping.delete
```
waxlamp, line 629: What does this decorator do?

jjnesbitt: This decorator tells DRF to mount the function under the same endpoint as the `star`, using the `DELETE` method.  However, you raise a good point, these functions differ only slightly, and should be unified.

waxlamp (approved):

jjnesbitt (approved):

kabilar:
Great work team.  This is a really nice feature.  I tested on staging and production, and it is working well.

PR #2125, opened 2024-12-31
Prevent duplicate serialization of dandiset in detail view

jjnesbitt:
On any endpoint that used `DandisetDetailSerializer` for its response, the dandiset information was being included twice. Once at the top level, and once nested under each nested version serializer (`draft_version` and `most_recent_published_version`).

### DIFF
--- a/dandiapi/api/views/serializers.py
+++ b/dandiapi/api/views/serializers.py
@@ -100,6 +100,12 @@ class VersionSerializer(serializers.ModelSerializer):
     dandiset = DandisetSerializer()
     # name = serializers.SlugRelatedField(read_only=True, slug_field='name')
 
+    def __init__(self, *args, child_context=False, **kwargs):
+        if child_context:
+            del self.fields['dandiset']
+
+        super().__init__(*args, **kwargs)
+
 
 class DandisetVersionSerializer(serializers.ModelSerializer):
     """The version serializer nested within the Dandiset Serializer."""
@@ -174,8 +180,8 @@ class DandisetDetailSerializer(DandisetSerializer):
     class Meta(DandisetSerializer.Meta):
         fields = [*DandisetSerializer.Meta.fields, 'most_recent_published_version', 'draft_version']
 
-    most_recent_published_version = VersionSerializer(read_only=True)
-    draft_version = VersionSerializer(read_only=True)
+    most_recent_published_version = VersionSerializer(read_only=True, child_context=True)
+    draft_version = VersionSerializer(read_only=True, child_context=True)
 
 
 class DandisetQueryParameterSerializer(serializers.Serializer):
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -360,13 +360,6 @@ def test_dandiset_rest_create(api_client, user):
             'asset_count': 0,
             'active_uploads': 0,
             'size': 0,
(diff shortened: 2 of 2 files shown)


dandiapi/api/views/serializers.py
```diff
@@ -100,6 +100,12 @@ class Meta:
     dandiset = DandisetSerializer()
     # name = serializers.SlugRelatedField(read_only=True, slug_field='name')
 
+    def __init__(self, *args, **kwargs):
+        if kwargs.get('context', {}).get('child', False):
```
waxlamp, line 104: Is there a reason to reflect into `kwargs` rather than just include `context=None` as a keyword arg?

waxlamp (commented):
Looks good overall, but technically this would be a breaking change (someone may be relying on the Dandiset info appearing in one of the version records). Not sure what to do about that (maybe nothing).

dandiapi/api/views/serializers.py
jjnesbitt: You raise a good point. I've reworked the implementation to pluck this kwarg out, no longer passing it to the underlying base serializer class.

jjnesbitt:
> Looks good overall, but technically this would be a breaking change (someone may be relying on the Dandiset info appearing in one of the version records). Not sure what to do about that (maybe nothing).

You're correct in that it's technically a breaking change, however there are no CLI integration failures, so it would appear the CLI makes no use of this nested duplicated field. @jwodder Are you able to confirm this?

jwodder:
@jjnesbitt We're talking about removing the `"dandiset"` field in each of the `*_version` fields below, right?  The CLI won't care.

```json
{
  "identifier": "000027",
  "created": "2020-07-08T21:54:42.543000Z",
  "modified": "2022-01-17T16:22:43.888215Z",
  "contact_person": "Halchenko, Yaroslav O.",
  "embargo_status": "OPEN",
  "most_recent_published_version": {
    "version": "0.210831.2033",
    "name": "Test dataset for testing dandi-cli.",
    "asset_count": 1,
    "active_uploads": 0,
    "size": 18792,
    "status": "Valid",
    "created": "2021-08-31T20:34:01.209046Z",
    "modified": "2021-08-31T20:34:01.283878Z",
    "dandiset": {
      "identifier": "000027",
      "created": "2020-07-08T21:54:42.543000Z",
      "modified": "2022-01-17T16:22:43.888215Z",
      "contact_person": "Halchenko, Yaroslav O.",
      "embargo_status": "OPEN"
    }
  },
  "draft_version": {
    "version": "draft",
    "name": "Test dataset for testing dandi-cli.",
    "asset_count": 1,
    "active_uploads": 0,
    "size": 18792,
    "status": "Valid",
    "created": "2020-07-08T21:54:42.543000Z",
    "modified": "2023-06-20T00:56:23.249993Z",
    "dandiset": {
      "identifier": "000027",
      "created": "2020-07-08T21:54:42.543000Z",
      "modified": "2022-01-17T16:22:43.888215Z",
      "contact_person": "Halchenko, Yaroslav O.",
      "embargo_status": "OPEN"
    }
  }
}
```

jjnesbitt:
> @jjnesbitt We're talking about removing the `"dandiset"` field in each of the `*_version` fields below, right? The CLI won't care.

Yep that's correct. Thanks for the confirmation :+1:.

mvandenburgh (approved):

PR #2126, opened 2025-01-02
Move Google analytics tag to environment variable to prevent DANDI clones using by accident

aaronkanzer:
Relates to discussion at https://github.com/dandi/handbook/pull/151/files#r1731341212

Cc @kabilar


web/src/main.ts
```diff
@@ -47,7 +47,7 @@ Sentry.init({
 });
 
 Vue.use(VueGtag, {
-  config: { id: 'UA-146135810-2' },
+  config: { id: import.meta.env.VITE_APP_GOOGLE_ANALYTICS_TAG || '' },
```
waxlamp, line 50: Is `''` a valid value? It seems like it would be better to wrap the whole `Vue.use()` statement in a conditional that checks for whether `VITE_APP_GOOGLE_ANALYTICS_TAG` is set.

waxlamp (changes requested):
This is a good improvement, but it doesn't prevent clones from pushing data to the wrong GA sink, right?

web/src/main.ts
aaronkanzer: Updated the code here

```
but it doesn't prevent clones from pushing data to the wrong GA sink, right?
```

By moving the GA tag value as an env var in the `netlify.toml`, we are using the same pattern in which a clone would also set their Sentry value, backend API, etc.

If not putting the value in `netlify.toml`, what would you suggest?

waxlamp: No, you're right to copy the same pattern as the other values. I was just reacting to the claim in the PR title.

The deeper solution will be something like moving to a Terraform provider for Netlify, and using Netlify-hosted secrets to control these values (see #2147 if you're curious).

waxlamp (approved):

PR #2127, opened 2025-01-02
Fix reactivity bug in meditor

### LINKED ISSUE #2081: Switching of the version on DLP does not lead to switching of metadata version meditor uses
Originally troubleshooted in
- https://github.com/dandi/dandi-archive/issues/2023#issuecomment-2489416822

I think it might potentially lead to (meta)data loss and confusion (at least it did for me).  Steps to reproduce

- go to default DLP of a released dandiset e.g. https://dandiarchive.org/dandiset/000029 (non draft -- released) or some older version e.g. https://dandiarchive.org/dandiset/000029/0.210712.1903
- open meditor - and mark how many contributors you see (in draft in 000029 now 3 ppl, in older version - 1)
- close meditor
- switch to draft
- open meditor -- in contributors still 3 or 1 person although should be 4!

if you refresh the page  -- url will be with the current draft version -- and open meditor, you will see updated view.

mvandenburgh:
Fixes #2081

This PR adds a `key` to the `Meditor` component consisting of the currently selected dandiset identifier and version. This will result in the meditor re-rendering when the user selects a new version, fixing the issue in #2081.

I initially attempted to fix this by making the `editorInterface` variable properly reactive (https://github.com/dandi/dandi-archive/commit/62f2bb2ce2bb01858331a050197bd4486fd76f7b), but that ended up breaking a bunch of other things in the meditor. Eventually after fixing a long trail of new bugs, I ran into this issue again https://github.com/dandi/dandi-archive/blob/master/web/src/components/Meditor/state.ts#L4-L7 and got stuck.

I think this code could use a refactor to be less brittle and fix whatever bizarre reactivity bug is going on with the `editorInterface` state, but this simple solution will fix the critical bug for the time being.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <Meditor v-if="currentDandiset" />
+    <Meditor v-if="currentDandiset" :key="`${currentDandiset.dandiset.identifier}/${currentDandiset.version}`" />
     <v-toolbar class="grey darken-2 white--text">
       <DandisetSearchField />
       <v-pagination


web/src/components/Meditor/Meditor.vue
```diff
@@ -283,16 +283,10 @@ const invalidPermissionSnackbar = ref(false);
...
-  basicSchema,
-  basicModel,
-  basicModelValid,
-  complexSchema,
-  complexModelValidation,
-  transactionTracker,
-} = editorInterface.value;
+watchEffect(() => {
+  editorInterface.value = new EditorInterface(schema.value, model.value as DandiModel);
+});
```
waxlamp, line 288: This is the crux of the fix, right? We stop using a single, one-time-defined instance of `EditorInterface` whose components were also statically "exported" to the rest of the module, and instead refresh it every time stuff changes under us?

A note explaining this in the PR description would be helpful to our future selves.

waxlamp (commented):
Nice, simple fix it seems. Thanks!

web/src/components/Meditor/Meditor.vue
mvandenburgh: Yes, that is the crux of the fix. But, doing that broke a few other things that I am currently debugging.

mvandenburgh: I updated the description, and also changed the approach to fixing the issue (see the PR description)

jjnesbitt (approved):
The solution is very concise, and I can confirm both the reproduction of this bug, and the resolution by this PR.

PR #2130, opened 2025-01-06
Add Code of Conduct

kabilar:
For LINC, we need the Netlify Pro Plan ($19/month) in order to use Netlify to create a DNS record for subdomains, which we need for our staging account since we are using same site cookies for Neuroglancer/CloudFront and Webknossos.  We tried to do this in AWS, but there was a conflict with Netlify DNS records.  To save on costs and since we qualify, I will be applying for the [Netlify sponsored open-source plan](https://www.netlify.com/legal/open-source-policy/) which includes the features of the Pro Plan. They require a link to netlify.com on the project homepage and a code of conduct, which I have included in this pull request.  I have selected to use the [Contributor Covenant](https://www.contributor-covenant.org/).  I am suggesting to push the code of conduct upstream to DANDI, since we have an active target to push all LINC features/pieces upstream to DANDI.

DANDI uses a Netlify Free Plan and continues to have the capability of creating DNS records for subdomains.  From what we can tell from the Netlify docs, this feature used to be free, is now part of the Pro Plan, and DANDI does not have to pay for this feature since they were a previously established user.

cc @aaronkanzer

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,4 +1,4 @@
-# Contributor Covenant Code of Conduct
+# Code of Conduct
 
 ## Our Pledge
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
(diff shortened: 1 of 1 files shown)


satra (approved):

yarikoptic:
also would need to add a codespell ignore for

```
CODE_OF_CONDUCT.md:8: socio-economic ==> socioeconomic
```

I will propose to origin to adopt "no dash" form...

kabilar:
> But shouldn't it also be included/pointed to from our landing page e.g. under version where we have ATM some empty space to fill for balance

Thanks @yarikoptic.  I will submit this as a separate pull request.  Perhaps the [Terms of use](https://www.dandiarchive.org/handbook/about/terms/) and [Policies](https://www.dandiarchive.org/handbook/about/policies/) should also be added to the landing page.

PR #2136, opened 2025-01-09
Rename `The DANDI Team` to `DANDI` in footer

kabilar:
If I understand correctly, `The DANDI Team`, `DANDI`, and `DANDI Archive` are synonyms in this context so I suggest this minor change to simplify the text.

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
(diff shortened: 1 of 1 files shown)


waxlamp:
Was this discussed in a meeting? If not, I'd advise going with [Chesteron's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) and leaving this alone until we have a good enough reason to change it.

kabilar:
After discussion with Roni, I am closing this pull request.  Will revisit in the future.

PR #2139, opened 2025-01-13
Rename `staging` to `sandbox`

### LINKED ISSUE #1030: Change name of "staging" server
Users think this deployment is for staging their data before it goes to production. We need a name more reflective of its purpose (a testbed, playground, etc.).

Then we'll need to figure out where this new name needs to be used.

kabilar:
Fix #1030 

- [ ] Depends on https://github.com/dandi/dandi-infrastructure/pull/220
- With these changes, I was only renaming the user facing URLs and did not change the variables with the term `staging`.


waxlamp:
This will require a much wider effort than just this PR. We should draw up a full plan if we're serious about this now; otherwise, I'd recommend closing this PR until we are ready (so it's not here in the PR list forever).

satra:
@waxlamp - please list the things that this would require. let's get this done. it's been on the todo list for a long time.

kabilar:
Closing since @mvandenburgh and @waxlamp have taken this over.

PR #2144, opened 2025-01-17
Move to 2025 from 2024 in the footer

yarikoptic:
minor.  I would merge Monday if nobody does it before or says we must not do it.

Ideally, at somepoint, we might just want to automate that value based on the year of the last commit in the codebase.

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
-          &copy; 2019 - 2024 The DANDI Team<br>
+          &copy; 2019 - 2025 The DANDI Team<br>
           version
           <a
             class="version-link"


web/src/components/DandiFooter.vue
```diff
@@ -15,7 +15,7 @@
       </cookie-law>
       <v-row>
         <v-col offset="2">
-          &copy; 2019 - 2024 The DANDI Team<br>
+          &copy; 2019 - 2025 The DANDI Team<br>
```
kabilar, line 18: Hi @jjnesbitt, could we use the `Date` object (`new Date().getFullYear()`)?  I presume we would need to declare the constant elsewhere and inject it in the footer.  [Stackoverflow post](https://stackoverflow.com/questions/6002254/get-the-current-year-in-javascript).

jjnesbitt: We could do that, sure.

waxlamp: I think automation is overkill for this. The yearly manual update should be just fine.

waxlamp (approved):

PR #2151, opened 2025-01-23
Add `/handbook` redirect to `docs.dandiarchive.org`

kabilar:
Netlify docs on redirects: https://docs.netlify.com/routing/redirects/


web/netlify.toml
```diff
@@ -37,5 +37,11 @@ to = "/index.html"
 status = 200
 
 
+[[redirects]]
+from = "/handbook"
+to = "docs.dandiarchive.org"
+status = 200
```
satra, line 43: i don't know what status means over here, but should be a `308` i think for permanent redirect which will tell search engines and bots to forget the source if they do so.

kabilar: Thanks.  I have updated.

waxlamp: We may want `301`, since it has better compliance from clients and its limitation (that it converts other methods into `GET` upon redirect) is fine for us, since all requests should be `GET` anyway.

waxlamp (approved):

kabilar:
Thanks Roni.  Confirming that this works as intended.

PR #2170, opened 2025-02-04
Check if dandiset contains zarr assets using asset list endpoint

### LINKED ISSUE #2149: refuses to publish dandiset claiming it has a zarr whenever it (no longer) does
on https://dandiarchive.org/dandiset/000029/draft

Image

likely I had zarr before so there might be "association back" but now it is clear of zarrs 

```
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/dandisets/000029/versions/draft/assets/?page_size=1000&metadata=false' -H 'accept: application/json' | jq '.results[].zarr'
null
null
null
null
null
```

and should be publishable, so I think the check should be fixed

jjnesbitt:
Closes #2149

This PR adds a simple `zarr` boolean query param to the `/assets` endpoint, and updates the client to use that endpoint when checking for zarrs in a dandiset (to prevent publishing of zarrs).

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -313,7 +313,7 @@ watchEffect(async () => {
       { params: { zarr: true, page_size: 1 } },
     );
 
-    containsZarr.value = res.count > 0;
+    containsZarr.value = res !== null && res.count > 0;
   }
 });
 
--- a/web/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/web/src/views/DandisetLandingView/DandisetPublish.vue
@@ -305,11 +305,15 @@ const alreadyBeingPublishedError = ref(false);
 const containsZarr = ref(false);
 watchEffect(async () => {
   if (currentDandiset.value) {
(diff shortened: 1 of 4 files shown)


web/src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -305,11 +305,15 @@ const alreadyBeingPublishedError = ref(false);
...
-    containsZarr.value = zarr.count > 0;
+    const { identifier } = currentDandiset.value.dandiset;
+    const { version } = currentDandiset.value;
+    const res = await dandiRest.assets(
+      identifier,
+      version,
+      { params: { zarr: true, page_size: 1 } },
+    );
+
+    containsZarr.value = res.count > 0;
```
waxlamp, line 316: Our `asset()` function returns `null` in case of a 404 error, so I think the following adjustment will solve the typechecking error.

```suggestion
    containsZarr.value = res !== null && res.count > 0;
```

waxlamp (changes requested):
With a fix for the typechecking error, this looks good to me!

waxlamp (approved):

PR #2171, opened 2025-02-05
Support setting Content-Type when uploading a Zarr entry

### LINKED ISSUE #1870: Extend Zarr entry upload procedure to support client-assigned Content-Types
(See https://github.com/dandi/dandidav/issues/90 for motivation)

Feature request: Support an optional `content_type: str | null` field in the objects in `POST /zarr/{zarr_id}/files/` requests.  When this field is present and non-null, the signed upload URL that the Archive produces for the entry should set the Content-Type field to the given value so that `GET` requests for the resulting S3 object will be replied to with responses that have that Content-Type.

Question: When the client performs the actual upload to the signed URL, must it specify the same Content-Type header again?

jwodder:
Closes #1870.

@dandi/dandiarchive Should I add tests for this?  If so, how?

### DIFF
--- a/dandiapi/api/storage.py
+++ b/dandiapi/api/storage.py
@@ -248,16 +248,23 @@ class VerbatimNameMinioStorage(VerbatimNameStorageMixin, DeconstructableMinioSto
         else:
             return response.etag
 
-    def generate_presigned_put_object_url(self, blob_name: str, _: str, content_type: str | None) -> str:
+    def generate_presigned_put_object_url(
+        self, blob_name: str, _: str, content_type: str | None
+    ) -> str:
         # Note: minio-py doesn't support using Content-MD5 headers
 
+        headers = {}
+        if content_type is not None:
+            headers['response-content-type'] = content_type
         # storage.client will generate URLs like `http://minio:9000/...` when running in
         # docker. To avoid this, use the secondary base_url_client which is configured to
         # generate URLs like `http://localhost:9000/...`.
-        return self.base_url_client.presigned_put_object(
+        return self.base_url_client.get_presigned_url(
+            method='PUT',
             bucket_name=self.bucket_name,
             object_name=blob_name,
             expires=timedelta(seconds=600),  # TODO: proper expiration
+            response_headers=headers,
         )
 
     def generate_presigned_head_object_url(self, key: str) -> str:
--- a/dandiapi/api/storage.py
+++ b/dandiapi/api/storage.py
@@ -248,7 +248,7 @@ class VerbatimNameMinioStorage(VerbatimNameStorageMixin, DeconstructableMinioSto
         else:
             return response.etag
 
-    def generate_presigned_put_object_url(self, blob_name: str, _: str, content_type: str) -> str:
+    def generate_presigned_put_object_url(self, blob_name: str, _: str, content_type: str | None) -> str:
         # Note: minio-py doesn't support using Content-MD5 headers
 
         # storage.client will generate URLs like `http://minio:9000/...` when running in
--- a/dandiapi/api/storage.py
+++ b/dandiapi/api/storage.py
@@ -182,15 +182,20 @@ class VerbatimNameS3Storage(VerbatimNameStorageMixin, TimeoutS3Storage):
                 return etag[1:-1]
             return etag
 
-    def generate_presigned_put_object_url(self, blob_name: str, base64md5: str) -> str:
+    def generate_presigned_put_object_url(
+        self, blob_name: str, base64md5: str, content_type: str | None
+    ) -> str:
+        params = {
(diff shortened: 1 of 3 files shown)


dandiapi/api/storage.py
```diff
@@ -243,7 +248,7 @@ def etag_from_blob_name(self, blob_name) -> str | None:
         else:
             return response.etag
 
-    def generate_presigned_put_object_url(self, blob_name: str, _: str) -> str:
+    def generate_presigned_put_object_url(self, blob_name: str, _: str, content_type: str) -> str:
```
yarikoptic, line 251:
```suggestion
    def generate_presigned_put_object_url(self, blob_name: str, _: str, content_type: str | None) -> str:
```

?

dandiapi/zarr/models.py
```diff
@@ -88,7 +88,9 @@ def s3_path(self, zarr_path: str) -> str:
 
     def generate_upload_urls(self, path_md5s: list[dict]):
         return [
-            self.storage.generate_presigned_put_object_url(self.s3_path(o['path']), o['base64md5'])
+            self.storage.generate_presigned_put_object_url(
+                self.s3_path(o['path']), o['base64md5'], content_type=o['content_type']
+            )
```
yarikoptic, line 93: while at it, should may be `path_md5s` to be renamed (it is internal, should be ok AFAIK) into e.g. `path_objs` or `path_recs` to signal that it is not just a list (or mapping) of paths to md5s?

dandiapi/api/storage.py
yarikoptic: re note below., according to https://min.io/docs/minio/linux/developers/python/API.html it seems it is possible to provide content-type with minio

```
# Get presigned URL string to upload 'my-object' in
# 'my-bucket' with response-content-type as application/json
# and one day expiry.
url = client.get_presigned_url(
    "PUT",
    "my-bucket",
    "my-object",
    expires=timedelta(days=1),
    response_headers={"response-content-type": "application/json"},
)
```

that should open the ways to test that it does get set in https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/zarr/tests/test_zarr_upload.py

yarikoptic:
> @dandi/dandiarchive Should I add tests for this? If so, how?

now that you added to minio interface, I think test should be adjusted and/or added to verify that functioning to https://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/zarr/tests/test_zarr_upload.py

jwodder:
@yarikoptic None of the tests in that file both upload to a Zarr and request the uploaded entry from minio, so I'm not sure exactly how to proceed.

@dandi/dandiarchive Please advise.

waxlamp (commented):
Overall, this looks good to me. Jake will advise on how to go about testing.

jwodder:
I tested this in the client using [this code](https://github.com/dandi/dandi-cli/pull/1576), with Docker Compose adjusted to point to a local build of this PR.  I determined that the client does need to set the "Content-Type" header when uploading to the signed URL, but I also found that sending the "Content-Type" to the archive for inclusion in the signed URL is unnecessary (for minio, at least).  If we can confirm that the same thing happens on S3, that would make this PR superfluous.

jjnesbitt:
> I tested this in the client using [this code](https://github.com/dandi/dandi-cli/pull/1576), with Docker Compose adjusted to point to a local build of this PR. I determined that the client does need to set the "Content-Type" header when uploading to the signed URL, but I also found that sending the "Content-Type" to the archive for inclusion in the signed URL is unnecessary (for minio, at least). If we can confirm that the same thing happens on S3, that would make this PR superfluous.

Regarding testing, Minio and S3 differ in some ways, which is part of the reason there can't really be good tests for these features. This is one of the friction points between our local and prod deployments.

Regarding the feature itself, according to my research, if `Content-Type` is provided as a param to the pre-signed URL generation, then the client using the pre-signed URL for upload _must_ include that header, to "match" the pre-signed URL (as is the case with other params like `ContentMD5`). However, if that param is _not_ included in the pre-signed URL params, the client may supply that header themselves, and it will be honored.

I'm unable to test this in S3 myself at the moment, but what I've found would seem to match the results of your test. So if that truly is the case, this is sort of a policy question, more than anything else. Do we want to require `Content-Type` to be included alongside other params like `ContentMD5` at URL generation time (and therefore enforced once the URL is "wielded")? Or should we leave that control up to the client?

To me, I agree it seems entirely superfluous to include that behavior here, since in either case, the responsibility of providing the `Content-Type` header lies with the client. The only difference is that with this code change, the client would need to additionally supply it at pre-signed URL generation time as well.

jwodder:
I was able to confirm that setting the Content-Type in the signed URL is unnecessary on S3.  Closing this PR as superfluous.

PR #2175, opened 2025-02-06
Fix dandiset ordering icons, default to descending

jjnesbitt:
The existing icons for sorting on any dandiset list page is backwards of what you'd expect (arrow pointing up is actually "descending"). This PR fixes that.

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -51,11 +51,11 @@
         >
           {{ option.name }}
           <v-icon right>
-            <template v-if="sortDir === 1 || sortOption !== i">
-              mdi-sort-ascending
+            <template v-if="sortDir === -1 || sortOption !== i">
+              mdi-sort-variant
             </template>
             <template v-else>
-              mdi-sort-descending
+              mdi-sort-reverse-variant
             </template>
           </v-icon>
         </v-chip>
@@ -204,7 +204,7 @@ export default defineComponent({
         sortDir.value *= -1;
       } else {
         sortOption.value = index;
-        sortDir.value = 1;
+        sortDir.value = -1;
       }
 
       page.value = 1;


web/src/components/DandisetsPage.vue
```diff
@@ -204,7 +207,7 @@ export default defineComponent({
         sortDir.value *= -1;
       } else {
         sortOption.value = index;
-        sortDir.value = 1;
+        sortDir.value = -1;
```
waxlamp, line 210: Nothing wrong here, once I decoded what this function is doing... but boy is it a confusing function! No need to do anything in this PR, but I feel like this function could use a comment explaining what it's doing.

web/src/components/DandisetsPage.vue
```diff
@@ -51,7 +51,10 @@
         >
           {{ option.name }}
           <v-icon right>
-            <template v-if="sortDir === 1 || sortOption !== i">
+            <!-- For some reason, MDI's icons are represented backwards of what you'd expect.
+             Descending has an arrow pointing upwards, which is usually meant to represent ascending.
+             We intentionally switch the values here, this is not a bug. -->
```
waxlamp, line 56: Oh goodness, I think I figured it out. Note that the icons both have three bars of different lengths representing data items. The lines get longer as you move down... so "ascending" is an arrow pointing down (towards the longer lines) while "descending" (of course) points up. That seems like bad icon design to me.

We could use the "wrong" icons as in this PR, or we could consider using these two ([ascending](https://pictogrammers.com/library/mdi/icon/sort-variant/), [descending](https://pictogrammers.com/library/mdi/icon/sort-reverse-variant/)) that don't feature the "backward" arrow. (I am leaning towards these "variant" icons myself, because using the "bad" icons, and putting the "bad" icons "backwards" both feel like bad options.)

waxlamp (commented):
This does indeed fix the perceived problem, but see my comment below on the different icons for some further discussion.

web/src/components/DandisetsPage.vue
jjnesbitt: Well I've changed the icons to your suggestion, as I think it does improve readability, although notably they are still backwards. `sort-variant`, which "should" correspond to ascending sort, shows the largest item at the top. However, visually this would correspond with descending sort. You can check out the deploy preview to see it, IMO I don't see many people preferring it the other way around.

web/src/components/DandisetsPage.vue
```diff
@@ -51,11 +51,14 @@
         >
           {{ option.name }}
           <v-icon right>
-            <template v-if="sortDir === 1 || sortOption !== i">
-              mdi-sort-ascending
+            <!-- For some reason, MDI's icons are represented backwards of what you'd expect.
+             Descending has an arrow pointing upwards, which is usually meant to represent ascending.
+             We intentionally switch the values here, this is not a bug. -->
```
waxlamp, line 56: I think we can get rid of the comment now. I know the variant icons still have a strange naming, but I think visually they make sense.

```suggestion
```

waxlamp (approved):

PR #2177, opened 2025-02-10
Metadata correction

candleindark:
This PR provides a solution to correct the corruptions in metadata documented in https://github.com/dandi/dandi-schema/issues/276.

The solution is implemented in two parts.
1. A general command, `correct_metadata`, to correct dandiset metadata.
2. A specific helper function which the general command calls to do the actual correction.

Extensive tests are provided for the helper function and its supporting function. However, because I am not familiar with this repo and Django in general, I am not able to provide tests for the command which interacts with the database. Advice and additional tests are very much appreciated.

The command can be run on a targeted dandiset at a particular version and run on all versions of all dandisets. Running on all dandiset versions will only correct the corrupted dandisets. If running only on targeted dandiset version is preferable, please let me know, and I will provide the list of corrupted dandiset versions. (Additionally, would changing the interface of the command to a file consisting of corrupted dandiset versions be better?)


dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if changes:
+        # Apply the changes to the database atomically
+        with transaction.atomic():
+            for ver, meta_corrected in changes:
+                ver.metadata = meta_corrected
+                ver.save()  # TODO: should I use `save()` in this case? It has some custom logic.
+
+        for ver, _ in changes:
+            write_manifest_files.delay(ver.id)  # TODO: Please check if this is needed
```
asmacdo, line 72: What happens if any of the file writes fail or time out? https://github.com/dandi/dandi-archive/blob/7728016558ab7ee77ee06534d461b119a200cbc4/dandiapi/api/tasks/__init__.py#L58-L62

Maybe we should at least keep the AsyncResults and monitor that they go as planned?

asmacdo (commented):
a couple of questions to start off with-- I'll have another look tomorrow.

dandiapi/api/management/commands/correct_metadata.py
candleindark: I actually don't know if I need to call `write_manifest_files.delay(ver.id)` at all. I was trying to mimic https://github.com/dandi/dandi-archive/blob/7728016558ab7ee77ee06534d461b119a200cbc4/dandiapi/api/management/commands/migrate_published_version_metadata.py#L52, but https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/management/commands/migrate_version_metadata.py didn't call `write_manifest_files()` at all. 

> Maybe we should at least keep the AsyncResults and monitor that they go as planned?

I think the answer is related to how important that statement gets executed successfully. In the example above, its execution status is not checked. If a success execution is important, I guess I can't interact with the DB the current way.

@dandi/dandiarchive

dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+    Correct corruptions in JSON objects with the `"schemaKey"` of `"Affiliation"`.
+
+    :param meta: The Dandiset metadata instance potentially containing the objects to be corrected.
+    :return: If there is correction to be made, return the corrected metadata; otherwise, return
+        `None`.
+
+    Note: This function corrects the corruptions described in
+        https://github.com/dandi/dandi-schema/issues/276
+    """
+    unwanted_fields = ['contactPoint', 'includeInCitation', 'roleName']
```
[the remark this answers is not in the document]
candleindark: In concept, you are right. I considered do that as well. However, I want to be careful for now. I want to remove only the ones that I know are causing problems. I don't want to remove anything that we haven't analyzed previously.

dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+import djclick as click
+
+from dandiapi.api.models import Version
+from dandiapi.api.tasks import write_manifest_files
+
+
+@click.command(
+    help='Correct corrupted metadata. If `--dandiset` and `--dandiset-version` are provided, '
+    'apply the correction to the specified Dandiset version. Otherwise, apply the correction to '
+    'all Dandiset versions.'
```
waxlamp, line 19: I'd like to reverse the default behavior: by default, require a Dandiset and version; to override, include a flag `--all` that would apply the change to all versions.

waxlamp (changes requested):
Thank you, @candleindark! One thing to change listed below. And I've asked @jjnesbitt to provide a more comprehensive review on this R as well.

dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+    # version to be changed and the new metadata
+    changes: list[tuple[Version, dict]] = []
+    for ver in vers:
+        try:
+            meta_corrected = correct_func(ver.metadata)
+        except Exception as e:
+            click.echo(
+                f'{ver.dandiset.identifier}/{ver.version}: Failed to correct metadata with '
+                f'`{correct_func.__name__}()`'
+            )
```
jjnesbitt, line 57: Have you actually observed any exceptions raised when performing the metadata correction?

dandiapi/api/management/commands/correct_metadata.py
candleindark: Done.

dandiapi/api/management/commands/correct_metadata.py
candleindark: No, I have not. The exception handling is put in as a protective programming measure, particularly to handle situation when `correct_func` is assigned to some other function, possibly in the future.

candleindark:
@jjnesbitt I have put in changes requested by @waxlamp regarding the command line interface. With the latest push, two unrelated tests failed. Can they be related to the latest changes in dandi-cli?

jjnesbitt:
I've pushed a commit which aims to simplify the logic in this management command. Namely:

* Remove the configuration of the correcting function.
	* We have no real use case for this yet, and no way to change this function is provided.
* Removal of verbose error handling logic
	* Along the same lines as the above point, since this is a single function, this is largely unnecessary. We have tests, and if the correcting function fails for some reason, the error will simply bubble up.
* Only support correction of the `Affiliation` schema key
	* Similarly again, we only have the use case of `Affiliation` at the moment, so all of the configuration surrounding other schema keys is unnecessary at the moment.
* Remove `find_objs` tests against generalized schema key
* Write manifest files synchronously
	*  Since we are not in a request context, it's better to write the manifest files to S3 synchronously, so that we can be sure it succeeds.
* Use transactions to couple logic of metadata save and manifest file generation


IMO this is ready, maybe @waxlamp should take another look.

waxlamp:
As per the meeting of 2025-03-10 (involving @yarikoptic, @candleindark, @jjnesbitt, @kabilar, and @waxlamp), we decided to reduce the scope of this change to just the draft versions; @jjnesbitt will work on adjusting the PR to that narrowed scope.

@candleindark, @yarikoptic: are we planning to release a new schema version that has restrictions on extra fields? If we apply the metadata correction to draft versions, then I suppose whether or not such a schema is released, there will not be any further validation errors.

candleindark:
> @candleindark, @yarikoptic: are we planning to release a new schema version that has restrictions on extra fields?

Yes, some minor changes to the schema will be introduced by this [PR](https://github.com/dandi/dandi-schema/pull/266)

> If we apply the metadata correction to draft versions, then I suppose whether or not such a schema is released, there will not be any further validation errors.
 
I can't say that after the corrections to the draft version that there will be no more validation errors in those versions. I can only say that if you make the corrections to the draft versions, there will not be validation errors in those version due to that particular change brought by that [PR](https://github.com/dandi/dandi-schema/pull/266) 

https://github.com/dandi/dandi-schema/pull/266#issuecomment-2603164223 provides an analysis of validation errors in the metadata instances caused by the changes in the PR but not an analysis of validation errors in the metadata instances in general.

waxlamp:
> > If we apply the metadata correction to draft versions, then I suppose whether or not such a schema is released, there will not be any further validation errors.
> 
> I can't say that after the corrections to the draft version that there will be no more validation errors in those versions. I can only say that if you make the corrections to the draft versions, there will not be validation errors in those version due to that particular change brought by that [PR](https://github.com/dandi/dandi-schema/pull/266)
> 

Sorry, this was what I meant.

yarikoptic:
do you think you would have time to work out solution for #2224 (it is an issue) or should someone else try to approach it to facilitate a potentially (not necessarily "likely" ;) ) faster resolution?

jjnesbitt:
> do you think you would have time to work out solution for #2224 (it is an issue) or should someone else try to approach it to facilitate a potentially (not necessarily "likely" ;) ) faster resolution?

I'm already working on #2224, it shouldn't take much time to complete.

jjnesbitt:
@waxlamp @mvandenburgh This is ready to go now.

dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,123 @@
...
+    if not apply_to_all:
+        dandiset = typing.cast(str, dandiset)
+        vers = vers.filter(dandiset=int(dandiset), version='draft')
+
+    if not vers.exists():
+        click.echo('No matching versions found')
+        return
+
+    # For each version, find and fix metadata corruptions, along with saving out manifest files
+    for ver in vers:
```
mvandenburgh, line 55:
```suggestion
    for ver in vers.iterator():
```

dandiapi/api/management/commands/correct_metadata.py
```diff
@@ -0,0 +1,123 @@
...
+                admin=True,
+                description='Apply metadata correction from https://github.com/dandi/dandi-schema/issues/276',
+            )
+
+    # Remaining check is not needed since no data was modified
+    if check:
+        return
+
+    # If we find any un-fixed instances, raise exception
+    remaining = [ver for ver in vers if correct_affiliation_corruption(ver.metadata) is not None]
```
mvandenburgh, line 86:
```suggestion
    remaining = [ver for ver in vers.iterator() if correct_affiliation_corruption(ver.metadata) is not None]
```

mvandenburgh (changes requested):
Just some suggested optimizations to memory usage. It's _probably_ not needed due to the number of `Versions` not being super high, but the `heroku run` dynos have pretty limited memory so it seems logical to do.

Otherwise, LGTM

mvandenburgh (approved):

yarikoptic:
So this was merged and now we got the command. Who is allowed to run it, as could I ?

PR #2193, opened 2025-02-21
Implement "atpath" asset endpoint

### LINKED ISSUE #2172: Implement "atpath" endpoint
See [design doc](https://github.com/dandi/dandi-archive/blob/master/doc/design/atpath.md) for motivation and design details.

Closes #1837.

jjnesbitt:
Closes #2172 

This implements the design doc laid out in #2155 ([link](https://github.com/dandi/dandi-archive/blob/master/doc/design/atpath.md)).

There are some additional changes included in this PR, that may seem unrelated, but were necessary for this feature's functionality. One such change is the slight restructuring of `urls.py`, which was done in order to mount two separate swagger pages, with their own respective visible endpoints. This is actually an overall code quality enhancement, as before, our URL and swagger configuration was very ad hoc.

@jwodder The only deviation from that design doc is that the endpoint is instead hosted at `api/webdav/assets/atpath`, with `dandiset_id` and `version_id` provided as query arguments. This is done to separate concerns from the public API, and what's needed for webdav. Let us know if you find any issues.


dandiapi/api/views/webdav/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    query_serializer.is_valid(raise_exception=True)
+
+    params = query_serializer.validated_data
+    version = get_object_or_404(
+        Version.objects.select_related('dandiset'),
+        dandiset_id=int(params['dandiset_id']),
+        version=params['version_id'],
+    )
+
+    if not is_dandiset_owner(version.dandiset, request.user):
```
jwodder, line 76: How does this work for unauthenticated requests?  They need to be supported, as that's what `dandidav` uses.

jjnesbitt:
I see there is a single test failure, which seems unrelated, but I'll investigate.

dandiapi/api/views/webdav/__init__.py
jwodder: `dandidav` only serves Dandisets that it can see via unauthenticated requests, which should be the OPEN ones, yes.

dandiapi/api/tests/test_audit.py
```diff
@@ -110,10 +110,11 @@ def user_info(u):
...
     verify_model_properties(rec, alice)
-    assert rec.details == {
-        'added_owners': [user_info(u) for u in [bob, charlie]],
-        'removed_owners': [user_info(u) for u in [alice]],
-    }
+
+    assert sorted(rec.details['added_owners'], key=lambda x: x['username']) == sorted(
+        [user_info(u) for u in [bob, charlie]], key=lambda x: x['username']
+    )
+    assert rec.details['removed_owners'] == [user_info(alice)]
```
mvandenburgh, line 117: Why did this change? Seems unrelated to this PR.

dandiapi/api/models/asset_paths.py
```diff
@@ -4,7 +4,7 @@
 
 
 class AssetPath(models.Model):
-    path = models.CharField(max_length=512)
+    path = models.CharField(max_length=512, db_collation='C')
```
mvandenburgh, line 7: I think we should have a comment similar to the one in `test_asset_path_ordering` describing why `db_collation` is set here.

dandiapi/api/views/webdav/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        return qs.none()
+
+    # Since path+version combinations are unique, we know we've matched exactly one path.
+    # Now see if we should extend this with it's children
+    if asset_path.asset is not None or not children:
+        return qs
+
+    # Now we know path is a folder, and we should show its children
+    children_paths = (
+        get_path_children(asset_path).select_related(None).select_related(*select_related_clauses)
```
mvandenburgh, line 53: Is the intent of the first `select_related` to "remove" the `select_related` that's added by `get_path_children`? And if so, have you tested that it works? I've never seen this pattern before, but it makes sense to me.

mvandenburgh (changes requested):
Just a couple of minor things.

dandiapi/api/views/webdav/__init__.py
jjnesbitt: Yes, that's the intention, and it does work as expected. It is mentioned [here](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#:~:text=without_relations%20%3D%20queryset.select_related(None)) in the django docs.

waxlamp (approved):

jwodder:
@jjnesbitt I've noticed that making an "atpath" request with a nonexistent path returns an empty list, yet my design doc states that it should result in a 404.  Was this change intentional?

jjnesbitt:
> @jjnesbitt I've noticed that making an "atpath" request with a nonexistent path returns an empty list, yet my design doc states that it should result in a 404. Was this change intentional?

My bad, it seems I misread that part of the design doc. I'll make a follow up PR.

PR #2198, opened 2025-03-05
Update `atpath` endpoint implementation

### LINKED ISSUE #2197: "atpath" implementation does not support paths ending in a slash
```
$ curl -fsSL https://api.dandiarchive.org/api/webdav/assets/atpath'?dandiset_id=000027&version_id=draft&children=true&path=sub-RAT123' | jq .
{
  "count": 2,
  "next": null,
  "previous": null,
  "results": [
    {
      "type": "folder",
      "resource": {
        "path": "sub-RAT123",
        "total_assets": 1,
        "total_size": 18792
      }
    },
    {
      "type": "asset",
      "resource": {
        "asset_id": "838bab7b-9ab4-4d66-97b3-898a367c9c7e",
        "path": "sub-RAT123/sub-RAT123.nwb",
        "created": "2023-03-02T22:10:45.985334Z",
        "modified": "2023-03-02T22:10:46.064360Z",
        "zarr": null,
        "blob": "2dbaf0fd-5003-4a0a-b4c0-bc8cdbdb3826",
        "size": 18792
      }
    }
  ]
}
$ curl -fsSL https://api.dandiarchive.org/api/webdav/assets/atpath'?dandiset_id=000027&version_id=draft&children=true&path=sub-RAT123/' | jq .
{
  "count": 0,
  "next": null,
  "previous": null,
  "results": []
}
```

While [the design doc](https://github.com/dandi/dandi-archive/blob/master/doc/design/atpath.md) does not explicitly state that trailing forward slashes must be supported, it is somewhat implied by the line "If the path ends in a forward slash and no folder exists at the given path, ...".

jjnesbitt:
Follow on to #2193 

This includes the following fixes that were erroneously not included in the original PR:
* Return 404 if no resource found at path
* Return 404 if trailing slash supplied and asset is found

Closes #2197

### DIFF
--- a/dandiapi/api/views/webdav/__init__.py
+++ b/dandiapi/api/views/webdav/__init__.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING
 from django.shortcuts import get_object_or_404
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import api_view
-from rest_framework.exceptions import PermissionDenied
+from rest_framework.exceptions import NotFound, PermissionDenied
 
 from dandiapi.api.asset_paths import get_path_children
 from dandiapi.api.models.asset_paths import AssetPath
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/webdav/__init__.py
```diff
@@ -80,8 +80,23 @@ def atpath(request: Request):
     if dandiset.embargoed and not is_dandiset_owner(dandiset, request.user):
         raise PermissionDenied
 
-    qs = get_atpath_queryset(version=version, path=params['path'], children=params['children'])
+    # This endpoint requires an ending slash for folders,
+    # but the atpath implementation requires the opposite
```
jwodder, line 84: Was this comment copied from somewhere else?  This *is* the atpath endpoint, and it doesn't require (or prohibit) an ending slash.

jjnesbitt: Good catch, that was a typo that I've now fixed. I meant to refer to the `AssetPath` implementation, which is what this endpoint utilizes.

jwodder (approved):

Issue #1272, opened 2022-09-08
provide robots.txt for api.dandiarchive.org disabling all compliant bots?


yarikoptic:
I do not think there is a point in allowing bots to traverse api.dandiarchive.org in its entirety.  I just spotted today in the logs entries such as 

```
(dandisets) dandi@drogon:/mnt/backup/dandi/heroku-logs/dandi-api$ grep petalbot 20220908-*.log | head -n 1
20220908-0301.log:2022-09-08T07:20:20.557120+00:00 app[web.1]: 10.1.62.48 - - [08/Sep/2022:07:20:20 +0000] "GET /api/dandisets/000021/versions/ HTTP/1.1" 200 12478 "https://api.dandiarchive.org/api/dandisets/000021/versions/draft/assets/paths/" "Mozilla/5.0 (Linux; Android 7.0;) AppleWebKit/537.36 (KHTML, like Gecko) Mobile Safari/537.36 (compatible; PetalBot;+https://webmaster.petalsearch.com/site/petalbot)"
```
it is not that many of those requests but they do traverse the website slowly and it seems even asking for robots.txt which we do not provide.  Here are the hits for today so far with most popular "paths" on the server presumably from announced bots:

```
(dandisets) dandi@drogon:/mnt/backup/dandi/heroku-logs/dandi-api$ grep bot 20220908-*.log | awk '{print $9;}' | sort | uniq -c | sort -n | nl | tail
    93        3 /static/drf-yasg/swagger-ui-dist/swagger-ui-standalone-preset.js
    94        3 /swagger/?format=openapi
    95        4 /api/dandisets/?page=1&page_size=1&ordering=-modified&draft=true
    96        4 /api/stats/
    97        4 /swagger/
    98        8 /
    99       10 /api/info/
   100       10 /robots.txt
   101       13 dyno=web.1
   102       18
```
and for petalbot 

```
(dandisets) dandi@drogon:/mnt/backup/dandi/heroku-logs/dandi-api$ grep petalbot 20220908-*.log | awk '{print $9;}' | sort | uniq -c | sort -n | nl | tail
    68        2 /api/dandisets/000238/users/
    69        2 /api/dandisets/000238/versions/
    70        2 /api/dandisets/000238/versions/draft/assets/paths/?path_prefix=&page=1&page_size=15
    71        2 /api/dandisets/000238/versions/draft/info/
    72        2 /api/dandisets/000239/versions/draft/assets/?page=3
[...]

PR #2200, opened 2025-03-06
Allow robots on web frontend, besides /dandiset/

### LINKED ISSUE #2199: We need less restrictive robots.txt
re forward of @satra 

ATM it is 

```
User-agent: *
Disallow: /
```

and that is why google says

Image

Should we just disallow `/dandiset/` ?  I will submit PR for that so we "could see"

yarikoptic:
I left api still with full Disallow, as it was the original concern here in #1272. It was unclear though on how it got to API -- via web UI or just somehow traversing api interfaces?

Original intent here -- is to make google "see" our website. 
- Hopefully closes #2199

### DIFF
--- a/web/public/robots.txt
+++ b/web/public/robots.txt
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow: /
+Disallow: /dandiset/


satra:
do we need to disable the dandiset as well? wouldn't it be good for google to index all dandisets? api is still disabled by the other part of the change in #1272

yarikoptic:
yeah -- that's a question, since we do want to disable https://dandiarchive.org/dandiset/..../files endpoint really when web frontend would be the one inquiring API, and in theory it is a different website and indeed it should check its robots...  I guess we would not discover unless we try. Agree ? ;) let's do that..

yarikoptic:
nah -- it would then be useless alone since `/dataset/` itself would need to inquire API to provide listing... I think we should allow some specific end point(s) only while disallowing "heavy" ones (listing of contents)...  Let's first hear from @dandi/archive-maintainers folks

web/public/robots.txt
```diff
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow: /
+Disallow: /dandiset/
```
yarikoptic, line 2: ok, let's just allow everything on frontend

```suggestion
Allow: /
```

yarikoptic:
@satra - I agreed that on this level let's just allow everything.  @waxlamp please review/bless.  Then we can see how much more google would discover and see if/how to tune up API robots.txt

waxlamp (approved):

Issue #605, opened 2021-11-16
Make DANDI admin email a Django setting


mvandenburgh:
See discussion at https://github.com/dandi/dandi-api/pull/594#issuecomment-969718985

The email address that admin-only messages are sent to should be a configurable Django setting instead of being hardcoded.

yarikoptic:
I guess it is still the case?

```
❯ git grep DJANGO.*@
.github/workflows/backend-ci.yml:          DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
.github/workflows/backend-ci.yml:          DJANGO_DANDI_DEV_EMAIL: test@example.com
.github/workflows/frontend-ci.yml:      DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
.github/workflows/frontend-ci.yml:      DJANGO_DANDI_DEV_EMAIL: test@example.com
.github/workflows/frontend-ci.yml:      DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
.github/workflows/frontend-ci.yml:      DJANGO_DANDI_DEV_EMAIL: test@example.com
dev/.env.docker-compose:DJANGO_DATABASE_URL=postgres://postgres:postgres@postgres:5432/django
dev/.env.docker-compose:DJANGO_DANDI_DEV_EMAIL=test@example.com
dev/.env.docker-compose-native:DJANGO_DATABASE_URL=postgres://postgres:postgres@localhost:5432/django
dev/.env.docker-compose-native:DJANGO_DANDI_DEV_EMAIL=test@example.com
```

yarikoptic:
ok, let's try to get this addressed

PR #2205, opened 2025-03-06
Move admin email into Django setting

asmacdo:
Fixes https://github.com/dandi/dandi-archive/issues/605

This will need a new value DJANGO_DANDI_ADMIN_EMAIL to be added to the production deployment.

I hand tested this by adding (uncommited) log messages to the `/api/users/me/` endpoint, but let me know if there's a better way to try this out.


asmacdo:
Is there a process for using a dandi-cli PR for those integration tests here?

jjnesbitt:
Thanks for this contribution. This will depend on https://github.com/dandi/dandi-infrastructure/pull/218 being merged, since that will add the required values to our heroku deploy. Once that's merged, this can follow suit (assuming no issues arise).

jjnesbitt:
> Is there a process for using a dandi-cli PR for those integration tests here?

Afaik, not in any way that's worthwhile. I think in this case the best course of action is just to allow them to fail for this PR, and merge in the order:

1. Infrastructure
2. API
3. CLI

jjnesbitt (approved):

mvandenburgh:
> > Is there a process for using a dandi-cli PR for those integration tests here?
> 
> Afaik, not in any way that's worthwhile. I think in this case the best course of action is just to allow them to fail for this PR, and merge in the order:
> 
> * [x]  Infrastructure
> * [ ]  API
> * [ ]  CLI

Wouldn't merging the CLI PR before the API PR solve the issue? I'm assuming adding an extra env var doesn't cause the CLI tests to fail.

jjnesbitt:
> Wouldn't merging the CLI PR before the API PR solve the issue? I'm assuming adding an extra env var doesn't cause the CLI tests to fail.

I was under the impression that it wouldn't be possible either way, due to the specifics of how the integration tests are run, but I may have been mistaken about that. Regardless, it doesn't hurt to try to merge the CLI PR first, the worst case scenario is the tests still fail. I'll update that comment.

yarikoptic:
ok we merged (and should get released)
- https://github.com/dandi/dandi-cli/pull/1588

mvandenburgh:
I reran the dandi-cli tests, and they're passing now.

jjnesbitt (approved):

PR #2211, opened 2025-03-12
Fix lingering issues with embargoed zarr implementation

jjnesbitt:
There were a couple of things left unaccounted for in the original [embargoed zarr PR](https://github.com/dandi/dandi-archive/pull/2069).

1. There was no ability to actually create embargoed zarrs, as the zarr create endpoint never passed a value to the zarr `embargoed` field, defaulting to false. Unlike with AssetBlobs, this field should correspond _exactly_ with the embargo status of a dandiset, and so this PR removes it entirely in favor of a property on the `ZarrArchive` model. 
	* From what I can tell, no "open" zarrs have been created on an embargoed dandiset due to this bug, but nonetheless, any such occurrences would be fixed by this change, as they will now assume the dandiset embargo status.
3. Creating a zarr on a dandiset that is in the middle of unembargo should be disallowed, yet there was no check for that. This PR adds such check.


mvandenburgh:
> Unlike with AssetBlobs, this field should correspond _exactly_ with the embargo status of a dandiset, and so this PR removes it entirely in favor of a property on the `ZarrArchive` model.

Just to clarify, is this true because zarrs are not de-duplicated, but asset blobs are? i.e. zarrs are always one-to-one with versions, while asset blobs might be one-to-many?

jjnesbitt:
> Just to clarify, is this true because zarrs are not de-duplicated, but asset blobs are? i.e. zarrs are always one-to-one with versions, while asset blobs might be one-to-many?

Correct, except that zarrs are one-to-one with dandisets, not versions (they're also technically many-to-one, since a dandiset can contain multiple zarrs).

mvandenburgh (approved):
Makes sense to me.

kabilar:
> 1. There was no ability to actually create embargoed zarrs, as the zarr create endpoint never passed a value to the zarr `embargoed` field, defaulting to false. Unlike with AssetBlobs, this field should correspond _exactly_ with the embargo status of a dandiset, and so this PR removes it entirely in favor of a property on the `ZarrArchive` model.
>    
>    * From what I can tell, no "open" zarrs have been created on an embargoed dandiset due to this bug, but nonetheless, any such occurrences would be fixed by this change, as they will now assume the dandiset embargo status.

Hi @jjnesbitt, thanks for being on top of this.  A user we asked to beta test this feature emailed us on March 11 (1 day prior to this pull request being opened) stating that they had successfully uploaded a Zarr to an embargoed Dandiset.  (I just forwarded you the email.)  So not sure why they were able to upload the Zarr if these changes hadn't yet been implemented.  Thanks.

jjnesbitt:
> Hi @jjnesbitt, thanks for being on top of this. A user we asked to beta test this feature emailed us on March 11 (1 day prior to this pull request being opened) stating that they had successfully uploaded a Zarr to an embargoed Dandiset. (I just forwarded you the email.) So not sure why they were able to upload the Zarr if these changes hadn't yet been implemented. Thanks.

Zarr upload to embargoed dandisets _was_ working, in the sense that you could create a zarr in an embargoed dandiset and point an asset to it. However, prior to this PR, the zarr model also had an `embargoed` flag, which was there to control if it is publicly visible or not (from the `/zarrs` endpoint or otherwise). The error was that this flag was always being set to `False`, without ever taking into consideration the embargo status of the dandiset it's contained within. So the consequence was that for a small amount of time, the zarr that the user uploaded to the embargoed dandiset was likely visible through the `/zarrs` endpoint, even though it should have been embargoed.

PR #2212, opened 2025-03-13
enh: inject datacite metadata as schema.org for dataset search

### LINKED ISSUE #752: Better indexing for search engines
At present the web ui is not well indexed by google. i've submitted a request. other options are:
1. to prerender:  i can enable prerendering on netlify or we can run this https://prerender.io/ for more custom capability
2. submit a sitemap separately. i'm thinking of generating this automatically using the datalad job for dandiarchive.org
 
also we need to inject the dandiset metadata into the html of the dandiset landing page so that google dataset search can pick it up.

satra:
This adds a little snippet to inject datacite metadata in schema.org for google dataset search.

closes #752


satra:
@mvandenburgh - in the sandbox environment, i’d like to use a different prefix for doi (`https://handle.stage.datacite.org/`) but should use the normal resolver in production (`https://doi.org/). where should i configure this prefix? this would be in the ui for now as the metadata doesn’t store the full url.

satra:
trying a check, but i would like to not use the hardcoded ui url so that this can be used in other sandboxes too.

mvandenburgh:
@satra I pushed some changes to move the DOI server into an env var + some other fixes. It works locally for me, but let's see if it works on the deploy preview.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -399,6 +400,17 @@ export default defineComponent({
       navigator.clipboard.writeText(value === 'dandiID' ? version : `https://doi.org/${meta.value?.doi}`);
     }
 
+    watchEffect(async () => {
```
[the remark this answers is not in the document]
mvandenburgh: The `watchEffect` callback will run whenever a reactive dependency changes (https://vuejs.org/guide/essentials/watchers.html#watcheffect). In this case, the only reactive dependency is `meta` (i.e., the currently displayed dandiset's metadata). So whenever the dandiset changes, it will send a new request to the DOI server.

satra:
i filed a bug about the handle server here: https://github.com/datacite/datacite/issues/2311 (i went and registered one of the test dois)

is there a way to try the ui also with the production server? i.e. could we have both previews?

web/netlify.toml
```diff
@@ -9,6 +9,7 @@ NODE_VERSION = "20"
 VITE_APP_OAUTH_API_ROOT = "https://api-staging.dandiarchive.org/oauth/"
 VITE_APP_OAUTH_CLIENT_ID = "Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl"
 VITE_APP_DANDI_API_ROOT = "https://api-staging.dandiarchive.org/api/"
+VITE_APP_DOI_SERVER = "https://handle.stage.datacite.org/"
```
yarikoptic, line 12: FTR note to myself: somehow Findable DOIs gather this URL for them on test fabric, e.g. here is Findable and draft to compare:

image

satra: yes either registered or findable will generate that handle, but right now the handle doesn't do what the doi version does.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -399,6 +400,17 @@ export default defineComponent({
...
+      // Inject datacite metadata into the page
+      if (meta.value?.doi) {
```
waxlamp, line 405: @mvandenburgh this seems to currently typecheck due to a catchall clause in the [definition of `DandisetMetadata`](https://github.com/dandi/dandi-archive/blob/c42f6201fabf7cf28a84b1eb52a259d080f90fab/web/src/types/schema.ts#L469). Would it be a good idea to make the `doi` field explicit there?

mvandenburgh: That type is automatically generated from the schema, see https://github.com/dandi/dandi-archive/blob/c42f6201fabf7cf28a84b1eb52a259d080f90fab/web/src/types/schema.ts#L3

waxlamp: Oh yeah, thanks for the reminder.

satra:
@mvandenburgh - a ping on this question?

> is there a way to try the ui also with the production server? i.e. could we have both previews?

given that their handle server doesn't seem to work currently, can i check the production site dandisets?

mvandenburgh:
> @mvandenburgh - a ping on this question?
> 
> > is there a way to try the ui also with the production server? i.e. could we have both previews?
> 
> given that their handle server doesn't seem to work currently, can i check the production site dandisets?

Unfortunately there is not. I copied a production DOI into a published staging dandiset, and temporarily pointed the deploy previews at the production DOI server, so you can test it now with https://deploy-preview-2212--gui-staging-dandiarchive-org.netlify.app/dandiset/216420?pos=1.

mvandenburgh:
Looks like we need to `JSON.stringify` the metadata object before writing it to a `script` tag.
Screenshot from 2025-03-18 09-52-03

satra:
at the moment it looks like our robot resolver (#2200) has not been quite resolved.  

https://search.google.com/test/rich-results/result?id=V_7BwHRu6j9AmjazGLaQuQ

but the deploy preview is generating the correct script tag now (thanks @mvandenburgh )

waxlamp:
Marked this as draft so we don't accidentally merge before reverting 5b876e6.

satra:
@waxlamp - to test the above bits out, we will also need to adjust the robots.txt on netlify.app in case it's different. it still says `Disallow: /` instead of `Allow: /`, which #2200 addressed. perhaps when deploying via netlify the same robots.txt as the ui endpoints can be used.

satra:
thanks @waxlamp - looks like we need more changes

image

don't know where the `X-Robots-Tag` is coming from.

waxlamp:
It seems that Netlify forces that header [deploy previews](https://answers.netlify.com/t/override-x-robots-tag-noindex-header-in-deploy-preview/38980/5), but not for [branch deploys](https://answers.netlify.com/t/override-x-robots-tag-noindex-header-in-deploy-preview/38980/8) so maybe we can use that to test this out.

I'm busy for a bit but will come back to try this out when I'm free.

waxlamp:
@satra, that seems to have worked: https://search.google.com/test/rich-results/result?id=aKGA18JvNMc_HEJiYtU5Rw

satra:
i think we should revert the special case for staging and merge this in and start monitoring if it starts showing up in google dataset search.

web/public/robots.txt
```diff
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow: /
+Allow: /
```
yarikoptic, line 2: FWIW -- this we already permitted in 
- https://github.com/dandi/dandi-archive/pull/2200

which was released 3 days ago.

web/src/views/DandisetLandingView/DandisetMain.vue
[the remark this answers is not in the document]
yarikoptic: that is per each client server, not server side, correct?

shouldn't this ideally be done/cached on backend since otherwise every our visitor (and robot?) would start bombarding those services potentially on the same DOIs, right?

web/public/robots.txt
satra: updated to main

web/src/views/DandisetLandingView/DandisetMain.vue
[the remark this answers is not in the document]
satra: that would be nice, but not a necessity for this. potentially could be enabled with server-side rendering in netlify, or if we generated the schema.org structure, but let's leave that as a future optimization, perhaps after we have minted dois for all dandisets.

web/public/robots.txt
waxlamp: @yarikoptic, I'm aware but "redoing" that small change here was easier than rebasing on master or merging it into this branch--it will not cause any problems when this is merged to master.

mvandenburgh (approved):

satra:
@waxlamp - feel free to merge this if it all looks good and any adjustments if we want to release this. this doesn't make much sense on staging to test given datacite's limitations.

waxlamp (approved):

PR #2216, opened 2025-03-14
Add search query parameter to dandiset list endpoint serializer

### LINKED ISSUE #2214: swagger search api
The swagger API doc (https://api.dandiarchive.org/swagger/) for /dandisets/ does not include "search":

Image

But the readme.io for the same endpoint (https://dandi.readme.io/reference/dandisets_list) does:

Image

I think including "search" is correct. Is there a reason this is missing in swagger?

waxlamp:
Despite the fact that the `/dandisets` endpoint accepts a query parameter `search` that is used to filter by the search terms, the parameter is not listed in the input serializer, and thus does not show up in the Swagger docs.

I built and tested this locally and it seems to work as expected.

Fixed #2214.


bendichter:
thanks @waxlamp !

PR #2223, opened 2025-03-19
Migrate puppeteer tests to playwright

### LINKED ISSUE #2209: Convert existing Puppeteer-based e2e tests to Playwright
Playwright/TypeScript has replaced Puppeteer/vanilla JS as our preferred browser-testing tool/framework, and all new E2E tests are written using it. We have [a collection of existing Puppeteer-based E2E tests](https://github.com/dandi/dandi-archive/tree/master/e2e) that need to be converted to Playwright.

### LINKED ISSUE #2210: Update E2E tests for Vuetify 3
#2186 upgrades us to Vuetify 3, which breaks many of our existing browser end-to-end tests. This means that our internal [`jest-puppeteer-vuetify`](https://github.com/kitware-resonant/jest-puppeteer-vuetify) helper library no longer works for selecting Vuetify elements. There may be other CSS selectors that are now outdated too. 

I propose that we take advantage of this to simultaneously complete #2209 and convert these tests over to playwright/typescript as part of updating them for Vuetify 3; not only will that let us drop Puppeteer, but writing Playwright E2E tests is much easier than Puppeteer with it's [code gen capabilities](https://playwright.dev/docs/codegen).

naglepuff:
- Closes #2209
- Closes #2210


mvandenburgh:
fyi, looks like the dandi-cli integration tests are failing for unrelated reasons - https://github.com/dandi/dandi-archive/pull/2285#issuecomment-2783878799

.github/workflows/frontend-ci.yml
```diff
@@ -31,228 +31,112 @@ jobs:
...
+          while ! nc -z localhost 8085; do
+            sleep 3
+          done
+
+          # start the dandi-api server
+          python ../../manage.py runserver &
+
+          # run the tests
+          npx playwright test
+      working-directory: e2e/playwright
```
mvandenburgh, line 135: I think we can move the contents of the `e2e/playwright/` directory to the top level `e2e/` directory now.

mvandenburgh (changes requested):
Looks good for the most part - couple of minor things related to file structure and git hygiene. I'll submit a separate review for the actual code changes.

- Can you rebase to remove the `celerybeat-*` files? Those should not be tracked in version control. #2280 will help prevent this from occurring going forward.
- https://github.com/dandi/dandi-archive/pull/2223/commits/1a70c9ec6a20290e14cd4ea988a9c20b16b929d5 shouldn't be necessary, as that variable was already added in #2205. Looks like it was erroneously removed in https://github.com/dandi/dandi-archive/commit/0a6832a84ced1d6180076c2f8f657b04838b9f96#diff-5cd76c2d3da116c3ec33f2a8ce86d739d0fe4e46c0563f2dde614ab8afc41f38L72 - can you fixup that commit and drop the redundant one?
- Can you squash https://github.com/dandi/dandi-archive/pull/2223/commits/5978fc73d1c19ee8f7ee7c1af5195e250250f550, https://github.com/dandi/dandi-archive/pull/2223/commits/6d557a07bc968e79605377ffa57a0dc152ee1f73, and https://github.com/dandi/dandi-archive/pull/2223/commits/5b0f739e3bc059efcbe948ff59ff9b7458fb0ad5
	- As part of the previous point, it looks like something might have went wrong during `yarn.lock` generation? https://github.com/dandi/dandi-archive/pull/2223/commits/5b0f739e3bc059efcbe948ff59ff9b7458fb0ad5 changes a bunch of the tar links from `yarnpkg.com` to `npmjs.org`. I was able to fix it locally by running `rm yarn.lock && yarn`.

e2e/.editorconfig
```diff
@@ -0,0 +1,22 @@
+root = true
```
mvandenburgh, line 1: Do we need this new file? If I'm understanding them correctly, the editorconfig docs suggest that a compliant editor should pick up the `.editorconfig` in the root of the repository - https://editorconfig.org/#file-location

e2e/tests/dandisetLandingPage.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Create a user to add as an owner later
+    const {
+      email: otherUserEmail,
+      firstname: otherUserFirstname,
+      lastname: otherUserLastName,
+    } = await registerNewUser(page);
+    const otherUserName = `${otherUserFirstname} ${otherUserLastName}`;
+    const otherUserInitials = `${otherUserFirstname.charAt(0)}${otherUserLastName.charAt(0)}`;
+    await page.getByRole("button", { name: otherUserInitials }).click();
+    // await page.getByRole("menuitem", { name: LOGOUT_BUTTON_TEXT }).click();
```
mvandenburgh, line 16:
```suggestion
```

e2e/tests/meditor.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  ).toBeVisible();
+};
+
+const dandisetsToTest = [
+  "000003",
+  "000018",
+  "000024",
+  // "000071",
+  "000106",
+  // "000107",
```
mvandenburgh, line 30: I wonder if there's value in testing at least one of `000071` or `000107`. Since there's no other dandisets being tested that have two simultaneous validation errors. What do you think?

e2e/tests/meditor.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  "000362",
+];
+
+const invalidDandisets = {
+  "000018": [noLicenseSelector],
+  "000024": [noContributorSelector, noLicenseSelector],
+  "000071": [noContributorSelector],
+  "000106": [invalidLicenseSelector],
+  "000107": [invalidLicenseSelector, noContributorSelector],
```
mvandenburgh, line 39: `000071` and/or `000107` can be removed here, right? (I say and/or because of my previous comment about testing one of them)

mvandenburgh (changes requested):
A couple minor things. Besides those, this looks great!

mvandenburgh:
Also, I noticed some weird formatting, but we're not doing any kind of linting on the e2e code so there's no set rules to adhere to. I filed #2306 so we can address that separately.

e2e/tests/meditor.spec.ts
naglepuff: Theses are commented out due to https://github.com/dandi/dandi-archive/issues/2281. Although, looking at this again, Dandiset 000024 also tests for `noContributorSelector`, so I'm a bit confused about that. I'll take a closer look.

e2e/.editorconfig
naglepuff: You're right. I'm new to editorconfig. I'll have to spend some time reading those docs, thanks for link. I'll remove the file.

e2e/tests/meditor.spec.ts
mvandenburgh: Yeah I agree a code comment would be appropriate here.

naglepuff: Added

e2e/tests/meditor.spec.ts
naglepuff: We can actually re-enable to tests for 000107. The actual assertion that fails is the icon indicating that the metadata is invalid. Both 000024 and 000107 have additional validation errors on the main tab that result in the correct icon being displayed. 000071 is special in that the ONLY validation error is on the contributors tab.

mvandenburgh (approved):

PR #2225, opened 2025-03-20
Add admin and description fields to AuditRecord model

### LINKED ISSUE #2224: Add admin audit records
This would add an `admin` boolean field to the audit record model, as well as a `description` text field. This would allow us to track whether an audit record in the system was performed by a normal user, or was done via an administrative action (e.g. a data migration or fix).

jjnesbitt:
Closes #2224 

This introduces two new fields to the `AuditRecord` model: `admin` and `description`. These fields are to be used in conjunction, when the modification of something in the system is done through administrative means, and not the normal workflow. An example of this would be if we were to fix some malformed data that existing in some versions or assets.


dandiapi/api/models/audit.py
```diff
@@ -43,5 +43,18 @@ class AuditRecord(models.Model):
...
+    # These fields are only used when these audit actions are done by admins
+    admin = models.BooleanField(default=False)
+    description = models.TextField(null=False, default='')
+
+    class Meta:
+        constraints = [
+            # Ensure that if admin=False, the description is also empty
+            models.CheckConstraint(
+                name='admin-description-pairing',
+                check=(models.Q(admin=False, description='') | models.Q(admin=True)),
```
waxlamp, line 55: Let's also require the description when `admin == True`.

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -53,7 +53,9 @@ def create_dandiset(
         draft_version.save()
 
         audit.create_dandiset(
-            dandiset=dandiset, user=user, metadata=draft_version.metadata, embargoed=embargo
+            params=audit.CommonAuditParams(dandiset=dandiset, user=user),
+            metadata=draft_version.metadata,
+            embargoed=embargo,
```
waxlamp, line 58: What do we gain from using this "common parameters" abstraction? It makes the definitions of the audit service functions harder to understand, and it makes the invocations (like here) more confusing than they need to be.

I'd prefer to back that out and just outfit the service functions with the admin and description arguments with default values (of `False` and `''`, respectively).

waxlamp (changes requested):
Looks good overall, but I want to strengthen the check constraint and reduce the "line noise" of the common params abstraction.

dandiapi/api/services/dandiset/__init__.py
waxlamp: > Well there's a couple reasons:
> 
> 1. Without this refactor, some of the audit functions now have many input arguments (6 or 7)

With the refactor, the functions still take that many arguments. It's just hidden behind a bit of a "legal fiction".

> 2. I think it's good to make a distinction between the "common" parameters between these audit functions, and the params that are unique to the function. You mention that it makes the invocations more confusing, but you could view it the other way. With this refactor, you _know_ that any audit function will take `CommonAuditParams` as the first argument.

I don't think much benefit comes from knowing that there is a common prefix of parameters; besides that, it's conceivable that future audit records *won't* have to supply all of these, in which case the abstraction would no longer be useful.

> I could go either way on this, there are pros/cons to each approach.

I agree in principle, but let's go with my recommendation since it's a smaller blast radius for this PR.

jjnesbitt:
> We had talked about making the `username` column nullable, but I had forgotten that column (along with the `user_email` and `user_fullname` columns) is text, not a foreign key to a user model. Does that mean we don't need these columns to be nullable at all, and can instead leave them blank or use stand-in data (like "admin", "[help@dandiarchive.org](mailto:help@dandiarchive.org)", and "DANDI admin team")?
> 
> If yes, then this is ready to merge; otherwise let's discuss what we should do.

Correct, my vote would be that we just leave them blank, instead of allow null. 

I'm realizing that we may want to require those fields be non-empty if `admin=False`. Although you could make the case it falls outside the scope of this PR (since if that's desired, the constraint should have already existed anyway).

If you agree we shouldn't address that here and now, then I agree with your previous comment that this is ready to merge.

mvandenburgh:
> Correct, my vote would be that we just leave them blank, instead of allow null.

I agree. In addition to the reasons already stated, there's also this rule - https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/.

waxlamp:
> I'm realizing that we may want to require those fields be non-empty if admin=False

Good point. Please go ahead and do that in a followup PR.

PR #2226, opened 2025-03-20
Remove quotes on DANDI_ADMIN_EMAIL env var

jjnesbitt:
Following from #2205, it seems the values supplied in the dev `.env` files contain quotes around them, which could cause errors when running the app locally, depending on the local environment. This is the only place this quoting occurs, which is why CI doesn't catch it.

### DIFF
--- a/dev/.env.docker-compose
+++ b/dev/.env.docker-compose
@@ -12,6 +12,6 @@ DJANGO_DANDI_DANDISETS_EMBARGO_LOG_BUCKET_NAME=dandiapi-embargo-dandisets-logs
 DJANGO_DANDI_WEB_APP_URL=http://localhost:8085
 DJANGO_DANDI_API_URL=http://localhost:8000
 DJANGO_DANDI_JUPYTERHUB_URL=https://hub.dandiarchive.org/
-DJANGO_DANDI_ADMIN_EMAIL="admin@test.mail"
+DJANGO_DANDI_ADMIN_EMAIL=admin@test.mail
 DJANGO_DANDI_DEV_EMAIL=test@example.com
 DANDI_ALLOW_LOCALHOST_URLS=1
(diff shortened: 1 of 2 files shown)


yarikoptic:
interesting. I thought that those are "shell" files and thus quotes would be good to have.  Do you know what / how reads those files besides doing `source` in some shell?  just to make sure we know syntax exactly there (e.g. what to do if there is a space in the value?)

jjnesbitt:
> interesting. I thought that those are "shell" files and thus quotes would be good to have. Do you know what / how reads those files besides doing `source` in some shell? just to make sure we know syntax exactly there (e.g. what to do if there is a space in the value?)

Some shells / methods of sourcing these files (in my case fish shell, which is why I discovered this) don't strip out the quotes, leaving the actual parsed value invalid. This is something I'm going to try to fix in my local environment, but regardless, there is no good reason to use quotes around this value, as there are no spaces and no other email address env vars do so.

yarikoptic:
in this case -- yes. But I wonder in general, since someone might run into a case where it is needed ("Last, First" name?), what is the assumption here on the syntax/format of such files?   Choices:
- Are they "POSIX compliant" in a sense that any POSIX compliant shell should be able to `source` them? (`fish` seems to be not, so YYMV indeed)
- Are they just ad-hoc `key=value` formatted set of lines with multiline values disallowed?

yarikoptic:
ok, then I think we better adopt some existing convention if not standard. There is

- [dotenv](https://www.npmjs.com/package/dotenv) : The most common library for loading .env files in Node.js
- [python-dotenv](https://pypi.org/project/python-dotenv/) : Similar library for Python

they seems to differ in their descriptions of the format unfortunately, but both support quotes if necessary (for spaces etc).  So for now this "fix" is ok as we do not need any, but IMHO it is something to keep in mind if we encounter the need in the future.

In particular from https://github.com/theskumar/python-dotenv?tab=readme-ov-file#file-format :

> The format is not formally specified and still improves over time. That being said, .env files should mostly look like Bash files.
> 
> Keys can be unquoted or single-quoted. Values can be unquoted, single- or double-quoted. Spaces before and after keys, equal signs, and values are ignored. Values can be followed by a comment. Lines can start with the export directive, which does not affect their interpretation.

Issue #2184, opened 2025-02-12
Create admin API endpoint(s) for sending out emails


yarikoptic:
We already are sending some emails e.g. when notifying users during registration.

DB has email information for dandiset owners and contacts, and all registered users of the archive.

I think it would make total sense to centralize ability to email DANDI users by use via API, so we could potentially use it in services outside of the dandi-archive, e.g. for admins

- dandi-hub to notify on home directories being reset (https://github.com/dandi/dandi-hub/issues/228) -- so for a specific user
- downtime notifications -- to all registered users
- validation errors reminders -- to corresponding authors of dandisets
- audit summaries -- to corresponding authors of dandisets

there could potentially be use cases even for authenticated users, although not immediately needed
 
- `dandi-cli reportbug {dandiset_id}` (does not exist but could) 

So having endpoint(s) allowing for those above would have been nice IMHO.  It could potentially also take care about proper re-formatting of markdown message text into html and txt versions, adding "branded" header/footnotes and other conveniences.

We could have per dandiset endpoint `/dandisets/{dandiset__pk}/contact` and then `/users/contact`  with data fields like

- content
- content-encoding (for markdown vs txt vs html)
- subject
- from

Related 

- I think it closes https://github.com/dandi/dandi-archive/issues/569

Current functionality which does emailing

```shell
❯ git grep 'def send_'
dandiapi/api/mail.py:def send_ownership_change_emails(dandiset, removed_owners, added_owners):
dandiapi/api/mail.py:def send_registered_notice_email(user: User, socialaccount: SocialAccount):
dandiapi/api/mail.py:def send_new_user_message_email(user: User, socialaccount: SocialAccount):
dandiapi/api/mail.py:def send_approved_user_message(user: User, socialaccount: SocialAccount):
dandiapi/api/mail.py:def send_rejected_user_message(user: User, socialaccount: SocialAccount):
dandiapi/api/mail.py:def send_pending_users_message(users: Iterable[User]):
dandiapi/api/mail.py:def send_dandiset_unembargoed_message(dandiset: Dandiset):
dandiapi/api/mail.py:def send_dandiset_unembargo_failed_message(dandiset: Dandiset):
dandiapi/api/tasks/scheduled.py:def send_pending_users_email() -> None:
```

[...]

PR #2227, opened 2025-03-20
WIP: Add endpoint to send mail to a user

asmacdo:
https://github.com/dandi/dandi-archive/issues/2184

- [x] Restrict to admins
- [ ] Improve Response
- [ ] Find a way to test
- [ ] Allow emails to multiple users?
- [ ] Issue mentioned "From:" but build_message does not support-- ok to just send from the admin email?
- [ ] Add option for html or plaintext

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING
 from allauth.socialaccount.models import SocialAccount
 from django.contrib.auth.models import User
 from django.core import mail
+from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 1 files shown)


jjnesbitt:
What is the point of having this exposed as an API endpoint? Why not expose this through the admin interface, as a custom dashboard perhaps?

asmacdo:
I'll close since this isnt making progress. (It was previously decided to open a design doc PR https://github.com/dandi/dandi-archive/pull/2370 which is also stuck) This could be useful as a reference, but its simple enough its probably not worth trying to directly apply these commits to modern dandi.

PR #2231, opened 2025-03-24
Fix a case where draft asset summaries could become stale

danlamanna:
There's an edge case where assets can be uploaded and the dandiset can transition into publishing before an appropriate asset summary for the draft version is computed. When generating the asset summary for the published version it can be copied to the draft to ensure the draft asset summary is always eventually correct.

### DIFF
--- a/dandiapi/api/services/publish/__init__.py
+++ b/dandiapi/api/services/publish/__init__.py
@@ -171,6 +171,9 @@ def _publish_dandiset(dandiset_id: int, user_id: int) -> None:
         # Add asset paths with new version
         add_version_asset_paths(version=new_version)
 
+        # Copy the finalized assetsSummary to the draft version in case it wasn't up to date
+        # before starting the publish.
+        old_version.metadata['assetsSummary'] = new_version.metadata['assetsSummary']
         # Set the version of the draft to PUBLISHED so that it cannot be published again without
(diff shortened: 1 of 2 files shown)


dandiapi/api/services/publish/__init__.py
```diff
@@ -173,6 +173,7 @@ def _publish_dandiset(dandiset_id: int, user_id: int) -> None:
 
         # Set the version of the draft to PUBLISHED so that it cannot be published again without
         # being modified and revalidated
+        old_version.metadata['assetsSummary'] = new_version.metadata['assetsSummary']
         old_version.status = Version.Status.PUBLISHED
         old_version.save()
```
mvandenburgh, line 178: The old comment needs to be moved to the next line, + I suggest adding a new comment along these lines describing why we're setting the assets summary
```suggestion
        # Synchronously update the assets summary of the draft version, since the
        # assets summary cron won't pick it up until it has been modified further.
        old_version.metadata['assetsSummary'] = new_version.metadata['assetsSummary']
        # Set the version of the draft to PUBLISHED so that it cannot be published again without
        # being modified and revalidated
        old_version.status = Version.Status.PUBLISHED
        old_version.save()
```

mvandenburgh (approved):
Sorry, thought I approved this

PR #2238, opened 2025-03-26
Convert remaining Vue components to `<script setup>` syntax

mvandenburgh:
All of our Vue components have been converted to TS/Composition API, but not all of them use `script setup`. 
This PR updates all those outstanding components, and also cleans up some unused variables that got detected by the better static analysis tooling that comes along with using `script setup`.

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -8,9 +8,9 @@
         v-if="!user"
         :close-on-content-click="false"
       >
-        <template #activator="{ props }">
+        <template #activator="{ props: menuProps }">
           <v-icon
-            v-bind="props"
+            v-bind="menuProps"
           >
             mdi-cog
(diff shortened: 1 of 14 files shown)


web/src/components/DandisetsPage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-page-title="pageTitle">
```
jjnesbitt, line 2: I guess this fixes a bug with the page title?

jjnesbitt (approved):

web/src/components/DandisetsPage.vue
mvandenburgh: Yeah, I described it in https://github.com/dandi/dandi-archive/pull/2238/commits/74452184f4a4ad714959ecbfb51073f7cfc3155e

PR #2241, opened 2025-03-26
update test doi server

satra:
Based on comment here: https://github.com/datacite/datacite/issues/2311#issuecomment-2752300648


yarikoptic (approved):
I also tried that curl invocation in referenced issue and indeed needs this fixed target service.  Tests also pass -- so likely is the right thing to do

mvandenburgh (approved):

mvandenburgh:
fyi I removed the `patch` label, since we don't want to release the Vue 3 client yet.

PR #2251, opened 2025-03-27
Remove gossip and mingling from Celery workers

aaronkanzer:
Stemming from behavior in @sandyhider 's work on EMBER Archive (a clone of DANDI Archive), their respective CloudAMQP instance quickly hits capacity

Similar behavior has been observed in LINC (Cc @kabilar )

From Sandy's correspondence with CloudAMQP support staff, a suggestion was received to disable gossip and mingling.

**Why can we not see current gossip and mingling:**

Current, our `Procfile` has --log-level at `INFO` -- gossip and mingling are only expressed at the `DEBUG` level

**Do we need mingling in DANDI Archive**

No -- `mingling` is meant to handle celery workers that have tasks defined in differing code bases. Our celery tasks are all prescribed in the same static codde

**Do we need gossip in DANDI Archive**

No -- gossip enables Celery workers to broadcast and listen to events (used for things like remote control, celery events, and dashboards like Flower).

Our Celery setup is primarily producer/consumer within a single deployment, so there’s no need for this cross-worker coordination.

This PR includes the appropriate flags to disable Celery functionality that does not seem integral, and should prevent bloat of our message broker

Cc @yarikoptic @satra

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -6,8 +6,8 @@ web: gunicorn --bind 0.0.0.0:$PORT dandiapi.wsgi --timeout 25
 # This is OK for now because of how lightweight all high priority tasks currently are,
 # but we may need to switch back to a dedicated worker in the future.
 # The queue `celery` is the default queue.
-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q celery -B
+worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q celery -B --without-gossip --without-mingle
 # The checksum-worker calculates blob checksums and updates zarr checksum files
-checksum-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q calculate_sha256,ingest_zarr_archive
+checksum-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO -Q calculate_sha256,ingest_zarr_archive --without-gossip --without-mingle
 # The analytics-worker processes s3 log files serially
-analytics-worker: REMAP_SIGTERM=SIGQUIT celery --app dandiapi.celery worker --loglevel INFO --concurrency 1 -Q s3-log-processing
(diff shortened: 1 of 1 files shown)


aaronkanzer:
**Note** In draft mode currently -- just doing some robust testing with `DEBUG` to evaluate frequency of gossip and mingling

aaronkanzer:
@waxlamp @mvandenburgh @jjnesbitt @satra @yarikoptic  -- Trying to replicate locally with `--log-level DEBUG` in the `celery`  (a bit tougher though since the RabbitMQ dashboard isn't present), but noticing some interesting behavior in the LINC logs (that lead me to have higher confidence that `mingle` and `gossip` disabled should help)

You can see below that without any messages being sent to the `ingest-zarr-achive`, `checksum`, `s3-log-processing`, about 5.4 messages per second are being consumed -- which scales per month to ~13M messages per month (~65% of 20M allocation per our current MQ instance quota)

I'd still like to be able to stress test how many messages are produced/consumed when some of our more intensive tasks are invoked. 

Screenshot 2025-03-27 at 4 26 04 PM

Screenshot 2025-03-27 at 4 27 25 PM

Curious if locally isn't as useful to find the root cause -- rather, a staging environment deployment might be best -- it is certainly possible that _both_ gossip/mingling _and_ something with our current infra<>tasks being sent are both ballooning message consumption

aaronkanzer:
https://sassy-cobalt-seahorse.rmq4.cloudamqp.com/#/

Seems DANDI has a similar scenario -- 6.1ish messages per second when none of the queues declared in our source code are being invoked

aaronkanzer:
Moving to ready for review per results observed by @ sandyhider on EMBER team: https://github.com/aplbrain/dandi-archive/commit/ab0188b1fdbb86dddf91b2c1bf7a91aec8c9a87c

kabilar:
Thank you, @sandyhider.

yarikoptic (approved):
@aaronkanzer said that @sandyhider  merged such a change and they didn't see the world flatten... so must be ok, we will give some time in staging (I just merged/released master)

PR #2258, opened 2025-03-27
Add Sentry back to Vue application

mvandenburgh:
I neglected to configure Sentry when I set up the Vue 3 app (#2186) :man_facepalming:. All the required env vars are already configured in `netlify.toml` from the Vue 2 app.


web/src/plugins/index.ts
```diff
@@ -17,4 +19,30 @@ export function registerPlugins (app: App) {
     .use(vuetify)
     .use(router)
     .use(pinia)
+
+  Sentry.init({
+    app,
+    dsn: import.meta.env.VITE_SENTRY_DSN,
+    environment: import.meta.env.VITE_APP_SENTRY_ENVIRONMENT,
+    integrations: [
```
jjnesbitt, line 27: The old setup included this integration:

https://github.com/dandi/dandi-archive/blob/1250202a25f01980fe01b3b8cc3cb00cc05108f3/web/src/main.ts#L29-L31

Do we no longer need that?

mvandenburgh: Thanks, I missed this one because it's not a part of the "Vue" sentry integration, but the "Javascript" one. I added it + added a configuration to also report `warn` logs.

jjnesbitt (approved):

PR #2267, opened 2025-03-31
Compute zarr checksums outside of transactions

danlamanna:
This should completely remove long lived database locks during zarr ingest. This should help prevent deadlocks like the one we saw recently.


jjnesbitt (approved):
Great idea! I originally thought this would present a race condition, if someone were to try to update the zarr while this was running. However, that shouldn't be possible (through the API at least), since its status must be `PENDING` for that to occur, and this function will exit early if that's the case.

PR #2272, opened 2025-04-03
Fix display of dandisets with many owners

### LINKED ISSUE #1761: cannot see >5 owners in DLP
[Here](https://dandiarchive.org/dandiset/000488) is an example of a dandiset landing page for dandiset 000488. I noticed that I am unable to see all of the dandiset owners. I can see the first 5 and "+1 more" but I don't see any way to expand the owners to see all of them.

image

Not a super high priority, but it would be nice to have some way in the UI to see all of the owners when there are >5 of them

bendichter:
fix #1761

Add management command to create test dandiset with multiple owners

Cline-assisted

image


bendichter:
image

yarikoptic:
@waxlamp could someone review this?

dandiapi/api/management/commands/create_dev_dandiset.py
```diff
@@ -6,28 +6,52 @@
...
-def create_dev_dandiset(*, name: str, email: str):
+@click.option(
+    '--num-extra-owners',
+    'num_extra_owners',
+    required=False,
+    type=int,
+    default=0,
+    show_default=True,
+    help='The number of extra owners to programmatically generate',
+)
```
mvandenburgh, line 30: Should we remove this? Or at least move it to its own management command if you feel its useful to keep? I think there's value in keeping `create_dev_dandiset` as simple as possible, since it's a key part of getting a dev environment set up without having to upload data with the CLI.

dandiapi/api/management/commands/create_dev_dandiset.py
```diff
@@ -6,28 +6,52 @@
 from django.contrib.auth.models import User
 from django.core.files.uploadedfile import SimpleUploadedFile
 import djclick as click
+import faker
```
mvandenburgh, line 9: This dependency isn't installed anywhere besides in `tox.ini` as part of testing, so this script would not run even if the user has run `pip install .`.

dandiapi/api/management/commands/create_dev_dandiset.py
jjnesbitt: This was previously a separate management command included in this PR, and I merged it into this command, in order to simplify things. I don't know how much value there is in keeping it. I don't think it warrants its own management command, since at that point it's essentially just adding owners to a dev dandiset. If you don't think it's valuable (or if it adds negative value) we can remove it from here.

My two cents is that it does provide value, since often times when developing locally, we have just a single dandiset that has just one owner (yourself). So having ways to extend that to provide a more realistic test dandiset is useful, I think. Although it does increase the complexity of the management command slightly.

dandiapi/api/management/commands/create_dev_dandiset.py
jjnesbitt: Included it in [15ca09b](https://github.com/dandi/dandi-archive/pull/2272/commits/15ca09bb8f4287cf7d97d0247fa96ad74a2e62f2) and force-pushed. If we decide not to keep the augmentation of that management command i'll just drop the commit.

bendichter:
The columns look nice! How does this look when the width of the browser window is narrow?

bendichter:
Also, what happens when there are more owners than can fit on the screen? I realize with columns this number is doubled (or more depending on the number of columns) but we should nevertheless make sure the display works even with more users than can fit. Easiest way would simply be confirming that scrolling works. We could even paginate this, but that seems overkill until we reach a point where we need it.

jjnesbitt:
> The columns look nice! How does this look when the width of the browser window is narrow?
It just stacks the columns on top of each other

image




> Also, what happens when there are more owners than can fit on the screen?

This is how it looks with 100 owners. We could put a lot of effort into making sure all columns are aligned perfectly, but I don't think it's worth it at all, since you'll only see that when you have over 30 owners. The important thing is that all will be visible.
image

dandiapi/api/management/commands/create_dev_dandiset.py
mvandenburgh: I don't have a strong opinion either way. I suppose I lean slightly towards removing it because I don't see myself using it. But, if you think you will use it / it could be useful in general for devs, then I'm not opposed. It does add some complexity, but imo it's so little additional complexity that someone finding it useful is a compelling enough reason to keep it. 

Although I think we should move `Faker` into `dev` dependencies, and only import it inside that conditional. Just to avoid pulling in a large dependency like that into production when it's not really needed

jjnesbitt: Fixed in [1472be2](https://github.com/dandi/dandi-archive/pull/2272/commits/1472be2cd644beede5612908163456a4ac643c43)

mvandenburgh (approved):

bendichter:
> This is how it looks with 100 owners. We could put a lot of effort into making sure all columns are aligned perfectly, but I don't think it's worth it at all, since you'll only see that when you have over 30 owners. The important thing is that all will be visible.

Agreed. I doubt this situation will come up very often so I am perfectly happy to move forward with a solution that works even if it isn't as pretty as it could be

PR #2282, opened 2025-04-06
Refactor StarButton layout and improve alignment in DandisetMain view

### LINKED ISSUE #2188: UI: star number not always next to star icon
Image

https://dandiarchive.org/dandiset/001031

When the title is long, the number is rendered below the star icon, which I think makes the design a bit confusing. Would it be possible to adjust the design so the number is always next to the star?

Also, when the title of a dandiset is short, the star is in line with the title:

Image

https://dandiarchive.org/dandiset/000108?page=1&sortOption=4&sortDir=-1&showDrafts=true&showEmpty=false&pos=7

I think it would look nicer if the star were right-aligned

bendichter:
fix #2188 

Cline-assisted

image

image

[cline_task_apr-6-2025_11-54-10-am.md](https://github.com/user-attachments/files/19622941/cline_task_apr-6-2025_11-54-10-am.md)

### DIFF
--- a/web/src/components/StarButton.vue
+++ b/web/src/components/StarButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="star-container">
+  <div class="d-flex align-center justify-end ml-auto">
     <v-btn
       variant="plain"
       icon
@@ -43,12 +43,3 @@ async function toggleStar() {
   }
 }
 </script>
-
-<style scoped>
-.star-container {
-  display: flex;
-  align-items: center;
-  justify-content: flex-end;
-  margin-left: auto;
-}
-</style>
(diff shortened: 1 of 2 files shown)


web/src/components/StarButton.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="star-container">
```
mvandenburgh, line 2: Instead of defining a custom css class, it is better practice to use Vuetify's built-in classes
```suggestion
  <div class="d-flex align-center justify-end ml-auto">
```

bendichter:
The latest change moves the start to the top, even if the title is multi-line

image

mvandenburgh (approved):
Confirmed this works on https://deploy-preview-2282--gui-staging-dandiarchive-org.netlify.app/dandiset/216486.

PR #2288, opened 2025-04-07
Lock dandiset when changing owners

### LINKED ISSUE #2000: Lock dandiset when changing owner list
_Originally posted by @mvandenburgh in https://github.com/dandi/dandi-archive/pull/1886#discussion_r1711706208_:

> This problem existed before this PR, but I'll just mention that by the time this `.save()` is called this `Dandiset` object might be stale. I think we need to lock the `dandiset` object with select_for_update.

https://github.com/dandi/dandi-archive/blob/0a225f4cfec3820b14a52a08bdc60f7a3fd8d182/dandiapi/api/views/dandiset.py#L416-L420

naglepuff:
Fixes #2000 

### Changes

The `users` endpoint for the Dandiset View now uses [select_for_update](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#select-for-update). This prevents accidentally saving stale data when calleding `dandiset.save()`.

### DIFF
--- a/dandiapi/api/views/dandiset.py
+++ b/dandiapi/api/views/dandiset.py
@@ -522,13 +522,14 @@ class DandisetViewSet(ReadOnlyModelViewSet):
 
             # All owners found
             with transaction.atomic():
+                dandiset_locked = Dandiset.objects.select_for_update().get(pk=dandiset__pk)
                 owners = user_owners + [acc.user for acc in socialaccount_owners]
-                removed_owners, added_owners = replace_dandiset_owners(dandiset, owners)
-                dandiset.save()
+                removed_owners, added_owners = replace_dandiset_owners(dandiset_locked, owners)
+                dandiset_locked.save()
 
                 if removed_owners or added_owners:
                     audit.change_owners(
-                        dandiset=dandiset,
+                        dandiset=dandiset_locked,
                         user=request.user,
                         removed_owners=removed_owners,
                         added_owners=added_owners,


dandiapi/api/views/dandiset.py
```diff
@@ -485,7 +485,10 @@ def unembargo(self, request, dandiset__pk):
     # TODO: move these into a viewset
     @action(methods=['GET', 'PUT'], detail=True)
     def users(self, request, dandiset__pk):  # noqa: C901
-        dandiset: Dandiset = self.get_object()
+        # This locks a row in the dandiset table until the function is finished
+        # This is likely fine here since this function doesn't make any network calls,
+        # or otherwise have any long-running operations.
+        dandiset: Dandiset = Dandiset.objects.select_for_update().get(pk=dandiset__pk)
```
mvandenburgh, line 491: I think this code comment can be classified as a "why, not what" comment. i.e. it is explaining something that should be obvious from reading the code. 
```suggestion
        dandiset: Dandiset = Dandiset.objects.select_for_update().get(pk=dandiset__pk)
```

mvandenburgh (changes requested):
In addition to the point about the code comment, as I look at this more closely I'm having second thoughts about locking the dandiset in this manner. TLDR: I think your original solution is actually better.

We discussed this before and decided to just lock the dandiset for the duration of the entire endpoint, as is implemented here, in order to eliminate an extra DB query. But I realize now that we only need the lock when the user is sending a PUT request to modify the dandiset owners. The vast majority of calls to this endpoint will not do that, but will instead be a GET request that simply retrieves the owners list (where no lock is needed). That GET request to `/users/` happens on every load of the DLP (i.e. probably 99% of calls to this endpoint are GET requests), so with the current approach we'd be adding the overhead of opening a transaction and holding a row lock to every DLP page load. 

So I think I'm now convinced that your original solution of re-querying for the dandiset with `SELECT FOR UPDATE` is better, @naglepuff. What do you think?

dandiapi/api/views/dandiset.py
danlamanna: Does this work since select for update needs to be done inside a transaction?

naglepuff: @danlamanna I'm not sure. I looked at the documentation for [select_for_update](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#select-for-update) and in their example, the `select_for_update` is used just outside of a transaction (at least that what it looks like to me)

mvandenburgh: The reason why it works in those docs is because querysets are evaluated lazily. In that example, `entries` is evaluated by the for loop, which is inside the transaction.atomic context manager so it works.
 
https://github.com/dandi/dandi-archive/pull/2288/files#diff-07074e43bf5319fe74c218e44c401e8152e75396cc6e478f6de0c751b1726fbdR524

So in your case, you will need a transaction as Dan suggested.

mvandenburgh (approved):
LGTM

PR #2290, opened 2025-04-07
Fix v-switch margins and color

### LINKED ISSUE #2249: `Show` options don't provide color distinctions between on/off
On staging after Vue 3 upgrade.

On production in Vue 2, when the option is selected the bar turns blue but this does not occur with the Vue 3 deployment:

Image

### LINKED ISSUE #2289: Dandiset list filter switches are cut off
Notice the left side of these is slightly cut off.

Image

mvandenburgh:
Fixes #2249
Fixes #2289

Before:

switch_broken

After:

switch_fixed2

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -25,6 +25,7 @@
               label="Drafts"
               density="compact"
               class="mx-2"
+              color="primary"
             />
           </v-list-item>
           <v-list-item>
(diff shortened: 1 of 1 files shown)


jjnesbitt:
Not to be a stickler but it seems the switch is still slightly cut off on the left side.
image

mvandenburgh:
Since switching to the `comfortable` variant doesn't fix the issue, and also changes the vertical margins to be larger than in the Vuetify 2 app, I just backed that out and added some additional horizontal margins.

jjnesbitt (approved):

PR #2303, opened 2025-04-14
Add an index for looking up pending assets

danlamanna:
Looking up pending assets needs to be fast for our asset metadata pipeline. Since there should always be a small number of pending assets, this index should be selective enough to always be used. There are some recent sentry issues showing this query slowing down because of having to do sequential scans.


dandiapi/api/models/asset.py
```diff
@@ -127,6 +127,16 @@ class Status(models.TextChoices):
 
     class Meta:
         ordering = ['created']
+        indexes = [
+            # Other statuses are likely too common to index, but pending assets are continually
+            # polled and being moved through the pipeline.
+            models.Index(
+                fields=['status'],
+                name='%(app_label)s_%(class)s_status_pending',
+                # TODO: refer to Pending via the Status enum
```
danlamanna, line 136: I'm working on a follow up PR for this. The tldr is we shouldn't define the enums inside of model classes because it otherwise becomes impossible to refer to them inside of meta classes. So `Asset.Status` needs to be moved to a top level `AssetStatus`. FYI @mvandenburgh

jjnesbitt (approved):
Looks good to me. 

I wonder how many pending assets we tend to have at once. I'd imagine it's very low, but I have no actual data.

danlamanna:
> I wonder how many pending assets we tend to have at once. I'd imagine it's very low, but I have no actual data.

At the moment it's ~100 out of ~500k :man_shrugging:

PR #2305, opened 2025-04-14
Move Asset.Status to top level AssetStatus class

danlamanna:
Without this, values of `AssetStatus` can't be used in indexes or constraints - see #2303 for an example.


danlamanna:
> I wonder if you could just set `Status = AssetStatus` in the Asset model? That way, there would be very little code change, but you would still get the desired benefit.

That works for everywhere else in the codebase - the diff is much smaller now.

jjnesbitt (approved):

PR #2307, opened 2025-04-14
Remove deprecated version field from docker compose .yml

mvandenburgh:
This field is deprecated and causes a warning to be logged whenever a `docker compose` command is invoked.
See https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md


danlamanna (approved):
I was about to remove this too :)

PR #2308, opened 2025-04-14
Denormalize species for faster searching

### LINKED ISSUE #2273: Sentry reports on /api/dandisets/search/ erroring out
this is at least 2nd instance of such an error report

```
   1   + Apr 03 Sentry           production: DANDI-API-MV - SystemExit: 1
   2   + Feb 24 Sentry           production: DANDI-API-MM - SystemExit: 1
```

implicating

```
[4]SystemExit /api/dandisets/search/
1

   ID: 95e9bef35faa49d896c49e077c4d7741
   April 3, 2025, 4:42:49 a.m. UTC
     Project   [5]dandi-api
   environment production
      Level    error

Exception

SystemExit: 1
(36 additional frame(s) were not displayed)
...
  File "dandiapi/api/views/dandiset.py", line 335, in search
    query_serializer = DandisetSearchQueryParameterSerializer(data=self.request.query_params)
  File "dandiapi/api/views/serializers.py", line 270, in __init__
    self.fields['species'].choices = list(
  File "psycopg_c/_psycopg/waiting.pyx", line 201, in psycopg_c._psycopg.wait_c

Request

    URL   [6]https://api.dandiarchive.org/api/dandisets/search/
   Method GET
   Query  page_size=8&page=1
```

so smth to checkout/fix to avoid errors -- seems to step from serializer of species values?  if Sentry contains header options provided to the query -- should be easy to reproduce **now** (might change later as metadata could change etc) to fix up.

NB oddly -- `grep` on the IP for heroku logs for these 2 days returns nothing!

danlamanna:
Fixes #2273 (as well as some "slow db query" sentry issues).

On page load we have to select the distinct species to populate the dropdown. Unfortunately it appears(?) that postgres can't do index-only scans on expressions (see https://stackoverflow.com/questions/53583257/is-there-a-workaround-to-ensure-index-only-scan-on-jsondoc-equality-que). So on page load we're currently iterating over the entire table - this should make it use an index-only scan and avoid reading from the table.


yarikoptic:
Given that our dandisets assets summaries which provide such metadata are updated not that frequently, shouldn't the target dictionaries for searches be composed/updated as a result of such updates and then reused without requiring any DB sweep across dandisets over metadata?

danlamanna:
> Given that our dandisets assets summaries which provide such metadata are updated not that frequently, shouldn't the target dictionaries for searches be composed/updated as a result of such updates and then reused without requiring any DB sweep across dandisets over metadata?

FWIW this PR doesn't change any existing behavior other than adding the species column to the materialized view for search. The migration is mostly identical to its previous migration, and unfortunately it has to be redefined because postgres doesn't allow tacking columns on to existing materialized views.

As for the original design, I think it was seen as a much simpler approach to keeping denormalized data synchronized than trying to make every write operation more costly, especially for an exploratory feature.

mvandenburgh (approved):

PR #2327, opened 2025-04-17
Update swagger/redocs urls to align with Resonant

mvandenburgh:
Typically a Resonant app will serve the redoc and swagger UI at `/api/docs/{swagger|redoc}` (see [cookiecutter](https://github.com/kitware-resonant/cookiecutter-resonant/blob/master/%7B%7B%20cookiecutter.project_slug%20%7D%7D/%7B%7B%20cookiecutter.pkg_name%20%7D%7D/urls.py#L33-L34)), but for historical reasons dandi serves them at the root of the API. This updates the routes to match the Resonant convention. We may want to come back to this later and deprecate/remove the old routes after some time, but for now the old routes are still available and simpy redirect to the new ones.

### DIFF
--- a/dandiapi/urls.py
+++ b/dandiapi/urls.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 from django.conf import settings
 from django.contrib import admin
-from django.urls import include, path, re_path, register_converter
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
Long needed!

PR #2329, opened 2025-04-18
Move banner with info blurb to top of all pages

### LINKED ISSUE #2302: Modify NIH banner
- [ ] Move to the top of the page
- [ ] Add to all pages
- [ ] Remove the `i` logo since it doesn't provide a 'tooltip'
- [ ] Lighten the background color so that the text is easier to read

naglepuff:
Fix #2302 

This moves the previously implemented `FooterBanner` component to the main `App.vue` component, making it visible on all pages.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -5,6 +5,7 @@
   >
     <AppBar />
     <v-main v-if="connectedToServer">
+      <InfoBanner />
       <UserStatusBanner />
       <router-view />
       <DandiFooter />
@@ -47,6 +48,7 @@ import { useDandisetStore } from '@/stores/dandiset';
 import AppBar from '@/components/AppBar/AppBar.vue';
 import DandiFooter from '@/components/DandiFooter.vue';
 import UserStatusBanner from '@/components/UserStatusBanner.vue';
+import InfoBanner from '@/components/InfoBanner.vue';
 
 const store = useDandisetStore();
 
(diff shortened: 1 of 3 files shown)


web/src/App.vue
```diff
@@ -47,6 +48,7 @@ import { useDandisetStore } from '@/stores/dandiset';
 import AppBar from '@/components/AppBar/AppBar.vue';
 import DandiFooter from '@/components/DandiFooter.vue';
 import UserStatusBanner from '@/components/UserStatusBanner.vue';
+import FooterBanner from '@/components/InfoBanner.vue';
```
mvandenburgh, line 51:
```suggestion
import InfoBanner from '@/components/InfoBanner.vue';
```

web/src/components/InfoBanner.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <v-banner
     v-if="bannerText"
-    color="grey-darken-1"
+    bg-color="grey-lighten-6"
+    sticky
+    :style="{ top: '64px' }"
```
mvandenburgh, line 6: Is this inline style needed? The banner stays the same when I remove it locally.
```suggestion
```

naglepuff: does it not scroll behind the app bar for you?

web/src/App.vue
naglepuff: Whoops! Should be fixed now.

web/src/components/InfoBanner.vue
mvandenburgh: oh, i didn't try scrolling - good catch :slightly_smiling_face:

mvandenburgh (approved):

PR #2331, opened 2025-04-18
Switch staging back to builtin oauth `Application`

### LINKED ISSUE #2319: Use oauth toolkit's `ALLOW_URI_WILDCARDS` instead of a custom `Application` model
Prerequisite to this issue: #1826 

`django-oauth-toolkit` now supports wildcard redirect URIs - https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#allow-uri-wildcards. We should drop the custom `Application` model we've defined to allow staging deploy previews to work (https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/oauth.py#L10), and switch to this.

mvandenburgh:
Fixes #2319

This commit switches the staging environment back to using the builtin `Application` model for OAuth2 authentication, while enabling the `ALLOW_URI_WILDCARDS` setting to allow for wildcard URIs to support Netlify deploy previews.

To be extra cautious, I did not remove the custom `StagingApplication` model yet. I think we should merge this, verify it works properly, and only then remove the custom model.

### DIFF
--- a/dandiapi/settings.py
+++ b/dandiapi/settings.py
@@ -201,11 +201,7 @@ class HerokuProductionConfiguration(DandiMixin, HerokuProductionBaseConfiguratio
     AUTO_APPROVE_USERS = False
 
 
-# NOTE: The staging configuration uses a custom OAuth toolkit `Application` model
-# (`StagingApplication`) to allow for wildcards in OAuth redirect URIs (to support Netlify branch
-# deploy previews, etc). Note that both the custom `StagingApplication` and default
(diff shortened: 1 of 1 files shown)


jjnesbitt:
This will temporarily break logins on staging until we go in and set the redirect URLs and other settings again, correct?

mvandenburgh:
> This will temporarily break logins on staging until we go in and set the redirect URLs and other settings again, correct?

Right. We'll have to create an `Application` in the django admin with the same values as the existing `StagingApplication`

jjnesbitt (approved):

PR #2332, opened 2025-04-18
Switch from `runtime.txt` to `.python-version`

### LINKED ISSUE #2325: Switch to `.python-version` for Heroku Python version
Using `runtime.txt` for setting the version of Python installed in Heroku is deprecated https://devcenter.heroku.com/articles/python-runtimes#selecting-a-runtime

We should switch to `.python-version` and only pin to a major version of Python as Heroku recommends. 

Upstream Resonant PR: https://github.com/kitware-resonant/cookiecutter-resonant/pull/215

mvandenburgh:
Fixes #2325

This switches us to the new file format, but does not resolve the issue pointed out by @waxlamp in [this comment](https://github.com/dandi/dandi-archive/issues/2325#issuecomment-2813468086), and instead maintains the current pinning behavior of `runtime.txt` (that is, including the patch version in the pinned version).

Another benefit of switching is that the `actions/setup-python` github action supports `.python-version`, so that eliminates a bit of code duplication there where we were hardcoding the python version across our CI workflow files.


waxlamp (approved):
This looks good, and I opened #2333 to capture the dependency pinning problem, so go ahead and close the original issue with this PR.

PR #2336, opened 2025-04-18
Remove custom `StagingApplication` model

mvandenburgh:
Follow on from #2331


mvandenburgh:
Due to unforeseen issues, we're reverting the changes this depends on and will come back to this later

PR #2339, opened 2025-04-18
Convert StagingApplication to a proxy model

jjnesbitt:
This switches the `StagingApplication` model to be a django [proxy model](https://docs.djangoproject.com/en/5.2/topics/db/models/#proxy-models). This means that any future oauth2_provider migrations would be applied on the standard application model, and would make moving away from the `StagingApplication` model altogether much more seamless in the future.


mvandenburgh (approved):
LGTM. We'll want to keep track of the current `StagingApplication` in the DB so we can recreate it, as it will get deleted when the migration runs.

jjnesbitt:
> LGTM. We'll want to keep track of the current `StagingApplication` in the DB so we can recreate it, as it will get deleted when the migration runs.

That's true, although I think we essentially did that on staging the other day when attempting to complete https://github.com/dandi/dandi-archive/pull/2331. Still we should ensure we have a copy of that data prior to merging.

PR #2340, opened 2025-04-20
Auto-allow people with `@nih.gov` and `@janelia.hhmi.org` email addresses

kabilar:


waxlamp (approved):
I opened #2349 to make this auto-approval list more of an administrative feature for the Archive, but this looks good for the near term.

PR #2341, opened 2025-04-21
Remove unneeded data from playwright test fixture

mvandenburgh:
#2223 modified the `playwright.json.xz` test fixture, and inadvertently added some additional DB tables into the test data that do not need to be included. These are mostly just some internal tables that Django/other upstream libraries create and maintain themselves. 

@naglepuff I missed this in my original review of #2223, I forgot we used the `--exclude` flag to ensure certain tables are excluded when running `manage.py dumpdata`. I added some brief documentation on how to generate that data so that we remember next time it needs to be updated.

I believe this is also why #2339 is failing CI - the sample data currently includes a `oauth2_provider.Application` instance which seems to be causing some issues.


naglepuff (approved):
LGTM

PR #2350, opened 2025-04-22
Introducing Dandiset DOIs

asmacdo:
- POC Implementation for https://github.com/dandi/dandi-archive/pull/2012

Example of Dandiset DOI (note -- no version) : 10.80507/dandi.000004

Remaining TODOs:
- [x] Update https://github.com/dandi/dandi-archive/pull/2012, which should be merged before this comes out of "draft PR"
- [x] Implement support in to_datacite within dandi-schema: https://github.com/dandi/dandi-schema/pull/297
  - [ ] figure out why tests do not fail now without https://github.com/dandi/dandi-schema/pull/297
- [x] Use DJANGO_DANDI_DOI_PUBLISH to determine whether DOIs are Findable
- [x] Handle Datacite validation failures (if not enough to make findable, revert to draft)
- [X] On update, if draft doi, "publish" to Findable
- [X] Reduce repetition in Datacite API usage, probably create a minimal client?
- [X] add tests
- [x] Move (and combine?) "mid level" functions from publish/ dandiset/ views/version/ unembargo/ to doi.py
- [x] Add Dandset DOI to vue for draft version
- [ ] management command ot populate DOIs for existing public dandisets
- [x] adjust web "e2e" (fronend) tests in `e2e/tests/dandisetLandingPage.spec.ts` to ensure we have DOIs
- [ ] Add hand-QA checklist to this PR
- [ ] For handle publication DOIs, catch datacite client exceptions since we are dealing with 2 DOIs (1 create,  1 update)
- [ ] Be more consistent about exception handling and logging
- [ ] consider adding `audit` related records on DOI operations
- [ ] reconsider using schema's Dandiset instead of constructing unvalidated
- [x] Decide how to handle "delete" there is no usage of the REST endpoint AFAICT! Admin delete uses a different endpoint

May be later as part of this depending on what comes in first
- [ ] Rebase schema PR on Isaac's changes

### DIFF
--- a/dandiapi/api/management/commands/inject_doi.py
+++ b/dandiapi/api/management/commands/inject_doi.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from django.core.management.base import BaseCommand
+
+from dandiapi.api.models import Dandiset, Version
+
+
+class Command(BaseCommand):
+    help = 'Inject a DOI into a dandiset version for testing'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'dandiset_identifier', type=str, help='Dandiset identifier (e.g., 000001)'
+        )
+        parser.add_argument(
+            '--dandiset-version', type=str, default='draft', help='Version (default: draft)'
+        )
+        parser.add_argument(
+            '--doi', type=str, help='DOI to inject (if not provided, will generate one)'
+        )
+
+    def handle(self, *args, **options):
+        dandiset_identifier = options['dandiset_identifier']
+        version = options['dandiset_version']
+        doi = options['doi']
+
+        try:
+            try:
+                dandiset = Dandiset.objects.get(id=int(dandiset_identifier))
+            except (ValueError, Dandiset.DoesNotExist):
+                numeric_id = (
+                    int(dandiset_identifier.lstrip('0')) if dandiset_identifier.lstrip('0') else 0
+                )
+                dandiset = Dandiset.objects.get(id=numeric_id)
+
+            version_obj = Version.objects.get(dandiset=dandiset, version=version)
+
+            if not doi:
+                # TODO: this prefix needs to be updated for non-dandi deployments
+                doi = f'10.80507/dandi.{dandiset_identifier}'
+
+            version_obj.metadata['doi'] = doi
+            version_obj.save()
+
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f'Successfully injected DOI "{doi}" into {dandiset_identifier}/{version}'
+                )
+            )
+
+        except Dandiset.DoesNotExist:
+            self.stdout.write(self.style.ERROR(f'Dandiset {dandiset_identifier} not found'))
+        except Version.DoesNotExist:
+            self.stdout.write(
+                self.style.ERROR(f'Version {version} not found for dandiset {dandiset_identifier}')
+            )
--- a/e2e/tests/dandisetLandingPage.spec.ts
+++ b/e2e/tests/dandisetLandingPage.spec.ts
@@ -1,6 +1,7 @@
 import { expect, test } from "@playwright/test";
 import { clientUrl, registerNewUser, LOGOUT_BUTTON_TEXT, registerDandiset } from "../utils.ts";
 import { faker } from "@faker-js/faker";
+import { execSync } from "child_process";
 
 test.describe("dandiset landing page", async () => {
   test("add an owner to the dandiset", async ({ page, browser }) => {
@@ -40,9 +41,42 @@ test.describe("dandiset landing page", async () => {
     await expect(newPage.getByText(otherUserName)).toHaveCount(1);
     await context.close();
   });
+
   test("navigate to an invalid dandiset URL", async ({ page }) => {
     await page.goto(`${clientUrl}/dandiset/1`);
     await page.waitForLoadState("networkidle");
     await expect(page.getByText("Error: Dandiset does not exist")).toHaveCount(1);
+  });
+
+  test("draft dandiset shows dandiset DOI", async ({ page }) => {
+    // Register a new user and create a dandiset
+    await registerNewUser(page);
+    const dandisetName = faker.lorem.words();
+    const dandisetDescription = faker.lorem.sentences();
+    const dandisetId = await registerDandiset(page, dandisetName, dandisetDescription);
+
+    // Inject a DOI directly using Django management command
+    const testDoi = `10.80507/dandi.${dandisetId}`;
+
+    // Execute the Django management command to inject DOI
+    try {
+      execSync(`cd .. && python manage.py inject_doi ${dandisetId} --dandiset-version=draft --doi="${testDoi}"`, {
+        stdio: 'inherit',
+        timeout: 10000
+      });
+    } catch (error) {
+      console.error('Failed to inject DOI:', error);
+    }
+
+    // Refresh the page to see the updated DOI
+    await page.reload();
+    await page.waitForLoadState("networkidle");
+
+    // The draft version should show the injected Dandiset DOI
+    await expect(page.getByText(testDoi)).toHaveCount(1);
+
+    // Should not show a version DOI (since it's a draft)
+    const versionDoiPattern = new RegExp(`10\\.(48324|80507)/dandi\\.${dandisetId}/`);
+    await expect(page.getByText(versionDoiPattern)).toHaveCount(0);
   });
 });


dandiapi/api/doi.py
```diff
@@ -24,22 +24,81 @@ def doi_configured() -> bool:
...
+        doi = f'{prefix}/dandi.{dandiset_id}'
     metadata = version.metadata
     metadata['doi'] = doi
-    return (doi, to_datacite(metadata, publish=publish))
+    # TODO(asmacdo) findable dois
+    # datacite_body = to_datacite(metadata, event=event)
+    datacite_body = to_datacite(metadata, version_doi=version_doi)
+    return (doi, datacite_body)
+
+def update_dandiset_doi(version: Version) -> str:
```
asmacdo, line 48: Maybe this whole chunk is unnecessary?  https://github.com/dandi/dandi-archive/issues/1709#issuecomment-1924227635

If we can just point to the Dandiset DOI to the DLP, rather than the latest version then we dont need to update.

CODE_OF_CONDUCT.md
```diff
@@ -3,7 +3,7 @@
 ## Our Pledge
 
 We as members, contributors, and leaders pledge to make participation in our
-community a harassment-free experience for everyone. We pledge to act and 
+community a harassment-free experience for everyone. We pledge to act and
```
yarikoptic, line 6: send a separate PR with unrelated changes

dandiapi/api/datacite.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class DataCiteClient:
+    """Client for interacting with the DataCite API."""
+
+    def __init__(self):
+        self.api_url = settings.DANDI_DOI_API_URL
+        self.api_user = settings.DANDI_DOI_API_USER
+        self.api_password = settings.DANDI_DOI_API_PASSWORD
+        self.api_prefix = settings.DANDI_DOI_API_PREFIX
```
asmacdo, line 39: Previously, we would use `settings.DANDI_DOI_API_PREFIX or '10.80507'`

If api prefix is not set, DOI API operations should be prevented by `is_configured()` so I think this is appropriate

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -69,13 +107,22 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
...
         # chance to grab the Dandiset information before it is destroyed.
         audit.delete_dandiset(dandiset=dandiset, user=user)
-
+        # Dandisets with published versions cannot be deleted
+        # so only the Dandiset DOI needs to be delete.
+        if draft_version and draft_version.doi is not None:
+            try:
+                doi.delete_or_hide_doi(draft_version.doi)
+            except Exception:
+                pass # doi operation should not stop deletion. delete_or_hide will log the exception.
```
yarikoptic, line 125: I would say that we should actually abort deletion of dandiset if something goes wrong with DOI deletion so we do not cause inconsistency

dandiapi/api/services/publish/__init__.py
```diff
@@ -106,6 +106,56 @@ def _build_publishable_version_from_draft(draft_version: Version) -> Version:
...
+            version_doi=False,
+            event='publish',  # Promote to Findable on first publication
+        )
+    else:
+        # For subsequent publications: update the metadata but keep as Findable
+        dandiset_doi, dandiset_doi_payload = doi.generate_doi_data(
+            version,
+            version_doi=False,
+            event='publish',  # Update existing Findable DOI
+        )
```
yarikoptic, line 143: both paths are identical... duplication is evil.

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -76,6 +85,12 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
         # chance to grab the Dandiset information before it is destroyed.
         audit.delete_dandiset(dandiset=dandiset, user=user)
 
+        # Dandisets with published versions cannot be deleted
+        # so only the Dandiset DOI needs to be deleted.
+        draft_version = dandiset.versions.filter(version='draft').first()
+        if draft_version and draft_version.doi is not None:
+            # Call DOI deletion prior to delete so if this raises, we do not proceed
+            doi.delete_or_hide_doi(draft_version.doi)
```
asmacdo, line 93: todo should still be a celery task, but we should make sure it passes or roll back transaction

yarikoptic (commented):
just posting an exception we observerd while trying locally.  Ideally should be more graceful I think

asmacdo:
+1 we can handle that more gracefully. This exception occurred because the Dandiset was created by the superuser without a name. So FWIW even if we relax how to_datacite sets that value it will still be an invalid DOI payload.

dandiapi/api/doi.py
```diff
@@ -1,86 +1,132 @@
...
+    Create a Draft DOI for a dandiset.
+
+    This is called during dandiset creation for public dandisets.
+    For embargoed dandisets, no DOI is created until unembargo.
+
+    Args:
+        draft_version: The draft version of the dandiset.
+    """
+    # Generate a Draft DOI (event=None)
+    dandiset_doi, dandiset_doi_payload = datacite_client.generate_doi_data(
```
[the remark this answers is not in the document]
asmacdo: I think we should ignore this in this PR, because I'd rather not change the functionality of `to_datacite` (this is actually a schema issue), and IMO we shouldn't worry about the developer-only case where the contributors do not have a name. (ie the superuser has been created without adding first and last name)

yarikoptic:
@waxlamp @jjnesbitt This item is on our weekly meeting notes "beyond 30 minutes of Roni" so we rarely get to it.  Could someone give the initial review? this "feature" was being in the kitchen for way too long with the design doc PR (#2012) and now this grew big and potentially growing conflicts.

dandiapi/api/services/dandiset/__init__.py
jjnesbitt: I don't agree that it must pass, or we should roll back the transaction. For starters, we would have to move all of this code into the celery task to ensure that. More practically, why should the outcome of the datacite API call influence this? Handling failure conditions for interacting with the datacite API should be done anyway. It wouldn't help anyone to prevent the deletion of a dandiset due to the datacite API being down (for example). We would just need to ensure that it is _eventually_ deleted.

For example, one way to approach this is to, upon dandiset deletion, place the DOI of the draft version into a table called `DeletableDOI` (or something). Then, a scheduled celery task could pull from values in this table, call out to datacite to delete the DOI, and then in the same transaction, delete that row (or mark is as "done") so it's not picked up on the next scheduled task.

This would ensure that the DOI _eventually_ gets deleted, even if there is a transient error during interaction with the datacite API. The same procedure could be implemented for DOI creation as well.

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -76,6 +81,12 @@ def delete_dandiset(*, user, dandiset: Dandiset) -> None:
         # chance to grab the Dandiset information before it is destroyed.
         audit.delete_dandiset(dandiset=dandiset, user=user)
 
+        # Dandisets with published versions cannot be deleted
+        # so only the Dandiset DOI needs to be deleted.
+        draft_version = dandiset.versions.filter(version='draft').first()
+        if draft_version and draft_version.doi is not None:
```
jjnesbitt, line 87: I have a couple of suggestions. First, we know a draft version will always exist on a dandiset, as it is created in the same transaction as the dandiset. So we don't need to check that it exists here.

Second, I actually think we should prevent deleting the dandiset until the DOI is generated. Imagine someone creates and then immediately deletes a dandiset, at the same time the celery task to create the DOI is running. This code block would be skipped, and the dandiset would still be deleted, but the DOI could have been created during that time. In that case, we would have "orphaned DOIs" to worry about.

I think it's acceptable to block deletion on draft DOI creation, and raise an appropriate error message from the request.

dandiapi/api/views/version.py
```diff
@@ -131,6 +135,18 @@ def update(self, request, **kwargs):
...
+                # to keep it in sync with the latest metadata
+                if not locked_version.dandiset.embargoed:
+                    transaction.on_commit(
+                        lambda: update_draft_version_doi_task.delay(locked_version.id)
+                    )
+                else:
+                    logger.debug(
+                        'Skipping DOI update for embargoed Dandiset %s.',
+                        locked_version.dandiset.identifier,
+                    )
```
jjnesbitt, line 148: If we're going to be keeping doi metadata up to date anytime a draft version is updated, then I'm even more in favor of what I proposed in my other comment (a table for keeping track of this). Doing it in this fashion is just a disaster waiting to happen in regards to synchronization of metadata.

jjnesbitt (changes requested):

dandiapi/api/datacite.py
```diff
@@ -0,0 +1,261 @@
...
+            )
+            response.raise_for_status()
+            # Return early on success
+            return doi  # noqa: TRY300
+        except requests.exceptions.HTTPError as e:
+            # HTTP 422 status code means DOI already exists
+            already_exists_code = 422
+            if e.response is not None and e.response.status_code == already_exists_code:
+                # Retry with PUT if DOI already exists
+                update_url = f'{self.api_url}/{doi}'
```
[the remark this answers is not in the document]
asmacdo: Yeah exactly

dandiapi/api/services/dandiset/__init__.py
asmacdo: My main concern was not to orphan the doi, but this seems like a reasonable approach

jjnesbitt:
Apologies for the delay, I've gotten back to working on this and have made significant progress. For ease of development I've made a separate branch based off this one, and will open that PR once it's ready (hopefully soon).

Issue #2296, opened 2025-04-10
import_dandisets management command was removed, still in dev docs


asmacdo:
The development docs reference this command several times https://github.com/dandi/dandi-archive/blob/master/DEVELOPMENT.md#import_dandisets

But the command was removed https://github.com/dandi/dandi-archive/pull/1335 

This command does seem useful for development to quickly reproduce a dandiset locally. Either we should add back or update the development docs. (Note: the PR states "This command is a bit broken and isn't used anymore." so if it is revived, does anyone know how it was a bit broken?)

kabilar:
cc @danlamanna

PR #1335, opened 2022-10-19
Delete unused management command import_dandisets.py

danlamanna:
This command is a bit broken and isn't used anymore.

PR #2351, opened 2025-04-23
Remove import_dandisets command from docs

asmacdo:
Fixes https://github.com/dandi/dandi-archive/issues/2296

Command itself was removed in https://github.com/dandi/dandi-archive/pull/1335


yarikoptic (approved):
"No brainer".  I think I would merge it in a day or so if nobody stops me

PR #2358, opened 2025-04-25
Update sorting controls for Dandisets page

### LINKED ISSUE #1764: Cant sort by size on mobile, button group bounces out of view
On mobile (firefox android) and try to sort by size. Trying to drag button group but it bounces back and hides the size button :(

Video:
https://github.com/dandi/dandi-archive/assets/12961499/8fb49d7a-5a7e-40b0-abe0-ad4ad49beb54

### LINKED ISSUE #2176: phone mode UI: no way to search on dandisets listing page
I can usually search, however the search bar is not accessible when the window is very narrow, as is the case when viewing the website on a phone:

Image

More broadly, is responsive design a priority for us? Do we care if the app is fully functional when viewed from a phone?

naglepuff:
Fix #1764 
Fix #2176 

### Changes

Replaces the current sorting controls on the Dandisets page with a more compact, mobile-friendly widget. Based on the design by @jtomeck outlined in #1764.

### Demo video
https://github.com/user-attachments/assets/153f07c2-dc81-4f8c-a239-b882aa40e0e6

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -11,6 +11,7 @@
         <template #activator="{ props: menuProps }">
           <v-icon
             v-bind="menuProps"
+            class="mr-6"
           >
             mdi-cog
           </v-icon>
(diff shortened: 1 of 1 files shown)


jjnesbitt:
Should these controls not be on the left side, instead of the right? Since it's already a prominent change in the dandisets page, it might be easier for people to adjust if it's around where the old sort options were. cc @jtomeck 

FWIW I really like the change.

naglepuff:
> Should these controls not be on the left side, instead of the right? Since it's already a prominent change in the dandisets page, it might be easier for people to adjust if it's around where the old sort options were. cc @jtomeck
> 
> FWIW I really like the change.

That's a fair point. I don't necessarily have a preference, so I'm interested in seeing what others have to say. It's a straightforward adjustment to make if there's a consensus that it should be moved.

web/src/components/DandisetsPage.vue
```diff
@@ -39,33 +40,53 @@
...
+          >
+            <span>Sorting Options</span>
+          </v-tooltip>
+          <v-menu activator="parent">
+            <v-list>
+              <v-list-item
+                v-for="(option, i) in sortingOptions"
+                :key="option.name"
+                :value="option"
+                :append-icon="option.name === sortingOptions[sortOption].name ? 'mdi-check' : ''"
```
mvandenburgh, line 64: Now that we're using `v-list` here, you should be able to use Vuetify's built in data binding capabilities (i.e., set `v-model="sortingOptions"` and let Vuetify set `sortOption` implicitly) instead of manually catching the `click` event and updating `sortOption` explicitly in `changeSort()`.

mvandenburgh, line 64: This is more subjective than the other two points, but you might consider using the [`append` slot](https://vuetifyjs.com/en/api/v-list-item/#slots-append) here instead of the `append-icon` prop. With that, you could avoid inlining a ternary expression into the template and instead just conditionally render the slot with `v-if`.

mvandenburgh (changes requested):

web/src/components/DandisetsPage.vue
naglepuff: See b2e7fee for my approach here. I couldn't seem to get `v-model` working with `v-list`, so it seemed like I had to wrap this bit in ` v-item-group`. It removes logic from the template and let me delete the `changeSort` function from the script, so there's definitely upside to using the approach, although it's still not obvious to me if this is _correct_ per se, or if I was missing something from the `VList` documentation.

naglepuff: Added to b2e7fee

web/src/components/DandisetsPage.vue
```diff
@@ -39,33 +40,61 @@
...
+        density="comfortable"
+        divided
+        class="btn-group--sort-options"
       >
-        <v-chip
-          v-for="(option, i) in sortingOptions"
-          :key="option.name"
-          @click="changeSort(i)"
+        <v-btn class="btn--sort-option">
+          <v-icon>mdi-sort-variant</v-icon>
```
mvandenburgh, line 51: Should the "sort" icon also vary based on whether it's set to ascending/descending? I'm not sure myself, as it seems like the arrow icon currently serves that purpose. But for some reason it looks off to me

i.e. should we do this?
```suggestion
          <v-icon v-if="sortDir === 1">
            mdi-sort-variant
          </v-icon>
          <v-icon v-else>
            mdi-sort-reverse-variant
          </v-icon>
```

@jtomeck was this intended in the original design?

mvandenburgh:
@naglepuff now that #2360 is merged, if you rebase this PR on `master` you should see those linting errors actually appear in CI

mvandenburgh (approved):

naglepuff:
@mvandenburgh I see you just approved so sorry to make you take another look, but I worked on implementing the suggestions discussed earlier today

web/src/components/DandisetsPage.vue
naglepuff: 601ac3d implements the design discussed earlier today, which swaps out the icon depending on the current sort order.

jjnesbitt:
Maybe we should make "Sort By" and "Order" slightly bigger and/or bold? It's a bit hard to tell the two sections apart since those texts are so small/light.
image

naglepuff:
@jjnesbitt how about this?
image

jjnesbitt:
That looks great!

PR #2361, opened 2025-04-28
Parametrize e2e tests in CI to run in both prod/dev modes

mvandenburgh:
Inspired by the discussion we had the other day about 12 factor and dev/prod parity - currently, we only run our e2e tests with the frontend served by the vite dev server and the backend server by `manage.py runserver`. While it's not possible to get anywhere near parity between CI and the Heroku prod deployment, we can at least (a) run the backend using gunicorn, and (b) serve the statically built Vue application with a static file server. With the latter, we've seen some divergence in the past between behavior with the Vite dev server and the final bundle produced by Rollup for prod, so this is still a strict improvement regardless.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -32,7 +32,16 @@ jobs:
         run: yarn run build
 
   test-e2e:
+    name: E2E tests (${{ matrix.django_server.name }})
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        django_server:
+          - name: manage.py runserver
+            command: python manage.py runserver
+          - name: gunicorn
+            command: python -m gunicorn --bind 0.0.0.0:8000 dandiapi.wsgi --timeout 25  # from /Procfile
     services:
       postgres:
         image: postgres:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/frontend-ci.yml
```diff
@@ -122,21 +134,20 @@ jobs:
...
           done
 
           # start the dandi-api server
-          python ../manage.py runserver &
+          ${{ matrix.mode.backend_server_command }} &
 
           # run the tests
-          npx playwright test
-      working-directory: e2e
+          cd e2e && npx playwright test
```
waxlamp, line 146: Just curious -- why this change?

waxlamp (commented):
This looks good to me. Withholding an explicit approval because I haven't actually tried this out. But you can either merge on your own authority, or (more likely) just wait until we have the new year's funding in place.

jjnesbitt (approved):
LGTM

naglepuff (approved):

.github/workflows/frontend-ci.yml
mvandenburgh: Since we also invoke `gunicorn` in this command, setting the working-directory to `e2e` makes it more complicated to invoke. My understanding is `gunicorn --bind 0.0.0.0:8000 ../dandiapi.wsgi` wouldn't work because `../dandiapi.wsgi` isn't a valid Python module path. So we'd need to set `PYTHONPATH` explicitly to call it from `/e2e`. It seemed cleaner to just set the working-directory to the root and `cd` into `e2e/` when needed.

PR #2367, opened 2025-04-30
Add Asset garbage collection design doc

mvandenburgh:
Outlines a design for Asset garbage collection. Implementing this will also unblock garbage collection from being run in general, so the idea is that we will begin running the `collect_garbage` script for assets/uploads/asset blobs after this feature lands.

### DIFF
--- a/doc/design/garbage-collection-assets.md
+++ b/doc/design/garbage-collection-assets.md
@@ -1,0 +1,34 @@
+# Asset Garbage Collection
+
+## Background
+
+Now that the [design for asset blob and upload garbage collection](./garbage-collection-uploads-asset-blobs-2.md) is deployed to staging, we are ready to implement Asset garbage collection. The garbage collection process in this design document only applies to “regular” assets; garbage collection of zarrs is not covered and is out of scope for this design doc.
+
+## What is an “Orphaned Asset”
+
+It’s important to precisely define what it means for an asset to be eligible for garbage collection (an “orphaned asset”).
+An orphaned asset has the following properties:
+- The asset is not currently associated with any Versions
+- The asset is older than 30 days
+
+## Integration with existing garbage collection routines
+
+Garbage collection routines for Asset Blobs and Uploads are already implemented and deployed (though not run yet). Currently, the number of Asset Blobs and Uploads that need to be garbage collected is not substantial. However, the act of running the Asset garbage collection routine will result in a significant amount of Asset Blobs being orphaned, which can then in turn be collected by the Asset Blob garbage collection routine.
+
+## Implementation notes
+
+A new module, `asset`, will be added to the `garbage_collection` service (see `../dandiapi/api/services/garbage_collection/`). This module will mirror the structure of the sibling `asset_blob` and `upload` modules and will be responsible for deleting orphaned assets and handling associated recordkeeping (i.e. adding a record to the `GarbageCollectionEvent` table for every delete event).
+
+We will introduce the feature initially by making it manually runnable via `manage.py collect_garbage` . This script [already exists](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/management/commands/collect_garbage.py) and currently supports asset blob and upload garbage collection, so it’s just a matter of integrating the `garbage_collection.asset` module into it once it’s implemented. Additionally, the script supports a dry run mode where it will log the number of items to be garbage collected, but not actually perform the delete operations. Once we’ve run it in production a few times and we’re ready, we’ll automate it with a cron via a Celery Beat task that calls the relevant service functions.
+
+As is the case with asset blobs and uploads, an orphaned asset will remain recoverable by DANDI Archive admins for 30 days after it is garbage collected.
+
+## Data
+
+The current amount of orphaned data in the system as of 4/29/25 is as follows:
+
+- Assets: 274,573 (354,040,196,605,150 bytes / 329,725.63 GB)
+- AssetBlobs: 88 (375,936,742,348 bytes / 350.12 GB)
+- Uploads: 962
+
+**Note that around 50% of the existing data in DANDI (About 329 TB out of 787 TB) consists of orphaned Assets and will be garbage collected.**


mvandenburgh:
@satra @yarikoptic PTAL and let me know if you have any questions or objections to this. I've also included the current counts of garbage-collectible assets and their total file sizes; it looks like we will be able to clean up a significant amount of stale data (almost 50%!).

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Data
+
+The current amount of orphaned data in the system as of 4/29/25 is as follows:
+
+- Assets: 274,573 (354,040,196,605,150 bytes / 329,725.63 GB)
+- AssetBlobs: 88 (375,936,742,348 bytes / 350.12 GB)
+- Uploads: 962
+
+**Note that around 50% of the existing data in DANDI (About 329 TB out of 787 TB) consists of orphaned Assets and will be garbage collected.**
```
satra, line 34: not sure i'm following this section since assets don't have bytes associated with them, just blobs do. an orphaned asset may be attached to a blob that is not orphaned (i.e. connected to another asset)

satra: but where are the numbers coming from? for example it says X of Y. i see X as the bytes associated, but where does Y come from?

yarikoptic: @mvandenburgh in above you mentioned "blobs", so your size estimate does not account for zarrs?

From https://github.com/dandi/example-notebooks/pull/114 it seems that we do have around 800TB of data across blobs and zarrs with some unknown (https://github.com/dandi/dandi-archive/issues/2380) amount for embargoed.  I just wanted to make sure that we have a better idea on what constitutes "About 329 TB out of 787 TB"

mvandenburgh: Zarrs are accounted for in the total archive size (787 TB at the time). I've updated the doc to reflect the current state of things (845 TB as of this morning).

waxlamp (approved):

doc/design/garbage-collection-assets.md
mvandenburgh: Correction to this - the "total data size" on the [dandiarchive.org](http://dandiarchive.org/) homepage excludes orphaned assets already. _Including_ orphaned assets, the total archive size is about 1.2 PB. So, removing the ~400TB of orphaned data will result in a final total assets size of ~800TB, which is in line with the number of NWB blobs and Zarrs.

mvandenburgh:
@satra @yarikoptic do you have any other questions or objections for this design?

satra:
overall this looks good to me. perhaps the one thing i would add in the manual process to start with is to ensure that after it's run perhaps we can do to ensure that every asset/blob of a published dandiset cannot be deleted. this would be equivalent to a terminate instance protection on aws ec2.

yarikoptic:
> we can do to ensure that every asset/blob of a published dandiset cannot be deleted.

do you mean @satra to similarly to our tagging for embargoed data/keys preventing any non-authtenticated access, we would add another tag, e.g. `read-only` and adjust S3 rules to Deny all Delete operations on keys with such a tag in https://github.com/dandi/dandi-infrastructure/blob/6287b22cb7d13ff12c6772b41ac41991eb9c9f7a/terraform/modules/dandiset_bucket/main.tf#L175 ?  FWIW if so -- sounds good to me. Besides that it might be yet another check to add to eventual `fsck` style set of commands to verify congruence between metadata in DB and state of the S3 (e.g. now it should raise errors for all blobs associated with released  assets, and potentially fix that if we add `fix` mode).

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Background
+
+Now that the [design for asset blob and upload garbage collection](./garbage-collection-uploads-asset-blobs-2.md) is deployed to staging, we are ready to implement Asset garbage collection. The garbage collection process in this design document only applies to “regular” assets; garbage collection of zarrs is not covered and is out of scope for this design doc.
+
+## What is an “Orphaned Asset”
+
+It’s important to precisely define what it means for an asset to be eligible for garbage collection (an “orphaned asset”).
+An orphaned asset has the following properties:
+- The asset is not currently associated with any Versions
```
yarikoptic, line 11:
```suggestion
- The asset is not currently associated with any Version
```

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+- The asset is older than 30 days
```
yarikoptic, line 12:
```suggestion
- The asset was not modified for more than 30 days
```

or

```suggestion
- The asset is older than 30 days based on `modified` datetime-stamp
```
as opposed to `created`, since we do have operations which modify assets in place and `modified` does differ from `created`

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Integration with existing garbage collection routines
+
+Garbage collection routines for Asset Blobs and Uploads are already implemented and deployed (though not run yet). Currently, the number of Asset Blobs and Uploads that need to be garbage collected is not substantial. However, the act of running the Asset garbage collection routine will result in a significant amount of Asset Blobs being orphaned, which can then in turn be collected by the Asset Blob garbage collection routine.
```
yarikoptic, line 16: interesting... so we do not even know on staging if blobs/uploads GC works?  

why not to deploy them in stages and not to enable already so we could decouple two changes and have more times for observing that first stage GC in action for longer?

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Implementation notes
+
+A new module, `asset`, will be added to the `garbage_collection` service (see `../dandiapi/api/services/garbage_collection/`). This module will mirror the structure of the sibling `asset_blob` and `upload` modules and will be responsible for deleting orphaned assets and handling associated recordkeeping (i.e. adding a record to the `GarbageCollectionEvent` table for every delete event).
```
yarikoptic, line 20: just to make sure -- we make conscious decision to **not** add any record to audit since it is not user/API invoked action, correct?

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+We will introduce the feature initially by making it manually runnable via `manage.py collect_garbage` . This script [already exists](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/management/commands/collect_garbage.py) and currently supports asset blob and upload garbage collection, so it’s just a matter of integrating the `garbage_collection.asset` module into it once it’s implemented. Additionally, the script supports a dry run mode where it will log the number of items to be garbage collected, but not actually perform the delete operations. Once we’ve run it in production a few times and we’re ready, we’ll automate it with a cron via a Celery Beat task that calls the relevant service functions.
```
yarikoptic, line 22: could we please add `asserts` into such a code that none of the to be removed assets have non-degenerate `publishedBy` and `datePublished` fields in their metadata?  might be that protection @satra [was talking about](https://github.com/dandi/dandi-archive/pull/2367#issuecomment-3155960243) in the main thread?

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+As is the case with asset blobs and uploads, an orphaned asset will remain recoverable by DANDI Archive admins for 30 days after it is garbage collected.
```
yarikoptic, line 24: will there be some interface and test implemented to verify that such "recovery" is actually possible? or otherwise - details of how recovery would be done?

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Data
+
+The current amount of orphaned data in the system as of 6/30/25 is as follows:
+
+- Assets: 278,186 (335,356.39 GB)
+- AssetBlobs: 103 (385.52 GB)
```
yarikoptic, line 31: numbers confuse me here abit, may be  this would clarify it

```suggestion
- AssetBlobs: 103 (385.52 GB)  (blobs in addition to those referenced by above Assets)
```

if correct?

yarikoptic (changes requested):
Mostly comments/questions with a single "typo" to fix

satra:
>FWIW if so -- sounds good to me.

@yarikoptic - yes, that would be the plan to put a tag "published" and deny removing any object that has such a tag.

doc/design/garbage-collection-assets.md
waxlamp: What's an example of degenerate values for those fields?

doc/design/garbage-collection-assets.md
waxlamp: This is pretty much how we plan to roll this out. First, run the blob GC on sandbox, and make sure it works as expected. Next, after asset GC is implemented and merged (as described in this design doc), we can run asset GC on sandbox, watch that it exposes the expected asset _blobs_, then run the blob GC again to make sure this works "end to end".

We can add a section to the design doc that describes this approach.

doc/design/garbage-collection-assets.md
waxlamp: The `GarbageCollectionEventRecord` table contains a JSON-serialized snapshot of the records being deleted, so we're keeping the maximum amount of information possible, which means it should be possible to recreate the deleted record. Currently, we're thinking of this as a manual intervention in case someone wants to "undelete" something, so we don't have immediate plans to create such an interface.

We will add a plan to carry out such a manual test to the design doc.

doc/design/garbage-collection-assets.md
waxlamp: If we follow Mike's plan in #2501, we won't need to worry about this (since deletion protection will be applied via legal hold to all objects that are part of published assets).

doc/design/garbage-collection-assets.md
```diff
@@ -0,0 +1,40 @@
...
+
+We will introduce the feature initially by making it manually runnable via `manage.py collect_garbage` . This script [already exists](https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/management/commands/collect_garbage.py) and currently supports asset blob and upload garbage collection, so it’s just a matter of integrating the `garbage_collection.asset` module into it once it’s implemented. Additionally, the script supports a dry run mode where it will log the number of items to be garbage collected, but not actually perform the delete operations. Once we’ve run it in production a few times and we’re ready, we’ll automate it with a cron via a Celery Beat task that calls the relevant service functions.
+
+As is the case with asset blobs and uploads, an orphaned asset will remain recoverable by DANDI Archive admins for 30 days after it is garbage collected.
+
+## Data
+
+The current amount of orphaned data in the system as of 6/30/25 is as follows:
```
waxlamp, line 28:
```suggestion
The current amount of orphaned data in the system as of 6/30/25 is as follows (bear in mind, deleting the Assets will likely expose more AssetBlobs to garbage collect; those AssetBlobs are _not_ reflected in this report):
```

PR #2368, opened 2025-04-30
Implement Asset garbage collection

mvandenburgh:
This PR implements asset garbage collection. See #2367 for the design doc/further explanation.


dandiapi/api/services/garbage_collection/asset.py
```diff
@@ -0,0 +1,69 @@
...
+            Version.objects.filter(
+                assets=OuterRef('id'),
+            ),
+        )
+    ).filter(
+        has_version=False,
+        published=False,
+        blob__isnull=False,  # only delete assets with blobs; zarrs are not supported yet
+        modified__lt=timezone.now() - ASSET_EXPIRATION_TIME,
+    )
```
jjnesbitt, line 42: Is there any downside to doing the following instead? This achieves the same effect using a `JOIN`, which is simpler if nothing else.

```suggestion
    return Asset.objects.filter(
        versions__isnull=False,
        published=False,
        blob__isnull=False,  # only delete assets with blobs; zarrs are not supported yet
        modified__lt=timezone.now() - ASSET_EXPIRATION_TIME,
    )
```

PR #2371, opened 2025-05-09
Fix broken/out-of-date docs links

NEStock:
- Fixed a few links that were referencing the old "numbered" pages
- Also replaced a few hardcoded references to https://docs.dandiarchive.org with the variable `dandiDocumentationUrl`

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -8,14 +8,14 @@
           <a
             target="_blank"
             rel="noopener"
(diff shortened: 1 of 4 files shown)


yarikoptic (approved):
Thank you! Looks good to me and on the preview.  Permitted CI runs (since then invited you to collaborators, I assume should run automagically in the future)

yarikoptic:
frontend CI seems unhappy since does not have yet a fix @NEStock gave in another PR... let's proceed.

PR #2373, opened 2025-05-13
Fix formatting of approved user message template

kabilar:
Since the current template results in the following message:

image

### DIFF
--- a/dandiapi/api/templates/api/mail/approved_user_message.txt
+++ b/dandiapi/api/templates/api/mail/approved_user_message.txt
@@ -1,8 +1,7 @@
 {% autoescape off %}
 Dear {{ greeting_name }},
 
-Your DANDI account has been approved. You can go to {{ dandi_web_app_url }} and login with your GitHub account
-to start creating dandisets and uploading data.
+Your DANDI account has been approved. You can go to {{ dandi_web_app_url }} and login with your GitHub account to start creating dandisets and uploading data.
 
 Please use the following links to post any questions or issues.
 


dandiapi/api/templates/api/mail/approved_user_message.txt
```diff
@@ -1,8 +1,7 @@
 {% autoescape off %}
 Dear {{ greeting_name }},
 
-Your DANDI account has been approved. You can go to {{ dandi_web_app_url }} and login with your GitHub account
-to start creating dandisets and uploading data.
+Your DANDI account has been approved. You can go to {{ dandi_web_app_url }} and login with your GitHub account to start creating dandisets and uploading data.
```
yarikoptic, line 4: note @NEStock : do you have vendorization patch for this by any chance? ;-)

yarikoptic (approved):

dandiapi/api/templates/api/mail/approved_user_message.txt
NEStock: @yarikoptic I believe we replaced instances of the text "DANDI" with "EMBER-DANDI" by hand.

And then `dandi_web_app_url` is set in dandi-infrastructure, so we were able to change that easily  (https://github.com/aplbrain/dandi-infrastructure/blob/f8c0f4b437aa0fcb6beee9d75602e8cae49f2c32/terraform/api.tf#L47)

Here is the EMBER version of this file for reference: https://github.com/aplbrain/dandi-archive/blob/527b818a664e1a4c8878b1131699ff5ec7376097/dandiapi/api/templates/api/mail/approved_user_message.txt#L4

PR #2377, opened 2025-05-14
BF: avoid clobbering of dict(..., **rec) type and always add a record for the author for new dandisets

### LINKED ISSUE #2376: Embargoed Dandisets missing ContactPerson
To reproduce:

Create embargoed ds, then unembargo.

`contributor: Value error, At least one contributor must have role ContactPerson`

It seems like for non-embargoed dandisets the ContactPerson is automatically populated from the creator, why isn't this the case for embargoed dandisets?

Image

yarikoptic:
- Requires_CLI_PR: https://github.com/dandi/dandi-cli/pull/1647

- Closes #2376

@asmacdo reported that Embergoed dandisets have a difference of not having their author automatically added to the metadata.  Identified via visual inspection this piece of code where there contributors list would be populated with the entry provided with the funder.  This code seems to be used only in `create_dandiset` so I do not see much of side-effects.

/end original

The CLI integration tests failed with the original bugfix due to the fixture dandiset behavior. When the `createsuperuser` management command is called, the user is created without a first_name or last_name. This can be seen in a development environment if `./manage.py create_dev_dandiset --owner <superuseremail>` is called before using the GUI to set a name for the user-- the resulting dandiset is invalid. After logging in to the GUI, the user gets a name, and subsequent calls to `create_dev_dandiset` usually (but not always) produce valid dandisets. 

In the CLI test fixtures, [this problem was overridden ](https://github.com/dandi/dandi-cli/pull/1647/files#diff-c81225ea8e187642e964c3f5b1ce7782143afa4f7022946d7285956b329dbcc5L581) by providing a complete list of contributors, excluding the superuser. However, with this bugfix, the original contributors are no longer clobbered, which left many of the dandisets with a contributor that had ", " as their name, which broke validation.

To get the tests to pass, the `createsuperuser` now accepts `DJANGO_SUPERUSER_FIRST_NAME` and `DJANGO_SUPERUSER_LAST_NAME`, and dandi-cli test fixture has been modified to use them, so fixture dandisets are now valid. I had to alter the expected values of the metadata to include the superuser in the cli tests as well.

There did not exist any mechanism to use a dandi-cli pull request in the tests, so I've added functionality to check the dandi-archive PR body for a required CLI-PR, and if present, checkout that branch and run the tests. I hope that this functionality will be useful for other cross-repo breaking changes, and can be extended for dandi-schema as well.

@yarikoptic this is ready for review but I am going to leave as a draft, because what should have just been a minor bugfix required such invasive changes, and the tests against `release` will be broken. IMO even if we choose to not keep this bugfix or the `createsuperuser` changes, I hope that "require dandi-cli pr" functionality will still be useful, and can be easily broken into a separate PR.

@asmacdo would continue to
- [x] add/adjust testing

### DIFF
--- a/dandiapi/api/management/commands/createsuperuser.py
+++ b/dandiapi/api/management/commands/createsuperuser.py
@@ -12,13 +12,12 @@ if TYPE_CHECKING:
     from composed_configuration._allauth_support.createsuperuser import EmailAsUsernameProxyUser
 
 
-def create_usermetadata(instance: EmailAsUsernameProxyUser, created: bool, **kwargs):
+def create_usermetadata(instance: EmailAsUsernameProxyUser, created: bool, **kwargs):  # noqa: FBT001
     if created and not hasattr(instance, '_usermetadata_created'):
         UserMetadata.objects.get_or_create(
-            user=instance, 
-            defaults={'status': UserMetadata.Status.APPROVED}
+            user=instance, defaults={'status': UserMetadata.Status.APPROVED}
         )
-        instance._usermetadata_created = True
+        instance._usermetadata_created = True  # noqa: SLF001
 
 
 class Command(createsuperuser.Command):
@@ -34,7 +33,7 @@ class Command(createsuperuser.Command):
         # Set first_name and last_name from environment variables if provided
         first_name = os.environ.get('DJANGO_SUPERUSER_FIRST_NAME')
         last_name = os.environ.get('DJANGO_SUPERUSER_LAST_NAME')
-        
+
         if first_name or last_name:
             # Find the user that was just created
             email = kwargs.get('email') or os.environ.get('DJANGO_SUPERUSER_EMAIL')


asmacdo:
I verified this against my branch, and it works as expected, still yet to adjust the testing though.

asmacdo:
@yarikoptic ready for review

dandiapi/api/tests/test_dandiset.py
```diff
@@ -571,7 +571,7 @@ def test_dandiset_rest_create_with_contributor(api_client, admin_user):
             'created': TIMESTAMP_RE,
             'modified': TIMESTAMP_RE,
         },
-        'contact_person': 'Jane Doe',
+        'contact_person': 'Doe, John',
```
asmacdo, line 574: This is a subtle behavior change-- Jane isnt "the" contact person anymore, shes 1 of 2, and John shows up first. @yarikoptic should there only be 1 contact person?

asmacdo:
I've made additional changes to the expected data in the CLI tests in

asmacdo:
Required PR logic only affects the master test for DANDI cli-integration, it doesn't affect release, so that failure is expected, but should serve as a warning that we should merge with caution.

.github/workflows/cli-integration.yml
```diff
@@ -78,9 +78,43 @@ jobs:
         if: matrix.dandi-version == 'release'
         run: pip install "dandi[test]"
 
-      - name: Install dev dandi
-        if: matrix.dandi-version == 'master'
-        run: pip install "dandi[test] @ git+https://github.com/dandi/dandi-cli"
+      - name: Extract CLI PR reference from description
```
yarikoptic, line 81: what is all this and why we need it for this fix testing?

PR #2387, opened 2025-05-19
TEMP: dummy change to assess current state of CI testing

yarikoptic:


yarikoptic:
ok, currently:
image

- for CI for frontend: #2388 
- dandi-cli: by accepting dandi-schema new release PR I forgot that dandi-cli could immediately install it and then server would refuse to accept those metadata records. That all results in `"message": "Metadata version 0.6.10 is not allowed. Allowed are: 0.6.9, 0.4.4, 0.5.1, 0.5.2, 0.6.0, 0.6.1, 0.6.2, 0.6.3, 0.6.4, 0.6.5, 0.6.6, 0.6.7, 0.6.8.",`  (note: 0.6.9 listed first, odd)

PR #2391, opened 2025-05-27
DONOTMERGE: small change to kick tests

asmacdo:


asmacdo:
Fails on released CLI, passes on master

PR #2395, opened 2025-05-28
Add script for CLI tests with local dandi-archive

asmacdo:
I found setting up CLI integration tests to be pretty fiddly, and so ultimately just created this script to do it for me. If others think this would be useful, it might make a nice dev tool.

Is scripts/ the right place for this?


kabilar:
Closing after discussion with @asmacdo.

PR #2396, opened 2025-06-04
Have the `api/info/` endpoint provide instance config info (replaced)

candleindark:
This issue is replaced by https://github.com/dandi/dandi-archive/pull/2430

This PR makes the instance config as defined in `dandischema.conf`, per https://github.com/dandi/dandi-schema/pull/294, available through the `api/info/` endpoint. The instance config, presented as serialized JSON object, can be reconstructed by calling `model_validate` or `model_validate_json` on `dandischema.conf.Config`.

## Release Notes

This PR makes the instance config as defined in `dandischema.conf`, per https://github.com/dandi/dandi-schema/pull/294, available through the `api/info/` endpoint.

### DIFF
--- a/dandiapi/api/tests/test_info.py
+++ b/dandiapi/api/tests/test_info.py
@@ -15,6 +15,7 @@ def test_rest_info(api_client):
     schema_url = f'http://{TEST_SERVER_NAME}{reverse("schema-dandiset-latest")}'
 
     assert resp.json() == {
+        'instance_config': {'instance_name': 'DANDI', 'doi_prefix': '10.80507'},
         'schema_version': settings.DANDI_SCHEMA_VERSION,
         'schema_url': schema_url,
         'version': __version__,


yarikoptic:
Testing failed since PR comes from fork.  @candleindark did you test locally?  should I resend from within this repo branch?

candleindark:
> Testing failed since PR comes from fork. @candleindark did you test locally? should I resend from within this repo branch?

@yarikoptic No, I haven't tested the solution yet (the solution is incomplete). Thanks for letting me know about the cause of the failure.

I would like to send a PR from this repo. However, I don't have write permission to this repo. Would you grant me the needed permission?

yarikoptic:
added you to https://github.com/orgs/dandi/teams/dandiarchive/members (attn note @dandi/archive-admin )

jjnesbitt:
Where is the branch/fork that adds this `get_instance_config` function to `dandischema`?

candleindark:
> Testing failed since PR comes from fork. @candleindark did you test locally? should I resend from within this repo branch?

I don't think the failures result from this PR being sent from another repo. #2430 produces the same failures. The failures are due to the `dandischema` dependency is not pointing the correct version, the `devendorize` branch, and, potentially, not setting the proper instance config via env vars that compatible with the existing tests.

candleindark:
> > Testing failed since PR comes from fork. @candleindark did you test locally? should I resend from within this repo branch?
> 
> I don't think the failures result from this PR being sent from another repo. #2430 produces the same failures. The failures are due to the `dandischema` dependency is not pointing the correct version, the `devendorize` branch, and, potentially, not setting the proper instance config via env vars that compatible with the existing tests.
> 
> Digging deeper ...

Never mind. @yarikoptic You were right initially. The `CI for backend` GitHub action does require the access to `secrets` and most likely requires the PR to be coming from this repo. I will use #2430 instead.

candleindark:
This is to be replaced by #2430 as discussed above.

PR #2397, opened 2025-06-04
fix: update robots.txt to allow Googlebot access to dandiset metadata

bendichter:
and disallow other bots from API endpoints

### DIFF
--- a/dandiapi/api/views/robots.py
+++ b/dandiapi/api/views/robots.py
@@ -4,5 +4,20 @@ from django.http import HttpResponse
 
 
 def robots_txt_view(request):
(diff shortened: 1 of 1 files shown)


yarikoptic:
I do not think it would have any (if much) of negative effects even if potentially not providing a full solution since I think our DLP also calls the `/assets/` endpoint for zarr for some reason (didn't check).  But with that , to check how better or worse google would start indexing our data, I will proceed with merge as soon as CI is green.

We also need to release dandi-archive finally to get both stating and this to use more recent dandischema, so adding the label

yarikoptic (approved):

PR #2400, opened 2025-06-04
Give explicit instructions on where to get code and check/set port for django DB

yarikoptic:
Otherwise people who are referred to this doc are unclear what to clone. And if they run DB locally, they run into a conflict, although docker-compose does not complain.

@tompollard please review/try out

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,7 +6,7 @@ You would need a local clone of the `dandi-archive` repository to develop on it.
 
 1. Run `git clone https://github.com/dandi/dandi-archive`
 2. Run `cd dandi-archive`
-3. Also make sure you are not already running a local PostgreSQL server on port 5432 (recommended), or
+3. Make sure your PostgreSQL port (5432) is available (recommended), or
 export `DOCKER_POSTGRES_PORT` environment variable to point to an alternative port.
 
 ## Develop with Docker (recommended quickstart)
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,5 +1,14 @@
 # Development Guidelines
 
+## Getting Started
+
+You would need a local clone of the `dandi-archive` repository to develop on it.
+
+1. Run `git clone https://github.com/dandi/dandi-archive`
+2. Run `cd dandi-archive`
+3. Also make sure you are not already running a local PostgreSQL server on port 5432 (recommended), or
+export `DOCKER_POSTGRES_PORT` environment variable to point to an alternative port.
+
 ## Develop with Docker (recommended quickstart)
 This is the simplest configuration for developers to start with.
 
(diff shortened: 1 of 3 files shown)


tompollard:
Looks good, thanks!

tompollard:
(perhaps nicer to suggest cloning `git@github.com:dandi/dandi-archive.git`, rather than https)

DEVELOPMENT.md
```diff
@@ -1,5 +1,14 @@
 # Development Guidelines
 
+## Getting Started
+
+You would need a local clone of the `dandi-archive` repository to develop on it.
+
+1. Run `git clone https://github.com/dandi/dandi-archive`
+2. Run `cd dandi-archive`
+3. Also make sure you are not already running a local PostgreSQL server on port 5432 (recommended), or
```
asmacdo, line 9: I dont think this makes sense here-- we arent running Postgress until `docker compose up` for quickstart, or ``docker compose -f ./docker-compose.yml up` for advanced setup.

dev/.env.docker-compose
```diff
@@ -1,5 +1,5 @@
 DJANGO_CONFIGURATION=DevelopmentConfiguration
-DJANGO_DATABASE_URL=postgres://postgres:postgres@postgres:5432/django
+DJANGO_DATABASE_URL=postgres://postgres:postgres@postgres:${DOCKER_POSTGRES_PORT}/django
```
asmacdo, line 2: Could this please default to 5432 so this "just works" for the simple case?

DEVELOPMENT.md
asmacdo: Nevermind, I misread. 

```suggestion
3. Make sure your PostgreSQL port (5432) is available (recommended), or
```

dev/.env.docker-compose
yarikoptic: yeap, thanks! will do/push the fix

asmacdo (approved):

yarikoptic:
> (perhaps nicer to suggest cloning `git@github.com:dandi/dandi-archive.git`, rather than https)

Some folks (e.g. on Window$) might still rely on http protocol and tokens. Moreover, I personally prefer cloning over http **always**, and then I have 

```
[url "git@github.com:"]
    pushinsteadOf = https://yarikoptic@github.com/
    pushinsteadOf = https://github.com/
    pushinsteadOf = http://github.com/
    pushInsteadOf = git://github.com/
```

in my `~/.gitconfig` --  this way for any repository I clone over http and if I decide to contribute somewhere (clone or not), I push over ssh.  a few more relevant configs could be found in https://github.com/con/catenate/blob/main/conboarding.md#conventions .  So, I will keep it this way for now, especially since some folks might do "DEVELOPMENT" not for contributing really.

PR #2402, opened 2025-06-05
Including trailing slash on assets frontend call to prevent redundancy

aaronkanzer:
Cc @yarikoptic


yarikoptic:
tests pass, and me testing on a preview instance -- all good, no original request, only the one with `/`  (console logs errors but datacite related, I do not think relate to this anyhow )

image

PR #2414, opened 2025-06-19
ENH: add filtering capabilities to NestedAssetViewSet

### LINKED ISSUE #2413: FEATURE - REST API: add ordering to /dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/
It would be great to get assets ordered by date modified, similar to how we can do this with dandisets.

This came up for me because I want to build a system that syncs metadata, with a third party system. I would like the sync command to be able to easily determine what assets have been modified since the last sync. This is already possible with dandisets. I think it's an oversight that this has not been implemented with assets.

bendichter:
fix #2413

The implementation involved:

1. Adding filter backends and the filterset class to the `NestedAssetViewSet` class in `dandiapi/api/views/asset.py`:
   ```python
   filter_backends = [filters.DjangoFilterBackend]
   filterset_class = AssetFilter
   ```
This change enables the existing `AssetFilter` class's ordering capability, which already supported ordering by:
   - `created` (creation date)
   - `modified` (modification date)
   - `path` (asset path)

2. I added comprehensive tests to verify the ordering functionality works correctly for different scenarios:
   - Ordering by created, modified, and path fields in both ascending and descending order
   - Testing with authenticated users
   - Testing with embargoed assets

Users can now order assets by date modified (and other fields) in the REST API endpoint `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/` by adding the `order` query parameter. For example:
- `/dandisets/123/versions/draft/assets/?order=modified` (ascending order by modification date)
- `/dandisets/123/versions/draft/assets/?order=-modified` (descending order by modification date)

### DIFF
--- a/dandiapi/api/tests/test_asset.py
+++ b/dandiapi/api/tests/test_asset.py
@@ -436,6 +436,132 @@ def test_asset_rest_list_ordering(api_client, version, asset_factory, order_para
     assert result_paths == ordering
 
 
+@pytest.mark.parametrize(
+    ('order_param', 'expected_order'),
+    [
+        ('created', ['asset1', 'asset2', 'asset3']),
+        ('-created', ['asset3', 'asset2', 'asset1']),
+        ('modified', ['asset1', 'asset2', 'asset3']),
+        ('-modified', ['asset3', 'asset2', 'asset1']),
+        ('path', ['asset1', 'asset3', 'asset2']),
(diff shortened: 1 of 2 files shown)


jjnesbitt (approved):
LGTM, works fine and arguably should have been in `NestedAssetViewSet` from the start.

PR #2420, opened 2025-06-27
Update to Python 3.13

mvandenburgh:
Updates the backend to require Python 3.13. We're currently running 3.11, so this will let us take advantage of new language features added since then.


mvandenburgh:
@waxlamp since you added the nix flake, can you test this branch and see if the change I made to it works?

yarikoptic:
>  this will let us take advantage of new language features added since then.

just wanted to alert that there is a good number of PRs to code which require reviews etc, so IMHO there should be no something like [pyupgrade](https://pypi.org/project/pyupgrade/) across codebase to raise potential for the conflicts.

mvandenburgh:
> > this will let us take advantage of new language features added since then.
> 
> just wanted to alert that there is a good number of PRs to code which require reviews etc, so IMHO there should be no something like [pyupgrade](https://pypi.org/project/pyupgrade/) across codebase to raise potential for the conflicts.

More specifically, I wanted to use https://docs.python.org/3/library/itertools.html#itertools.batched (which was added in 3.12) to avoid pulling in an external library or implementing it myself. Agreed that we don't want to make sweeping changes while other PRs are open. `ruff` actually includes all of `pyupgrade`'s rules, and no changes were suggested by it for this PR.

jjnesbitt (approved):
Installed and run locally without any issue :+1:

waxlamp (approved):
> waxlamp since you added the nix flake, can you test this branch and see if the change I made to it works?

I added a change the Flake so that it works with Python 3.13. LGTM!

PR #2426, opened 2025-07-02
Remove `ProcessedS3Log` model

mvandenburgh:
Follow up to #2425.

As of #2425, this model is no longer used. Note: **merging this PR will cause the associated DB table to be dropped**.


mvandenburgh:
> @mvandenburgh Will you have to change anything in the schema regarding the 'download_count'?

I missed that the first time around. Fixed https://github.com/dandi/dandi-archive/pull/2426/commits/a60dbf3feaba16e112ccdab70cceff575e88b267. Thanks!

mvandenburgh:
> Since there's essentially no use for the `analytics` app anymore, is this PR mainly to remove the `ProcessedS3Log` model from the database, so that the app itself can be removed entirely afterwards?

Exactly. The `analytics` app needs to exist for the migrations in this PR to run, so once this is merged and deployed we can follow up with a third and final PR that removes the rest of the app.

CodyCBakerPhD:
> which we will store in S3 alongside the rest of the backups we've taken in the past

Under which subdirectory of which S3 bucket can I find these?

mvandenburgh:
> > which we will store in S3 alongside the rest of the backups we've taken in the past
> 
> Under which subdirectory of which S3 bucket can I find these?

I DM'd it to you on slack. The bucket isn't publicly accessible, but I figured it's still worth keeping the name private :man_shrugging:

yarikoptic:
FWIW, no objections on my end

jjnesbitt (approved):

yarikoptic:
@satra pinged on slack and we have  @dandi/ember (I invited more there)  on migration needed.

PR #2427, opened 2025-07-02
Set `statement_timeout` in materialized view query

mvandenburgh:
Using `statement_timeout` will ensure that this SQL query finishes executing before the Celery task exits and allows other queries to run. Alongside #2425, this should prevent the `another command is already in progress` error that we've been seeing in Sentry from occurring.


jjnesbitt:
When I run this locally I immediately get the following error:

```
django.db.utils.OperationalError: canceling statement due to statement timeout
```

jjnesbitt:
Ah, this may be the culprit. From the postgres docs:

>  If this value is specified without units, it is taken as milliseconds

If I set this value using milliseconds, it seems to behave as expected.

dandiapi/api/tasks/scheduled.py
```diff
@@ -127,7 +128,15 @@ def refresh_materialized_view_search() -> None:
     updated without locking the table.
     """
     with connection.cursor() as cursor:
+        # This query may take a long time, so we explicitly set a timeout with
+        # postgres that is slightly less than the Celery soft_time_limit to
+        # avoid it running too long.
+        # Note that setting Celery time limits alone is not sufficient. If the task times out,
+        # Celery will stop the Python execution, but not the DB query.
+        cursor.execute('BEGIN;')
+        cursor.execute('SET LOCAL statement_timeout = %s;', [REFRESH_MATERIALIZED_VIEW_TIMEOUT - 5])
```
jjnesbitt, line 137:
```suggestion
        cursor.execute('SET LOCAL statement_timeout = %s;', [(REFRESH_MATERIALIZED_VIEW_TIMEOUT - 5) * 1000])  # Value must be in milliseconds
```

mvandenburgh:
> Ah, this may be the culprit. From the postgres docs:
> 
> > If this value is specified without units, it is taken as milliseconds
> 
> If I set this value using milliseconds, it seems to behave as expected.

Ah, that makes sense. I was testing this out with extremely short `REFRESH_MATERIALIZED_VIEW_TIMEOUT` values so it went unnoticed. Fixed

dandiapi/api/tasks/scheduled.py
```diff
@@ -127,7 +128,17 @@ def refresh_materialized_view_search() -> None:
...
     """
     with connection.cursor() as cursor:
+        # This query may take a long time, so we explicitly set a timeout with
+        # postgres that is slightly less than the Celery soft_time_limit to
+        # avoid it running too long.
+        # Note that setting Celery time limits alone is not sufficient. If the task times out,
+        # Celery will stop the Python execution, but not the DB query.
+        cursor.execute('BEGIN;')
+        cursor.execute(
+            'SET LOCAL statement_timeout = %s;', [(REFRESH_MATERIALIZED_VIEW_TIMEOUT * 1000) - 5]
```
jjnesbitt, line 138: Just so you know this is setting the timeout to 5 milliseconds shorter than `REFRESH_MATERIALIZED_VIEW_TIMEOUT`, not 5 seconds. Not sure if that's what you intended.

mvandenburgh: Good catch. It was not intended, I was bouncing between a couple PRs and did it quickly without thinking enough :sweat_smile:

jjnesbitt (approved):

PR #2429, opened 2025-07-07
Use asset path aggregation for Asset.total_size

### LINKED ISSUE #2416: Move `Asset.total_size` to be implemented with AssetPaths
The `/stats` endpoint makes use of the `Asset.total_size` function, which does a summation over essentially all assets in the system. While this isn't terribly slow, since the asset paths implementation already has those number pre-computed, it would be very cheap to instead implement `total_size` using those models.

jjnesbitt:
Closes #2416

Since we already pre-compute asset size and count aggregations with the `AssetPath` model, `Asset.total_size` can sum over all root paths in the system (draft versions only), instead of over all assets.

This seems to reduce the value of `total_size` in the system by about 20TB. Prior to this change we were summing over all distinct asset blobs and zarrs that were contained within _any_ version, which can contain inconsistencies, such as assets that were removed between versions.

### DIFF
--- a/dandiapi/api/models/asset.py
+++ b/dandiapi/api/models/asset.py
@@ -16,6 +16,7 @@ from django.db.models import Q
 from django.urls import reverse
 from django_extensions.db.models import TimeStampedModel
 
+from dandiapi.api.asset_paths import get_root_paths_many
 from dandiapi.api.models.metadata import PublishableMetadataMixin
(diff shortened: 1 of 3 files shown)


jjnesbitt:
Closing as superseded by https://github.com/dandi/dandi-archive/pull/2435. While this improvement could still be made, the runtime of the aggregation no longer poses a significant issue, and there are implementation challenges to overcome.

Issue #2382, opened 2025-05-15
De/re-vendorize dandi-archive


yarikoptic:
- NB Adapted from https://github.com/dandi/dandi-archive/blob/401108fb6f052ec9a71db39ed4bf415a67eb3366/doc/design/vendored-schema-1.md

Work is ongoing in [dandi-schema:PR#294](https://github.com/dandi/dandi-schema/pull/294) to make vendorization of the schema configurable.
That would result in `dandi/schema` JSONSchema serializations becoming generally de-vendorized.
And it will be `dandi-archive` instance responsibility to re-vendorize, which would primarily consist in changing regular expressions more restrictive, via configuration/environment-variables.

ATM we still have some hardcoded vendorization in dandi-archive code:

#### Backend

Excluding some where we might want to vendorize too (e.g. email subjects etc):

```shell
❯ git grep DANDI: -- dandiapi | grep -v -e test_ -e 'subject=' -e 'verbose_name'
dandiapi/api/models/version.py:            'identifier': f'DANDI:{self.dandiset.identifier}',
dandiapi/api/models/version.py:            'id': f'DANDI:{self.dandiset.identifier}/{self.version}',
dandiapi/api/services/metadata/__init__.py:            f'DANDI:{publishable_version.dandiset.identifier}/{publishable_version.version}'
dandiapi/api/tests/fuzzy.py:DANDISET_SCHEMA_ID_RE = Re(r'DANDI:\d{6}')
dandiapi/api/views/dandiset.py:            if identifier.startswith('DANDI:'):
```

#### Web frontend

```shell
❯ git grep DANDI: -- web | grep -v -e test_ -e 'subject=' -e 'verbose_name'
web/src/components/DandisetList.vue:        DANDI:<b>{{ item.dandiset.identifier }}</b>
web/src/stores/dandiset.ts:      schema['properties']['identifier']['pattern'] = '^DANDI:\\d{6}$'
web/src/views/DandisetLandingView/DownloadDialog.vue:  // Use the special 'DANDI:' url prefix if appropriate.
web/src/views/DandisetLandingView/DownloadDialog.vue:  const dandiUrl = `DANDI:${identifier}`;
```

Should rely on information to come from API, hence:
- https://github.com/dandi/dandi-archive/pull/2396

[...]

PR #2430, opened 2025-07-08
Have the `api/info/` endpoint provide instance config info

candleindark:
This PR is a replacement of #2396, which can't be properly tested since it is coming from another repo.

This PR makes the instance config as defined in `dandischema.conf`, per https://github.com/dandi/dandi-schema/pull/294, available through the `api/info/` endpoint. The instance config, presented as serialized JSON object, can be reconstructed by calling `model_validate` or `model_validate_json` on `dandischema.conf.Config`.

**TODOs**:

- [ ] Remove the "tmp: temporarily set the dandischema dependency to the devendorize branch" commit which is a temporary measure for testing this solution.

##Release Notes

This PR makes the instance config as defined in `dandischema.conf`, per https://github.com/dandi/dandi-schema/pull/294, available through the `api/info/` endpoint.

### DIFF
--- a/dandiapi/api/tests/test_info.py
+++ b/dandiapi/api/tests/test_info.py
@@ -15,6 +15,7 @@ def test_rest_info(api_client):
     schema_url = f'http://{TEST_SERVER_NAME}{reverse("schema-dandiset-latest")}'
 
     assert resp.json() == {
+        'instance_config': {'instance_name': 'DANDI', 'doi_prefix': '10.80507'},
         'schema_version': settings.DANDI_SCHEMA_VERSION,
         'schema_url': schema_url,
         'version': __version__,
--- a/dandiapi/api/views/info.py
+++ b/dandiapi/api/views/info.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from dandischema.conf import get_instance_config
 from django.conf import settings
 from django.urls import reverse
 from drf_yasg.utils import no_body, swagger_auto_schema
@@ -8,6 +9,8 @@ from rest_framework.decorators import api_view
 from rest_framework.response import Response
 
 from dandiapi import __version__
+
+_INSTANCE_CONFIG = get_instance_config()
 
 
 def get_schema_url(request):
@@ -40,6 +43,9 @@ class ApiInfoSerializer(serializers.Serializer):
             }
         )
 
+    # Instance Configuration
+    instance_config = serializers.JSONField()
+
     # Schema
     schema_version = serializers.CharField()
     schema_url = serializers.URLField()
@@ -61,6 +67,13 @@ def info_view(request):
     api_url = f'{settings.DANDI_API_URL}/api'
     serializer = ApiInfoSerializer(
         data={
+            'instance_config': _INSTANCE_CONFIG.model_dump(
+                # Not excluding any `None` values in this object because the `None` values are
+                # needed to set the instance configuration in dandi-cli properly since
+                # a corresponding environment variable is used to set the instance configuration if
+                # the argument for a particular field is not provided.
+                mode='json'
+            ),
             'schema_version': settings.DANDI_SCHEMA_VERSION,
             'schema_url': get_schema_url(request),
             'version': __version__,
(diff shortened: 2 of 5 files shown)


dandiapi/api/tests/test_info.py
```diff
@@ -15,6 +15,7 @@ def test_rest_info(api_client):
     schema_url = f'http://{TEST_SERVER_NAME}{reverse("schema-dandiset-latest")}'
 
     assert resp.json() == {
+        'instance_config': {'instance_name': 'DANDI', 'doi_prefix': '10.80507'},
```
yarikoptic, line 18: @mvandenburgh could you please look at this PR and this change to the `/info/` API endpoint output?
It is to be included in the larger 
- #2386 

The point here is to expose configuration relevant to the instance of dandi-archive, so client and others could properly discover instance names etc.

Those two variables are the ones we are to define ATM but there could be more later on.  But now I started to wonder if we should just add them flat, instead of nesting into `instance_config`?  wdyt @candleindark

candleindark: > But now I started to wonder if we should just add them flat, instead of nesting into `instance_config`?

I prefer have all the instance config variables nested in the `instance_config` field in the JSON response at the `api/info/ endpoint. In that way,
1. The value of the `instance_config` field can be readily used construct a `dandischema.conf.Config` define in https://github.com/dandi/dandi-schema/pull/294.
2. Validation of the value of the `instance_config` field is delegated to the `dandischema.conf.Config` Pydantic model. On the other hand, transferring the definition of the variables in the `instance_config` field to the upper level of the `api/info/` response would requires redefinition of the fields in `dandischema.conf.Config` using Django's lingo, which is a duplication of logic that can easily get out of hand when add more fields to `dandischema.conf.Config` in the future.

yarikoptic:
please rebase to resolve conflicts and ensure that there is no new test fails

yarikoptic:
as we decided not to tune dandi-cli ATM for specific instance configs, let's postone this altogether

candleindark:
Reopening this one per https://github.com/dandi/dandi-archive/issues/2382#issuecomment-3176060536

pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
   "celery",
   "dandi>=0.60.0",  # minimal version is also provided in API /info
   # Pin dandischema to exact version to make explicit which schema version is being used
-  "dandischema==0.11.1",  # schema version 0.6.10
+  "dandischema @ git+https://github.com/dandi/dandi-schema.git@devendorize",
```
jjnesbitt, line 15: This is temporary, correct? We'll want to set this back to the necessary dandischema version before this is merged.

jjnesbitt:
> @jjnesbitt do you see anything to be changed in this PR or should we absorb into that
> 
> * [Local runtime schema serialization endpoints #2386](https://github.com/dandi/dandi-archive/pull/2386)
>   so we could approach web frontend with both endpoints available ?

I left one comment. Besides that, I'd like to see the multiple `tmp` commits squashed into one commit that we can drop once we're ready to merge the main branch into master. Also, the temp `dandi-schema` branch being pointed at needs to be fixed (`devendorize-init-config-with-field-name` has been merged into `devendorize`).


What is the timeline for https://github.com/dandi/dandi-schema/pull/294 being merged?

yarikoptic:
I think we should list a clear order of PR merges across all repos in e.g. 
- https://github.com/dandi/dandi-schema/issues/299

@candleindark please adjust original issue description with clear stages, as e.g. we should get this and #2386 get merged/released/tested, infrastructure ones merged, and then do final testing of https://github.com/dandi/dandi-schema/pull/294 and merge/release it too, IIRC the steps correctly.

dandiapi/api/views/info.py
```diff
@@ -11,6 +12,8 @@
 
 from dandiapi import __version__
 
+_INSTANCE_CONFIG = get_instance_config()
```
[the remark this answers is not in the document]
candleindark: We need to modify this solution a bit since the return of `get_schema_instance_config()` is not recommended to be stored in ‎`dandiapi/settings/base.py` per https://github.com/dandi/dandi-archive/pull/2584/files#r2436938965.  I will try to do some monkey patching in https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/checks.py instead to implement the solution.

PR #2434, opened 2025-07-09
Fix Meditor rendering bugs

### LINKED ISSUE #2432: Clarifying relationship between resources and dandisets in help text
From the help text in the "Resource relation" metadata field, it is not clear the "direction" of the relation:
Image

I think it might be due to a incorrect rendering of:
https://github.com/dandi/dandi-schema/blob/f1c937037cfe578ec3c19dc9ed9b44d40c8ab519/dandischema/models.py#L1051

### LINKED ISSUE #2433: Meditor help text does not render correctly in development
When running the frontend in development mode, the help text in the meditor does not render properly when the icons are clicked.

At first glance, they appear to do nothing. But upon closer examination, the popups _are_ appearing, just underneath the parent meditor popup (see screenshot below) 
Image

Strangely, this appears to happen in development only. https://dandiarchive.org/ and https://gui-staging.dandiarchive.org/ do not have this issue.

mvandenburgh:
Fixes #2433 
Fixes #2432 

I discovered #2433 when trying to debug #2432 locally. See code comments for more details on each bug.

### DIFF
--- a/web/src/utils/schema.ts
+++ b/web/src/utils/schema.ts
@@ -1,0 +1,38 @@
+import type { JSONSchema } from "@apidevtools/json-schema-ref-parser";
+
+/**
+ * Applies various miscellaneous fixes/massages the Dandiset metadata JSON Schema
+ * to make it compatible with VJSF.
+ *
+ * @param schema The Dandiset metadata JSON Schema
+ */
+export function fixSchema(schema: JSONSchema) {
+  let deepCopySchema = JSON.parse(JSON.stringify(schema));
+
+  // TODO: Fix this upstream in the schema
+  deepCopySchema['properties']['identifier']['pattern'] = '^DANDI:\\d{6}$'
+  deepCopySchema['$defs']['Project']['properties']['wasAssociatedWith']['items']['oneOf'][2]['properties']['identifier']['pattern'] = '^RRID:.*'
+
+  // vjsf renders < and > as HTML tags, which breaks rendering of the schema
+  // in places like help text. For example, given this description for a field:
+  //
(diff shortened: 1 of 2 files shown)


satra:
thanks @mvandenburgh - i can confirm that both issues are no longer present in the deployment.

web/src/utils/schema.ts
```diff
@@ -0,0 +1,38 @@
...
+  //
+  // <p>This relation should satisfy: dandiset <relation>resource</relation>.</p>
+  //
+  // To fix this, we escape the < and > characters that look like HTML tags.
+  deepCopySchema = JSON.parse(
+    JSON.stringify(schema)
+      .replace(
+        /<\/?[a-zA-Z][^>\s]*[^>]*>/g, t => t.replace(/</g, '\\\\<').replace(/>/g, '\\\\>')
+      )
+  );
```
jjnesbitt, line 35: Because you're assigning `deepCopySchema` here, but passing in `schema` to `JSON.stringify`, aren't the above two assignments erased? Should this be:
```suggestion
  deepCopySchema = JSON.parse(
    JSON.stringify(deepCopySchema)
      .replace(
        /<\/?[a-zA-Z][^>\s]*[^>]*>/g, t => t.replace(/</g, '\\\\<').replace(/>/g, '\\\\>')
      )
  );
```

mvandenburgh: Good catch. Now I'm confused as to how the meditor was rendering properly without those fixes being applied. Maybe they aren't needed anymore. I'll file an issue to look into that more closely

PR #2438, opened 2025-07-11
Remove the rest of `dandiapi.analytics`

mvandenburgh:
Now that we've removed the associated database tables (#2426), we can remove the rest of the `dandiapi.analytics` app.


jjnesbitt (approved):
LGTM

PR #2439, opened 2025-07-14
Redirect `gui-staging` URLs to `sandbox`

mvandenburgh:
Configures the existing netlify `redirector` app to also redirect gui-staging URLs to sandbox.

Docs: https://docs.netlify.com/configure-builds/file-based-configuration/#redirects

Notes:
- The `gui-staging-dandiarchive-org` Netlify app will need to be reconfigured to point to the `redirector/` directory once we are ready to switch over to sandbox
- The docs don't have any examples of specifying a full domain for the `from` URL as we do here, but this post does have one https://answers.netlify.com/t/need-to-redirect-custom-domain-netlify/75184/3 so there is some reason to think this will work, but we won't know for sure until this is merged.

### DIFF
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -3,3 +3,7 @@ from = "https://gui.dandiarchive.org/*"
 to = "https://dandiarchive.org/:splat"
 status = 301
 
+[[redirects]]
+from = "https://gui-staging.dandiarchive.org/*"
+to = "https://sandbox.dandiarchive.org/:splat"
+status = 301
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -1,4 +1,5 @@
 [[redirects]]
-from = "/*"
+from = "https://gui.dandiarchive.org/*"
 to = "https://dandiarchive.org/:splat"
 status = 301
+


redirector/netlify.toml
```diff
@@ -1,4 +1,9 @@
 [[redirects]]
-from = "/*"
```
jjnesbitt, line 2: I don't quite understand why this line is being changed / what this line change actually does.

mvandenburgh: Sorry, I thought I put the detailed execution plan somewhere on github but I can't find it now. I added it here https://github.com/dandi/dandi-archive/issues/1030#issuecomment-3070458070.

We want `gui-staging.dandiarchive.org` to redirect to `sandbox.dandiarchive.org` going forward, so that means that the `gui-staging-dandiarchive-org` Netlify app will turn into a simple redirector like we currently do for the `gui-dandiarchive-org` project. So, since both `gui-staging-dandiarchive-org` and `gui-dandiarchive-org` will be deployed from `redirector/netlify.toml`, it's necessary to differentiate the two by specifying full URLs instead of just `/*`

jjnesbitt (approved):

PR #2441, opened 2025-07-14
Update hardcoded staging URL

mvandenburgh:
I missed this in #2439. We should probably move this to a env var and set it via Terraform using a reference to the associated `aws_route53_record` resource, but this fixes it for the time being.


mvandenburgh:
Filed an issue #2441

PR #2229, opened 2025-03-21
Add staging rename design doc

waxlamp:

PR #2445, opened 2025-07-21
Update staging rename redirect plan

mvandenburgh:
The consensus was to not bother setting up a new redirection service for the old API staging URL (see GitHub discussion here - https://github.com/dandi/dandi-archive/pull/2229#discussion_r2010211743) but this was never encoded in the design document. Since I am now proceeding with the implementation for this (https://github.com/dandi/dandi-infrastructure/pull/231), I would like to make sure our exact intentions are codified before making any changes to URLs.


waxlamp:
Because it's very low cost to do so and may save some pain, we'll follow the language of the design doc as-written and put in a redirection for staging API as well. (@mvandenburgh signed off on this decision.)

PR #2449, opened 2025-07-24
Add HTTP redirect for staging server

mvandenburgh:
Per the design doc, it uses code 308 to preserve the request method.

### DIFF
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -7,3 +7,8 @@ status = 301
 from = "https://gui-staging.dandiarchive.org/*"
 to = "https://sandbox.dandiarchive.org/:splat"
 status = 301
+
+[[redirects]]
+from = "https://api-staging.dandiarchive.org/*"
+to = "https://api.sandbox.dandiarchive.org/:splat"
+status = 308


redirector/netlify.toml
```diff
@@ -7,3 +7,8 @@ status = 301
 from = "https://gui-staging.dandiarchive.org/*"
 to = "https://sandbox.dandiarchive.org/:splat"
 status = 301
+
+[[redirects]]
+from = "https://api-staging.dandiarchive.org/*"
+to = "https://api.sandbox.dandiarchive.org/:splat"
+status = 308
```
jjnesbitt, line 14: I don't understand how this would ever take effect, since netlify doesn't host the API, it only hosts the GUI?

jjnesbitt: I was confused by that infra PR still, but I think I understand now. There is a _new_ project in Netlify, called `api-staging-dandiarchive-org`, which also serves the `redirector` directory. The infra PR points https://api-staging.dandiarchive.org to that newly created netlify app, which uses this redirector, and redirects all requests for https://api-staging.dandiarchive.org to https://api.sandbox.dandiarchive.org.

Is that correct?

mvandenburgh: Yes, that's exactly correct

jjnesbitt (approved):

PR #2451, opened 2025-07-28
Add middleware/logger to log username on every request

mvandenburgh:
Configures Django's logging to append usernames to HTTP request logs if the request is associated with an authenticated user.

There isn't a clean way to get access to the `request.user` object globally, so I've stored it in a `ContextVar` as the Python logging cookbook outlines here https://docs.python.org/3/howto/logging-cookbook.html#use-of-contextvars.


mvandenburgh:
I updated this to modify the existing logger instead of adding a new one.

PR #2456, opened 2025-08-03
Extend allowed time to fulfill PUT upload to half-hour from 10 minutes

yarikoptic:
Both numbers are arbitrary. It could be of security hazard to make them arbitrarily long happen such URLs leak/get shared publicly.  But in both cases durations are still reasonably short.

On some network(s) though we can run into situation that we do not upload "fast enough" and thus running into 403s which could be (I think) avoided by allowing for longer period of time to initiate uploads.  See

   https://github.com/dandi/dandi-cli/issues/1662

Sure thing a proper solution is TODO (as good old comment says) - for client to react and to re-request a new pull of URLs etc.  That is to be done, but I think this PR also would make a world a better place meanwhile ;)

### DIFF
--- a/dandiapi/api/storage.py
+++ b/dandiapi/api/storage.py
@@ -22,6 +22,8 @@ from storages.backends.s3 import S3Storage
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
+PRESIGNED_URL_TIMEOUT = timedelta(minutes=30)
+
 
 class ChecksumCalculatorFile:
     """File-like object that calculates the checksum of everything written to it."""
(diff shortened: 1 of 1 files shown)


jjnesbitt:
@yarikoptic Are you running into the case where you request many URLs at once, and then by the time a URL at the "end of the line" starts uploading data, it's almost expired?

If that's the case, I'd perhaps suggest making the "chunk size" of the URLs you're requesting smaller, to mitigate this issue. Nevertheless, I don't see a huge issue with this. While I don't think it's ideal to increase the timeout, I think the risk is fairly small.

So, I approve this. @mvandenburgh any objections?

yarikoptic:
> @yarikoptic Are you running into the case where you request many URLs at once, and then by the time a URL at the "end of the line" starts uploading data, it's almost expired?
> 
> If that's the case, I'd perhaps suggest making the "chunk size" of the URLs you're requesting smaller, to mitigate this issue. Nevertheless, I don't see a huge issue with this. While I don't think it's ideal to increase the timeout, I think the risk is fairly small.

They do expire and indeed there are multiple ways to mitigate. We even added now retry cycles which would remint new URLs when getting such 403s.  But I still think extending the lifetime for those URLs could be of benefit to avoid avoidable API calls, so thanks for merging!

PR #2457, opened 2025-08-04
Adjust robots.txt for backend to fixup google dataset search

yarikoptic:
Despite our previous efforts google dataset search does not search us. We need to figure it out.

This initial commit is just an empty temporary one for us to get our PR/preview-URL

TODOs
- [ ] rewrite/remove the initial commit in this branch


jjnesbitt:
@yarikoptic It seems this PR is stale, contains no changes, and is not being worked on. If that's not the case, feel free to re-open.

PR #2462, opened 2025-08-05
Fix embargoed dandiset creation form

### LINKED ISSUE #2459: fix embargoed dandiset creation form
Currently, the embargoed dandiset creation formform looks like this:

Image

This has some problems:
* It requires the award to be NIH (via regex)
* It does not provide a place for users to indicate with the grant ends
* It does not allow for projects that don't correspond to research grants.

Proposed solution (discussed with @satra):
* Toggle for have award vs. don't have award. Have award should be default.
* If award:
    * Form fields:
        * Funding source (required)
        * Grant number (not required, no regex)
        * Grant end date (required)
    * If Funding source and Grant number, funding source is added as a Contributor to the Dandiset.
* If not award, automatically show the embargo end date, which is 2 years from the current date.

bendichter:
fix #2459 

The interface now looks like this:

image

It checks that the date is in bounds:

image

image

We also support users the are not funded by a grant. Then embargo is 2 years

image

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -147,7 +147,8 @@
               Grant/Award Number
             </div>
             <div class="mb-3">
-              Provide the grant or award number. This field is optional but recommended.
+              Provide the grant or award number. For awards without a grant number, please
+              provide the project name.
             </div>
             <v-text-field
               v-model="awardNumber"


bendichter:
Also, I tried getting the dandiset metadata using the REST API and it contains the data:

```json
"access": [
      {
        "status": "dandi:EmbargoedAccess",
        "schemaKey": "AccessRequirements",
        "embargoedUntil": "2027-08-05"
      }
    ],
```

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -20,7 +20,7 @@ def create_dandiset(
     *,
     user,
     identifier: int | None = None,
-    embargo: bool,
+    embargo_config: dict,
```
mvandenburgh, line 23: Making this an opaque `dict` significantly decreases the readability of this code, I think we should just pull the dict contents into separate function arguments -


```suggestion
    embargo: bool,
    funding_source: str | None,
    award_number: str | None,
```

However, I don't think this is that important because I think we have a bigger problem here - the `create_dandiset` function has become bloated due to the increasing divergence between open/embargoed logic. I propose that we break this out into two separate functions (like we [already do](https://github.com/dandi/dandi-archive/blob/a85249e45b3f005ba3d7596cba091e16e9a4d265/web/src/rest.ts#L213-L243) for the equivalent functions in the frontend). See #2507 for my suggested change here

mvandenburgh (dismissed):

bendichter:
@mvandenburgh I added some backend logic that hadn't gotten merged yet

dandiapi/api/services/dandiset/__init__.py
bendichter: I was trying to implement the feature with minimal changes. I'd be fine with refactoring to expose these args and to create two separate entry points for creating public and embargoed dandisets

dandiapi/api/views/serializers.py
```diff
@@ -78,6 +81,17 @@ def get_is_starred(self, dandiset):
...
+    award_number = serializers.CharField(required=False, allow_blank=True)
+    embargo_end_date = serializers.DateField(
+        default=lambda: timezone.now().date() + timedelta(days=730)
+    )
+
+    def validate_embargo_end_date(self, value: date) -> date:
+        max_end_date = timezone.now().date() + timedelta(days=730)
+        if value > max_end_date:
+            raise ValidationError('Embargo end date cannot be more than 730 days in the future')
+        return value
```
mvandenburgh, line 94: @bendichter how far in the future should we allow embargo end dates? I went with 730 days to match the default, but let me know if there's a different value we should use.

mvandenburgh: I see this is already set in the frontend as 5 years https://github.com/dandi/dandi-archive/pull/2462/files#diff-5c8140317c91f14905bc0bb134aeadbb668ec81c2b341dad8ff27e2c0a8e061aR299-R302 - I'll update the backend logic accordingly

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -113,23 +113,85 @@
...
           </div>
-          <v-text-field
-            v-model="awardNumber"
-            label="Award number"
-            :counter="120"
-            :required="embargoed"
-            variant="outlined"
-            density="compact"
-            class="my-4"
-            :rules="awardNumberRules"
```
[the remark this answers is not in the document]
mvandenburgh: Added this back tentatively - https://github.com/dandi/dandi-archive/pull/2462/commits/de0918b08b4e7208a44f680dfc0e5b1db37f5422

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -180,18 +243,69 @@ const name = ref('');
...
+);
+
+const grantEndDateRules = computed(() => [
+  (v: string) => !!v || 'Grant end date is required',
+  (v: string) => {
+    if (!v) return true; // Skip validation if empty (handled by required rule)
+
+    const selectedDate = new Date(v);
+    const today = new Date();
+    const fiveYearsFromNow = new Date(today.getFullYear() + 5, today.getMonth(), today.getDate());
```
satra, line 293: @mvandenburgh - the default duration is the end date of the grant if it's supported by nih, or a maximum of 2 years if not an nih award. if you want to simplify set it to 2 years by default if no grant end date is specified, not 5.

mvandenburgh:
I think this is about ready, but because there are a few specific combinations of inputs here with differing implications, I want to make sure the implementation does what we want. Here is my current understanding -

Scenarios for embargoed dandiset creation:
- User supplies no funding source and no award number
	- Result: Successful, embargo end date set to two years in the future (automatic/not changeable by user)
- User supplies a funding source but no award number
    - Result: Failure, supplying a funding source with no award number is not valid (**TODO**: should this be valid?)
- User supplies an award number but no funding source
    - Result: Failure, supplying a funding source with no award number is not valid
- User supplies both a funding source and an award number
	- Result: Successful, embargo end date is set to whatever the user specified in form (defaults to two years, but user can specify up to a maximum of 5 years)


@bendichter @satra @yarikoptic can you confirm if this behavior is correct or not? In particular, I'm unsure about the second point, but it would be great to get consensus on the exact specifics here before I finish up this implementation.

bendichter:
@mvandenburgh thanks for this breakdown. From my perspective I would be fine with the conditions as you laid them out here. I think it might be nice to allow for a funding source and no award number. This could be useful e.g. in the case of funding from a foundation that may not have proper award numbers, but I don't feel strongly about that, because in that case users could just use the project name as an award number.

dandiapi/api/services/dandiset/__init__.py
```diff
@@ -52,10 +58,81 @@ def create_dandiset(
...
+            if contributors is None or not isinstance(contributors, list):
+                contributors = []
+
+            kwargs = {
+                'name': funding_source,
+                'schemaKey': 'Organization',
+                'roleName': RoleType.Funder.value,
+            }
+            if award_number:
+                kwargs['award_number'] = award_number
```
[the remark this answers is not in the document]
bendichter: I think it doesn't make sense to allow for an award number and no funding source

PR #2465, opened 2025-08-06
Improvements to GitHub Actions

mvandenburgh:
* Removes references to custom secrets in favor of `permissions` block
* Removes cron jobs (unused)
* Remove unused auto-add-issues job

### DIFF
--- a/.github/workflows/backend-ci.yml
+++ b/.github/workflows/backend-ci.yml
@@ -8,8 +8,8 @@ on:
       - master
     paths-ignore:
       - "web/**"
-  schedule:
-    - cron: "0 0 * * *"
+permissions:
+  contents: read
 jobs:
   test:
     runs-on: ubuntu-22.04
@@ -23,10 +23,18 @@ jobs:
         env:
           POSTGRES_DB: django
           POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd "pg_isready --username postgres"
+          --health-start-period 30s
+          --health-start-interval 2s
         ports:
           - 5432:5432
       rabbitmq:
-        image: rabbitmq:management
+        image: rabbitmq:management-alpine
+        options: >-
+          --health-cmd "rabbitmq-diagnostics ping"
+          --health-start-period 30s
+          --health-start-interval 2s
         ports:
           - 5672:5672
       minio:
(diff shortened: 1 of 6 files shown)


brianhelba:
Once this is merged we should remove `secrets.AUTO_ADD_ISSUES` and `secrets.GH_TOKEN` from the repo.

.github/workflows/backend-production-deploy.yml
```diff
@@ -19,7 +22,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # fetch history for all branches and tags
-          token: ${{ secrets.GH_TOKEN }} # use PAT with permissions to push to master
```
jjnesbitt, line 22: I'm curious why the other workflow using the auto generated `GITHUB_TOKEN` secret, while this one uses nothing.

.github/workflows/backend-ci.yml
```diff
@@ -8,8 +8,8 @@ on:
       - master
     paths-ignore:
       - "web/**"
-  schedule:
-    - cron: "0 0 * * *"
```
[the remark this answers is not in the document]
brianhelba: I believe the nightly CI was still executing, but I don't think it provides much value. For active repos (like this one), CI is already being run very frequently (on every PR).

For inactive repos, when we'd actually want periodic CI to run, [GH Actions behaves as](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-workflows):
> When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days.

Worse, once disabled, a scheduled workflow remains disabled even for events (like PRs), which essentially breaks the repo's CI for everyone, until a repo administrator re-enables the workflow.

So, due to the lack of value and the pitfalls of scheduled CI, I think it should be removed.

.github/workflows/backend-production-deploy.yml
brianhelba: Where is this other code?

jjnesbitt: https://github.com/dandi/dandi-archive/blob/796a6ae83749b66a2fae422b58533f6d4a84ed09/.github/workflows/release.yml#L28

jjnesbitt (approved):

waxlamp:
> @waxlamp Are you actually relying on `.github/workflows/auto-add-issues.yml` or has it become cruft?
> 
> If it's being used, we can keep it and hopefully change it to use the auto-generated `GITHUB_TOKEN`, instead of a secret that we manually manage.

Still using it.

waxlamp (commented):
I'd like to restore the auto-add-issues action, but otherwise if Mike is good with this then so am I.

PR #2470, opened 2025-08-06
Ensure `requirements.txt` is included in sdist

mvandenburgh:
Follow up to #2469. Heroku requires this file to be present in the sdist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,13 @@ packages = [
   "dandiapi",
 ]
 
+# When building an sdist for upload to Heroku, also include "requirements.txt"
+[tool.hatch.build.targets.sdist]
+only-include = [
+  "dandiapi",
+  "/requirements.txt",
+]
+
 [tool.hatch.version]
 source = "vcs"
 raw-options = {version_scheme = "no-guess-dev"}


pyproject.toml
```diff
@@ -78,6 +78,7 @@ test = [
 packages = [
   "dandiapi",
 ]
+force-include = { "requirements.txt" = "requirements.txt" }
```
brianhelba, line 81: Does `include` work: https://hatch.pypa.io/1.9/config/build/#patterns

Or is that incompatible with the use of `packages`?
```suggestion
include = [ "requirements.txt" ]
```

brianhelba: I see the issue:
> `packages` option is semantically equivalent to `only-include`

> You can use the `only-include` option to prevent directory traversal starting at the project root and only select specific relative paths to directories or files. Using this option ignores any defined `include patterns`.

---

We should be able to do this instead:
```suggestion

# When building an sdist for upload to Heroku, also include "requirements.txt"
[tool.hatch.build.targets.sdist]
only-include = [
  "dandiapi",
  "/requirements.txt",
]
```

Technically, we could put this into `tool.hatch.build` (so that it applies when building any kind of install), but this makes it obvious that this is an explicit workaround for sdists being sent to Heroku.

brianhelba (approved):

pyproject.toml
mvandenburgh: Thanks, can confirm that `requirements.txt` is properly bundled into the sdist with this change :+1:

PR #2471, opened 2025-08-06
Retain focus on meditor text input fields

### LINKED ISSUE #2464: Can't edit metadata, input text box loses focus immediately
Video shows what I'm describing. As soon as the keyboard input stops, focus is lost in the input text item and I can no longer continue typing. 

https://github.com/user-attachments/assets/00c2d9d9-68a6-4869-9699-1ebd7af50c62

jjnesbitt:
Closes #2464 

The cause of the issue was due to [this](https://github.com/dandi/dandi-archive/blob/f86e27b01f53e058f9182a604a4bb7d2ad5b7d6e/web/src/components/Meditor/Meditor.vue#L190-L192) line, which specifies that the key of the `vjsf` component is dependent on the current "transaction" from our transaction tracker (the class we use to control undo/redo). Because of that, any time that value changes (i.e. any time data in the basic model changes), the form is re-rendered, which removes the focus on whatever element was being edited.

This PR simply waits for that to happen, and then re-focuses the element right afterwards. This is sort of a stop-gap measure, until we can more concretely re-factor how we handle updating data and transactions in the meditor.

### DIFF
--- a/web/src/components/Meditor/Meditor.vue
+++ b/web/src/components/Meditor/Meditor.vue
@@ -324,7 +324,44 @@ const disableUndo = computed(
 const disableRedo = computed(
   () => readonly.value || !transactionTracker.areTransactionsAhead(),
 );
-const vjsfListener = () => transactionTracker.add(basicModel.value, false);
+const vjsfListener = () => {
+  // Since the vjsf form will be re-rendered, retrieve the active element, and store its properties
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
Agreed we should refactor this more generally. But this works/solves the current problem from what I can tell

PR #2474, opened 2025-08-07
Fix broken behavior when text includes single quotes

### LINKED ISSUE #2464: Can't edit metadata, input text box loses focus immediately

jjnesbitt:
Follow up to #2471. 
Closes #2464

If single quotes were included in the text field, the buggy behavior persisted.

### DIFF
--- a/web/src/components/Meditor/Meditor.vue
+++ b/web/src/components/Meditor/Meditor.vue
@@ -346,11 +346,25 @@ const vjsfListener = () => {
   if (typeof newValue !== 'string') {
     return;
   }
-  
+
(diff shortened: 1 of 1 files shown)


jjnesbitt:
The only situation I'm aware of where this will still malfunction is when two fields have the exact same contents, in which case this will always focus the first one. However, that's a pretty reasonable limitation of this temporary solution imo.

PR #2476, opened 2025-08-08
Fix "union-attr" type errors

brianhelba:


dandiapi/api/models/asset.py
```diff
@@ -166,45 +166,43 @@ class Meta:
...
         ]
 
-    @property
-    def is_blob(self):
-        return self.blob is not None
-
-    @property
-    def is_zarr(self):
-        return self.zarr is not None
-
```
mvandenburgh, line 176: Why remove these? Does mypy not properly evaluate the return type of these as `bool`?

brianhelba: > Or, is it because `mypy` can't infer that `self.blob` is `None` if just `.is_blob` is called?

ExactlY.

Given that, these functions aren't being used and (IMO) provide little value except to obscure the real data structure.

PR #2483, opened 2025-08-12
Switch `django-composed-configuration` to flat settings

mvandenburgh:
This PR rips out the deprecated `django-composed-configuration` (https://github.com/kitware-resonant/django-composed-configuration/) library in favor of a flat settings model that makes use of the new `django-resonant-settings` package (https://github.com/kitware-resonant/cookiecutter-resonant/tree/master/django-resonant-settings).


dandiapi/api/tests/test_version.py
```diff
@@ -73,9 +73,7 @@ def test_draft_version_metadata_computed(draft_version: Version):
 
     expected_metadata = {
         **original_metadata,
-        'manifestLocation': [
-            f'{settings.DANDI_API_URL}/api/dandisets/{draft_version.dandiset.identifier}/versions/draft/assets/'
-        ],
+        'manifestLocation': [HTTP_URL_RE],
```
jjnesbitt, line 76: The changes in this file seem unrelated? And seem to change the behavior of the tests?

dandiapi/settings/_sentry_utils.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from sentry_sdk._types import SamplingContext
+
+
+def get_sentry_performance_sample_rate(sampling_context: SamplingContext) -> float:
```
jjnesbitt, line 10: It seems this function still exists in `apps.py`

dandiapi/urls.py
```diff
@@ -133,7 +133,7 @@ def to_url(self, value):
     # this url overrides the authorize url in oauth2_provider.urls to
     # support our user signup workflow
     re_path(r'^oauth/authorize/$', authorize_view, name='authorize'),
-    path('oauth/', include('oauth2_provider.urls', namespace='oauth2_provider')),
+    path('oauth/', include('oauth2_provider.urls')),
```
jjnesbitt, line 136: Curious why this is being done

jjnesbitt (changes requested):

dandiapi/api/tests/test_version.py
brianhelba: Attempting to hardcore a full URL in the test is failing and it's probably a sufficient guarantee to just ensure that it structurally looks like an HTTP URL.

If you think we need a stranger assertion, we could add a new assertion to ensure that this URL ends with the appropriate file name.

dandiapi/settings/_sentry_utils.py
brianhelba: It should have been removed there and moved here. @mvandenburgh

PR #2500, opened 2025-08-19
Switch the default storage to always use `S3Storage`

brianhelba:
This depends on https://github.com/dandi/dandi-infrastructure/pull/240


dandiapi/zarr/tests/test_zarr_upload.py
```diff
@@ -11,20 +11,20 @@
 
 @pytest.mark.django_db
 def test_zarr_rest_upload_start(
-    authenticated_api_client, user, zarr_archive: ZarrArchive, storage, monkeypatch
+    authenticated_api_client, user, embargoed_zarr_archive_factory, storage, monkeypatch
 ):
+    zarr_archive: ZarrArchive = embargoed_zarr_archive_factory.create(
```
brianhelba, line 16: Beyond switching to a factory, this needs to be an *embargoed* Zarr, or the test won't pass. Only embargoed Zarr should actually have tags applied, as we assert.

dandiapi/zarr/models.py
```diff
@@ -89,16 +93,6 @@ def s3_path(self, zarr_path: str) -> str:
         """Generate a full S3 object path from a path in this zarr_archive."""
         return zarr_s3_path(str(self.zarr_id), zarr_path)
 
-    def generate_upload_urls(self, path_md5s: list[dict]):
```
brianhelba, line 92: This was a bad API, as the `dict` in the parameter was tightly coupled to the one and only place it was invoked from a view. I just moved it inline to the view.

dandiapi/settings/development.py
```diff
@@ -44,10 +43,20 @@
...
+        'region_name': 'us-east-1',  # This is MinIO's default region
+        'access_key': _minio_url.username,
+        'secret_key': _minio_url.password,
+        'bucket_name': _minio_url.path.lstrip('/'),
+        'signature_version': 's3v4',
+        'querystring_expire': int(timedelta(hours=6).total_seconds()),
+    },
 }
-DANDI_DANDISETS_BUCKET_NAME = MINIO_STORAGE_MEDIA_BUCKET_NAME
+DANDI_DANDISETS_BUCKET_NAME = STORAGES['default']['OPTIONS']['bucket_name']  # TODO: remove
```
brianhelba, line 59: This can probably be removed soon.

dandiapi/storage.py
```diff
@@ -0,0 +1,216 @@
...
+    # Storage.get_valid_name uses django.utils.text.get_valid_filename,
+    # which cleans spaces and other characters.
+    # Since these are designed around filesystem safety, not S3 key safety, it's
+    # simpler to do sanitization before saving.
+    def generate_filename(self, filename: str) -> str:
+        return filename
+
+    def _url_unsigned(self, name: str) -> str:
+        if self.endpoint_url:
+            # TODO: correct URL when inside Docker
```
brianhelba, line 87: This still needs to be fixed.

dandiapi/api/storage.py
```diff
@@ -1,360 +1,20 @@
...
-    allowing either to be used as an additional non-default Storage.
-    """
-    # For production, calling django.core.files.storage.get_storage_class is fine
-    # to return the storage class of S3Storage.
-    default_storage_class = get_storage_class()
-
-    if issubclass(default_storage_class, S3Storage):
-        storage = VerbatimNameS3Storage(bucket_name=bucket_name)
-        # Required to upload to the sponsored bucket
-        storage.default_acl = 'bucket-owner-full-control'
```
brianhelba, line 312: All this ACL stuff is no longer necessary, once https://github.com/dandi/dandi-infrastructure/pull/240 is merged.

dandiapi/zarr/tests/test_zarr_upload.py
brianhelba: Obsoleted by #2503

dandiapi/storage.py
```diff
@@ -0,0 +1,116 @@
...
+        return f'https://{self.bucket_name}.s3.{self.region_name}.amazonaws.com/{name}'
+
+    def url(
+        self,
+        name: str,
+        *,
+        parameters: Mapping[str, Any] | None = None,
+        expire: int | None = None,
+        http_method: str | None = None,
+        signed: bool = True,
```
brianhelba, line 100: We should make more use of `signed`, as anything that isn't embargoed can be unsigned.

dandiapi/api/multipart.py
```diff
@@ -0,0 +1,76 @@
...
+        return resp['UploadId']
+
+    # Forked to accept a "tags" parameter
+    def initialize_upload(
+        self,
+        object_key: str,
+        file_size: int,
+        content_type: str,
+        *,
+        tags: Mapping[str, str] | None = None,
```
brianhelba, line 60: @brianhelba Upstream to S3FF?

dandiapi/storage.py
```diff
@@ -0,0 +1,216 @@
...
+    ) -> str:
+        # Just calling "S3Storage.url(..., http_method=='PUT')" doesn't work, as "get_object" can't
+        # accept "Tagging" options, even if its method is overridden.
+        if expire is None:
+            expire = self.querystring_expire
+        optional_params = {}
+        if content_md5 is not None:
+            optional_params['ContentMD5'] = content_md5
+        if tags is not None:
+            optional_params['Tagging'] = urlencode(tags)
```
mvandenburgh, line 122:
```suggestion
        # Just calling "S3Storage.url(..., http_method=='PUT')" doesn't work, as "get_object" can't
        # accept "Tagging" options, even if its method is overridden. This is a bug in boto3.
        # TODO: remove this logic when it is fixed upstream
        if expire is None:
            expire = self.querystring_expire
        optional_params = {}
        if content_md5 is not None:
            optional_params['ContentMD5'] = content_md5
        if tags is not None:
            optional_params['Tagging'] = urlencode(tags)
```

dandiapi/api/tests/test_tasks.py
```diff
@@ -15,62 +16,30 @@
...
-    # Pretend like AssetBlob was defined with the given storage
-    AssetBlob.blob.field.storage = storage
-
-    asset_blob = asset_blob_factory(sha256=None)
-
-    h = hashlib.sha256()
-    h.update(asset_blob.blob.read())
-    sha256 = h.hexdigest()
+def test_calculate_checksum_task():
+    asset_blob = AssetBlobFactory.create(blob__data=b'known-content', size=13, sha256=None)
```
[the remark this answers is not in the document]
brianhelba: No, it's the same behavior by default. However, calling the actual constructor seems to really freak out type checkers, whereas `create` just mildly confuses them. I'd like to get https://github.com/kitware-resonant/cookiecutter-resonant/pull/332 working, but absent that, this is the less-problematic way to handle typing with factories.

PR #2503, opened 2025-08-21
Only include tagging header in zarr upload URL if zarr is embargoed

jjnesbitt:
Setting the tagging header to `''` when generating the pre-signed URL doesn't affect the upload flow, as if set to `''` it is not required to be included in the `PUT` request headers. However, the `x-amz-tagging` substring _is_ still included in the pre-signed URL in that case, which, for better or worse, is an easy way to distinguish between open and embargoed URLs.

This PR only includes the tagging header if the value is not an empty string. It also fixes the test, to ensure the behavior is as we expect.

### DIFF
--- a/dandiapi/zarr/tests/test_zarr_upload.py
+++ b/dandiapi/zarr/tests/test_zarr_upload.py
@@ -18,19 +18,17 @@ def test_zarr_rest_upload_start(
     zarr_archive = zarr_archive_factory(
         dandiset__embargo_status=Dandiset.EmbargoStatus.EMBARGOED
         if embargoed
-        else Dandiset.EmbargoStatus.OPEN
+        else Dandiset.EmbargoStatus.OPEN,
+        # Set as complete, to mimic past upload
+        status=ZarrArchiveStatus.COMPLETE,
+        checksum=EMPTY_CHECKSUM,
+        file_count=1,
+        size=100,
     )
     add_dandiset_owner(zarr_archive.dandiset, user)
 
     # Pretend like our zarr was defined with the given storage
     monkeypatch.setattr(ZarrArchive, 'storage', storage)
-
-    # Set as complete, to mimic past upload
-    zarr_archive.status = ZarrArchiveStatus.COMPLETE
-    zarr_archive.checksum = EMPTY_CHECKSUM
-    zarr_archive.file_count = 1
-    zarr_archive.size = 100
-    zarr_archive.save()
 
     # Request upload files
     resp = authenticated_api_client.post(


brianhelba:
Do you want to use this change instead: https://github.com/dandi/dandi-archive/pull/2500/files#r2291895047 ?

dandiapi/zarr/tests/test_zarr_upload.py
```diff
@@ -33,8 +40,11 @@ def test_zarr_rest_upload_start(
     )
     assert resp.status_code == 200
     assert resp.json() == [HTTP_URL_RE]
-    if isinstance(storage, S3Storage):
+
+    if embargoed and isinstance(storage, S3Storage):
         assert 'x-amz-tagging' in resp.json()[0]
+    else:
+        assert 'x-amz-tagging' not in resp.json()[0]
```
brianhelba, line 47: If we're going to make this stronger assertion about things being absent, I think we should test for the absence of `"embargoed"` instead.

While I expect that `"x-amz-tagging"` wouldn't be included in a non-embargoed case, it isn't essential for correctness. In other words, if `"x-amz-tagging"` starts appearing for some other reason, this test shouldn't fail.

For symmetry, maybe look for the presence of `"embargoed"` in the embargoed case too.

dandiapi/zarr/tests/test_zarr_upload.py
```diff
@@ -4,15 +4,22 @@
...
 from dandiapi.zarr.models import ZarrArchive, ZarrArchiveStatus
 
 
 @pytest.mark.django_db
+@pytest.mark.parametrize('embargoed', [False, True])
 def test_zarr_rest_upload_start(
-    authenticated_api_client, user, zarr_archive: ZarrArchive, storage, monkeypatch
+    authenticated_api_client, user, zarr_archive_factory, storage, monkeypatch, embargoed
 ):
+    zarr_archive = zarr_archive_factory(
```
brianhelba, line 18: If you're going to start using a factory, you can move the mutations to `zarr_archive` below into it too.

brianhelba (changes requested):
I agree that the improvement to test both cases is good.

Feedback inline.

dandiapi/zarr/tests/test_zarr_upload.py
brianhelba: I mean test for the string `"embargoed=true"` (which is the actual tagged value), rather than test for the string `"x-amz-tagging"`.

Saying "this URL shouldn't be setting tags for embargoing" is what we actually want to assert, rather "this URL shouldn't be setting any tags".

If the URL happens to include `"x-amz-tagging"` (as it did when we were passing the empty string before), that alone isn't really a problem.

jjnesbitt: Oh, I completely agree. However, the tag itself isn't actually included in the pre-signed URL, as far as I can tell. In any pre-signed URL that I've created, it never includes `embargoed=true` or even just `embargoed` anywhere in the string. It just includes `x-amz-tagging`. This appears to also be the case with other headers as well, like `content-md5` and `x-amc-acl`.

I probably should have explained that searching for `x-amz-tagging` was done out of necessity. I would much rather search for just `embargoed=true`.

brianhelba: Ah, I see, sorry for my confusion. This is fine then.

PR #2504, opened 2025-08-22
Fix avatar and new dandiset button line break issue in header

### LINKED ISSUE #2362: Header line break issue
Image

The header either needs to be tightened or swapped to mobile styles earlier to prevent issues like what is in the screenshot.

jtomeck:
I'm proposing a slight redesign of the navigation to close #2362.

Screenshot 2025-08-26 at 11 27 29 AM

In the screenshot you'll see that the nav items have been converted to `v-tabs` and the external links now live under a dropdown menu labeled "External Resources." I'm proposing this change because I do not believe decreasing paddings or font sizes in the menu is the right solution, nor do I like the idea of writing custom CSS to do so (Vuetify's methods, like `density="compact"` only change height and do not save horizontal space).

### DIFF
--- a/e2e/tests/dandisetsPage.spec.ts
+++ b/e2e/tests/dandisetsPage.spec.ts
@@ -46,7 +46,7 @@ test.describe("dandisets page", async () => {
         const name = `name ${id}`;
         const description = `description ${id}`;
         const identifier = await registerDandiset(page, name, description);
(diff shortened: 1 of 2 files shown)


mvandenburgh (approved):
Agreed that this is a great improvement :+1:

PR #2512, opened 2025-08-26
Fix broken links: Terms and Policies

NEStock:
https://docs.dandiarchive.org/about/terms/ and https://docs.dandiarchive.org/about/policies/ lead to 404 page.


jjnesbitt (approved):
Thanks for the improvement!

PR #2516, opened 2025-08-27
Apply embargo functionality to dandiset manifest files

jjnesbitt:
It appears embargo tags were never applied to manifest files. This PR ensures going forward, the proper embargo tags are applied to the manifest files of embargoed dandisets, when generated. It also includes the corresponding unembargo functionality, as well as a management command to apply the tags to embargoed dandisets.

### DIFF
--- a/dandiapi/api/management/commands/rectify_manifest_tags.py
+++ b/dandiapi/api/management/commands/rectify_manifest_tags.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+
+import logging
+
+from django.conf import settings
+import djclick as click
+
+from dandiapi.api.manifests import all_manifest_filepaths
+from dandiapi.api.models import Version
+from dandiapi.api.models.dandiset import Dandiset
+from dandiapi.api.storage import get_boto_client
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument(
+    'dandisets',
+    type=click.INT,
+    nargs=-1,
+)
+@click.option(
+    '-a',
+    '--all',
+    'include_all',
+    is_flag=True,
+    help='Run on all dandisets.',
+)
+def tag_embargoed_manifests(dandisets, include_all):
+    """Apply embargoed tags to dandiset manifest files."""
+    if bool(dandisets) == include_all:
+        raise click.ClickException("Must specify exactly one of 'dandisets' or --all")
+
+    embargoed_versions = Version.objects.select_related('dandiset').filter(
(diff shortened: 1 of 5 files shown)


dandiapi/api/services/embargo/utils.py
```diff
@@ -77,8 +78,27 @@ def _delete_zarr_object_tags(client: S3Client, zarr: str):
...
+    version = dandiset.draft_version
+
+    paths = all_manifest_filepaths(version)
+    logger.info('Removing tags from dandiset %s', dandiset.identifier)
+    for path in paths:
+        try:
+            client.delete_object_tagging(
+                Bucket=settings.DANDI_DANDISETS_BUCKET_NAME,
+                Key=path,
+            )
```
waxlamp, line 91: I don't think this is correct, long term, because it deletes all tags on the object. That works now since we just use one tag  (`embargoed`) but if we decide to use others for other purposes, this may cause bugs in the future.

I think the correct thing is to use `get_object_tagging()`, remove the `embargoed` tag, then `put_object_tagging()`.

dandiapi/api/management/commands/rectify_manifest_tags.py
```diff
@@ -0,0 +1,60 @@
...
+                    Key=path,
+                    Tagging={
+                        'TagSet': [
+                            {
+                                'Key': 'embargoed',
+                                'Value': 'true',
+                            },
+                        ]
+                    },
+                )
```
waxlamp, line 57: Another instance where this will work just fine today, but will clobber other valid tags if run in the future.

waxlamp (commented):
Looks good overall, but a major concern about clobbering tags (see individual comments inline).

dandiapi/api/services/embargo/utils.py
jjnesbitt: While I agree, the reason I didn't worry about that here is because that assumption (that the only tag applied to an S3 object is the embargoed tag) is already made in other parts of the code, as well as the CLI (see https://github.com/dandi/dandi-cli/pull/1685). Changing those other parts would be out of scope for this PR, and I've tested that this PR operates as expected. I can make a follow up PR to change everything in a backwards compatible way.

jjnesbitt: #2517

waxlamp (commented):
LGTM; please defer to Mike's review for merging.

mvandenburgh (approved):
I'd still like to discuss the comment I made, but since this is somewhat urgent I don't want to block on that. The actual logic looks good to me

PR #2517, opened 2025-08-27
Don't modify existing object tags when adding/removing embargoed tags

### LINKED ISSUE #2518: Don't clobber tags on S3 objects
See discussion from a specific [tag-clobbering PR](https://github.com/dandi/dandi-archive/pull/2516#discussion_r2304797117).

The codebase makes an assumption that `embargoed` is the only tag we use on S3 objects. While this assumption is true currently, we are planning on possibly using another tag to protect published asset deletion, and there might be other use cases still. We should examine the codebase to see if we can remove that assumption and do a better job of surgically altering tags instead of doing so wholesale.

The tradeoff is that manipulating tags "surgically" will involve two requests: a "get" to grab to tag state followed by a "put" to change it. But doing it this way future-proofs us against what could be severe bugs.

jjnesbitt:
Depends on #2516
Closes #2518

This doesn't directly affect anything in the code base at the moment, since we only make use of the `embargoed` tag. However, it's a good fail-safe to have.

### DIFF
--- a/dandiapi/api/management/commands/rectify_manifest_tags.py
+++ b/dandiapi/api/management/commands/rectify_manifest_tags.py
@@ -41,7 +41,12 @@ def tag_embargoed_manifests(dandisets, include_all):
         paths = all_manifest_filepaths(version)
         for path in paths:
             try:
-                default_storage.put_tags(path, {'embargoed': 'true'})
+                existing_tags: dict[str, str] = default_storage.get_tags(path)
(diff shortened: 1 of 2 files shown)


mvandenburgh (commented):
I think we should defer this until #2500 is merged to prevent introducing merge conflicts for that larger PR. And once we have that storage refactor in, we can just update the `DandiS3Storage.put_tags()` method to do this.

PR #2542, opened 2025-09-18
Fix intermittent login error

jjnesbitt:
There has been an issue lately that would cause a GUI error when a user attempted to load the GUI with expired credentials. The ad-hoc solution was to reset the browser cache, or initiate a login manually. This behavior is fixed by https://github.com/kitware-resonant/resonant-oauth-client/releases/tag/v1.0.3.


mvandenburgh:
FYI @sandyhider @NEStock this will fix that "server error" issue we saw intermittently during our recent debugging sessions.

PR #2545, opened 2025-09-22
Clean up factories

brianhelba:


dandiapi/api/tests/factories.py
```diff
@@ -75,21 +76,38 @@ def extra_data(self):
...
     class Meta:
+        model = Version
+        skip_postgeneration_save = True
         abstract = True
 
     dandiset = factory.SubFactory(DandisetFactory)
     name = factory.Faker('sentence')
+    # Don't use "Version.next_published_version" to create versions, as that will require
+    # additional database queries.
+    version = factory.Sequence(
```
brianhelba, line 102: In tests, `Version` created at the exact same time via `VersionFactory` will no longer have the same `.version` value. I think this is actually an improvement as it's more realistic. If a duplicate version is desired, it can still be created with `VersionFactory(version=...)`.

dandiapi/api/tests/factories.py
```diff
@@ -116,32 +134,32 @@ def metadata(self):
...
-            **version.metadata,
-            'publishedBy': version.published_by(now),
-            'datePublished': now.isoformat(),
-        }
-        version.save()
-        return version
+    doi = factory.LazyAttribute(
+        lambda self: f'10.80507/dandi.{self.dandiset.identifier}/{self.version}'
+    )
+    status = Version.Status.PUBLISHED
```
brianhelba, line 162: Previously, `PublishedVersionFactory` wasn't setting the `status` to `PUBLISHED`, but I think this makes sense. Note, I had to fix some tests accordingly.

dandiapi/api/tests/factories.py
```diff
@@ -200,38 +223,45 @@ def metadata(self):
...
+        skip_postgeneration_save = True
 
     upload_id = factory.Faker('uuid4')
     multipart_upload_id = factory.Faker('uuid4')
-    blob = factory.django.FileField(data=factory.Faker('binary', length=100))
+    blob = factory.django.FileField(
+        filename=factory.Faker('file_name', extension='dat'),
+        data=factory.Faker('binary', length=factory.SelfAttribute('...size')),
+    )
+    size = 100
```
brianhelba, line 249: I'm actually not sure the ideal way to structure this. I can optimize this to set a user-defined `size`, whereby it will auto-generate the `blob`:
```python
UploadFactory.create(size=...)
```
Or I can optimize this to set some content, whereby it will auto-populate the `size`:
```python
UploadFactory.create(blob__data=b'abcd')
```

I don't think I can easily support both. Which is a more useful pattern?

dandiapi/api/tests/test_version.py
```diff
@@ -37,30 +37,25 @@ def test_version_next_published_version_nosave(dandiset):
 
 
 @pytest.mark.django_db
-def test_version_next_published_version_save(mocker, dandiset, published_version_factory):
+def test_version_next_published_version_save(dandiset, published_version_factory):
```
brianhelba, line 40: Should this test be using `freeze_time` too?

dandiapi/api/tests/factories.py
jjnesbitt: I'd tend to think a user-defined `size` field would be more commonly used, although I think they'd both be used.

dandiapi/api/tests/factories.py
```diff
@@ -32,6 +32,7 @@ class Meta:
 class UserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = User
+        skip_postgeneration_save = True
```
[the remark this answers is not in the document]
brianhelba: The presence of a `RelatedFactory` triggers the warning too. It's maybe a false positive, but adding the `skip_postgeneration_save` guard will suppress it.

jjnesbitt (approved):

PR #2549, opened 2025-09-24
Limit number of threads in remove_dandiset_embargo_tags

jjnesbitt:
The high number of threads was causing the celery worker to crash, since each thread creates a new boto3 resource object, which consumes a non-trivial amount of memory in the process.

### DIFF
--- a/dandiapi/api/services/embargo/utils.py
+++ b/dandiapi/api/services/embargo/utils.py
@@ -67,7 +67,9 @@ def _delete_zarr_object_tags(client: S3Client, zarr: str):
         Bucket=settings.DANDI_DANDISETS_BUCKET_NAME, Prefix=zarr_s3_path(zarr_id=zarr)
     )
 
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
Should we do https://github.com/dandi/dandi-archive/pull/2549/files#diff-1dcbb9672e28d6fc8d0ed064854dc9e8551e4958c712a43cab7dabf4bfe13fb7L70 too? I imagine that would have the same issues if it is called, right?

jjnesbitt:
> Should we do [#2549 (files)](https://github.com/dandi/dandi-archive/pull/2549/files#diff-1dcbb9672e28d6fc8d0ed064854dc9e8551e4958c712a43cab7dabf4bfe13fb7L70) too? I imagine that would have the same issues if it is called, right?

Good point, updated.

mvandenburgh (approved):

PR #2550, opened 2025-09-24
Add rudimentary mailmap file

yarikoptic:
For better looking   git shortlog -sn  if you ever interested in a summary of contributions


yarikoptic:
actually I just merged it since it doesn't break anything etc. Feel welcome to extend, but let's discuss if you want to remove it ;-)

PR #2551, opened 2025-09-24
Don't retry _delete_object_tags

jjnesbitt:
Boto3 does automatically retries failed requests, and so including the `retry` decorator on the `_delete_object_tags` function can cause many more retries/requests to occur than is desirable.

### DIFF
--- a/dandiapi/api/services/embargo/utils.py
+++ b/dandiapi/api/services/embargo/utils.py
@@ -53,7 +53,6 @@ def retry(times: int, exceptions: tuple[type[Exception]]):
     return decorator
 
 
-@retry(times=3, exceptions=(Exception,))
 def _delete_object_tags(blob: str):
     existing_tags: dict[str, str] = default_storage.get_tags(blob)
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
This LGTM, although I am curious if you were running into problems caused by this. I would expect the botocore retry engine to have operated transparently over the call to `put_object_tagging()` (inside `put_tags()`), and for our homebrew retry decorator to only kick in after three such failures (in which case, our decorator would not be causing redundant extra tries).

jjnesbitt:
> This LGTM, although I am curious if you were running into problems caused by this. I would expect the botocore retry engine to have operated transparently over the call to `put_object_tagging()` (inside `put_tags()`), and for our homebrew retry decorator to only kick in after three such failures (in which case, our decorator would not be causing redundant extra tries).

We weren't running into problems, we just realized it's unnecessary, and can lead to an unreasonable amount of wasted requests being made in some cases (like zarrs).

PR #2557, opened 2025-09-26
Objects in test are uploaded directly, not via multipart

brianhelba:
Given the current usage of test factories to populate objects via `Storage.save`, objects uploaded in tests use a simple MD5 for their ETag value. In production, all non-Zarr objects are uploaded via multipart upload and use a different ETag format (which DANDI labels as `"dandi:dandi-etag"`).

Eventually, test uploads should also use a multipart upload, which will cause them to have the correct ETag format. For now, work around this problem by explicitly expecting a different ETag format during test and disallowing all usage of `Storage.save` in production (which would cause a objects to incorrectly have an MD5 ETag).


dandiapi/api/tests/factories.py
```diff
@@ -187,12 +186,17 @@ def sha256(self) -> str:
         return sha256
 
     @factory.lazy_attribute
-    def etag(self):
-        etagger = DandiETag(self.size)
-        for part in etagger._part_gen:
-            etagger.update(self.blob.read(part.size))
+    def etag(self) -> str:
+        # In production, files woule be uploaded with a multipart ETag:
```
jjnesbitt, line 190:
```suggestion
        # In production, files would be uploaded with a multipart ETag:
```

dandiapi/api/tests/factories.py
```diff
@@ -259,7 +263,14 @@ class Meta:
     dandiset = factory.SubFactory(DandisetFactory)
 
     @factory.lazy_attribute
-    def etag(self):
+    def etag(self) -> str:
+        # In production, files woule be uploaded with a multipart ETag:
```
jjnesbitt, line 267:
```suggestion
        # In production, files would be uploaded with a multipart ETag:
```

jjnesbitt (approved):

PR #2561, opened 2025-09-26
Remove use of `social_account` and `user` test fixtures

brianhelba:


brianhelba:
It'll probably be better to review this commit-by-commit. However, most are straightforward, except the last commit, which deserves much closer attention.

dandiapi/api/tests/factories.py
```diff
@@ -19,58 +18,55 @@
...
-            'created_at': created,
+    extra_data = factory.Dict(
+        {
+            'name': factory.LazyAttribute(
+                lambda self: (
+                    f'{self.factory_parent.user.first_name} {self.factory_parent.user.last_name}'
+                )
+            ),
+            'login': factory.SelfAttribute('..user.username'),
+            'email': factory.SelfAttribute('..user.email'),
```
brianhelba, line 66: Note, this is a change in policy. Before, a code comment indicated that the `SocialAccount` intentionally set different values (e.g. for `'email'`) than the `User`, since these could differ in the database.

However, since all users should be now created exclusively via social login, I think it makes more sense for these to just match. This makes it easier to access the authoritative values directly from the user object.

If this is reasonable, I might want to do a little more follow-up to ensure this change is respected everywhere. If we should expect that the `SocialAccount` and `User` have different metadata, than I'll also need to make some updates.

dandiapi/api/tests/test_users.py
```diff
@@ -162,38 +128,33 @@ def test_user_search_multiple_matches(api_client, user, user_factory, social_acc
...
-        serialize_social_account(social_account) for social_account in social_accounts[:10]
+        serialize_user(user) for user in users[:10]
     ]
 
 
 @pytest.mark.django_db
-def test_user_search_extra_data(api_client, user, social_account, social_account_factory):
+def test_user_search_extra_data(api_client):
     """Test that searched keyword isn't caught by a different field in `extra_data`."""
+    user = UserFactory.create(social_account__extra_data__test='odysseus')
```
brianhelba, line 154: Note this syntax. This is why it's nice when we can use `factory.Dict` for properties, as it can be easily overridden selectively.

dandiapi/api/tests/factories.py
jjnesbitt: We currently set the `username` field to the email supplied from the social account. So in reality, due to the github login flow, the username is almost always set to the `email` value from `SocialAccount`. 

It it sometimes (rarely) an even different value, due to differences in how people can specify emails on Github, changing github profile info after first login, etc.

While that's something that ought to be fixed in its own right, I think for now our tests would be more robust if we didn't make that assumption.

jjnesbitt (approved):

mvandenburgh (approved):

PR #2562, opened 2025-09-26
Create a `ZarrFileFactory`

brianhelba:


dandiapi/zarr/tests/factories.py
```diff
@@ -28,3 +35,34 @@ def ensure_draft_version(obj: ZarrArchive, *args, **kwargs):  # type: ignore  #
...
+
+class ZarrFileFactory(PostgenerationAttributesFactory):
+    class Meta:
+        model = ZarrArchiveFile
+
+    class Params:
+        blob = factory.Faker('binary', length=100)
+        zarr_archive = factory.SubFactory(ZarrArchiveFactory)
+
+    path = factory.Faker('file_path', absolute=False, extension='nwb')
```
jjnesbitt, line 48: Due to the structure of zarrs, a realistic path for a zarr file (relative to the zarr root) would be something like 

```
1/0/10/10
```

I think to achieve this behavior, we'd probably need to create a custom faker provider. If nothing else though, I think we should just go with no file extension.

```suggestion
    path = factory.Faker('file_path', absolute=False, extension='')
```

dandiapi/zarr/tests/utils.py
```diff
@@ -1,31 +1,18 @@
...
 
-from dandiapi.api.storage import get_boto_client
+class _PostgenerationAttributesFactoryOptions(factory.base.FactoryOptions):
+    def use_postgeneration_results(self, step, instance, results):
+        self.factory._after_postgeneration(
+            instance,
+            create=step.builder.strategy == factory.CREATE_STRATEGY,
+            results=results,
+            attributes=step.attributes,
+        )
```
jjnesbitt, line 13: I don't really understand why this function is necessary, or where it's called. A comment explaining this might be helpful, since it seems like this is making use of undocumented behavior.

jjnesbitt (changes requested):

dandiapi/zarr/tests/utils.py
brianhelba: Sure, I'll better document this class.

PR #2564, opened 2025-09-29
Fix a drf-yasg warning

brianhelba:
This doesn't change any behavior, but it required by the upstream package to avoid warnings.


brianhelba:
I've tested this and the docs still render, but please feel free to verify the same.

mvandenburgh (approved):
LGTM. Confirmed that https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/views.py#L99-L111 is the only place this setting is referenced upstream

PR #2569, opened 2025-09-30
Fix `ValidationError` error handling

brianhelba:
A Django `ValidationError` can contain either an `error_list` or an `error_dict`. Instead of relying on these lower-level APIs, just use `ValidationError.messages` to obtain the actual message content.


brianhelba:
This has the potential to cause 500 errors, given the wrong type of `ValidationError`.

mvandenburgh (approved):

PR #2574, opened 2025-09-30
Flag all warnings as test failures

brianhelba:
This ensures that all warnings (which were important enough for a library to emit) are affirmatively acted upon. False positive or WONTFIX warnings can always be ignored via `filterwarnings`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,9 +236,9 @@ addopts = [
   "--ds=dandiapi.settings.testing",
 ]
 filterwarnings = [
-  # "error",
+  "error",
   # pytest often causes unclosed socket warnings
   "ignore:unclosed <socket\\.socket:ResourceWarning",
-  # https://github.com/celery/kombu/issues/1339
-  "ignore:SelectableGroups dict interface is deprecated:DeprecationWarning:kombu",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -236,9 +236,9 @@ addopts = [
...
 ]
 filterwarnings = [
-  # "error",
+  "error",
   # pytest often causes unclosed socket warnings
   "ignore:unclosed <socket\\.socket:ResourceWarning",
-  # https://github.com/celery/kombu/issues/1339
-  "ignore:SelectableGroups dict interface is deprecated:DeprecationWarning:kombu",
+  # TODO: Remove this with pytest_factoryboy
+  "ignore:Using a <class 'dict'> as model type:UserWarning:pytest_factoryboy.fixture",
```
jjnesbitt, line 243: Is this warning not fixed via https://github.com/dandi/dandi-archive/pull/2571?

brianhelba: No, unrelated. This is purely caused by `pytest_factoryboy` and use of `factory.Dict`.

jjnesbitt (approved):

PR #2584, opened 2025-10-03
Utilize vendor specific info from schema instance config

candleindark:
This PR closes https://github.com/dandi/dandi-archive/issues/2382

### DIFF
--- a/dandiapi/api/models/version.py
+++ b/dandiapi/api/models/version.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 
 from dandischema.models import AccessType
+from dandischema.conf import get_instance_config as get_schema_instance_config
 from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.validators import RegexValidator
@@ -18,6 +19,7 @@ from .dandiset import Dandiset
 
 logger = logging.getLogger(__name__)
 
+_SCHEMA_INSTANCE_CONFIG = get_schema_instance_config()
 
 class Version(PublishableMetadataMixin, TimeStampedModel):
     VERSION_REGEX = r'(0\.\d{6}\.\d{4})|draft'
@@ -217,9 +219,9 @@ class Version(PublishableMetadataMixin, TimeStampedModel):
             ),
             'manifestLocation': manifest_location(self),
             'name': self.name,
-            'identifier': f'DANDI:{self.dandiset.identifier}',
+            'identifier': f'{_SCHEMA_INSTANCE_CONFIG.instance_name}:{self.dandiset.identifier}',
             'version': self.version,
-            'id': f'DANDI:{self.dandiset.identifier}/{self.version}',
+            'id': f'{_SCHEMA_INSTANCE_CONFIG.instance_name}:{self.dandiset.identifier}/{self.version}',
             'repository': settings.DANDI_WEB_APP_URL,
             'url': (
                 f'{settings.DANDI_WEB_APP_URL}/dandiset/{self.dandiset.identifier}/{self.version}'


candleindark:
@mvandenburgh Why do we need the following line?

https://github.com/dandi/dandi-archive/blob/2ff1bc7a954d7b2ec467fbbdf31f8c1a058d157c/web/src/utils/schema.ts#L13

The JSON schema of `Dandiset` already specifies the exact same pattern.

https://github.com/dandi/schema/blob/c90cd41cb18a29e75abaa13728c13125f38d8236/releases/0.6.10/dandiset.json#L1389

candleindark:
@mvandenburgh I don't think the `DANDI:` hardcoded at https://github.com/dandi/dandi-archive/blob/9aa8b1f2b4510d2163f06c864da3c1126c72f3bd/web/src/views/DandisetLandingView/DownloadDialog.vue#L141

needs to be replaced because it is a shortened URL for the `https://dandiarchive.org` instance.  All instances of DANDI at other domain will just use the long URL.

candleindark:
@yarikoptic Should be bother to replace the hardcoded `'spdx:CC0-1.0'` in the following?

https://github.com/dandi/dandi-archive/blob/9aa8b1f2b4510d2163f06c864da3c1126c72f3bd/dandiapi/api/management/commands/create_dev_dandiset.py#L35

It is one of the default supported licenses, and `create_dev_dandiset` is just for development.

dandiapi/api/models/metadata.py
```diff
@@ -19,7 +21,8 @@ def published_by(self, now: datetime.datetime):
             'wasAssociatedWith': [
                 {
                     'id': uuid4().urn,
-                    'identifier': 'RRID:SCR_017571',
+                    # TODO: handle the case where `instance_identifier` is None
+                    'identifier': settings.DANDI_SCHEMA_INSTANCE_CONFIG.instance_identifier,
```
yarikoptic, line 25: make it optional and don't add if `not` .

below, please add also `version` to match the version from `/api/info` -- it is not per se API version but would be useful info to know with what version of archive we did it. I think it is much better than some ad-hoc fixed version since API etc change and reflect in the version of server implementation.  May be even change name to "{instance_name} API Server", which would disambiguate instances, and also that it is not just API version but the version of the server.

mvandenburgh:
> so it seems that the 0.6.10 had fixed the pattern and we can just remove that ad-hoc fixup of the schema. @mvandenburgh please confirm but meanwhile we proceed under such an assumption

Yes, those can be removed now that they are fixed in the schema. I tested locally to confirm, and the meditor works without that code.

dandiapi/api/models/metadata.py
```diff
@@ -3,13 +3,20 @@
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
+from django.conf import settings
+
+from dandiapi import __version__
```
mvandenburgh: #2597 removed `__version__` in favor of `importlib.metadata.version` as a best practice.
```suggestion
import importlib
```

mvandenburgh (changes requested):

dandiapi/api/management/commands/create_dev_dandiset.py
```diff
@@ -30,9 +30,18 @@
...
+    supported_licenses.sort()
+
+    license_to_use = (
+        'spdx:CC0-1.0' if 'spdx:CC0-1.0' in supported_licenses else supported_licenses[0]
+    )
+
     version_metadata = {
         'description': 'An informative description',
-        'license': ['spdx:CC0-1.0'],
+        'license': [license_to_use],
```
jjnesbitt: Why not just this (built off my other suggestion):
```suggestion
license = get_instance_config().licenses[0]
version_metadata = {
'description': 'An informative description',
        'license': license,
```

jjnesbitt (changes requested):

dandiapi/settings/base.py
```diff
@@ -173,6 +174,8 @@
     'DJANGO_DANDI_SCHEMA_VERSION', default=_DEFAULT_DANDI_SCHEMA_VERSION
 )
 
+DANDI_SCHEMA_INSTANCE_CONFIG = get_schema_instance_config()
```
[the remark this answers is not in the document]
candleindark: I will however `from dandischema.conf import get_instance_config as get_schema_instance_config` to distinguish the return as settings/config originated from the schema instance instead of the DANDI instance.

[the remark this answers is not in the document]
yarikoptic: > This file is for configuring settings that will be used elsewhere in the code base. This variable isn't a setting at all, it's more of a constant, or something similar.

@jjnesbitt I would argue that it is more of a *setting* than any of those `ROOT_URLCONF`, `INSTALLED_APPS` etc defined in this module: its value does depend and effected by settings of some of those `DJANGO_*` variables AFAIK as well, so to that degree I would say this file is the best location for it really. Indeed might be worth a comment that the values there are affected by DJANGO_ variables as described e.g. in README.md of https://github.com/dandi/dandi-schema/ (ATM in a branch of https://github.com/dandi/dandi-schema/pull/294)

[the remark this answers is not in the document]
candleindark: Done. Requested change implemented.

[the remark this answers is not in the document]
candleindark: @yarikoptic I applied the requested change already. So I am proceeding forward for now.

dandiapi/api/models/metadata.py
```diff
@@ -20,10 +28,10 @@ def published_by(cls, now: datetime.datetime):
             'wasAssociatedWith': [
                 {
                     'id': uuid4().urn,
-                    'identifier': 'RRID:SCR_017571',
-                    'name': 'DANDI API',
+                    **({'identifier': instance_identifier} if instance_identifier else {}),
+                    'name': f'{instance_name} API Server',
                     # TODO: version the API
-                    'version': '0.1.0',
+                    'version': importlib.metadata.version('dandiapi'),
```
yarikoptic: this one though indeed would be "an improvement" so let's keep the library version here.

dandiapi/api/tests/test_asset.py
```diff
@@ -124,10 +131,10 @@ def test_publish_asset(draft_asset: Asset):
             'wasAssociatedWith': [
                 {
                     'id': URN_RE,
-                    'identifier': 'RRID:SCR_017571',
-                    'name': 'DANDI API',
+                    **({'identifier': instance_identifier} if instance_identifier else {}),
+                    'name': f'{instance_name} API Server',
```
yarikoptic: per above revert
```suggestion
                    'name': f'{instance_name} API',
```

dandiapi/api/models/metadata.py
```diff
@@ -20,10 +27,10 @@ def published_by(cls, now: datetime.datetime):
             'wasAssociatedWith': [
                 {
                     'id': uuid4().urn,
-                    'identifier': 'RRID:SCR_017571',
-                    'name': 'DANDI API',
+                    **({'identifier': instance_identifier} if instance_identifier else {}),
+                    'name': f'{instance_name} API Server',
```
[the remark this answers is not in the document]
candleindark: Done. Change applied with adjustment to tests.

dandiapi/api/tests/test_asset.py
candleindark: Done.

candleindark:
@mvandenburgh @jjnesbitt @waxlamp This is ready to review. Beware of the remaining TODOs regarding the pinning of `dandischema` dependency before the merge.

.github/workflows/frontend-ci.yml
```diff
@@ -78,6 +78,7 @@ jobs:
       DJANGO_DANDI_WEB_APP_URL: http://localhost:8085
       DJANGO_DANDI_API_URL: http://localhost:8000
       DJANGO_DANDI_JUPYTERHUB_URL: https://hub.dandiarchive.org/
+      DJANGO_DANDI_INSTANCE_NAME: DANDI
```
waxlamp: Is there a reason why only `DJANGO_DANDI_INSTANCE_NAME` appears here? Is it because the other values are not tested in CI?

It might be good to include all the variables here so that the setup is complete (even if some aspects of the setup are not being tested or relied on).

web/src/components/DandisetList.vue
```diff
@@ -10,7 +10,7 @@
       :to="{
         name: 'dandisetLanding',
         params: { identifier: item.dandiset.identifier },
-        query: { ...$route.query, pos: getPos(index) },
+        query: { ...route.query, pos: getPos(index) },
```
waxlamp: Confirm that this is correct. This change seems like it shouldn't be here.

pyproject.toml
```diff
@@ -9,12 +9,11 @@ license = "Apache-2.0"
 license-files = ["LICENSE", "NOTICE"]
 classifiers = ["Private :: Do Not Upload"]
 dependencies = [
-  # Runtime dependencies, always needed
   "boto3",
   "celery",
   "dandi",  # minimal version is also provided in API /info
   # Pin dandischema to exact version to make explicit which schema version is being used
-  "dandischema==0.11.1",  # schema version 0.6.10
+  "dandischema @ git+https://github.com/dandi/dandi-schema.git@devendorize",
```
waxlamp: Change this to the new version of `dandischema` (i.e. `0.12.0`).

waxlamp (changes requested):
Thanks, @candleindark, for all the work on this, and for working with us on the strategy for getting it merged and released.

I've requested a bunch of changes in this review, and @mvandenburgh and @jjnesbitt will go ahead and make those changes.

dandiapi/api/management/commands/create_dev_dandiset.py
```diff
@@ -31,16 +31,12 @@
...
-    license_to_use = (
-        'spdx:CC0-1.0' if 'spdx:CC0-1.0' in supported_licenses else supported_licenses[0]
-    )
+    # The licenses field is a set, sort values to ensure consistent behavior
+    licenses = sorted(x.value for x in get_schema_instance_config().licenses)
 
     version_metadata = {
         'description': 'An informative description',
-        'license': [license_to_use],
+        'license': licenses[0],
```
candleindark: @jjnesbitt The verbose code for picking the `license` value I put in was an attempt to keep the original behavior as much as possible. I want you to be aware that with https://github.com/dandi/dandi-schema/pull/294, `dandischema.models.LicenseType` is different across different instances of DANDI as defined in https://github.com/dandi/dandi-schema/blob/33ac78163c8b92470cac2197aa06912a8a18f0de/dandischema/models.py#L122-L133. The `'spdx:CC0-1.0'` license is not guaranteed to be one of the support licenses since supported licenses are specified via https://github.com/dandi/dandi-schema/blob/33ac78163c8b92470cac2197aa06912a8a18f0de/dandischema/conf.py#L162-L171. But of course, this consideration doesn't matter if there is nothing particular about the `'spdx:CC0-1.0'` license.

However, the value for the `license` key was a `list[str]` not a `str`.

.github/workflows/frontend-ci.yml
candleindark: @waxlamp Sorry for accidentally deleting your comment about this line you posted earlier.

Your question was regarding why other environment variables are not included to set [`dandischema.conf.Config`](https://github.com/dandi/dandi-schema/blob/33ac78163c8b92470cac2197aa06912a8a18f0de/dandischema/conf.py#L89-L234).

There are five fields in `dandischema.conf.Config` to set. The `DJANGO_DANDI_INSTANCE_NAME` env var sets `instance_name`. There is already `DJANGO_DANDI_WEB_APP_URL` env var that sets `instance_url`. `licenses` defaults to the value for DANDI Archive. The remaining fields to be set are `instance_identifier` and `doi_prefix`, and they can be set by `DJANGO_DANDI_INSTANCE_IDENTIFIER` and `DJANGO_DANDI_DOI_API_PREFIX` env var respectively. However, the frond end code doesn't seem to require those values.

If you want to input set the remaining field, you may want to pick from those in https://github.com/dandi/dandi-infrastructure/pull/224/files#diff-1a87a9189dbcd06318fab560eeb95c27d4c7dd16ef4574ffdf368cb2b63d7e30.

dev/.env.docker-compose-native
```diff
@@ -5,3 +5,4 @@ DJANGO_MINIO_STORAGE_URL=http://minioAccessKey:minioSecretKey@localhost:9000/dan
 DJANGO_DANDI_WEB_APP_URL=http://localhost:8085
 DJANGO_DANDI_API_URL=http://localhost:8000
 DJANGO_DANDI_JUPYTERHUB_URL=https://hub.dandiarchive.org/
+DJANGO_DANDI_INSTANCE_NAME=DANDI
```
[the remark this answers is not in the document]
candleindark: See https://github.com/dandi/dandi-archive/pull/2584#discussion_r2478966404 for info.

dandiapi/api/models/metadata.py
```diff
@@ -1,15 +1,23 @@
...
+import importlib
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
+from dandischema.conf import get_instance_config as get_schema_instance_config
+
 if TYPE_CHECKING:
     import datetime
 
+_SCHEMA_INSTANCE_CONFIG = get_schema_instance_config()
```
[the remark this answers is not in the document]
candleindark: Indeed, current arrangement does have a tradeoff. Do it in a way that you prefer.

dandiapi/api/views/info.py
```diff
@@ -71,6 +77,13 @@ def info_view(request):
     api_url = f'{settings.DANDI_API_URL}/api'
     serializer = ApiInfoSerializer(
         data={
+            'instance_config': _INSTANCE_CONFIG.model_dump(
+                # Not excluding any `None` value fields in this object because the `None` values are
+                # needed to reconstitute a `dandischema.conf.Config` instance properly  in any
+                # receiving client since a corresponding environment variable to a field is used to
+                # set the value of the field if the argument for a particular field is not provided.
```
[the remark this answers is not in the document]
jjnesbitt: I understand your concern Isaac. I updated this line to explicitly set that argument (even though it's already set by default), and reworded the comment.

.github/workflows/frontend-ci.yml
candleindark: @yarikoptic Is making `DJANGO_DANDI_INSTANCE_IDENTIFIER` required an issue if one of your goals is to support people running DANDI instances in isolation without support for publishing?

yarikoptic: IMHO it should not be required, as we would just default to the adhoc one and instance should just rely on what schema tells it to be.
But overall I think it is fine as long as the `infrastructure` changes are up to date (or beyond) with the server for which they are intended for. So infrastructure could/should just introduce them **before** (ie. now) even adopting a new version of the dandi-archive which requires it. And for DOI we have other more DOI-specific settings to enable or disable it.

yarikoptic:
FWIW - https://pypi.org/project/dandischema/0.12.0/ is now out. Test against it?!

PR #2587, opened 2025-10-07
Use new OAuth-Toolkit setting to support wildcard redirects

brianhelba:
To deploy, on the staging environment, we'll need to set `DJANGO_OAUTH2_ALLOW_URI_WILDCARDS=True`.

---

**For the EMBER Archive:**
This PR has two follow up PRs (#2602 and #2603), as the release of this didn't initially succeed. Instead of merging these one by one, you should probably just jump to #2603, merging them all at once. 

cc @sandyhider @NEStock


jjnesbitt (approved):
Looks good, will be nice to get rid of that workaround. Just had one question, not blocking.

jjnesbitt:
For some reason, github errors out when I try to leave a comment on the deleted file `staging_heroku_production.py`. My question was:

Just to be clear, it seems this file was never actually in use, since https://github.com/dandi/dandi-infrastructure/pull/237 just switched the settings file to `heroku_production`, correct?

dandiapi/api/migrations/0001_stagingapplication.py
```diff
@@ -5,19 +5,11 @@
...
-# This migration is needed to support wildcards in OAuth redirect URIs in staging.
-# The model created here is only actually used in staging.
-
 
 class Migration(migrations.Migration):
     # This has no dependencies, but isn't considered "initial" by Django's definition
     initial = False
 
-    # This is creating a swappable model, so it needs to exist before the OAuth2 models are created
-    run_before = [
```
brianhelba, line 17: This model doesn't matter anymore, so there's no need to keep a particular ordering with respect to OAuth-Toolkit.

dandiapi/api/migrations/0028_default_oauth_application.py
```diff
@@ -0,0 +1,48 @@
...
+        skip_authorization=True,
+    )
+
+
+def reverse_create_application(apps, schema_editor):
+    Application = apps.get_model(oauth2_settings.APPLICATION_MODEL)
+    Application.objects.filter(client_id=DEFAULT_CLIENT_ID).delete()
+
+
+class Migration(migrations.Migration):
```
brianhelba, line 33: This is essentially a rename of `0003_default_oauth_application`, but it needs to run late in the `api` app's ordering, so that reverting migrations from the `oauth2_provider` app won't also revert many other `api` migrations.

dandiapi/api/migrations/0028_default_oauth_application.py
```diff
@@ -0,0 +1,48 @@
...
+        authorization_grant_type='authorization-code',
+        client_secret='',
+        name=DEFAULT_NAME,
+        # This can be turned off in production if appropriate.
+        skip_authorization=True,
+    )
+
+
+def reverse_create_application(apps, schema_editor):
+    Application = apps.get_model(oauth2_settings.APPLICATION_MODEL)
```
brianhelba, line 29: I added a true reverse step here. I'm not sure what I was thinking before, but if we're going to actually reverse this step, we should accept the consequences that it'll delete the record.

brianhelba:
I've pushed some additional changes, all related to migrations. I made some comments about rationale for change details inline.

When this is merged, my expectation is that on production and for any long-lived developer instances, only `0027_delete_stagingapplication` will actually run.

On staging, I'm not certain exactly what will happen; the Postgres database on that (and only that, to my knowledge) instance has a number of other OAuth-Toolkit tables with inbound ForeignKey references to the table `api_stagingapplication` (which `0027_delete_stagingapplication` is going to try to drop). Regardless of whether the migration explicitly fails or succeeds and leaves the database in an invalid state, my plan is to immediately:
* run `./manage.py migrate oauth2_provider zero`; this will drop all OAuth-Toolkit tables
* run `./manage.py migrate`; this will create all the OAuth-Toolkit tables again, but this time, all will reference the internal `Application` model (since that's what the default value of OAuth-Toolkit setting indicates); this will also (re?)run `0028_default_oauth_application`
* use the Django Admin to correctly reconfigure the application to work with Netlify, with the same values it had before

None of this should impact other third-party deployments of this codebase, unless they were explicitly configured to opt-in to using `StagingApplication` at some point in the past.

The takeaway is that this should fix the failures with branch previews on staging, while simplifying DANDI's code.

brianhelba:
> Just to be clear, it seems this file was never actually in use, since [dandi/dandi-infrastructure#237](https://github.com/dandi/dandi-infrastructure/pull/237) just switched the settings file to `heroku_production`, correct?

Yes, it was an oversight that staging never ended up actually using `staging_heroku_production`. This essentially fixes that bug, so staging should start working as intended.

PR #2589, opened 2025-10-08
Upgrade `djangorestframework`

### LINKED ISSUE #1896: DRF 3.15.0 breaks tests
Example failing test run using DRF 3.15.0 - https://github.com/dandi/dandi-archive/actions/runs/8283015197/job/22853304575

We should read through the [release notes](https://www.django-rest-framework.org/community/release-notes/#315x-series) and [DRF issues](https://github.com/encode/django-rest-framework/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) to see what we need to change to use the latest DRF release.

brianhelba:
Fixes #1896


dandiapi/urls.py
```diff
@@ -45,13 +45,13 @@
     .register(
         r'versions',
         VersionViewSet,
-        basename='version',
+        basename='dandiset-version',
```
brianhelba, line 48: [DRF 3.15 changes](https://www.django-rest-framework.org/community/release-notes/#3150):
> Raise ImproperlyConfigured exception if `basename` is not unique [[#8438](https://github.com/encode/django-rest-framework/pull/8438)]

I think this is a sensible new name, since the viewsets are nested.

dandiapi/zarr/tests/test_zarr.py
```diff
@@ -57,12 +57,12 @@ def test_zarr_rest_dandiset_malformed(authenticated_api_client, user):
...
 @pytest.mark.django_db
-def test_zarr_rest_create_not_an_owner(authenticated_api_client, zarr_archive):
+def test_zarr_rest_create_not_an_owner(authenticated_api_client, dandiset):
     resp = authenticated_api_client.post(
         '/api/zarr/',
         {
-            'name': zarr_archive.name,
-            'dandiset': zarr_archive.dandiset.identifier,
+            'name': 'New Zarr',
+            'dandiset': dandiset.identifier,
```
brianhelba, line 65: This might be an information leakage problem. Look at [the underlying view](https://github.com/dandi/dandi-archive/blob/2ff1bc7a954d7b2ec467fbbdf31f8c1a058d157c/dandiapi/zarr/views/__init__.py#L137-L145). Previously, if a non-owned `ZarrArchive` (in that the underlying `Dandiset` wasn't owned) actually existed, calling this would get all the way to [the permission check](https://github.com/dandi/dandi-archive/blob/2ff1bc7a954d7b2ec467fbbdf31f8c1a058d157c/dandiapi/zarr/views/__init__.py#L144-L145C19) before failing with a 403.

After [the DRF 3.16 changes](https://www.django-rest-framework.org/community/release-notes/#3160):
> [This] may change the behaviour of existing features and pre-existing behaviour. Specifically, some fixes were added to around the support of UniqueConstraint with nullable fields which will improve built-in serializer validation.
>
> Fix raising on nullable fields part of `UniqueConstraint` in [#9531](https://github.com/encode/django-rest-framework/pull/9531)

This means that since the request's new `ZarrArchive` violates the model's uniqueness constraint, the view will quickly fail [when validating the input `ZarrSerializer`](https://github.com/dandi/dandi-archive/blob/2ff1bc7a954d7b2ec467fbbdf31f8c1a058d157c/dandiapi/zarr/views/__init__.py#L138), returning a 400.

Technically, this 400 reveals information about the existence of a non-owned Dandiset. I don't know whether this actually matters (maybe the full information is readable anyway?), but it might also have implications for embargo. Perhaps this is actually fine, if all it's revealing is that "yes, a hidden `Dandiset` with this identifier exists".

Finally, note that I rewrote the test to ensure "non-owners can't create new `ZarrArchive`", which does still return 403. We could make the test instead ensure "non-owners can't create new _duplicate_ `ZarrArchive`", which is closer to the old test, but would now return 400.

dandiapi/zarr/views/__init__.py
```diff
@@ -151,16 +151,9 @@ def create(self, request):
...
-            try:
-                with transaction.atomic():
-                    zarr_archive.save()
-            except IntegrityError as e:
-                raise ValidationError('Zarr already exists') from e
-
-            audit.create_zarr(dandiset=dandiset, user=request.user, zarr_archive=zarr_archive)
+        zarr_archive: ZarrArchive = ZarrArchive.objects.create(name=name, dandiset=dandiset)
+
+        audit.create_zarr(dandiset=dandiset, user=request.user, zarr_archive=zarr_archive)
```
brianhelba, line 156: Background on this code is here: https://github.com/dandi/dandi-archive/pull/1886#discussion_r1698990052

Now that the serializer enforces the uniqueness constraint, it seems like it'd have to be an rare race condition for the save to fail here. In that case, I think it's fine to allow the `IntegrityError` to just bubble up. That seems rare enough that the defensive code isn't worthwhile (given that I'm sure there are lots of other places where an rare race condition could cause an constraint to be violated and a save to fail).

Alternatively, most of the function should be wrapped in a transaction, so that the uniqueness query done by the serializer happens in the same transaction as the save.

dandiapi/zarr/tests/test_zarr.py
```diff
@@ -79,7 +79,9 @@ def test_zarr_rest_create_duplicate(authenticated_api_client, user, zarr_archive
         },
     )
     assert resp.status_code == 400
-    assert resp.json() == ['Zarr already exists']
+    assert resp.json() == {
+        'non_field_errors': ['The fields dandiset, name must make a unique set.']
+    }
```
brianhelba, line 95: This is the new error raised by the serializer.

jjnesbitt (dismissed):

dandiapi/zarr/views/__init__.py
brianhelba: Re-added the explicit capture of `IntegrityError`.

dandiapi/zarr/views/__init__.py
```diff
@@ -49,7 +51,35 @@ class Meta:
         ]
         fields = ['name', 'dandiset', *read_only_fields]
 
-    dandiset = serializers.RegexField(f'^{Dandiset.IDENTIFIER_REGEX}$')
+    dandiset = serializers.PrimaryKeyRelatedField(
+        queryset=Dandiset.objects.all(),
+        pk_field=DandisetIdentifierField(),
+    )
```
jjnesbitt: Does this automatically raise an exception if the specified dandiset ID is not found? And if so, what's the response code (seems like it might be 400 instead of 404)?

jjnesbitt (changes requested):

dandiapi/zarr/views/__init__.py
```diff
@@ -49,7 +51,35 @@ class Meta:
...
+
+    def __init__(self, instance=None, data=serializers.empty, **kwargs):
+        super().__init__(instance=instance, data=data, **kwargs)
+
+        # If this was loaded via ".get_serializer", then the request is available in context
+        self.user: User | AnonymousUser | None = (
+            self._context['request'].user if 'request' in self._context else None
+        )
```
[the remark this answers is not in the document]
brianhelba: Note slightly below in `validate_dandiset`. We do an additional sanity check for `self.user` being set.

dandiapi/api/views/serializers.py
```diff
@@ -43,7 +44,27 @@ class UserDetailSerializer(serializers.Serializer):
...
+    def to_internal_value(self, data):
+        if not isinstance(data, str | int):
+            self.fail('invalid')
+        if isinstance(data, str):
+            if not re.fullmatch(Dandiset.IDENTIFIER_REGEX, data):
+                self.fail('invalid')
+            try:
+                return int(data)
+            except (ValueError, TypeError):
+                self.fail('invalid')
```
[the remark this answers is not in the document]
brianhelba: Good point, I'll fix.

dandiapi/zarr/views/__init__.py
brianhelba: Yes, `PrimaryKeyRelatedField` validates the existence of the reference (that's why `queryset` is needed).

If the `dandiset` argument is not found or is invisible to the user when calling `POST /api/zarr/`, it now raises a 400, rather than a 404. I'd consider this to be a bugfix, since it's really just communicates that the particular argument to the REST request was invalid.

dandiapi/api/views/serializers.py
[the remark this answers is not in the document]
brianhelba: I've added range checking. Note, the regex actually permits `000000` to be a valid dandiset identifier (even though we know it'll never be a PK), so I allowed `DandisetIdentifierField` to accept that.

What it really just means is that `dandiset=000000` will 404 instead of 400.

I'm fine to also reject `0`, but it'll mean that failure behavior might be different depending on whether `Dandiset.IDENTIFIER_REGEX` is used directly to validate or not (and it still is used directly in some places).

dandiapi/api/views/serializers.py
```diff
@@ -43,7 +44,26 @@ class UserDetailSerializer(serializers.Serializer):
     status = serializers.CharField()
 
 
+class DandisetIdentifierField(serializers.Field[int, str | int, str, Any]):
+    default_error_messages = {'invalid': 'A valid Dandiset identifier is required.'}
+
+    def to_internal_value(self, data: Any) -> int:
+        if not isinstance(data, str | int):
```
jjnesbitt, line 51: I was not aware you could use this syntax 👍

jjnesbitt (approved):

dandiapi/api/views/serializers.py
brianhelba: @jjnesbitt Yup, as of 3.10: https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator

yarikoptic:
@jjnesbitt approved 5 days ago -- merge?

waxlamp:
> @jjnesbitt approved 5 days ago -- merge?

We'd like to resolve the integration test failures first. @yarikoptic, we need a new release of the CLI to fix that issue.

yarikoptic:
it was released, I restarted failed jobs

PR #2593, opened 2025-10-09
Update tooltip width so that it is readable

### LINKED ISSUE #2393: `Embargo this Dandiset` tooltip is difficult to read
On Chrome version 129.0.6668.101 and Safari version 18.0.1:

Image

kabilar:
Fix #2393

### DIFF
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -13,7 +13,7 @@
               Embargo this Dandiset
               <v-tooltip
                 location="right"
-                max-width="400px"
+                width="30vw"
               >
                 <template #activator="{ props }">
                   <div


mvandenburgh:
> Is it possible to utilize the deploy preview for updates like this one where I need to log in with github? I am receiving the following error:

That was a bug in our staging deployment, but it's fixed now. I confirmed logging into https://deploy-preview-2593--sandbox-dandiarchive-org.netlify.app/ works

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -13,7 +13,7 @@
               Embargo this Dandiset
               <v-tooltip
                 location="right"
-                max-width="25%"
+                max-width="400px"
```
mvandenburgh: While this does fix the problem, it sets the max width of this tooltip to a fixed value, irrespective of the user's display size/resolution. I think the intent of the original `max-width=25%` was to have it expand up to 25% of the display; we can accomplish this by setting `width` instead of `max-width` and using `vw` instead of pixels: 

```suggestion
                width="30vw"
```

I tested this with various screen sizes, and ended up bumping it to 30% of the viewport width to make it look good on small screens.

mvandenburgh (approved):

PR #2605, opened 2025-10-16
Switch from Yarn to npm

brianhelba:
This provides equivalent speed and capabilities, but works better with CI caching and is easier for users to obtain.


.gitattributes
```diff
@@ -1,2 +1,3 @@
+package-lock.json linguist-generated=true
 uv.lock linguist-generated=true
-web/yarn.lock linguist-generated=true
```
mvandenburgh: Just to be clear, we can remove `web` here because simply specifying `package-lock.json` will match _any_ file with that name in the repo, correct?

mvandenburgh (approved):

Issue #185, opened 2021-03-25
503 Application Error in HTML response upon POST to https://api.dandiarchive.org/api/uploads/initialize/


yarikoptic:
In follow up to #184 decided to reupload and it failed with Application Error

```
drogon:/tmp/000029
*$> DANDI_DEVEL=1 dandi upload -i dandi-api --allow-any-path  --validation skip --jobs 2:10 largeuploads/blobs/f5d/aac/79d9c6c1ba2439200d45680f1cb941087750a000e484ea2973232709d2 
2021-03-25 14:02:06,376 [    INFO] Found 1 files to consider
PATH                                                                                  SIZE    ERRORS UPLOAD STATUS                       MESSAGE                                                                                       
largeuploads/blobs/f5d/aac/79d9c6c1ba2439200d45680f1cb941087750a000e484ea2973232709d2 3.6 TB                ERROR                        Error 503 while sending POST request to https://api.dandiarchive.org/api/uploads/initialize/  
Summary:                                                                              3.6 TB                1 ERROR                      1 Error 503 while sending POST request to https://api.dandiarchive.org/api/uploads/initialize/
2021-03-25 14:02:39,326 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210325180158Z-12041.log
(dandi-devel) 
#92 !1223 [0].....................................:Thu Mar 25 14:02:39:.
drogon:/tmp/000029
$> cat /home/yoh/.cache/dandi-cli/log/20210325180158Z-12041.log
2021-03-25T14:01:58-0400 [INFO    ] dandi 12041:139975036532096 [META] sys.argv = ['/home/yoh/miniconda3/envs/dandi-devel/bin/dandi', 'upload', '-i', 'dandi-api', '--allow-any-path', '--validation', 'skip', '--jobs', '2:10', 'largeuploads/blobs/f5d/aac/79d9c6c1ba2439200d45680f1cb941087750a000e484ea297
3232709d2']
2021-03-25T14:01:58-0400 [INFO    ] dandi 12041:139975036532096 [META] os.getcwd() = /tmp/000029
2021-03-25T14:02:06-0400 [INFO    ] dandi 12041:139975036532096 Found 1 files to consider
2021-03-25T14:02:39-0400 [ERROR   ] dandi 12041:139974440777472 Error 503 while sending POST request to https://api.dandiarchive.org/api/uploads/initialize/: <!DOCTYPE html>
        <html>
          <head>
[...]

PR #2606, opened 2025-10-20
Add release notes functionality to dandiset publishing process

### LINKED ISSUE #189: publish: add "description" parameter / db column
- to relate to git `commit` or `tag -a` - it would be useful to everyone to let users annotate their published versions: what has changed and why
- should probably be some freeform TEXT blob of some smallish size
- it would allow to establish a "changelog" for each dandiset (not just some not-so-informative versions)

- probably should not be mandatory but I am ok to make it such: that could help to avoid users to unnecessarily mint versions
- like git commit message it is IMHO ok to have it immutable thus no need to expose some "edit" functionality via API

support for it should be used by web ui (yet to file a report)

bendichter:
fix #189

- Updated publish functions to accept optional release notes.
- Introduced PublishVersionSerializer for handling release notes in requests.
- Enhanced UI to allow users to input release notes during publishing.
- Display release notes for published versions in the UI.

https://github.com/user-attachments/assets/cd060bad-4d94-4453-8fd2-bd6be33c2593

The published metadata uses the `releaseNotes` field added to the schema in https://github.com/dandi/dandi-schema/pull/185, which is included in dandischema 0.12.1 (schema 0.7.0), the version already pinned in this repo.


dandiapi/api/views/serializers.py
```diff
@@ -162,12 +166,25 @@ class Meta:
...
+        """Extract release notes from version metadata."""
+        return version.metadata.get('releaseNotes')
+
+    def to_representation(self, instance):
+        """Remove release_notes from representation if it's None."""
+        representation = super().to_representation(instance)
+        if representation.get('release_notes') is None:
+            representation.pop('release_notes', None)
+        return representation
+
```
mvandenburgh: We should be able to access the `releaseNotes` directly from the metadata in the client, which would remove the need for this extra logic in this serializer. First, we'll need https://github.com/dandi/dandi-schema/pull/185/ to be released.

mvandenburgh (commented):
We may want to rethink how we're passing `release_notes` around, but I'll get back to you on that. Aside from that, this generally looks good.

Note - we will need https://github.com/dandi/dandi-schema/pull/185 to be released before this PR can be merged.

yarikoptic (approved):
oh, @bendichter we forgot about this one and I just got confused by some info box while trying to publish a dandiset but was not given a box to provide release notes -- I thought we already merged and released that functionality long ago.

Given that kitware folks already provided some reviews, and even contributed, I think we should be in a good shape.  Hence approving and giving you a chance for final look/merge.

NB: might be worth rebasing/merging master and ensuring CI is still green

bendichter:
thanks for the ping, @yarikoptic . I am doing another pass on this now.

bendichter:
Rebased on `master` and pushed a few follow-up changes after testing the publish + release-notes flow end-to-end against a local full stack (Django + Vue, real `publish_dandiset` path).

### Changes

**1. Merged `master` and resolved a migration collision.** The branch's `0031_version_release_notes` collided with `master`'s `0031_dandiset_embargo_end_date`, leaving two leaf nodes in the migration graph. Renamed it to `0033_version_release_notes` and rechained its dependency onto `0032_remove_upload_embargoed_upload_zarr_and_more` so the graph is linear again.

**2. Source release notes from `metadata.releaseNotes` instead of the `release_notes` column.** Previously `VersionSerializer.release_notes` read the model column. Because `publish_dandiset` writes that column on the **draft** (it's the transport mechanism into the publish task), the draft version echoed the last-published notes, and the UI showed a release-notes button on the draft. The serializer now reads `obj.metadata.get('releaseNotes', '')` — the canonical, schema-defined field (per dandi-schema #185). Published versions are unchanged; the draft correctly reports no notes. Updated `test_version_rest_list_with_release_notes` accordingly. This also addresses @mvandenburgh's earlier note about rethinking how `release_notes` is passed around — the column is now purely an internal publish-time detail.

   Verified safe for releases without notes: `Version.metadata` is `JSONField(default=dict)` (never `None`), so `.get(..., '')` returns `''`, and the frontend button is guarded by `v-if="version.release_notes"`, so no button/dialog and no error.

**3. Tidied the version-list layout in the sidebar.** The view button was wedged between the version arrow and the version number, cramping the row. The button is now a right-aligned trailing action with a "View release notes" tooltip, and the date is stacked under the version. The version column truncates with an ellipsis when the row is too narrow so the button always stays visible rather than being pushed off-screen.

### Verification

Ran a local docker-compose stack on this branch, published a dandiset with release notes through the real publish service, and confirmed via the API and UI that:
- the published version's `metadata.releaseNotes` and the serialized `release_notes` both carry the notes;
- the draft returns `release_notes: ""` and shows no button;
- the release-notes dialog renders the correct version's notes;
- the 6 `release_notes` backend tests pass via tox.

bendichter:
screenshots, since the preview doesn't allow you to view the backend changes. The addition is the box with lines in it on the sidebar section for versions:

image

image

image

image

PR #2612, opened 2025-10-24
Remove additional pytest-factoryboy fixtures

brianhelba:


dandiapi/api/tests/test_asset.py
```diff
@@ -491,9 +490,7 @@ def test_nested_asset_ordering(api_client, asset_factory, order_param, expected_
 def test_nested_asset_ordering_with_authenticated_user(api_client, asset_factory):
     """Test that ordering works with an authenticated user."""
     user = UserFactory.create()
-    draft_version = DraftVersionFactory.create()
-    # Add user as owner
-    add_dandiset_owner(draft_version.dandiset, user)
+    draft_version = DraftVersionFactory.create(dandiset__owners=[user])
```
jjnesbitt: I'm sure you're aware of this, but at some point we should really move away from calling `DraftVersionFactory` directly, and just have `DandisetFactory` always automatically create a draft version.

PR #2626, opened 2025-11-03
Update version regex in tests

jjnesbitt:
The Ember project had the following version string in their tests:

```
0.0.post1.dev7292+g6b9062efc
```
This broke the tests as the regex expects exactly **three** digits, but their version string only has **two**.

I also noticed that locally, my tests failed against this regex. I had the following version string:

```
0.11.5.post6+g54d98073.d20250709
```
This version string is missing the `dev` section, as well as has a different digit after `post` besides `1` . I updated the version regex to make that dev portion optional, and to generalize the `post` section.

@brianhelba I'm not sure if this breaks from the `no-guess-dev` versioning scheme.


jjnesbitt:
Superseded by #2635

PR #2630, opened 2025-11-04
feat: implement AND logic for multi-word dandiset search

### LINKED ISSUE #2629: DANDI search requires exact text match
DANDI search currently requires an exact text match. Take for example DANDI:000986: "Auditory cortex Neuropixels recordings and pupil diameter traces from mice during passive exposure to pure tones", if you search ["pure tones"](https://dandiarchive.org/dandiset/search?search=pure+tones) your results contain this dataset. However if you search ["mice pure tones"](https://dandiarchive.org/dandiset/search?search=mice+pure+tones) it does not, even though all of those words are in the title of the dataset. This leads me to believe that the search requires an exact match for the entire string, which is a pretty big deviation from expectations and also makes the search very limiting. I think this is the primary reason users think our search is bad.

I think the standard syntax that most people would expect is that this query would be translated as:

CONTAINS("mice") AND CONTAINS("pure") AND CONTAINS("tones").

I know there has been a lot of effort thinking about how to improve search, and there are many ways this could be improved, but in the meantime can we please just fix this specifically?

bendichter:
fix #2629

Split search terms into individual words and apply AND filtering logic so that all words must be present in any order. Previously, the search required an exact phrase match. This change enables more flexible searching by allowing words to appear in any position within the dandiset metadata.

### DIFF
--- a/dandiapi/api/tests/test_dandiset.py
+++ b/dandiapi/api/tests/test_dandiset.py
@@ -1268,6 +1268,26 @@ def test_dandiset_rest_search_many_versions(api_client):
 
 
 @pytest.mark.django_db
+def test_dandiset_rest_search_multiple_words_and_logic(api_client):
+    """Test that search splits words and applies AND logic."""
+    # Create a dandiset with specific metadata
+    draft_version = DraftVersionFactory.create()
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
LGTM. Tested locally. Seems like a good incremental improvement.

PR #2627, opened 2025-11-03
S3 Backup design doc

mvandenburgh:
This PR lays out a design for S3 backup using S3 Replication and the Glacier Deep Storage class. Related #524

PR #2637, opened 2025-11-11
S3 backup design doc (NESE)

CodyCBakerPhD:
NESE Tape analog/duplicate of #2627 

Based only on rough estimates at this point; need to follow up and get a more concrete commitment before

TL;DR though, somewhere between 1/4-1/3 of the cost of the S3 Replication/Deep Glacier approach


doc/design/s3-backup-nese.md
```diff
@@ -0,0 +1,82 @@
...
+- files < 100 MB copied to tape, but also remain on disk.
+
+## Limitations and Considerations
+
+- **Upper bound on total storage**: No quota for an upper limit of long-term storage has yet been provided, aside from the grand 70 PB total of the entire NESE store; the providers claim expansion is possible as need arises.
+- **Replication is eventually consistent**: No guarantees about replication speed (the time between an object finishing upload into the primary bucket and when it is available in the backup bucket) are provided. Using previous bandwidth experience to the old Dropbox backup, multi-gigabit speeds should be possible and is expected to keep up with ingest rates on the primary S3 bucket.
+
+## Cost
+
+Note that unlike the [Deep Glacier approach](https://github.com/dandi/dandi-archive/blob/b3e0a9df4188533723fb2ad4a95506aa724fc089/doc/design/s3-backup.md), no egress cost would be induced from the data transfer as all operations are strictly under the open data bucket.
```
mvandenburgh: Do we know this for sure? AWS would probably take notice if our data transfer costs suddenly started spiking on a recurring basis due to a backup script downloading the entire multi-PB bucket.
As far as I understand, this isn't a _technical_ limitation of the open data account; from our perspective, the billing for the entire AWS account is paid for by the AWS Open Data program. 

So, we could apply the same thinking to the Deep Glacier approach - if we put the backup bucket in the open data AWS account, the storage costs would also theoretically be covered by the open data program. In practice, it's more of a policy issue - is AWS willing to pay for backup? Framed that way, I don't think there is a clear winner between the two approaches in terms of AWS transfer costs.

doc/design/s3-backup-nese.md
```diff
@@ -0,0 +1,82 @@
...
+- all newly written files > 2 hr of modification time are copied to tape.
+- if fileset quota > 99%; files are stubbed (replaced with a small pointer) down to 75% quota.
+- files with access time age > 2 weeks are stubbed.
+- files < 100 MB copied to tape, but also remain on disk.
+
+## Limitations and Considerations
+
+- **Upper bound on total storage**: No quota for an upper limit of long-term storage has yet been provided, aside from the grand 70 PB total of the entire NESE store; the providers claim expansion is possible as need arises.
+- **Replication is eventually consistent**: No guarantees about replication speed (the time between an object finishing upload into the primary bucket and when it is available in the backup bucket) are provided. Using previous bandwidth experience to the old Dropbox backup, multi-gigabit speeds should be possible and is expected to keep up with ingest rates on the primary S3 bucket.
+
```
mvandenburgh: I think there are two additional considerations worth thinking about here.

1) If we're considering a cost basis purely based on monthly bills for cloud services, the NESE backup approach is clearly cheaper. But, as mentioned further up in the document, there is likely going to be some manual work to perform the transfer. This will consume engineering time and add to the maintenance burden of the system much more than the Deep Glacier approach; outside of the initial work to set it up and do the initial transfer, the Deep Glacier approach is largely hands-off and automatic.

2) One of the motivations for backup is obviously data security - the data collected on DANDI is the most valuable part of the project, and any data loss bugs could be catastrophic for the project as a whole. This is one of the reasons #2367 is currently on hold; we want to get an ironclad backup system into place before we start deleting "garbage" from the bucket, in case there is a bug in our code. With the Deep Glacier design, there is essentially no hand-rolled code being run - it is simply configuring various options in AWS to facilitate backup and then deferring the actual backup/replication logic to AWS. With the NESE design, there would presumably need to be at least some custom code/scripts written to facilitate the download of data from S3 followed by the upload to Dartmouth's system; this theoretically leaves more room for bugs to sneak in.

**Note**, I'm not necessarily arguing for or against either design at this point (I haven't read this enough times/thought it through enough yet), I just want to make sure these points are discussed.

PR #2639, opened 2025-11-12
Update devcontainer section in readme

mvandenburgh:
- Adds commands for running `celery` and the frontend (currently, we only specify the command to run the Django server)
- Moves devcontainer configuration to the top of the readme (swapping with "Develop with Docker"), labeling it as the "recommended quickstart".
- Adds step to account for the permission issue that happens when attempting to run devcontainers on a system that previously followed the "Develop with Docker" configuration
	- I went back and forth on whether this should be included, but I think it makes sense since we're still officially supporting the "Develop with Docker" configuration alongside devcontainers.


jjnesbitt (approved):

danlamanna (approved):
LGTM.

I think we should look at switching to eager celery execution for development in the future.

PR #2642, opened 2025-11-13
Add create_test_user management command

jjnesbitt:
Sometimes we have to create local users that are _not_ super users, for the purposes of testing various features. This PR adds a management command that does so.


mvandenburgh (changes requested):
LGTM, I think we should wrap the insert operations in a transaction, though

mvandenburgh (approved):

PR #2644, opened 2025-11-13
Remove `DANDI_SCHEMA_VERSION` as a setting

jjnesbitt:
Depends on #2584

For historical reasons, we have a way to specify the dandi schema version as an env var. However, this is never used and is rather nonsensical, since we'd be defining the dandi schema version in conflict with whatever the actual schema version is. As it stands now, the version is pulled from `dandischema.consts` anyway. This PR simply imports that value directly from `dandischema`, replaces all uses of the setting with that value, and removes the setting.

### DIFF
--- a/dandiapi/api/management/commands/create_dev_dandiset.py
+++ b/dandiapi/api/management/commands/create_dev_dandiset.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from uuid import uuid4
 
 from dandischema.conf import get_instance_config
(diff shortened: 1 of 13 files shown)


mvandenburgh (approved):

jjnesbitt:
Putting in draft until #2584 is merged.

yarikoptic:
attn @jjnesbitt - #2584 was merged, should we proceed ?

PR #2647, opened 2025-11-17
Release note suggested changes

mvandenburgh:
This PR functions as a "requested changes" review for #2606. 

Changes from the design in the original PR:

- In the version response serializer, instead of making the presence of the `release_notes` field conditional on whether a Version has release notes or not, it _always_ includes it, simply returning an empty string if there are no release notes. This is a cleaner API design and makes things such as Swagger/OpenAPI specs easier to generate.

- Moved release notes to be stored on the `Version` model directly, and then write it to the metadata upon publish (very similar to how `doi` works currently, but obviously slightly different semantics). This simplifies the code significantly, as we're no longer having to pass the release_notes string through several layers of function calls. Additionally, this avoids having to place the entire `release_notes` string on the RabbitMQ queue; from what I understand, doing that is _technically_ fine, but usually we try to avoid that pattern and only place database primary keys on the queue.

### DIFF
--- a/dandiapi/api/tests/test_version.py
+++ b/dandiapi/api/tests/test_version.py
@@ -1069,48 +1069,13 @@ def test_version_publish_with_release_notes(api_client: APIClient, draft_asset_f
 
 
 @pytest.mark.django_db
-def test_version_publish_without_release_notes(api_client: APIClient, draft_asset_factory):
-    """Test publishing a dandiset without release notes."""
-    user = UserFactory.create()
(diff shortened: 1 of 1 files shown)


yarikoptic:
FWIW we have already released 0.12.0 (and 0.12.1) of dandi-schema after which should support release notes in metadata and master has that since 
- 1aed085706008464d9fefdf9566c9d9ef929dfd4

so the "foundation" should be ready

bendichter:
These changes sound fine to me. This feature came up in a meeting with @magland today. I just want to make sure it's still on your radar.

Issue #2648, opened 2025-11-21
Migrate identifiers into instance ones from hardcoded DANDI:


yarikoptic:
sandbox now runs "de/re-vendorized" version which allows for control of the identifier.
So sandbox became DANDI-SANDBOX not just DANDI.  That leads to validation errors even at the meditor level 

<img width="3334" height="1070" alt="Image" src="https://github.com/user-attachments/assets/db635898-383b-4786-a5a4-039a80c449b0" />

and then even more at the validation stage of previously published (on staging) dandiset:

<img width="1952" height="1036" alt="Image" src="https://github.com/user-attachments/assets/30fbc380-3758-42ec-8a85-b82f72623f2f" />

Since we have no means for users to edit that field, I wonder how we should address it.

- we likely should not care about what is already published, as we do not care to adjust/revalidate ATM older json records
- in draft version:
    - dois are likely to be just wiped out upon upgrade -- new publication would get one
    - id: and identifier: could be automagically remapped if should be anything but DANDI

jjnesbitt:
Upon further investigation, the issue isn't that the underlying data isn't being migrated. Rather, it's that the underlying data _is_ changing (since those fields are generated on save), but the `schemaVersion` listed in the metadata is _not_ changing. This means that we're trying to validate metadata that contains our new schema values against the old schema (and its static `DANDI` values), causing it to fail.

We can address this by updating the `schemaVersion` values on all draft versions in the archive, and explicitly saving them. That will ensure that validation is re-run, and any old validation errors are cleaned out.

Ironically, this won't really affect production too much, but it will affect any other instances, like sandbox, ember, and linc. So this will be done as a management command that can be run from the respective archive's deployment.

jjnesbitt:
> idea [#2650](https://github.com/dandi/dandi-archive/pull/2650) inspired: add option to `migrate_version_metadata` to migrate those identifiers. note that might also want an option to go through published ones too, not just `draft`.

I've split these up into two different issues (see #2653), and renamed this issue.

PR #2650, opened 2025-11-24
Update management command for schema migration

jjnesbitt:
Closes https://github.com/dandi/dandi-archive/issues/2648

The vendorization update takes effect within the archive immediately, while any existing data is validated against the _old_ schema, since the `schemaVersion` of any dandiset hasn't yet been incremented. Since the data we pass to the validation logic in `dandischema` is populated at runtime, it introduces values that are based on the _new_ schema, while `dandischema` is trying to validate that data against the  _old_ schema, causing validation errors.

This PR augments the existing `migrate_version_metadata` management command, to both fix (last modified 4 years ago) and improve the behavior.

This will be required to run on any DANDI deployment that isn't https://dandiarchive.org, as that is the only deployment where the values haven't actually changed.

### DIFF
--- a/dandiapi/api/management/commands/migrate_version_metadata.py
+++ b/dandiapi/api/management/commands/migrate_version_metadata.py
@@ -53,8 +53,6 @@ def migrate_version_metadata(dandisets: tuple[int, ...], *, include_all: bool):
 
             try:
                 metanew = migrate(locked_version.metadata, skip_validation=True)
-                # We set this manually due to https://github.com/dandi/dandi-schema/issues/353
-                metanew['schemaVersion'] = DANDI_SCHEMA_VERSION
             except Exception as e:
                 logger.exception('Failed to migrate %s', version, exc_info=e)
                 failed_count += 1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,9 +12,9 @@ dependencies = [
   # Runtime dependencies, always needed
   "boto3",
   "celery",
-  "dandi",  # minimal version is also provided in API /info
+  "dandi", # minimal version is also provided in API /info
   # Pin dandischema to exact version to make explicit which schema version is being used
-  "dandischema==0.12.0",  # schema version 0.7.0
+  "dandischema==0.12.1", # schema version 0.7.0
   "django[argon2]",
   "django-allauth",
   "django-auth-style",
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.13"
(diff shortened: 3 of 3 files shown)


dandiapi/api/management/commands/migrate_schema_versions.py
```diff
@@ -0,0 +1,58 @@
...
+            locked_version.status = Version.Status.PENDING
+            locked_version.save()
+
+            audit.update_metadata(
+                dandiset=locked_version.dandiset,
+                metadata=locked_version.metadata,
+                user=None,
+                admin=True,
+                description=f'Update schema version to {DANDI_SCHEMA_VERSION}',
+            )
```
jjnesbitt: @waxlamp I opted to make use of the `admin` feature of our audit record system. Does the inclusion of an audit record at all make sense here? I would think so, since the underlying metadata is still changing, but I'd like a second opinion.

dandiapi/api/management/commands/migrate_schema_versions.py
```diff
@@ -0,0 +1,58 @@
...
+    if dandisets:
+        versions = versions.filter(dandiset_id__in=dandisets)
+
+    logger.info('Running migration for %s versions', versions.count())
+
+    for version in versions.iterator():
+        logger.info('-----------------------------------------')
+        with transaction.atomic():
+            locked_version = Version.objects.select_for_update().get(id=version.id)
+            locked_version.metadata['schemaVersion'] = DANDI_SCHEMA_VERSION
```
yarikoptic: I do not see here an actual [`dandischema.metadata.migrate`](https://github.com/dandi/dandi-schema/blob/master/dandischema/metadata.py#L340) call which would migrate the model... we cannot just slap a new version onto an older one - some fields might need to change.

dandiapi/api/management/commands/migrate_version_metadata.py
```diff
@@ -1,28 +1,92 @@
...
+            )
+
+            migrated_count += 1
+            logger.info('Metadata migrated for version %s', version)
+
+        # Validate outside of transaction, since this function uses `select_for_update` itself
+        validate_version_metadata(version=version)
+
+    logger.info(
+        'Migrated %s out of %s total selected versions',
```
yarikoptic:
```suggestion
        'Migrated %d out of %d total selected versions, no changes in %d, failed on %d',
```

yarikoptic:
@jjnesbitt could you please check 

- https://github.com/dandi/dandi-archive/pull/2658

to see if you would like to adopt it one way (as is) or another (craft an "inspired" test) so we ensure that this functionality is at least somewhat tested while we have it in the codebase?

PR #2659, opened 2025-12-02
Progress dandi-schema library forward (to 0.12.1) for fixed migrations

yarikoptic:
Would be beneficial (bug fixed) for
https://github.com/dandi/dandi-archive/pull/2650


jjnesbitt:
This looks good. I think I will incorporate this change into #2560 directly, so that there's no need for a workaround.

Issue #1902, opened 2024-03-22
Design and implement rate limiting policy


yarikoptic:
- https://github.com/dandi/dandi-archive/issues/1891
showed the need for means to either balance out or just to limit "heavy users" of the API server. A quick attempt is being worked on in 
- https://github.com/dandi/dandi-archive/pull/1899
to provide a quick mitigation to that issue, but it is just a stop gap measure.

As many other public (or paid) services have rate limiting, we can and should have one too. 
 
- it needs to be designed with current utilization patterns in mind to not cause unnecessary slow downs etc. May be it could be somehow made to be enforced only/when server experiencing high load etc. The point is that it should not unnecessarily complicate/negatively effect access to our services.
- well documented and described in the [Terms & Polices](https://www.dandiarchive.org/handbook/about/terms/)
- dandi-cli should be ensured to continue working, gracefully handling rate limiting, and possibly alerting users about need to authenticate to gain better access, or consider composition of more efficient scripts, etc.

yarikoptic:
Additional freshier references:
-  ran into an article on medium https://medium.com/@mathur.danduprolu/throttle-traffic-to-your-django-app-implementing-a-rate-limiting-algorithm-afc58643ed4a ("Throttling Traffic to Your Django App: Implementing a Rate Limiting Algorithm") stating 
> Implementing rate limiting in Django is a critical step in ensuring the security, performance, and fair usage of your web application. 
- Happen we want some unified reverse proxy ahead of all our components -- https://doc.traefik.io/traefik/ seems to be quite well known and has https://doc.traefik.io/traefik/middlewares/http/ratelimit/ 
- An "alternative" to rate limiting could be some kind of QoS, filed recently https://github.com/dandi/dandi-archive/issues/2085 

@waxlamp what are you thoughts on re-approaching implementing rate limiting?  It might not provide some ultimate solution to address all our issues, but it could add some guards.

PR #1899, opened 2024-03-20
Rate limit assets list endpoint for unauthenticated users

mvandenburgh:
This PR adds a rate limit of 60 requests per minute to the asset list endpoint for unauthenticated users. Logged-in users are not rate-limited. This is a short-term fix for #1891.

Issue #1891, opened 2024-03-15
Slow response times from asset list endpoint


mvandenburgh:
The asset list endpoint (`/api/dandisets/{identifier}/versions/{version}/assets`) can sometimes take multiple seconds to run. This seemingly occurs on dandisets with a large number of assets, regardless of what `page` is queried for.

mvandenburgh:
The culprit appears to be the database query. Relevant sentry issue: https://dandiarchive.sentry.io/issues/5060924930/?end=2024-03-15T04%3A40%3A12&project=5266078&query=&referrer=performance-related-issues-issue-stream&start=2024-03-15T03%3A14%3A38&utc=true

mvandenburgh:
I did some more exact tests. I used a script to iterate through each page of the assets using that endpoint, and these were the results:

![response_times](https://github.com/dandi/dandi-archive/assets/37340715/de77b783-6b6a-4527-a061-7ef6f50b880f)


Some response times approached 1 second, but it never got anywhere near 10 seconds like Heroku was suggesting. So, I went a little further and did the same thing, but I made the requests in parallel instead of one after the other. The results were much worse:
![response_times](https://github.com/dandi/dandi-archive/assets/37340715/8808a238-bf3f-4bff-a26d-58eab9ca64ff)

At one point there was a nearly 12 seconds (!) response time. 


Here's the (AI-generated) scripts I used - 
https://gist.github.com/mvandenburgh/9cb9810a2abcc5d3f6c70c5268ade0f0 (sequential requests)
https://gist.github.com/mvandenburgh/96c1f422810cbf5da39979e68bea1e2b (parallel requests)

satra:
thanks for digging into this @mvandenburgh - if there are quick solutions let's do that for the short term.

ps. for the longer term, i think we need to have a new design for the directory browser, effectively combining unified/faceted search and that component together. for large dandisets it's going to become unwieldy to humanly navigate them. (cc: @kabilar)

yarikoptic:
is there any chance to provide some basic caching (instead of rate limit) for such DB invocations or overall listing request -- would it potentially help after saturating the cache and thus leading to no DB interactions until one of the listings expiring, e.g. in a minute or so?

yarikoptic:
are there any options on DB node -- can we address problem partially by more powerful DB node with faster IO/throughput?

yarikoptic:
[...]

PR #2665, opened 2025-12-08
Add basic rate limit implementation

jjnesbitt:
Partially addresses https://github.com/dandi/dandi-archive/issues/1902

This PR adds basic rate limiting via DRF [throttling](https://www.django-rest-framework.org/api-guide/throttling/). This is a very permissive version of rate limiting, imposing a rate limit only for anonymous users. This limit is also purposefully set _above_ any number I could obtain through local usage of the GUI. The point of this first step of implementation is not to be a full solution, but to protect against the most extreme of use cases, with no impact on existing legitimate users, anonymous or otherwise.

This is very similar in implementation to a [previous attempt](https://github.com/dandi/dandi-archive/pull/1899) we had made to introduce rate limiting, although that was sort of made in response to a [crisis](https://github.com/dandi/dandi-archive/issues/1891). The current circumstances and implementation are slightly different, in the following ways:
1. The crux of the issue in the previous implementation was that the asset list endpoint was very slow, which meant that a trivial amount of requests to it could cause a DOS. Rate limiting cannot prevent this effectively. The asset list endpoint has been fixed, so this limitation no longer exists.
2. The previous implementation only applied to the asset list endpoint, while this implementation applies to all API requests.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -105,6 +105,9 @@ jobs:
     - name: Apply migrations to API server
       run: ./manage.py migrate
 
+    - name: Create any cache tables
+      run: ./manage.py createcachetable
+
     - name: Install test data
       run: ./manage.py loaddata playwright
 


jjnesbitt:
Need to resolve failing tests.

Issue #2663, opened 2025-12-04
Google search result displays error message


kabilar:
`We're sorry but the DANDI Archive doesn't work properly without JavaScript enabled. Please enable it to continue.`

<img width="916" height="729" alt="Image" src="https://github.com/user-attachments/assets/a2da4ec6-4a2f-4eb9-b2c3-fda23b57930b" />

PR #2669, opened 2025-12-12
Add sitemap.xml

### LINKED ISSUE #752: Better indexing for search engines

### LINKED ISSUE #2230: Even better indexing for search engines: sitemap.xml
- #752 
was closed but I think we still might want to do better since after we did fix also allowing robots to crawl it , google indexes again but it looks imho suboptimal:

Image

(21 is apparently for 000021 dandiset ;) ). Compare e.g. to much more succinct but IMHO better result for openneuro

Image

We even lack the favicon/logo!

I think re the record and better discovery, sitemap.xml could help.

waxlamp:
This PR adds a sitemap.xml view to the DANDI backend that should help with Google presenting search results for individual Dandisets. Together with activating prerendering through Netlify (see #2663), this should solve most of our (current) SEO problems.

### Details
- The sitemap.xml view is generated by Django's sitemaps framework
- A new robots.txt view presents a `Sitemap` record to visiting crawlers
- Both of these views are generated serverside and proxied to the frontend (using vite server proxy config for dev, and Netlify redirects for the sandbox and production deployments)
- While the sitemap is always generated, there's a new setting to control whether the sitemap *entries* are generated. This is because we will not want all DANDI deployments to be search engine indexed. The most salient example is the sandbox deployment: we only want Google to turn up references to production, so we'll leave this turned off for sandbox. This setting also controls whether the `Sitemap:` directive appears in the robots.txt view.

### Deployment
When this PR is merged/released, it will not have any immediate effect since the feature defaults to inactive. To activate it on select deployments, we need to set the `DJANGO_DANDI_ENABLE_SITEMAP_XML` environment variable to `True`; to do so for our prod deployment, the best way would be to send a PR to the `dandi-infrastructure` repo.

### TODO
- [ ] after merge/release, set up a dandi-infrastructure PR to activate for prod
- [ ] after merge/release, submit site to Google for reindexing

Closes #752 (this was already closed, but the current PR addresses it)
Closes #2230


waxlamp (commented):
I think this is pretty straightforward, so I'm just looking for obvious pitfalls.

dandiapi/urls.py
```diff
@@ -163,6 +165,16 @@ def to_url(self, value):
...
 
+if settings.DANDI_ENABLE_SITEMAP_XML:
+    urlpatterns += [
+        path(
+            'sitemap.xml',
+            sitemap,
+            {'sitemaps': sitemaps},
+            name='django.contrib.sitemaps.views.sitemap',
+        ),
+    ]
```
waxlamp: Forgot to mention it in my self-review, but is there any need to cache this view?

My guess is there's no need, at least not at the scale we're currently running at. Apparently, past 50000 entries, you have to split your sitemap.xml file up or risk Google ignoring you, but of course we're nowhere near that number.

dandiapi/sitemaps.py
```diff
@@ -0,0 +1,25 @@
...
+    priority = 0.5
+
+    def items(self):
+        return Dandiset.objects.filter(embargo_status=Dandiset.EmbargoStatus.OPEN)
+
+    def location(self, obj: Dandiset) -> str:
+        return f'/dandiset/{obj.identifier}/'
+
+    def get_domain(self, site):
+        return settings.DANDI_WEB_APP_DOMAIN
```
jjnesbitt: Since this is only used in one place, maybe we can avoid adding a new setting:
```suggestion
        return urlparse(settings.DANDI_WEB_APP_URL).netloc
```

dandiapi/urls.py
jjnesbitt: I don't see a need, not at the moment anyway.

jjnesbitt (changes requested):
Looks good overall. My only lingering concern is that we're generating sitemaps for a different domain than the one the API is hosted at (`api.dandiarchive.org` vs `dandiarchive.org`). Is there any issue with this?

dandiapi/sitemaps.py
waxlamp: Good idea; see 25352f5b2dc6f8d7bf36b14a4fd6d65e7a1c3290.

web/.env.development
```diff
@@ -1,3 +1,4 @@
+VITE_APP_DANDI_BACKEND_ROOT=http://localhost:8000
 VITE_APP_DANDI_API_ROOT=http://localhost:8000/api/
 VITE_APP_OAUTH_API_ROOT=http://localhost:8000/oauth/
```
waxlamp: I think these can be refactored now to inject the `DANDI` and `API` variables into the config at runtime, but it seemed like it was out of scope for this PR.

waxlamp:
@jjnesbitt, this is ready for another review. There are some choices in the PR now that may seem a bit odd, so I'm happy to discuss any time.

jjnesbitt (approved):

PR #2681, opened 2025-12-30
Don't return `null` names or usernames from user search endpoint

jjnesbitt:
There is a bug in the API, where if a user has a value of `null` in their social account `name` field, their user isn't properly displayed in the user search (used mainly in the GUI). This in theory could also exist for the `login` field, but since the response from github is guaranteed to have that field filled out with the user's github username, it should be impossible.

### DIFF
--- a/dandiapi/api/views/users.py
+++ b/dandiapi/api/views/users.py
@@ -67,12 +67,12 @@ def social_account_to_dict(social_account: SocialAccount):
 def serialize_user(user: User):
     """Serialize a user that's been annotated with a `social_account_data` field."""
     username = user.username
-    name = f'{user.first_name} {user.last_name}'.strip()
+    name = user.get_full_name()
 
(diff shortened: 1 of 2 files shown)


dandiapi/api/views/users.py
```diff
@@ -67,12 +67,12 @@ def social_account_to_dict(social_account: SocialAccount):
 def serialize_user(user: User):
     """Serialize a user that's been annotated with a `social_account_data` field."""
     username = user.username
-    name = f'{user.first_name} {user.last_name}'.strip()
+    name = user.get_full_name()
```
jjnesbitt: This [function](https://github.com/django/django/blob/c496a7df64ec0a80c2bab7f155aad567155fb37b/django/contrib/auth/models.py#L501-L506) on the `User` model is exactly the same as what we're doing here.

PR #2683, opened 2025-12-30
Ensure contact person and dandiset author are always present on new dandiset creation

### LINKED ISSUE #2682: New dandisets can be missing contact person
Supersedes #2376

If you supply a contributor field in the metadata when creating a dandiset, it will override our [metadata filtering process](https://github.com/dandi/dandi-archive/blob/ef6b10e08ffdc2606793473143dd306dd79ed08a/dandiapi/api/services/version/metadata.py#L19-L33), potentially resulting in no contributor with a `ContactPerson` role.

jjnesbitt:
Closes #2682 

If a user explicitly doesn't want themselves or a contact person on the dandiset, for whatever reason, they can explicitly change that after dandiset creation, by modifying the metadata.

### DIFF
--- a/dandiapi/api/services/version/metadata.py
+++ b/dandiapi/api/services/version/metadata.py
@@ -38,20 +38,7 @@ def _normalize_contributor(version_metadata: dict, name: str, email: str) -> dic
 
     contributor = version_metadata['contributor']
 
-    # Check that no contact person exists. If they do, we don't need to do anything
-    contact_person = next(
-        (
-            cont
-            for cont in contributor
-            if isinstance(cont.get('roleName'), list)
-            and CONTACT_PERSON_ROLE in cont.get('roleName', [])
-        ),
-        None,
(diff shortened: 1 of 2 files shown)


dandiapi/api/services/version/metadata.py
```diff
@@ -1,9 +1,63 @@
...
+
+    # Ensure that the contributor field is of the correct shape
+    contributor = version_metadata.get('contributor')
+    if not isinstance(contributor, list):
+        version_metadata['contributor'] = [default_contributor]
+        return version_metadata
+
+    # We know contributor field is a list, but the values could be anything.
+    # Filter the value to be only dicts.
+    version_metadata['contributor'] = [cont for cont in contributor if isinstance(cont, dict)]
```
jjnesbitt: Don't overwrite data, only append to list or update the author contributor.

PR #2696, opened 2026-01-19
Fix dandiset ID parsing during upload initialization

williamjallen:
Closes https://github.com/dandi/dandi-archive/issues/2649

### DIFF
--- a/dandiapi/api/tests/test_upload.py
+++ b/dandiapi/api/tests/test_upload.py
@@ -102,6 +102,26 @@ def test_upload_initialize(api_client, embargoed):
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


dandiapi/api/tests/test_upload.py
```diff
@@ -101,6 +101,26 @@ def test_upload_initialize(api_client, embargoed):
     assert upload.blob.name == f'blobs/{upload_id[:3]}/{upload_id[3:6]}/{upload_id}'
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize('dandiset_id', ['DANDI: abc'])
```
danlamanna: Is there a reason for parameterizing even if it's one parameter?

Issue #2697, opened 2026-01-19
Asset metadata missing `embargoedUntil` field causing validation errors


h-mayorquin:
I am getting validation errors on [dandiset 01693](https://dandiarchive.org/dandiset/001693/draft)

To create this dandiset I used the machinery created by @bendichter in PR #2462 to create the dandiset with embargo just by date:

<img width="1146" height="617" alt="Image" src="https://github.com/user-attachments/assets/089624b4-b9be-49f0-baf2-bfb757ab9634" />

But I am still getting errors like this:

<img width="1082" height="476" alt="Image" src="https://github.com/user-attachments/assets/b883a3a1-ff33-45d1-95a2-9ebef6a85140" />

What I think is happening.

PR #2462 correctly sets the `embargoedUntil` date in the dandiset version metadata when creating embargoed dandisets. However, asset-level metadata does not include this field, causing dandischema validation to fail with: "An embargo end date is required for NIH awards to be in compliance with NIH resource sharing policy."

The issue is in [`dandiapi/api/models/asset.py`](https://github.com/dandi/dandi-archive/blob/33384746bc76ec928f85a773b816de4d559e317e/dandiapi/api/models/asset.py#L248-L255) in the `full_metadata` property. When building asset metadata, the `access` field only includes `schemaKey` and `status`, but omits `embargoedUntil`:

```python
'access': [
    {
        'schemaKey': 'AccessRequirements',
        'status': AccessType.EmbargoedAccess.value
        if self.is_embargoed
        else AccessType.OpenAccess.value,
    }
],
```

The validation occurs in [`dandiapi/api/services/metadata/__init__.py`](https://github.com/dandi/dandi-archive/blob/1aed085706008464d9fefdf9566c9d9ef929dfd4/dandiapi/api/services/metadata/__init__.py#L48-L86) in `validate_asset_metadata()`, which calls `asset.published_metadata()` and then `dandischema.validate()`. The dandischema `AccessRequirements` pydantic validator requires `embargoedUntil` when `status == EmbargoedAccess`.

An easy fix is to propagate `embargoedUntil` from the dandiset version metadata to the asset metadata. That said, I am not sure why we have `embargoedUntil` at the asset level at all since this is only a dandiset parameter. Duplicating it lends ourselves to bugs like this.

PR #2698, opened 2026-01-19
Fix propagation of embargoed date from dandisets to assets

h-mayorquin:
Fixes https://github.com/dandi/dandi-archive/issues/2697

### DIFF
--- a/dandiapi/api/tests/test_asset_access_metadata.py
+++ b/dandiapi/api/tests/test_asset_access_metadata.py
@@ -183,19 +183,3 @@ def test_access_metadata_embargoed_blob_no_embargoed_until(
 
     for asset in assets:
         assert 'embargoedUntil' not in asset.access_metadata()
-
-
-@pytest.mark.django_db
-def test_access_metadata_embargoed_zarr_not_in_draft_version(
-    embargoed_zarr_archive_factory, draft_asset_factory
-):
-    """Embargoed zarr asset not in any draft version has no embargoedUntil."""
-    zarr = embargoed_zarr_archive_factory()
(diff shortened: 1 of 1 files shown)


satra:
@jjnesbitt - thanks for the summary. i think it highlights some of the very elements that we need to have around licensing (why it's a list in the metadata) and the flows.  the fact that an open and embargoed blob is part of a dandiset should be indicated to dataset maintainers in some way (we don't). aggregating access information would be one way of starting this. 

we already have steps in place when a blob is opened, it stops being "embargoed" in any dandiset. so if the conceptual notion of embargo is at a dandiset level, the practical notion of any access has always been at the underlying asset+blob level. our asset model does not allow sharing assets across dandisets at this point, every addition is a new asset. so as you say the access really boils down to the blob. 

so we could say this dandiset is embargoed, but practically it may contain all open blobs. the only place to record this information is in the asset metadata. this requires propagation of information, but we have already have various tasks that propagate changes of various kinds. 

regarding the question of medium or long term, many of these use cases and designs were borne by my experience with releasing data and doing research. so at least from this user's perspective it would be nice to not change the model and figure out the proper solution, as the model i think is still accurate about flow in scientific spaces. in fact some of the linc datasets would benefit from having a mixed model. they are currently copying between embargoed and open datasets as a solution. 

if in the short term we need to patch something, it could be the validation to understand which failed components are ok for embargoed dandisets. so i would rather patch it at the control layer for now not the model layer. also i think you already identified what we need to do in terms of access propagation, and the reverse lookup from blob to dandiset versions is already something that people have requested (at least internally).

PR #2719, opened 2026-02-26
Improve use of asset validation errors

### LINKED ISSUE #1214: Dandiset info endpoint large response sizes

### LINKED ISSUE #1653: Remove use of Asset `VALIDATING` status if not used

### LINKED ISSUE #2540: problem rendering many asset validation errors
For this dataset: https://dandiarchive.org/dandiset/001169

We are unable to publish due to asset and metadata errors.

When I click the orange buttons to see the validation errors, it takes a long time for this modal window to show up. Sometimes, the button does not work and the page dies.

Here are the error pages:

Image

Image

A few problems here:
1. Why aren't those times valid? Is it because there should be a "T" between the day and the time? That's not clear from the error. It would be nice to understand how to make this a valid datetime.
2. I think this is heavy to load because it is trying to render thousands of asset validation errors. Can we paginate this to avoid that problem?
3. Why is access being reported as an asset-level error? Shouldn't that be a dandiset-level error?
4. There is no space between the name of the file and the error string. This is not a big deal but it would be nice if we could make this look a bit better.

jjnesbitt:
- Closes #2540 
- Closes #1653
- Closes #1214 

The primary purpose of this PR is to address an issue with asset validation errors for zarr assets. The existing behavior is to share the same status between assets that are in the process of validation, and zarr assets that cannot be validated yet (due to the underlying zarr not being finalized). 

This is observed mostly on dandiset `000108`, however other dandisets have experienced this as well. This change won't actually address the underlying problem, as the solution to that is simply to finalize the zarrs backing those assets (something I will address separately). However, it will fix the asset validation error that is shown, to more accurately describe the problem.

There are several secondary improvements:
* Remove the deprecated `VALIDATING` status on the Asset model
* Fix the broken asset validation UI
* Separate out asset validation errors into their own endpoint, removing them from the version info endpoint.
* Enabled by the above improvement, remove the truncation of asset validation errors (originally introduced in https://github.com/dandi/dandi-archive/pull/2548)


dandiapi/api/models/version.py
```diff
@@ -94,29 +97,43 @@ def publishable(self) -> bool:
...
+            .values('field', 'message', 'path')
+        )
+
+        # Finally, get all INVALID assets. Each of these should have one or more
+        #   validation errors stored in the database.
         # For performance reasons, we truncate the list of INVALID assets such
-        # that we only display errors for the 50 assets with the most errors.
-        invalid_assets: models.QuerySet[Asset] = (
+        #   that we only display errors for the 50 assets with the most errors.
+        invalid_assets: Iterable[dict] = (
```
mvandenburgh: [`QuerySet.values()` returns a `QuerySet` of `dicts`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.values). 
```suggestion
        invalid_assets: QuerySet[dict] = (
```

dandiapi/api/models/version.py
```diff
@@ -94,29 +97,43 @@ def publishable(self) -> bool:
...
-            .values('field', 'message', 'path')[:50]
+            .values('field', 'message', 'path')
         )
 
-        # Next, get all INVALID assets. Each of these should have one or more
-        # validation errors stored in the database.
+        # Next, get all zarr assets which have not been finalized. These also are not stored
+        #   explicitly as errors in the database, so we need to construct them manually.
+        # For performance reasons, we only return the first 50 assets.
+        incomplete_zarr_assets: Iterable[VersionAssetValidationError] = (
```
mvandenburgh: [`QuerySet.values()` returns a `QuerySet` of `dicts`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.values). 
```suggestion
        incomplete_zarr_assets: QuerySet[VersionAssetValidationError] = (
```

mvandenburgh (approved):

dandiapi/api/models/version.py
jjnesbitt: Good catch, I was trying to fix a variety of type errors, and got carried away.

.github/workflows/cli-integration.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - name: Run dandi-api tests in dandi-cli
         if: matrix.dandi-version == 'master'
         run: >
-          uvx --with "dandi[test] @ git+https://github.com/dandi/dandi-cli"
+          uvx --with "dandi[test] @ git+https://github.com/dandi/dandi-cli@use-new-asset-validation-errors"
```
yarikoptic, line 78: should not be merged with this AFAIK... I wish there was a comment about PR which to check if this branch was merged/released

```suggestion
          uvx --with "dandi[test] @ git+https://github.com/dandi/dandi-cli"
```

yarikoptic (changes requested):
remove testing against some dev branch of dandi-cli

jjnesbitt:
This PR is still blocked on CLI integration. I do not have time to address this in the short term, so I will put it back into draft.

yarikoptic:
I do not see 'restart' button somehow but should be ok if redone now (dandi-cli vise)

.github/workflows/cli-integration.yml
yarikoptic: Found it (draft ATM):

- https://github.com/dandi/dandi-cli/pull/1823

yarikoptic:
note: I tuned description to put "Closes #XXX" into itemized list so we could see the titles of those issues to be closed

> However, it will fix the asset validation error that is shown, to more accurately describe the problem.

you left me hanging there ;)  could you visualize how it looked before in UI and now?  

Screenshot for added new endpoint in swagger? 

I am yet to grasp why we need a new endpoint  (and thus changes to dandi-cli) as we already have per asset one:
image

and not some option for an existing endpoint to filter if so desired for UI -- was such an option considered? (we already have many good options for other endpoints, as eg. filtering in `/dandisets/` etc)

jjnesbitt:
> I am yet to grasp why we need a new endpoint (and thus changes to dandi-cli) as we already have per asset one: image

The per asset validation error endpoint is not helpful if you want to view all asset validation errors.

> and not some option for an existing endpoint to filter if so desired for UI -- was such an option considered? (we already have many good options for other endpoints, as eg. filtering in `/dandisets/` etc)

The dandisets endpoint you mentioned is perhaps the most clunky and inefficient endpoint in the entire API (and to which we've had numerous past bugs), precisely because it attempts to do too many things at once via query params and filtering. 

Right now, the asset validation errors are bundled into the version info endpoint, which is used in several ways, and is overall an important endpoint. Since there may be many asset validation errors for a given version, if all errors are returned, this can can cause timeouts or otherwise slow response times. To avoid this, we artificially limit the number of asset validation errors that are returned from the version info endpoint. However, this causes issues, namely an incomplete and misleading picture of what asset validation errors are present on a version.

Separating out the asset validation errors into their own endpoint removes this burden from the main version info endpoint, and allows for the full list of asset validation errors to be returned from the new asset validation errors endpoint, without affecting the return of other data.

PR #2726, opened 2026-03-08
Truncate long filenames in file browser

### LINKED ISSUE #2725: UI: rendering of long filenames on the asset navigation window
Image

bendichter:
## Summary
- Fixes #2725
- Moves filename text from the `#prepend` slot (which has no width constraints) into a `v-list-item-title` element, which has built-in CSS for `text-overflow: ellipsis` + `overflow: hidden` + `white-space: nowrap`
- Adds a `title` attribute so the full filename is visible on hover

## Before
Image

## Test plan
- [x] Navigate to a dandiset with long filenames (e.g. dandiset 000409, sub-CSH-ZAD-001)
- [x] Verify long filenames truncate with ellipsis instead of overflowing
- [x] Verify full filename appears on hover tooltip
- [x] Verify folder names and ".." navigation still work correctly

### DIFF
--- a/web/src/views/FileBrowserView/FileBrowser.vue
+++ b/web/src/views/FileBrowserView/FileBrowser.vue
@@ -110,8 +110,8 @@
                   >
                     mdi-folder
                   </v-icon>
-                  ..
                 </template>
+                <v-list-item-title>..</v-list-item-title>
               </v-list-item>
 
               <v-list-item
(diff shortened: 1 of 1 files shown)


yarikoptic:
@bendichter ping... we should find some examples at least if not to add a test in e2e for this

bendichter:
Hmm yes, that's helpful. Is there a reason we don't render a front-end preview against the prod backend?

bendichter:
https://deploy-preview-2726--sandbox-dandiarchive-org.netlify.app/dandiset/213842/draft/files?location=sub-LongSubjectIdentifier001&page=1

image

yarikoptic:
> Hmm yes, that's helpful. Is there a reason we don't render a front-end preview against the prod backend?

because it is tandem and we can't really use master frontend with released (what is in prod) backend. And using devel (master) backend on prod S3 would be reckless in so many ways ;-)

yarikoptic (approved):

PR #2731, opened 2026-03-13
Add `embargo_end_date` field to Dandiset model

### LINKED ISSUE #1286: upon unembargo set embargoedUntil
Ref https://github.com/dandi/dandi-schema/pull/143 and motivation I have expressed yesterday in our dandi-standup - to be able to discover which dandisets became public from first being embargoed.  Going through dandisets which have embargoedUntil set would provide an indirect answer.   Also IMHO it would be more factually correct than e.g. some hypothetical date user or us set in metadata while provisionally planning the study: seeing some future date for unembargo in a public dataset would be pretty much incorrect metadata IMHO.

### LINKED ISSUE #2116: Embargo end date should not be a user populated field

### LINKED ISSUE #2730: Store `embargo_end_date` on dandisets directly
Related to https://github.com/dandi/dandi-archive/issues/1286 and https://github.com/dandi/dandi-archive/pull/2729

That issue and PR made me realize that if we're going to start relying on the embargo end date of a dandiset, we should probably store it in a field that the user can't modify. At the moment, since the embargo end date is _only_ stored within the metadata, it could be overwritten by someone simply using the API. 

I think we should instead create a new `embargo_end_date` field on the Dandiset directly, which would not be modifiable by the user directly. Instead, the value would be set initially on embargoed dandiset creation, and updated upon dandiset unembargo.

This way, the `embargoedUntil` value within the `access` metadata field can be computed from the source of truth (the field on the dandiset). It would be trivial to migrate the existing values from `embargoedUntil` to this new field.

Thoughts @satra @yarikoptic @waxlamp? If nobody objects to this, I could make a PR that supersedes https://github.com/dandi/dandi-archive/pull/2729

jjnesbitt:
Closes #2730 
Closes #1286
Closes #2116 
Supersedes #2729

This PR creates a field on the Dandiset model directly, to store the embargo end date for embargoed dandisets. This massively improves the ergonomics around embargoed dandisets, as now, we can guarantee that an embargoed dandiset has an embargo end date, at the database level. It also removes the ability for a user to change the embargo end date, which was easily accomplished via a PUT request. Now, the `embargoedUntil` metadata field is populated from the dandiset `embargo_end_date` field.

This also has additional ramifications for asset metadata. Now we can guarantee that if an embargoed asset is associated with any versions, it will have an embargoed end date.

### DIFF
--- a/dandiapi/api/migrations/0031_dandiset_embargo_end_date.py
+++ b/dandiapi/api/migrations/0031_dandiset_embargo_end_date.py
@@ -11,6 +11,15 @@ def populate_embargo_end_date(apps, schema_editor):
     Dandiset = apps.get_model('api', 'Dandiset')
     Version = apps.get_model('api', 'Version')
 
+    # Ensure that all embargoed dandisets have a value in the embargoedUntil field
+    problem_dandisets = Dandiset.objects.filter(
+        embargo_status='EMBARGOED', versions__metadata__access__0__embargoedUntil__isnull=True
+    )
+    if problem_dandisets.exists():
+        raise ValueError(
+            f"Found {problem_dandisets.count()} dandisets without 'embargoedUntil' value!"
+        )
+
     embargoed_dandisets = Dandiset.objects.filter(
         embargo_status='EMBARGOED', versions__metadata__access__0__embargoedUntil__isnull=False
     )
--- a/dandiapi/api/migrations/0031_dandiset_embargo_end_date.py
+++ b/dandiapi/api/migrations/0031_dandiset_embargo_end_date.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+
+from datetime import date
+
+from django.conf import settings
+from django.db import migrations, models
+from tqdm import tqdm
+
+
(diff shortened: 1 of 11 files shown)


waxlamp (approved):
Looks good!

waxlamp (approved):

PR #2744, opened 2026-03-19
Return 404 on orphaned assets

### LINKED ISSUE #2733: Disallow direct access of orphaned assets
Originally, we used to require that you access an asset through the dandiset and version it belonged to. For a while now, we've had an endpoint for accessing assets directly, via their asset ID. However, this leads to a situation where you can easily access an orphaned asset, which can be misleading, as you might assume it's an "active" asset. We also intend to garbage collect orphaned assets, which would prevent access to that asset anyways.

I think we should either update the direct asset endpoint to only return assets that have an associated version (and thus would 404 on orphaned asset access), or return some other 4xx level response.

@yarikoptic @satra Do you know of anyone that make significant use of orphaned assets?

jjnesbitt:
Depends on #2731
Closes #2733

Orphaned assets can still be accessed via the `/assets/{asset_id}/` endpoint. This PR filters out orphaned assets from that endpoint, resulting in a `404 Not Found` when attempting to access an orphaned asset.


yarikoptic:
merged master to redo CI and see where we stand ATM on this

jjnesbitt:
I believe the blocker here is still CLI compatibility

PR #2755, opened 2026-03-26
Display data sizes on the front-end using the decimal system

### LINKED ISSUE #2753: Display data sizes on the front-end using the decimal system
Over the past year the front-end switched to displaying data sizes using the binary system.  And the DANDI Client still reports sizes in the decimal system.

I would suggest that we switch back to using the decimal system on the front-end which is more commonly used in day to day life.  Thanks.

kabilar:
Fix #2753

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -16,7 +16,7 @@
         "axios": "^1.8.1",
         "core-js": "^3.41.0",
         "dompurify": "^3.2.3",
-        "filesize": "^10.1.6",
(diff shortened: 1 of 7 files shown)


yarikoptic (approved):
it seems that there was agreement in the issue cited on changing 

@kabilar having a few screenshots of before/after could have helped to assess the change. e.g.

on sandbox:
image

on preview:
image

and 

Image

vs

Image


if no objections voiced by @dandi/archive-maintainers I will merge in few days

PR #2765, opened 2026-04-02
Replace hardcoded DANDI identifiers with dynamic instance config

### LINKED ISSUE #2762: Bug: Identifier under How to Cite does not appear to use "vendorized" value
The Identifier uses just "DANDI" rather than "EMBER-DANDI"

Note: the identifier that gets copied here, is correct
Image

But this is not
Image

yarikoptic:
## Summary

- Replace all hardcoded `DANDI:` identifier prefixes, `DANDI Archive` publisher names, and `SCR_017571` RRID in frontend code with dynamic values from the backend `/api/info/` endpoint (`DJANGO_DANDI_INSTANCE_NAME`, `DJANGO_DANDI_INSTANCE_IDENTIFIER`)
- Add a shared Pinia store (`stores/instance.ts`) that caches instance config across components
- Add a grep-based lint check (`scripts/check-no-hardcoded-dandi.sh`) to prevent re-introducing hardcoded identifiers, wired into pre-commit and `tox -e lint`
- Update E2E tests to fetch instance config from the API instead of asserting hardcoded values
- Closes #2762
- Fixes some AI slop left for listing of identifiers and composition of them for bibtex key and notes

<details>
<summary>Files changed overview</summary> 

| File | Change |
|------|--------|
| `web/src/stores/instance.ts` | **New** — shared Pinia store for instance config |
| `web/src/utils/cff.ts` | All citation format functions accept `instanceName` parameter |
| `web/src/components/DLP/HowToCiteTab.vue` | Dynamic identifier, archive name, RRID |
| `web/src/views/DandisetLandingView/DownloadDialog.vue` | Dynamic identifier prefix and instance URL check |
| `web/src/components/DandisetList.vue` | Use shared instance store |
| `web/src/directives/index.ts` | Dynamic page title |
| `scripts/check-no-hardcoded-dandi.sh` | **New** — lint check for hardcoded identifiers |
| `.pre-commit-config.yaml` | Add `no-hardcoded-dandi-identifiers` hook |
| `tox.ini` | Run the check in `lint` environment |
| `e2e/utils.ts` | Add `fetchInstanceConfig()` helper |
| `e2e/tests/dandisetLandingPage.spec.ts` | Use API instance config in assertions |
| `e2e/tests/dandisetsPage.spec.ts` | Use API instance config in assertions |

</details>

## Tests etc plan

- [x] `npm run lint` passes (web/)
- [x] `npm run type-check` passes (web/)
- [x] `uv run tox -e lint` passes (includes hardcoded identifier check)
- [ ] Seek feedback from @dandi/archive-maintainers on the overall approach
- [x] Pre-commit hooks pass
- [x] Frontend CI passes
- [x] E2E tests pass with `DJANGO_DANDI_INSTANCE_NAME=DEV-DANDI`
- [x] Backend CI passes (no backend changes)
- [x] Move from Draft after CI turns green

<details>
<summary>I also fixed/adjusted for some left-over AI slop, as duplicate identifiers and inconsistent presentation</summary> 

e.g. compare bibtex cite as on https://dandiarchive.org/dandiset/000690?pos=5  having 
image

and

image

to https://deploy-preview-2765--sandbox-dandiarchive-org.netlify.app/dandiset/217838?pos=2

image

and two separate copy pasteable sections for identifiers

image

there is also a fix to bibtex record key. Can also look current sandbox at e.g. https://sandbox.dandiarchive.org/dandiset/217838?pos=2
</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,11 @@ repos:
   rev: v2.3.0
   hooks:
   - id: codespell
+- repo: local
+  hooks:
+  - id: no-hardcoded-dandi-identifiers
+    name: Check for hardcoded DANDI identifiers in frontend
+    entry: scripts/check-no-hardcoded-dandi.sh
+    language: script
+    files: '\.(vue|ts)$'
+    exclude: '(node_modules|__tests__|\.test\.|\.spec\.)'
--- a/e2e/tests/dandisetLandingPage.spec.ts
+++ b/e2e/tests/dandisetLandingPage.spec.ts
@@ -1,5 +1,5 @@
 import { expect, test } from "@playwright/test";
-import { clientUrl, registerNewUser, LOGOUT_BUTTON_TEXT, registerDandiset } from "../utils.ts";
+import { clientUrl, registerNewUser, LOGOUT_BUTTON_TEXT, registerDandiset, fetchInstanceConfig } from "../utils.ts";
 import { faker } from "@faker-js/faker";
 
 test.describe("dandiset landing page", async () => {
@@ -48,6 +48,14 @@ test.describe("dandiset landing page", async () => {
 
   test.describe("how to cite tab", () => {
     let dandisetId: string | undefined;
+    let instanceName: string;
+    let instanceIdentifier: string | null;
+
+    test.beforeAll(async () => {
+      const config = await fetchInstanceConfig();
+      instanceName = config.instance_name;
+      instanceIdentifier = config.instance_identifier;
+    });
 
     test.beforeEach(async ({ page }) => {
       await registerNewUser(page);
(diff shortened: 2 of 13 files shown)


bendichter (commented):
## Review

Overall approach is sound — centralizing instance config in a Pinia store is the right call, and the lint script + pre-commit hook is a nice guardrail. The BibTeX/RIS double-prefix fix is a real bug fix too. A few issues to address:

### Issues

**1. Race condition in instance store** (`web/src/stores/instance.ts`)

Multiple components (`HowToCiteTab`, `DandisetList`, `DownloadDialog`) all call `fetchInstanceInfo()` on mount. Since the guard is just `if (this.loaded) return`, concurrent callers before the first fetch completes will fire multiple `/api/info/` requests. Should track an in-flight promise so concurrent callers share the same request.

**2. Remaining hardcoded URL in `HowToCiteTab.vue:~369`**

The `dandiUrl` computed property still has a hardcoded fallback:
```ts
return `https://dandiarchive.org/dandiset/${identifier}`;
```
This is used in "Materials and Methods" and "Data Availability Statement" sections. Should use `instanceStore.instanceUrl` instead, otherwise non-primary instances link to the wrong archive.

**3. Double API call in `DownloadDialog.vue`**

`onMounted` calls `dandiRest.info()` directly (for CLI version info) and then `instanceStore.fetchInstanceInfo()` which internally also calls `dandiRest.info()`. Should consolidate to avoid the redundant request.

**4. No app-level initialization of instance store**

`directives/index.ts` reads from the store but never triggers a fetch. On routes that don't mount `HowToCiteTab`, `DandisetList`, or `DownloadDialog`, the page title stays at the fallback `"DANDI Archive"` forever. Consider fetching once at app startup (e.g., in `App.vue` or a router guard).

**5. `instanceName` defaults to empty string**

Before the fetch completes, the identifier briefly renders as `:<id>/draft` (empty prefix). A default like `'DANDI'` would prevent the flash of malformed text.

### Minor

- snake_case local variables (`dandiset_version_identifier`, `instance_name`) in `HowToCiteTab.vue` — should be camelCase per project convention.
- Lint script uses `\s` in `grep -E` which is not POSIX-portable (BSD grep on macOS). Consider `[[:space:]]`.
- `fetchInstanceConfig()` in `e2e/utils.ts` doesn't check `resp.ok` before parsing JSON.

PR #2771, opened 2026-04-04
fix: use datePublished for citation year on published versions

### LINKED ISSUE #2769: Citations are missing year
The How to Cite citations are missing the year/date field. Previous discussions and screenshots of previous citations show the year field but it's not there anymore. https://github.com/dandi/dandi-archive/issues/1555 and https://github.com/dandi/dandi-archive/pull/2671

https://dandiarchive.org/dandiset/000409?search=000409&pos=1
Image

https://dandiarchive.org/dandiset/000003?search=000003&pos=1
Image

bendichter:
## Summary
- Published dandiset versions have `datePublished` set but `dateModified` is null, causing the "How to Cite" citations to be missing the year field in all formats (APA, MLA, Chicago, etc.)
- Prefer `datePublished` over `dateModified` when generating the CFF `date-released` field

Fixes #2769

## Test plan
- [x] Navigate to a published dandiset (e.g. `/dandiset/000409`) and open the "How to Cite" tab
- [x] Verify the year appears in all citation formats (APA, MLA, Chicago, Harvard, Vancouver, IEEE, BibTeX, RIS)
- [x] Verify draft dandisets still show dates correctly (they use `dateModified`)

### DIFF
--- a/web/src/utils/cff.ts
+++ b/web/src/utils/cff.ts
@@ -160,9 +160,10 @@ export function dandisetToCFF(metadata: DandisetMetadata, doi?: string): CFF {
     cff.version = metadata.version;
   }
 
-  // Add release date
-  if (metadata.dateModified) {
-    cff['date-released'] = metadata.dateModified.split('T')[0]; // Extract date part
+  // Add release date (prefer datePublished for published versions, fall back to dateModified)
+  const releaseDate = (metadata.datePublished || metadata.dateModified) as string | undefined;
+  if (releaseDate) {
+    cff['date-released'] = releaseDate.split('T')[0]; // Extract date part
   }
 
   // Add related resources as references
--- a/web/src/utils/cff.ts
+++ b/web/src/utils/cff.ts
@@ -160,8 +160,8 @@ export function dandisetToCFF(metadata: DandisetMetadata, doi?: string, versionM
     cff.version = metadata.version;
   }
 
-  // Add release date (datePublished for published versions, version modified date for drafts)
-  const releaseDate = (metadata.datePublished || versionModified) as string | undefined;
+  // Add release date (version modified date is most current, fall back to datePublished)
+  const releaseDate = (versionModified || metadata.datePublished) as string | undefined;
   if (releaseDate) {
     cff['date-released'] = releaseDate.split('T')[0]; // Extract date part
   }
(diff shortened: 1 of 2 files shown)


web/src/utils/cff.ts
```diff
@@ -85,7 +85,7 @@ function organizationToCFFAuthor(org: Organization): CFFAuthor {
 /**
  * Convert DANDI dandiset metadata to CFF format
  */
-export function dandisetToCFF(metadata: DandisetMetadata, doi?: string): CFF {
+export function dandisetToCFF(metadata: DandisetMetadata, doi?: string, versionModified?: string): CFF {
```
yarikoptic: multiline diff didn't work out for me here but I would suggest

```suggestion
export function dandisetToCFF(metadata: DandisetMetadata, doi?: string, dateModified?: string): CFF {
```
since otherwise it suggests that it is a version not a date.

but I failed to trace this typescript to why we need this extra option here and why it is not in `metadata`?  may be it could be even more descriptive name  like

```suggestion
export function dandisetToCFF(metadata: DandisetMetadata, doi?: string, dateDraftModified?: string): CFF {
```

otherwise unclear what is this version and how it is different from `metadata`

web/src/utils/cff.ts
```diff
@@ -160,9 +160,10 @@ export function dandisetToCFF(metadata: DandisetMetadata, doi?: string): CFF {
     cff.version = metadata.version;
   }
 
-  // Add release date
-  if (metadata.dateModified) {
-    cff['date-released'] = metadata.dateModified.split('T')[0]; // Extract date part
+  // Add release date (version modified date is most current, fall back to datePublished)
+  const releaseDate = (versionModified || metadata.datePublished) as string | undefined;
```
yarikoptic: again since I do not yet grasp what versionModified is, shouldn't datePublished take precedence ?  but also "explicit better than implicit" if `modified` for draft and `published` for published -- check explicitly on either published and assign using correct field. or we endup with citation for some elderly dandiset with buggy metadata without published date, assigned date of current draft modified?

bendichter:
Before:
https://sandbox.dandiarchive.org/dandiset/214527
image



https://deploy-preview-2771--sandbox-dandiarchive-org.netlify.app/dandiset/214527
image

bendichter:
@yarikoptic could you please take another look at this? I'd like to get this fix in.

@satra, I'm still missing some info on how you envisioned bringing these different citation mechanisms together.

yarikoptic (approved):
my initial concerns seems to be addressed, and if we find that smth is still not fully kosher -- we could fix, not end of the world in particular for this niche functionality. Thank you @bendichter !

PR #2778, opened 2026-04-09
Fix instance store race condition, hardcoded URL, and minor issues

bendichter:
## Summary

Follow-up fixes for #2765:

- **Race condition in instance store**: Concurrent `fetchInstanceInfo()` calls from multiple components now share a single in-flight promise instead of firing duplicate `/api/info/` requests
- **Hardcoded URL in HowToCiteTab**: `dandiUrl` computed property was still falling back to `https://dandiarchive.org` — now uses `instanceStore.instanceUrl`
- **App-level store initialization**: Instance store is now fetched at startup in `main.ts` so page titles work on all routes (not just those mounting components that call `fetchInstanceInfo()`)
- **Double API call in DownloadDialog**: Reordered to use cached instance store fetch before the separate `dandiRest.info()` call for CLI version info
- **snake_case → camelCase**: Fixed local variable naming in HowToCiteTab to match project conventions
- **POSIX-portable grep**: Replaced `\s` with `[[:space:]]` in lint script for BSD grep compatibility
- **E2E error handling**: Added `resp.ok` check in `fetchInstanceConfig()`

## Test plan

- [x] `npm run lint` passes
- [x] `npm run type-check` passes
- [x] Hardcoded identifier lint check passes
- [ ] CI passes

### DIFF
--- a/web/src/components/DLP/HowToCiteTab.vue
+++ b/web/src/components/DLP/HowToCiteTab.vue
@@ -346,18 +346,15 @@ const formattedCitations = computed<Record<CitationFormat, string>>(() => {
       cff: '',
     };
   }
-  const dandiset_version_identifier = dandisetVersionIdentifier.value;
-  const instance_name = instanceName.value;
-
   return {
-    apa: cffToAPA(cffObject.value, instance_name),
-    mla: cffToMLA(cffObject.value, instance_name),
-    chicago: cffToChicago(cffObject.value, instance_name),
-    harvard: cffToHarvard(cffObject.value, instance_name),
-    vancouver: cffToVancouver(cffObject.value, instance_name),
-    ieee: cffToIEEE(cffObject.value, instance_name),
-    bibtex: cffToBibTeX(cffObject.value, dandiset_version_identifier, instance_name),
-    ris: cffToRIS(cffObject.value, dandiset_version_identifier, instance_name),
+    apa: cffToAPA(cffObject.value, instanceName.value),
+    mla: cffToMLA(cffObject.value, instanceName.value),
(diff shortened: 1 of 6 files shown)


web/src/components/DLP/HowToCiteTab.vue
```diff
@@ -346,18 +346,15 @@ const formattedCitations = computed<Record<CitationFormat, string>>(() => {
...
-    bibtex: cffToBibTeX(cffObject.value, dandiset_version_identifier, instance_name),
-    ris: cffToRIS(cffObject.value, dandiset_version_identifier, instance_name),
+    apa: cffToAPA(cffObject.value, instanceName.value),
+    mla: cffToMLA(cffObject.value, instanceName.value),
+    chicago: cffToChicago(cffObject.value, instanceName.value),
+    harvard: cffToHarvard(cffObject.value, instanceName.value),
+    vancouver: cffToVancouver(cffObject.value, instanceName.value),
+    ieee: cffToIEEE(cffObject.value, instanceName.value),
+    bibtex: cffToBibTeX(cffObject.value, dandisetVersionIdentifier.value, instanceName.value),
+    ris: cffToRIS(cffObject.value, dandisetVersionIdentifier.value, instanceName.value),
```
yarikoptic, line 357: imho it was better before - shorter variables made it easier to read... rationale is not listed among items -- do you see what it was ?

yarikoptic:
I will merge, although that dance with variables isn't clear to me

web/src/components/DLP/HowToCiteTab.vue
bendichter: snake_case → camelCase

PR #2784, opened 2026-04-22
Add support for zarr multi-part uploads

### LINKED ISSUE #2758: Support multi-part uploads for Zarrs
Prompted by 

- https://github.com/dandi/dandi-cli/issues/1821

ATM we use single part "put_object" for zarrs:

[`92 `](https://github.com/dandi/dandi-archive/blob/6ad43337cac40ffe2e16985ea490a56927ed2fbf/dandiapi/zarr/views/__init__.py#L93) `class ZarrViewSet(ReadOnlyModelViewSet):`
[`283 `](https://github.com/dandi/dandi-archive/blob/6ad43337cac40ffe2e16985ea490a56927ed2fbf/dandiapi/zarr/views/__init__.py#L284) `    def create_files(self, request, zarr_id):`
[`286 `](https://github.com/dandi/dandi-archive/blob/6ad43337cac40ffe2e16985ea490a56927ed2fbf/dandiapi/zarr/views/__init__.py#L287) `        with transaction.atomic():`
[`301 `](https://github.com/dandi/dandi-archive/blob/6ad43337cac40ffe2e16985ea490a56927ed2fbf/dandiapi/zarr/views/__init__.py#L302) `            urls = [`
**[`302:`](https://github.com/dandi/dandi-archive/blob/6ad43337cac40ffe2e16985ea490a56927ed2fbf/dandiapi/zarr/views/__init__.py#L303) `                zarr_archive.storage.generate_presigned_put_object_url(`**

The immediate use case of LINC project (attn @kabilar @satra) apparently has multi-GB files within zarrs!

Max single-part upload size is 5GB and there dandi-cli had succeeded with 2.6GB upload and potentially others were even larger (apparently we do not depict error properly there if that was the case).

In any case -- with v3 sharding it is logical to expect large files within zarr "filesets". Hence we need to implement support for multi-part uploads, and likely as consistent with regular assets uploads as possible/feasible.

Components affected and requiring analysis/changes:

- [ ] dandi-archive -- 
- [ ] dandi-cli -- 
- [ ] backups2datalad -- likely we use for minting keys and/or checks 
- [ ] dandidav -- "not sure" -- it might in particular in case of manifests and checksums/ETags there
- [ ] https://github.com/dandi/dandi-archive/pull/2702 -- same as per dandidav
- [ ] ...?

Also worth checking aiming forward:

- [ ] what are the file sizes within nwb-zarr's AIND works with (@alejoe91 ?)

jjnesbitt:
Closes #2758

This is a backwards compatible change that adds new endpoints for performing zarr multi-part upload. It adds the following endpoints:

1. `/uploads/zarr/initialize`
2. `/uploads/zarr/{zarr_id}/complete`
3. `/uploads/zarr/{zarr_id}/validate`

These endpoints operate largely the same as the standard upload endpoints, with the following differences:

1. The zarr upload initialize endpoint accepts a request of the following structure:

```json

{
  "contentSize": "12312313123",
  "digest": {
    "algorithm": "dandi:dandi-etag",
    "value": "16308760353355915d657a6f32d6d6f1-1"
  },
  "zarr_id": "5d784322-a40e-4641-89a1-32f36610e2c8",
  "chunk_key": "/0/0/1",
}
```

2. The zarr upload validate endpoint doesn't return any data in the 200 response, since there is nothing minted, like in the asset blob upload workflow.

Here is what the upload section of the swagger page now looks like:

image


yarikoptic:
As there is no design document for this feature formalized, would be nice to at least get a PR description to describe the 
- new API endpoint(s)
- principle of operation
- effect (or absence) on existing zarrs in the archive

yarikoptic:
> 1. The zarr upload initialize endpoint accepts a request of the following structure:
> 
> ```json
> {
>   "contentSize": "int",
>   "digest": "md5",
> ```

here -- we are talking about any file or only the one over 5GB?  

for multipart, shouldn't it be not "md5" but "dandi-etag" so that there is verification of the upload at the end according to ETag?

jjnesbitt:
> here -- we are talking about any file or only the one over 5GB?

These endpoints are just the new endpoints, so only for over 5GB. The single part upload process for files under 5GB has not changed at all.

> for multipart, shouldn't it be not "md5" but "dandi-etag" so that there is verification of the upload at the end according to ETag?

You're correct, I made a mistake in creating that example. I've updated it.

yarikoptic:
what happens if size less than 5G -- we still proceed to multipart upload or error response?

IMHO should be error so operation is consistent and "malfunctioning" client would not violate overall consistency of zarrs by e.g. uploading large (but smaller than 5G) files as multiparts which would then not align with checksum compute etc.

dandiapi/api/views/upload.py
```diff
@@ -268,3 +282,98 @@ def upload_validate_view(request: AuthenticatedRequest, upload_id: str) -> HttpR
...
+    method='POST',
+    request_body=ZarrUploadInitializationRequestSerializer,
+    responses={200: UploadInitializationResponseSerializer},
+)
+@api_view(['POST'])
+@parser_classes([JSONParser])
+@permission_classes([IsApproved])
+def zarr_upload_initialize_view(request: AuthenticatedRequest) -> HttpResponseBase:
+    """
+    Initialize a multipart zarr upload.
```
yarikoptic:
```suggestion
    Initialize a multipart upload of a large file in a zarr
```

since "multipart zarr" is not per se making sense since zarr is a "folder", and there is no "folder upload"s (multi- or single- part)

dandiapi/api/views/upload.py
```diff
@@ -268,3 +282,98 @@ def upload_validate_view(request: AuthenticatedRequest, upload_id: str) -> HttpR
...
+    method='POST',
+    request_body=UploadCompletionRequestSerializer,
+    responses={200: UploadCompletionResponseSerializer},
+)
+@api_view(['POST'])
+@parser_classes([JSONParser])
+@permission_classes([IsApproved])
+def zarr_upload_complete_view(request: AuthenticatedRequest, upload_id: str) -> HttpResponseBase:
+    """
+    Complete a multipart zarr upload.
```
yarikoptic:
```suggestion
    Complete a multipart upload of a file in a zarr.
```

dandiapi/api/models/upload.py
```diff
@@ -40,16 +48,42 @@ class Upload(models.Model):  # noqa: DJ008
...
+        ]
+
+    @property
+    def embargoed(self):
+        dandiset = self.dandiset or self.zarr.dandiset
+        return dandiset.embargoed
+
+    @classmethod
+    def initialize_zarr_multipart_upload(cls, etag, size, zarr: ZarrArchive, chunk_key: str):
+        object_key = zarr.s3_path(chunk_key.lstrip('/'))
```
yarikoptic, line 71: I think here (or outside) we need to verify that `size` is >= 5GB since , as I made a comment in the main thread, we are inviting inconsistency by bad client doing multipart uploads of "small" files in zarrs

yarikoptic (dismissed):
Needs explicit error and clarification in API docstring

dandiapi/api/views/upload.py
jjnesbitt: The term for this is a zarr chunk, I'll update it.

yarikoptic: nope.  "chunk" is part of data array. in v2 it is indeed a separate file. in zarr v3 multiple chunks could be placed into a single shard (a file)

Also there are metadata files which are not chunks at all (although unlikely to indeed be over 5GB ;) ) ... 

I do not know if there is any universal term really for those "files" rather than "a file", and here semantic is not really about what kind of file it is but rather its size, even if unlikely to be metadata file -- could be a chunk or could be a shard or who knows what it would be in some zarr v4 (may be hdf5 pack ;-) )

jjnesbitt:
Successful integration test run: https://github.com/dandi/dandi-archive/actions/runs/25181766500/job/73829881539?pr=2784

PR #2785, opened 2026-04-24
Move `vue-router` to runtime dependencies

brianhelba:


jjnesbitt (approved):
I'm curious how the site was functioning at all without this. Maybe it was installed as a child dependency.

brianhelba:
@jjnesbitt Unless we're publishing this as a library package, it'll still work in either group, since `npm install` will install both groups into a flat `node_modules` structure. Still, it's helpful to separate runtime from build-time deps.

I cared about this so that Renovate would place it into the correct update group.

PR #2792, opened 2026-04-26
Fix incompatibility with RabbitMQ 4.3

brianhelba:
RabbitMQ 4.3 disallows a queue configuration (non-exclusive, non-durable) that Celery is still using by default. Celery creates such queues for:
* remote control (`pidbox`)
* events (`celeryev`)

To disable the (remote) control queue:
* `--without-mingle`
* `worker_enable_remote_control = False`

To disable the events queue:
* `--without-heartbeat`
* `--without-gossip`
* `worker_send_task_events = False` (already the default)

Note, gossip, mingle, and heartbeat features provide no value. They generate extra traffic and are only used for task revocation (which DANDI doesn't use) and worker monitoring.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,7 +27,7 @@ This is the simplest configuration for developers to start with.
 ### Run Application
 1. Run the following commands in three separate VSCode built-in-terminals:
    1. `./manage.py runserver_plus 0.0.0.0:8000`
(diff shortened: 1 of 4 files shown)


brianhelba:
This bug is currently breaking the CLI integration tests on all branches.

jjnesbitt (approved):

jjnesbitt:
It seems the [issue](https://github.com/celery/kombu/issues/2237) was recently closed, so hopefully there will be a hotfix release soon.

brianhelba:
@jjnesbitt It's a bit more complicated than just that one issue, but yes, I expect it'll be fixed within a few months at most.

PR #2794, opened 2026-04-26
[codex] Fix citation author filtering

### LINKED ISSUE #2793: problem with citation: non-author can be included
https://dandiarchive.org/dandiset/000690/0.260330.2144, the citations reads:

> National Institutes of Health (NIH), Mehta, M. R., Purandare, C., Jha, S., Lecoq, J., Durand, S., Gillis, R., Belski, H., Bawany, A., Carlson, M., Peene, C., Wilkes, J., Johnson, T., Naidoo, R., Suarez, L., Han, W., Amaya, A., Nguyen, K., Ouellette, B., Swapp, J., & Williford, A.. Allen Institute Openscope - Vision2Hippocampus project (Version 0.260330.2144) [Data set]. DANDI Archive. https://doi.org/10.48324/dandi.000690/0.260330.2144

"National Institutes of Health (NIH)" should not have been included

bendichter:
## Summary

Fixes #2793 by making the frontend CFF/citation conversion include only contributors explicitly marked with `includeInCitation: true`.

The backend citation path and the landing-page overview already use that stricter behavior, but the frontend citation formatter treated missing `includeInCitation` values as included. That allowed non-author contributors such as funders to appear in formatted citations.

This also handles no-author citation formatting more gracefully across exported citation formats so empty author lists do not produce malformed punctuation or empty BibTeX author fields.

## Checks

- `npm run type-check`
- `npx eslint src/utils/cff.ts --max-warnings=0`
- One-off formatter check against the reported Dandiset metadata shape confirmed `National Institutes of Health (NIH)` is excluded

`npm run build` was attempted, but local Vite startup hit `EMFILE: too many open files, watch` before code compilation.

### DIFF
--- a/web/src/utils/cff.ts
+++ b/web/src/utils/cff.ts
@@ -264,6 +264,15 @@ export function cffToBibTeX(cff: CFF, identifier: string): string {
  * Convert CFF to APA format (7th edition)
  */
 export function cffToAPA(cff: CFF): string {
+  const year = cff['date-released'] ? ` (${cff['date-released'].split('-')[0]})` : '';
+  const version = cff.version ? ` (Version ${cff.version})` : '';
+  const doi = cff.doi ? ` https://doi.org/${cff.doi}` : '';
+
+  // APA 7: when there is no author, the title moves into the author position.
+  if (cff.authors.length === 0) {
+    return `${cff.title}${year}${version} [Data set]. DANDI Archive.${doi}`;
+  }
+
   const authors = cff.authors.map((author) => {
     if (author.name) {
       return author.name;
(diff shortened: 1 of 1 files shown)


bendichter:
Before:

Image

After:

image

https://deploy-preview-2794--sandbox-dandiarchive-org.netlify.app/dandiset/218300?pos=1

web/src/utils/cff.ts
```diff
@@ -276,7 +278,9 @@ export function cffToAPA(cff: CFF): string {
   });
 
   let authorString = '';
-  if (authors.length === 1) {
+  if (authors.length === 0) {
+    authorString = cff.title;
```
yarikoptic: what is this about?  in line 295 then there is some odd , if no authors make title empty to compensate for this now... overall feels a bit too aisloppy

bendichter: Fair point — that was confusing. Refactored in b29fd936: each formatter now early-returns on the no-author case with its own template, and `authorString` is strictly about authors (no more stuffing the title in and then blanking it out downstream). Also tightened punctuation in the no-author path so it matches each style's with-author conventions (period inside the closing quote for MLA/Chicago, comma inside the quote for IEEE).

yarikoptic: uff, my duplication allergies kicking in, but all that AI sloppy code I guess could tollerate that. ok.

yarikoptic (approved):

Issue #2759, opened 2026-03-31
Manifests on S3 are not up-to-date! DOIs missing, and overall could be outdated


yarikoptic:
### Originally -- was about missing DOIs:

While working on linkml-ification and trying to figure out on why 'schemaKey' of PublishedDandiset's is actually just a 'Dandiset' we mentioned that all recent Dandiset manifests lack "doi" field and there is no consistency besides that most of the recent ones do not have it!!!

```shell
dandi@drogon:/mnt/backup/dandi/dandiset-manifests$ for f in */0.*/dandiset.jsonld ; do echo -en "$f:\t"; jq -r  -r '"\(.schemaVersion) \(.doi)"' $f; done | sort -V -t/ -k 2
000029/0.210730.1538/dandiset.jsonld:	0.4.4 10.48324/dandi.000029/0.210730.1538
000029/0.210805.2047/dandiset.jsonld:	0.4.4 10.48324/dandi.000029/0.210805.2047
000029/0.210806.1511/dandiset.jsonld:	0.4.4 10.48324/dandi.000029/0.210806.1511
000029/0.210806.2112/dandiset.jsonld:	0.4.4 10.48324/dandi.000029/0.210806.2112
000003/0.210812.1448/dandiset.jsonld:	0.4.4 10.48324/dandi.000003/0.210812.1448
000067/0.210812.1457/dandiset.jsonld:	0.4.4 10.48324/dandi.000067/0.210812.1457
000059/0.210812.1514/dandiset.jsonld:	0.4.4 10.48324/dandi.000059/0.210812.1514
...
000167/0.220928.1306/dandiset.jsonld:	0.6.0 10.48324/dandi.000167/0.220928.1306
000064/0.221025.1735/dandiset.jsonld:	0.6.0 10.48324/dandi.000064/0.221025.1735
000029/0.221107.2344/dandiset.jsonld:	0.4.4 null
000147/0.221122.2256/dandiset.jsonld:	0.6.0 null
000363/0.221127.2345/dandiset.jsonld:	0.6.3 null
000398/0.221208.1803/dandiset.jsonld:	0.6.3 null
000350/0.221219.1506/dandiset.jsonld:	0.6.3 null
000363/0.230124.1456/dandiset.jsonld:	0.6.3 null
000363/0.230214.1833/dandiset.jsonld:	0.6.3 10.80507/dandi.123456/0.123456.1234
000233/0.230223.0815/dandiset.jsonld:	0.6.2 null
[...]

Issue #2342, opened 2025-04-21
Fake DOI was injected into published versions of various dandisets, needs a fix


yarikoptic:
While working on addressing
- https://github.com/dandi/dandi-archive/issues/1953

we ran into a realization that many dandisets have fake DOIs in their metadata records!  Looking at manifests (if those are "up to date!?")

<details>
<summary>I see the test prefix DOI in over 28 dandiset versions</summary> 

```
(dandiarchive-logs-parser) dandi@drogon:/mnt/backup/dandi/dandiset-manifests-s3cmd$ grep -l 10.80507 */*/dandiset.yaml | grep -v 'draft/dandiset' | uniq  | nl
     1  000029/0.210712.1903/dandiset.yaml
     2  000029/0.230317.1541/dandiset.yaml
     3  000037/0.240209.1623/dandiset.yaml
     4  000236/0.230316.2031/dandiset.yaml
     5  000237/0.230316.1655/dandiset.yaml
     6  000363/0.230214.1833/dandiset.yaml
     7  000363/0.230613.1608/dandiset.yaml
     8  000363/0.230808.2053/dandiset.yaml
     9  000363/0.230822.0128/dandiset.yaml
    10  000363/0.231012.2129/dandiset.yaml
    11  000462/0.230416.2132/dandiset.yaml
    12  000469/0.240122.1800/dandiset.yaml
    13  000469/0.240122.1812/dandiset.yaml
    14  000469/0.240123.1806/dandiset.yaml
    15  000473/0.230417.1502/dandiset.yaml
    16  000481/0.230417.2148/dandiset.yaml
    17  000541/0.240611.1947/dandiset.yaml
    18  000549/0.230519.2043/dandiset.yaml
    19  000568/0.230705.1633/dandiset.yaml
    20  000576/0.231010.1811/dandiset.yaml
    21  000623/0.240121.2058/dandiset.yaml
    22  000623/0.240130.2108/dandiset.yaml
    23  000629/0.240627.0646/dandiset.yaml
    24  000686/0.240515.1909/dandiset.yaml
    25  000727/0.240106.0043/dandiset.yaml
    26  000768/0.231221.2138/dandiset.yaml
    27  000769/0.231221.2138/dandiset.yaml
    28  000891/0.240215.0831/dandiset.yaml

```

</details>

[...]

PR #2799, opened 2026-04-28
feat: DataCite DOI lifecycle — concept DOI at creation, version DOI at publish

satra:
## Summary

- Register a **Draft concept DOI** on DataCite when a dandiset is created, stored in `Dandiset.concept_doi`
- Compute **real version DOI** deterministically at publish (eliminates fake DOI injection pattern `*.123456/0.123456.1234`)
- Create **Findable version DOI** on DataCite at publish, promote concept DOI to Findable with `HasVersion` link
- Track DOI lifecycle via `Version.doi_state` field (`pending` → `draft`/`findable`/`failed`/`registered`)
- All DataCite operations via async Celery tasks on `doi` queue with retry + exponential backoff
- `datacite_session()` context manager with timeouts (5s connect/30s read), HTTP-level retry (429/5xx), proper cleanup
- `IsVersionOf` relation on version DOIs pointing to concept DOI (when dandischema supports `concept_doi` param)
- Suppress DOI display for draft versions in frontend; filter placeholder DOIs
- **Hide** (not delete) Findable version DOIs on version deletion (transition to Registered per DataCite best practices)
- Update Draft concept DOI metadata on metadata changes (pre-publication only)
- Mint Draft concept DOI on unembargo
- Management commands: `remediate_dois` (fix historical fake/null DOIs with `--dry-run` and `--delay`), `check_doi_health` (find stuck/failed DOIs)
- `doi` queue added to Procfile and docker-compose for production + dev
- DOI string always set (deterministic, needed for validation); DataCite registration gated by `doi_configured()`

Builds on DOI service layer architecture from @jjnesbitt's `rf-enh-dandiset-dois` branch (dandi/dandi-archive#2350). Service functions ported onto current master with vendorization preserved.

**Depends on**: 
- dandi/dandi-schema#401 (adds `concept_doi` param to `to_datacite()`, `dates` field, DANDI identifier)

**Related issues**: 
- dandi/dandi-archive#1319, 
- dandi/dandi-archive#1709,
- dandi/dandi-archive#1710,
- dandi/dandi-archive#2342,
- dandi/dandi-archive#2759,
- dandi/dandi-schema#259,
- dandi/dandi-schema#292

**Prior attempt(s)**:
- dandi/dandi-archive#2350

**Design document**: `doc/design/doi-generation-2.md` (PR #2012) — all requirements implemented

<details>
<summary>DOI Lifecycle Sequence Diagrams</summary>

### 1. Dandiset Creation — Draft Concept DOI

```mermaid
sequenceDiagram
    participant User
    participant API as Django API
    participant DB as PostgreSQL
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API

    User->>API: POST /api/dandisets/ (create open)
    API->>DB: Create Dandiset + draft Version
    API->>API: concept_doi = format_doi(id)
    API->>DB: Set dandiset.concept_doi
    API->>DB: Set draft_version.doi = concept_doi
    alt DOI configured
        API->>DB: Set doi_state = 'pending'
        API->>DB: COMMIT
        DB-->>Celery: on_commit → create_dandiset_doi_task
        Celery->>DC: POST /dois (Draft, minimal metadata)
        alt Success
            DC-->>Celery: 201 Created
            Celery->>DB: UPDATE doi_state = 'draft'
        else Failure (retries exhausted)
            Celery->>DB: UPDATE doi_state = 'failed'
        end
    else DOI not configured
        API->>DB: doi_state remains NULL
        API->>DB: COMMIT
        Note over API: No DataCite call — DOI string set for schema validation only
    end
    API-->>User: 200 OK
```

### 2. Metadata Update — Sync Draft DOI (Pre-Publication)

```mermaid
sequenceDiagram
    participant User
    participant API as Django API
    participant DB as PostgreSQL
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API

    User->>API: PUT /api/dandisets/{id}/versions/draft/
    API->>DB: Update version metadata
    alt Open, unpublished, has concept_doi, DOI configured
        API->>DB: COMMIT
        DB-->>Celery: on_commit → update_dandiset_doi_task
        Celery->>DC: PUT /dois/{concept_doi} (update Draft metadata)
        DC-->>Celery: 200 OK
    else Embargoed, or published, or no concept_doi
        API->>DB: COMMIT
        Note over API: No DOI update (no-op)
    end
    API-->>User: 200 OK
```

### 3. Publish — Findable Version DOI + Concept DOI Promotion

```mermaid
sequenceDiagram
    participant User
    participant API as Django API
    participant DB as PostgreSQL
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API

    User->>API: POST .../versions/draft/publish/
    API->>DB: Lock dandiset, set PUBLISHING
    API->>DB: COMMIT
    DB-->>Celery: on_commit → publish_dandiset_task

    Celery->>DB: Build publishable version from draft
    Celery->>Celery: version_doi = format_doi(id, version)
    Celery->>DB: Set metadata.doi = version_doi
    alt DOI configured
        Celery->>DB: Set doi_state = 'pending'
    end
    Celery->>DB: Validate metadata, COMMIT

    alt DOI configured
        DB-->>Celery: on_commit → create_published_version_doi_task
        Celery->>DC: PUT /dois/{version_doi} (Findable, full metadata + IsVersionOf)
        DC-->>Celery: 200 OK
        Celery->>DC: PUT /dois/{concept_doi} (promote Draft→Findable, HasVersion)
        DC-->>Celery: 200 OK
        Celery->>DB: UPDATE doi_state = 'findable'
    end
    DB-->>Celery: on_commit → write_manifest_files
```

### 4. DOI Promotion Failure + Retry

```mermaid
sequenceDiagram
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API
    participant DB as PostgreSQL

    Celery->>DC: PUT /dois/{doi} (promote to Findable)
    DC-->>Celery: 503 Service Unavailable
    Note over Celery: urllib3 retries 3x (backoff 1s)
    DC-->>Celery: 503 (all urllib3 retries fail)
    Note over Celery: RequestException raised
    Note over Celery: Celery autoretry 1/5 (backoff)
    Celery->>DC: PUT /dois/{doi} (Celery retry 1)
    DC-->>Celery: 200 OK
    Celery->>DB: UPDATE doi_state = 'findable'

    Note over Celery: If ALL 5 Celery retries fail:
    Celery->>DB: UPDATE doi_state = 'failed'
```

### 5. Version Deletion — Hide DOI

```mermaid
sequenceDiagram
    participant Admin
    participant API as Django API
    participant DC as DataCite API
    participant DB as PostgreSQL

    Admin->>API: DELETE /api/dandisets/{id}/versions/{ver}/
    alt Version has DOI
        API->>DC: PUT /dois/{version_doi} {event: "hide"}
        DC-->>API: 200 OK (Findable → Registered)
        Note over DC: DOI still resolves but metadata not public
    end
    API->>DB: Delete version
    API-->>Admin: 204 No Content
```

### 6. Dandiset Deletion — Delete Draft DOI

```mermaid
sequenceDiagram
    participant Admin
    participant API as Django API
    participant DB as PostgreSQL
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API

    Admin->>API: DELETE /api/dandisets/{id}/
    Note over API: Only allowed if no published versions
    API->>DB: Record concept_doi before deletion
    API->>DB: Delete versions + dandiset
    API->>DB: COMMIT
    alt concept_doi exists
        DB-->>Celery: on_commit → delete_dandiset_doi_task
        Celery->>DC: DELETE /dois/{concept_doi}
        alt 204 (Draft deleted)
            DC-->>Celery: 204 No Content
        else 404 (never registered)
            DC-->>Celery: 404 (logged, ignored)
        end
    end
```

### 7. Unembargo — Mint Concept DOI

```mermaid
sequenceDiagram
    participant System
    participant DB as PostgreSQL
    participant Celery as Celery Worker (doi queue)
    participant DC as DataCite API

    System->>DB: Set embargo_status = OPEN
    System->>System: concept_doi = format_doi(id)
    System->>DB: UPDATE dandiset.concept_doi
    alt DOI configured
        System->>DB: UPDATE draft_version.doi_state = 'pending'
        DB-->>Celery: on_commit → create_dandiset_doi_task
        Celery->>DC: POST /dois (Draft concept DOI)
        DC-->>Celery: 201 Created
        Celery->>DB: UPDATE doi_state = 'draft'
    end
```

### 8. Remediation — Fix Historical DOIs

```mermaid
sequenceDiagram
    participant Op as Operator
    participant Cmd as remediate_dois
    participant DB as PostgreSQL
    participant DC as DataCite API

    Op->>Cmd: ./manage.py remediate_dois --delay 2
    loop For each version with null/fake DOI
        Cmd->>Cmd: real_doi = format_doi(id, version)
        Cmd->>DB: Set doi + doi_state='pending'
        Cmd->>DC: PUT /dois/{real_doi} (Findable)
        DC-->>Cmd: 200 OK
        Cmd->>DB: Set doi_state = 'findable'
        Cmd->>Celery: write_manifest_files (async)
        Note over Cmd: sleep(delay)
    end
    loop For each dandiset without concept_doi
        Cmd->>DB: Set concept_doi
        Cmd->>DC: POST /dois (Draft)
        Note over Cmd: sleep(delay)
    end
```

### DOI State Machine

```mermaid
stateDiagram-v2
    [*] --> NULL: DOI not configured
    [*] --> pending: creation / publish (DOI configured)
    pending --> draft: concept DOI registered on DataCite
    pending --> findable: version DOI created + concept promoted
    pending --> failed: all retries exhausted
    draft --> findable: publish (concept DOI promoted)
    findable --> registered: hide (version deletion)
    failed --> findable: manual remediation
    NULL --> pending: DOI later configured + remediation
```

</details>

## Changes by area

**New files:**
- `dandiapi/api/services/doi/` — DOI service layer (`__init__.py`, `utils.py`, `exceptions.py`)
- `dandiapi/api/management/commands/remediate_dois.py` — historical DOI remediation (AI-generated)
- `dandiapi/api/management/commands/check_doi_health.py` — DOI health monitoring (AI-generated)
- `dandiapi/api/migrations/0032_*`, `0033_*` — model fields + unique constraint

**Modified files:**
- `dandiapi/api/services/dandiset/__init__.py` — concept DOI at creation + deletion cleanup
- `dandiapi/api/services/publish/__init__.py` — real DOI replaces fake injection, async version DOI
- `dandiapi/api/services/embargo/__init__.py` — concept DOI on unembargo
- `dandiapi/api/tasks/__init__.py` — 5 DOI Celery tasks with retry on `doi` queue
- `dandiapi/api/views/version.py` — DOI update on metadata change, hide DOI on version delete
- `dandiapi/api/models/version.py` — `doi_state` field with `DoiState` choices
- `dandiapi/api/models/dandiset.py` — `concept_doi` field (unique)
- `web/src/components/DLP/HowToCiteTab.vue` — filter placeholder DOIs, suppress draft DOIs
- `Procfile` — add `doi` queue to worker
- `docker-compose.override.yml` — add `doi` queue to dev worker

**Deleted files:**
- `dandiapi/api/doi.py` — replaced by `services/doi/` package

## Test plan

- [x] 800 existing tests pass with 0 failures (local)
- [x] Lint clean (`ruff check` + `ruff format`)
- [x] Design doc compliance verified (all 12 requirements from `doi-generation-2.md`)
- [x] Reviewed by tech stack, DataCite, design, DevOps, SysOps, informatician, and devil's advocate experts
- [ ] Docker integration test with DataCite test instance
- [ ] Verify concept DOI registration on dandiset creation
- [ ] Verify version DOI + concept DOI promotion on publish
- [ ] Run `remediate_dois --dry-run` against staging


yarikoptic:
I have edited the description to place all issues and PRs into itemized lists so their titles get cited so it is easier to understand what is cited.  We should create overall project CLAUDE.md or actually CONTRIBUTING.md (to be referred to in CLAUDE.md) with such instructions IMHO.  I also added prior attempt of 

- #2350

by @asmacdo so I will invite him to review this work as well.

satra:
@yarikoptic - the pointer to 2350 was there, no? or are you talking about styling it specifically?

yarikoptic-gitmate (changes requested):
# PR #2799 — DOI lifecycle review

Reviewed against the design doc `doc/design/doi-generation-2.md` and the user's
two specific concerns: design-doc compliance, and test quality (parametrization,
no slop). Reviewed commits `a829e9a04..5107988ae` (head `5107988a`) on
`origin/enh/dandiset-dois-phase4`.

## TL;DR

- **Design doc compliance**: ~10 of 12 requirements implemented; 2 are partial
  (concept-DOI hide on dandiset deletion, and the design's claim that "no DB
  migration will be needed" — PR adds two new fields and two migrations).
- **Tests for new behavior**: **none added**. The PR introduces ~850 lines of
  new logic across services, tasks, views, models and management commands, and
  changes 0 lines of test code. Existing tests cover the new code only as far
  as `doi_configured()` short-circuits in test settings.
- **Code-quality issues** worth addressing before merge: three
  CLAUDE.md-violating inline imports, a synchronous DataCite call inside the
  destroy view, a state-machine bug that leaves 4xx-failed DOIs stuck in
  `pending` forever, and a `Procfile` change that doesn't deliver the
  "dedicated queue" claim from the description.

<details>
<summary>Details: click to expand</summary> 

---

## 1. Design doc requirements vs. implementation

Cross-checked each bullet of the "Proposed Solution" section
(`doc/design/doi-generation-2.md` lines 53–84) against the code at the PR
head. Status: ✅ implemented, ⚠️ partial / deviates, ❌ missing.

### Public dandisets

| # | Design requirement | Status | Where (PR head) |
|---|---|---|---|
| 1 | On creation, mint Draft Dandiset DOI w/ minimal metadata + DLP URL | ✅ | `services/dandiset/__init__.py:83-94` (`create_open_dandiset` schedules `create_dandiset_doi_task` on commit); `services/doi/__init__.py:64-85` (`create_dandiset_doi`) |
| 2 | If minting fails: log and continue, don't block creation | ✅ | task is async on the `doi` queue; failure only updates `doi_state='failed'` after retries; creation already committed |
| 3 | On metadata update of unpublished open dandiset: update Draft DOI | ✅ | `views/version.py:138-139` schedules `update_dandiset_doi_task` only if `not embargoed`, no published version, and `concept_doi` set |
| 4 | If update validation fails: log and continue | ✅ | `update_dandiset_doi_task` only retries on `RequestException`; `DataCiteAPIError` (from 4xx/5xx) propagates and is logged by Celery |
| 5 | On dandiset deletion (pre-publication): delete Draft DOI | ✅ | `services/dandiset/__init__.py:175-177` schedules `delete_dandiset_doi_task` |
| 6 | On first publication: mint Findable Version DOI | ✅ | `services/publish/__init__.py:191-200` + `services/doi/__init__.py:34-61` |
| 7 | On first publication: update Dandiset DOI metadata to match published version | ✅ | `services/doi/__init__.py:159-167` (`create_published_version_doi` calls `update_dandiset_doi(publish=True)`) |
| 8 | On first publication: promote Dandiset DOI to Findable | ✅ | same as #7 — `publish=True` triggers DataCite `event=publish` via `to_datacite` |
| 9 | On metadata update **after** first publication: no-op | ✅ | `views/version.py:138` guard `most_recent_published_version is None` |
| 10 | On deletion of published version: hide Version DOI to Registered | ⚠️ Partial — see "Issues" below | `views/version.py:184-195` calls `hide_published_version_doi` synchronously, NOT through the doi Celery queue |
| 11 | On dandiset deletion: "hide Dandiset DOI if Findable, delete if Draft" | ❌ Missing the "hide if Findable" path | `delete_dandiset` only schedules DELETE; in practice this is unreachable because dandiset deletion is blocked when published versions exist, but the design also covers the case where all published versions have been deleted/hidden |
| 12 | On subsequent publications: mint new Version DOI + update Dandiset DOI | ✅ | same flow as #6/#7; `update_dandiset_doi(publish=True)` is idempotent against an already-Findable concept DOI (DataCite accepts `event=publish` repeatedly) |

### Embargoed dandisets

| # | Design requirement | Status | Where |
|---|---|---|---|
| 13 | On creation: no DOI | ✅ | `create_embargoed_dandiset` does not set `concept_doi` |
| 14 | On metadata changes/deletion: no DOI ops | ✅ | no code paths call DataCite for embargoed |
| 15 | On unembargo: mint Draft Dandiset DOI | ✅ | `services/embargo/__init__.py:76-86` |

### Cautions section (lines 86–98)

| # | Design requirement | Status | Notes |
|---|---|---|---|
| 16 | If `DANDI_DOI_PUBLISH=false`: disable Findable creation/promotion | ✅ | `_validate_datacite_configuration` raises `DataCitePublishNotEnabledError` for `event in ['publish','hide']` |
| 17 | If DOI config absent: **no DOI string set on version** | ❌ Deliberately deviated | PR sets the deterministic DOI string regardless of config (see "Deviations" below) |

### Migration / DB changes

The design doc explicitly states (line 185-186):
> No DB migration will be needed, as no new field will be added to `Dandiset`
> model, and instead, the `Dandiset DOI` will be stored in the "draft" `Version`.

PR adds **two new fields** (`Dandiset.concept_doi`, `Version.doi_state`) and
two migrations (0032, 0033). This is a deliberate departure for two reasons,
neither documented in the PR description or design doc:
- `concept_doi` on `Dandiset` provides a stable reference even when the draft
  version is deleted or replaced — sensible.
- `doi_state` is needed for the `pending → draft → findable / failed`
  state machine introduced by the async-Celery design.

These departures are reasonable but **the design doc should be updated** (or
an addendum added) to reflect the actual schema.

---

## 2. Specific issues

### 2.1 No tests for any of the new behavior

The PR description claims "800 existing tests pass with 0 failures (local)",
"Design doc compliance verified" — but:

```
$ git diff --name-only origin/master...origin/enh/dandiset-dois-phase4 | grep -i test
(empty)
```

Zero test files changed. The new code paths are exercised only insofar as
`doi_configured()` returns `False` in `dandiapi/settings/testing.py` (no
`DJANGO_DANDI_DOI_API_URL/USER/PASSWORD` set), which causes every public
DOI service function to early-return.

What is therefore **not tested**:
- Concept DOI creation on dandiset creation (the `transaction.on_commit`
  branch is dead in tests).
- DOI state machine transitions (`pending → draft / findable / failed`).
- `update_dandiset_doi_task` scheduling on metadata update.
- `delete_dandiset_doi_task` scheduling on dandiset deletion.
- `hide_published_version_doi` on version deletion.
- Concept DOI minting on unembargo.
- Idempotency claim for PUT-based version DOI creation.
- Filter for placeholder DOIs and concept-DOI suppression in the frontend
  (`HowToCiteTab.vue`).
- Behavior of `to_datacite` `concept_doi` parameter and the `TypeError`
  fallback (`services/doi/utils.py:96-103`).
- Both management commands (`remediate_dois`, `check_doi_health`).

The existing `test_publish_task` (`tests/test_tasks.py:347`) does exercise
`_publish_dandiset` end-to-end and verifies that `published_version.doi`
matches what's in metadata — so it indirectly covers the
"`format_doi` replaces fake injection" path. That is the only meaningful new
coverage.

#### Suggested test additions (parametric where possible)

A reasonable minimum set, all using `monkeypatch.setattr(settings, ...)` to
flip `doi_configured()` on/off and `responses` / `pytest-mock` to stub
DataCite:

1. **`test_create_open_dandiset_concept_doi`** — parametrized over
   `(doi_configured: True | False)`:
   - configured → `dandiset.concept_doi` set, `draft_version.doi_state ==
     'pending'`, `create_dandiset_doi_task` was scheduled on_commit (use
     `django_capture_on_commit_callbacks`).
   - unconfigured → `concept_doi` set (deterministic), `doi_state is None`,
     no task scheduled.

2. **`test_create_dandiset_doi_task_state_transitions`** — parametrized over
   the DataCite response shape:
   ```python
   @pytest.mark.parametrize(
       ('datacite_response', 'expected_state', 'expected_retries'),
       [
           ({'status': 201}, 'draft', 0),
           ({'status': 503}, 'failed', 5),  # urllib3+celery retries exhausted
           ({'status': 422}, ..., 0),       # 4xx — no retry
       ],
   )
   ```

3. **`test_publish_creates_version_doi_and_promotes_concept`** — parametric
   over `is_first_publication: bool`. Verifies `_create_version_doi` PUT
   + `update_dandiset_doi(publish=True)` call sequence and final `doi_state ==
   'findable'`.

4. **`test_version_destroy_hides_doi`** — parametric over
   `(version.doi: None | str, doi_configured: bool, datacite_succeeds: bool)`.
   When DOI is set and DataCite call fails, verify the version is still
   deleted and the failure is logged but not raised.

5. **`test_metadata_update_schedules_doi_update`** — parametric over the
   matrix `(embargoed, has_published, has_concept_doi)`; only the
   `(False, False, True)` case should schedule `update_dandiset_doi_task`.

6. **`test_unembargo_mints_concept_doi`** — verify `format_doi` set, task
   scheduled, doi_state pending.

7. **`test_remediate_dois_command`** — parametric over the version's existing
   DOI value `[None, '.123456/0.123456.1234', 'real-doi']`; verify only the
   first two trigger remediation in dry-run output.

Most of these naturally take 2–4 parametric arguments and consolidate into 6–7
test functions covering 25+ behaviors. That's a much smaller test footprint
than what the design doc warrants and would catch any regression in the state
machine.

### 2.2 Synchronous DataCite call in the destroy view (design doc §10)

`dandiapi/api/views/version.py:184-195`:

```python
if version.doi is not None:
    try:
        hide_published_version_doi(version)
    except Exception:
        import logging  # ← inline import, see 2.4
        logging.getLogger(__name__).exception(...)
version.delete()
```

Issues:
- The PR description and architecture (sequence diagrams §1, §2, §3)
  promise: "All DataCite operations via async Celery tasks on dedicated
  `doi` queue with retry + exponential backoff." This call is synchronous
  inside the request handler, blocking on DataCite for up to ~3 minutes
  in the worst case (5s connect + 30s read × 3 urllib3 retries).
- `except Exception` is broader than needed; `except DataCiteAPIError`
  (and possibly `RequestException`) would be tighter and let unexpected
  bugs propagate.
- Recommend: introduce `hide_published_version_doi_task` on the `doi`
  queue and `transaction.on_commit(...)` it after `version.delete()`.

### 2.3 `Procfile` change merges `doi` into the default celery queue

```diff
-worker: ... -Q celery -B ...
+worker: ... -Q celery,doi -B ...
```

The PR description says DOI ops run on a "dedicated `doi` queue", but the
`Procfile` change makes the default web/api worker consume it as well.
Two consequences:
- A burst of DOI tasks (e.g., from `remediate_dois`) will compete with
  every other `celery` queue task on the same worker.
- Celery beat (`-B`) is on this worker, so under DOI-task bursts a
  long-running DOI HTTP call (60s soft-limit) may delay scheduled beat
  ticks on a single-worker deploy.

If "dedicated" is the goal, the Procfile should add a separate `doi-worker`
process (analogous to `checksum-worker`) and remove `,doi` from the default
worker. If the goal is just "isolate from `manifest-worker`", the current
change is fine but the description is misleading.

### 2.4 Inline imports (CLAUDE.md violation)

The user's global CLAUDE.md says: "All imports MUST be at the module level".
Three new inline imports in this PR:

- `dandiapi/api/views/version.py:188` — `import logging` inside `destroy`.
  No legitimate reason; module has no top-level `logging` import — just add one.
- `dandiapi/api/services/embargo/__init__.py:83` —
  `from dandiapi.api.tasks import create_dandiset_doi_task` inside
  `unembargo_dandiset`. The same module already imports `unembargo_dandiset_task`
  from `dandiapi.api.tasks` at the top (line 19), so there's no circular-import
  excuse here — move it up.
- `dandiapi/api/services/doi/utils.py:48` —
  `from dandischema.conf import get_instance_config` inside `format_doi`.
  `dandischema` is a third-party package, so no circular-import excuse — move
  it to the module level.

(Other `from ... import ...` statements inside functions in this codebase
predate this PR and exist to break genuine `services ↔ tasks` cycles, e.g.
`services/publish/__init__.py:208`. Those are acceptable. The three above are not.)

### 2.5 State-machine bug: 4xx responses leave DOIs stuck in `pending`

Three of the new Celery tasks (`create_dandiset_doi_task`,
`create_published_version_doi_task`, `delete_dandiset_doi_task`) declare:

```python
autoretry_for=(requests.exceptions.RequestException,)
```

and inside `create_*_doi_task` (`tasks/__init__.py:115-141`):

```python
except Exception:
    if self.request.retries >= self.max_retries:
        Version.objects.filter(...).update(doi_state='failed')
    raise
```

When DataCite returns a 4xx (e.g., 422 invalid metadata), the code raises
`DataCiteAPIError` — which is **not** a `RequestException`, so Celery does
not retry (correct: don't retry invalid metadata). But the `except Exception`
handler then evaluates `self.request.retries >= self.max_retries` as
`0 >= 5 → False`, so `doi_state` is **never** flipped to `failed`.

Net effect: 4xx-failed DOIs stay at `doi_state='pending'` forever.
`check_doi_health` only flags pending older than its threshold *and* failed,
so these would surface as "STUCK PENDING" but never as "FAILED" — operators
would have to manually re-trigger.

Fix: gate the state transition on the retry decision, not the retry counter.
E.g., flip to `failed` if the exception is **not** a `RequestException`, or
if it is and retries are exhausted.

### 2.6 Minor

### 2.5 Minor

- `services/doi/utils.py:96-103` — the `TypeError` fallback for older
  `dandischema` versions silently drops `concept_doi`. PR description lists a
  hard `Depends on: dandi/dandi-schema#401`. If that's a hard dependency,
  the fallback is dead code that masks future regressions; either drop the
  fallback or document why it's load-bearing.
- `services/doi/__init__.py:115` (`update_dandiset_doi`) — when called
  on a dandiset whose draft version's `doi is None`, raises
  `VersionDOIMissingError`. There is a `# TODO: DOI: Remove once DOI is
  required in all versions` comment but no companion field validation
  (`doi` on `Version` is still `null=True`). The contract is "always set
  the DOI string"; consider following up with a migration that backfills
  + a `NOT NULL` constraint, or keep the TODO actionable in an issue.
- `tasks/__init__.py:90` — `delete_doi_task` is documented as a "Legacy
  wrapper" but `git grep` finds zero remaining callers anywhere in the
  repo (only the def). Delete it.
- `models/version.py:50-59` — the `DoiState` choices include `REGISTERED`
  but no code path transitions to it (the synchronous hide path in 2.2
  doesn't update `doi_state` either). Either `hide_published_version_doi`
  should set `doi_state='registered'` after success, or `REGISTERED` should
  be dropped from choices until used.
- `tasks/__init__.py:152-155` — `update_dandiset_doi_task` performs zero
  `doi_state` bookkeeping on success or failure. Less severe than 2.5
  (concept-DOI updates are a metadata refresh, not a state transition),
  but it leaves the operator with no signal when concept-DOI updates fail
  silently. Worth at least a structured log line on each branch.
- `services/dandiset/__init__.py:84-86` — `dandiset.save(update_fields=
  ['concept_doi'])` is fine; the surrounding `transaction.atomic` plus the
  PK uniqueness on `Dandiset.id` already prevents duplicate concept DOIs.
  Given `concept_doi = format_doi(identifier)` is a deterministic function
  of the PK, the **`unique=True` constraint added in migration 0033 is
  redundant**. Either drop it (one less constraint to maintain), or
  document why it's load-bearing.

---

## 3. Test-quality audit (the user's second concern)

Since the PR adds **no tests**, there's no slop to remove. The slop is in
the *absence* of tests for the new code. The recommended additions in
§2.1 are designed to be parametric (each covers 3–4 cases per function),
keeping the new test surface tight.

For the **existing** tests, the only DOI-touching ones (`test_publish_task`
in `test_tasks.py:347`, `test_version_publish_version` in
`test_version.py:310`) still use `'doi': fake_doi` literal strings and
`fake_doi = 'doi'` magic values. They pass only because
`_build_publishable_version_from_draft` doesn't compute the DOI any more;
`_publish_dandiset` does — and `test_version_publish_version` bypasses
`_publish_dandiset` entirely by calling the helper directly. The moment
DOI assignment moves into `_build_publishable_version_from_draft` (the
natural next refactor), the test silently starts comparing `'doi'` to
`format_doi(...)` and breaks. Tighten these to assert the deterministic
format now, while the meaning of the assertion is still obvious.

---

## 4. Recommendations

In priority order:

1. **Fix the 4xx-stuck-pending state-machine bug** (§2.5). Single-line fix,
   prevents silent operator pain in production.
2. **Add the test set in §2.1** — at minimum tests #1, #3, and #4 (creation,
   publish, destroy), and add explicit coverage of the §2.5 4xx path.
   Without these, any regression in the state machine ships silently.
3. **Move `hide_published_version_doi` to a Celery task** on the `doi` queue
   (§2.2). Same for the dandiset-deletion "hide if Findable" gap (§ table
   row 11).
4. **Resolve the Procfile/`doi`-queue contradiction** (§2.3) — either add a
   dedicated `doi-worker` line or update the PR description to remove the
   "dedicated queue" claim.
5. **Fix the three inline imports** (§2.4) — small but a CLAUDE.md guideline.
6. **Update the design doc** to reflect the new fields and state machine
   (or add a "Deviations" section to the PR description), since the doc
   currently says "No DB migration will be needed".
7. Address the minor items in §2.6 as low-priority follow-ups.

The architectural design is sound; the gap is in shipping it without a
safety net of tests for code paths that, by definition, can't be exercised
in the test environment without DOI configuration.

</details>

yarikoptic:
> @yarikoptic - the pointer to 2350 was there, no? 

right -- missed that.  But I think it is misattributed to @jjnesbitt whenever it was @asmacdo (didn't fix since may be I misunderstood)

> or are you talking about styling it specifically?

overall about styling - instead of list of numbers like it was 

image

it became more immediately human accessible

image

asmacdo:
I've done a very rough first pass review. A more in-depth review is needed, but my major concerns should reveal themselves during manual testing.

**`concept_doi` on `Dandiset`**

This was explicitly ruled out by the design doc to avoid a migration. I'm not opposed. IMO that's the correct place for it, and it avoids the awkwardness of multiple DOI patterns to check for in the pydantic model. But the design doc should be updated in this PR with explicit reasoning for the change (e.g. enabling the backfill workflow, and probably other awkardness). 

**Schema dependency is incomplete**

This PR passes `concept_doi` to `to_datacite`, which only exists in dandi-schema #401. That's fine as a declared dependency, but #401 itself seems incomlete. It adds the `concept_doi` parameter, a `dates` field, and a DANDI altIdentifier. It does **not** touch the `PublishedDandiset(**meta)` validation. Specifically I remember failing validation on the `meta.datePublished.year` line, a freshly created draft (no `datePublished`) will raise `ValidationError` inside `to_datacite`.  This is the same problem #297 addressed with `construct_unvalidated_dandiset` and a try-validate-then-fallback pattern. That work is was ugly but necessary (at least it seemed so at the time) and there's no equivalent in #401. 

I'm suspicious of `try: to_datacite(..., concept_doi=...) except TypeError` fallback in `generate_doi_data`, I think it risks silent failures on version mismatch, but I'll have a closer look next time.

Because of those issues, I assume that this PR has not been tested against a real datacite API, so I expect a handful of errors. that would be the next step before more detailed review. 

**WARNING**: Prior to testing against a real datacite API, be sure not to alter the doi so its not just dandiset 000001, because that doi will never be able to be deleted and will cause issues on future tests.

satra:
Thank you all for the thorough reviews. Addressing each concern:

## Attribution

@yarikoptic — correct, the original PR #2350 was by @asmacdo. The `rf-enh-dandiset-dois` branch (which we ported the service layer from) was @jjnesbitt's follow-up. I'll fix the attribution in the description.

## `concept_doi` on `Dandiset` (design doc deviation)

@asmacdo — agreed, this is a deliberate deviation from the design doc's "no new field on Dandiset." The reasoning:

1. **Concept DOI is semantically a dandiset-level attribute**, not a version-level one. Storing it on the draft Version creates awkward semantics when the draft is deleted/replaced.
2. **`doi_state` tracking** is needed for the async Celery workflow to detect stuck/failed states — impossible without a field.
3. **The `remediate_dois` backfill command** needs to set concept DOIs on dandisets that pre-date this feature — much cleaner with a dedicated field.

Will update the design doc in this PR with a "Deviations" addendum.

## Schema dependency / `to_datacite` validation on drafts

@asmacdo — this is a key concern. You're right that `to_datacite(PublishedDandiset(**meta))` will fail for a freshly created draft (no `datePublished`, etc.). The current code path for concept DOIs calls `generate_doi_data(dandiset, version=None, publish=False)` which passes the draft metadata through `to_datacite`. For a brand-new draft with minimal metadata, this will indeed fail inside `to_datacite` when trying to construct `PublishedDandiset`.

This is the exact problem #297 solved with `construct_unvalidated_dandiset`. We need either:
- A) The try-validate-then-fallback pattern from #297
- B) A separate code path for Draft DOIs that sends minimal metadata without going through `PublishedDandiset` validation

This needs to be fixed before merging. Will address.

## `try/except TypeError` for `concept_doi`

@asmacdo — your concern about silent failures is valid. This was a backward-compatibility shim for the transition period where the archive might run with an older dandischema. Since #401 is a hard dependency, the fallback is dead code that could mask regressions. Will remove it and make the `concept_doi` parameter unconditional (fail loudly if the schema doesn't support it).

## Testing against real DataCite API

@asmacdo — correct, this has NOT been tested against a real DataCite API. The 800 passing tests all run with `doi_configured() == False`, so every DOI service function early-returns. The next step before merge is DataCite test instance integration testing. Will be careful about DOI suffix uniqueness per your warning.

## No tests added (TDD question)

@yarikoptic-gitmate — this is the most important concern. Here's the honest accounting:

**What happened**: TDD was used during development on the earlier `enh/datacite-doi-minting` branch (Phases 1-3), which had dedicated test files (`test_doi.py`, `test_publish_doi_workflow.py`). When we pivoted to building on Jake's service layer architecture (this branch), those test files were on the other branch and the ported service layer code from `rf-enh-dandiset-dois` was integrated without porting the tests.

**Why the gap exists**: The fundamental challenge is that `doi_configured()` returns `False` in the test environment, so all DOI service functions no-op. Testing the actual DOI logic requires either:
1. Mocking `doi_configured()` to return `True` and stubbing `datacite_session()`
2. Setting DOI env vars in tests and using `responses` library to mock HTTP

Neither was set up on this branch. The existing 800 tests pass because the `doi_configured()` guards silently skip all new code.

**The plan**: Will add the 7 parametrized test functions suggested in the review (§2.1) before this is merge-ready. These cover creation, state transitions, publish, destroy, metadata update, unembargo, and remediation — all using mocked DataCite responses.

## Synchronous `hide_published_version_doi` in destroy view

Agreed — should be async. Will create a `hide_published_version_doi_task` on the `doi` queue.

## Procfile / "dedicated queue" claim

Fair point — adding `doi` to the default worker's `-Q` list makes it a shared queue, not dedicated. Two options:
1. Add a separate `doi-worker` process (truly dedicated)
2. Keep current approach but update the description to say "separate queue" instead of "dedicated queue"

Option 2 is simpler for single-worker deploys. Will update the description and add a comment in the Procfile.

## 4xx stuck-in-pending state machine bug

Good catch. The `except Exception` handler checks `self.request.retries >= self.max_retries`, but for `DataCiteAPIError` (not a `RequestException`), Celery doesn't retry, so `retries` stays at 0 and the condition is never true. Fix: mark as `failed` if the exception is NOT in `autoretry_for`.

## Inline imports

Will move the 3 inline imports to module level. The `format_doi` one in `utils.py` was to avoid a potential circular import but `dandischema.conf` is a third-party package — no cycle possible.

## `delete_doi_task` (legacy wrapper)

Confirmed no callers — `git grep` shows zero references outside its own definition. Will remove.

## `REGISTERED` state unused

`hide_published_version_doi` should set `doi_state='registered'` after success. Will fix.

---

## Summary of planned fixes

1. Fix 4xx stuck-pending state machine bug
2. Add 7 parametrized test functions (mocked DataCite)
3. Handle `to_datacite` validation failure for Draft DOIs (minimal metadata path)
4. Remove `TypeError` fallback (make `concept_doi` param unconditional)
5. Move `hide_published_version_doi` to async task
6. Set `doi_state='registered'` after successful hide
7. Fix inline imports
8. Remove dead `delete_doi_task`
9. Update design doc with deviations addendum
10. Update PR description re: queue terminology
11. Fix attribution (@asmacdo for #2350)

dandiapi/api/tasks/__init__.py
```diff
@@ -134,11 +134,15 @@ def create_published_version_doi_task(self, version_id: int) -> None:
...
         Version.objects.filter(id=version_id).update(doi_state='findable')
-    except Exception:
-        # Only mark as 'failed' on the final retry attempt
+    except requests.exceptions.RequestException:
+        # Transient — mark failed only on final retry
         if self.request.retries >= self.max_retries:
             Version.objects.filter(id=version_id, doi_state='pending').update(doi_state='failed')
         raise
+    except Exception:
+        # Non-retryable (4xx, validation error) — mark failed immediately
```
yarikoptic, line 150: how do we know it is 4xx?  note that there is also

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429

which is "429 Too Many Requests" and it would have Retry-After. Anything here (and similar above) would handle those as to sleep accordingly and retry again?

yarikoptic-gitmate (changes requested):
# Follow-up review of PR #2799 — second pass

Re-reviewed at head `9ba7376f0` (commit "fix: address all reviewer feedback —
schema dep, state machine, cleanup"), comparing against the prior review at
`5107988ae`. Single new commit, 6 files, 35 insertions / 33 deletions.

## TL;DR

- **Addressed (5 of 13 items)**: 4xx-stuck-pending state machine bug ✅,
  TypeError fallback removed ✅, two of three inline imports moved to module
  level ✅, dead `delete_doi_task` removed ✅, `DoiState.REGISTERED` now set
  after successful hide ✅.
- **Still open (8 of 13 items)** — all higher-priority than the items above:
  zero new tests for ~850 LOC of new logic, synchronous DataCite call still
  in destroy view, Procfile still doesn't deliver "dedicated `doi` queue",
  third inline import left in place with an incorrect "circular import"
  justification, redundant `unique=True` constraint, no state bookkeeping
  on `update_dandiset_doi_task`, "hide if Findable" gap on dandiset deletion,
  design doc still says "No DB migration will be needed".
- **Re: yarikoptic's 429 line comment**: 429 *is* retried (urllib3 hoists
  it via `status_forcelist`), but the Celery retry layer ignores the
  `Retry-After` header during sustained rate-limit. Details + fix sketch
  in §3 below.
- **Verdict**: still **request changes**. The state-machine fix is exactly
  right; the rest of the priority-1/priority-2 items from §4 of the prior
  review remain untouched.

<details>
<summary>1. Item-by-item status against prior review</summary>

Numbering follows prior review's §4 Recommendations (priority order).

| # | Item | Prior status | This pass |
|---|---|---|---|
| 1 | Fix 4xx-stuck-pending state machine bug (§2.5) | OPEN | ✅ **FIXED** — `tasks/__init__.py:107-119` and `:137-145` now split `except RequestException` (retry-aware) from generic `except Exception` (immediate `failed`). Verified in diff. |
| 2 | Add tests in §2.1 (creation, publish, destroy minimum) | OPEN | ❌ **NOT ADDRESSED** — `git diff merge-base..HEAD --name-only \| grep -i test` is still empty. Commit message claims "800 tests pass" — same number as before, confirming no additions. |
| 3 | Move `hide_published_version_doi` to a Celery task | OPEN | ❌ **NOT ADDRESSED** — `views/version.py:188-194` still calls `hide_published_version_doi(version)` synchronously inside `destroy`. Only change is that `doi_state='registered'` is now set on success (good — addresses the §2.6 sub-item) and the `import logging` was hoisted (good). The sync-blocking-DataCite issue stands. |
| 3b | Same gap for "hide if Findable" on dandiset deletion (§ table row 11) | OPEN | ❌ **NOT ADDRESSED** — `services/dandiset/__init__.py:176-178` still only schedules `delete_dandiset_doi_task` (DELETE). No path handles a Findable concept DOI. |
| 4 | Resolve Procfile / `doi`-queue contradiction (§2.3) | OPEN | ❌ **NOT ADDRESSED** — `Procfile` line 9 unchanged: `-Q celery,doi -B`. PR description still claims "dedicated `doi` queue". Pick one. |
| 5 | Fix the three inline imports (§2.4) | OPEN | ⚠️ **PARTIAL** — 2 of 3 fixed: `views/version.py` (`logger` now at module level, line 4 + 35) ✅; `services/doi/utils.py` (`get_instance_config` now at line 8) ✅. The third (`services/embargo/__init__.py:84`) is **NOT** fixed — see callout below. |
| 6 | Update design doc (`doi-generation-2.md`) | OPEN | ❌ **NOT ADDRESSED** — design doc line 185-186 still reads "No DB migration will be needed, as no new field will be added to `Dandiset` model". PR adds two fields and two migrations. Either fix the doc or add a "Deviations" section to the PR description. |
| 7a | `services/doi/utils.py` `TypeError` fallback (§2.6) | OPEN | ✅ **FIXED** — removed; now a hard call to `to_datacite(..., concept_doi=concept_doi)`. Coupled with the new `git+` dependency below. |
| 7b | `tasks/__init__.py` `delete_doi_task` legacy wrapper (§2.6) | OPEN | ✅ **FIXED** — definition removed in diff. |
| 7c | `models/version.py` unused `DoiState.REGISTERED` (§2.6) | OPEN | ✅ **FIXED** — `views/version.py:191` now sets `version.doi_state = 'registered'` after successful hide. |
| 7d | `update_dandiset_doi_task` zero `doi_state` bookkeeping (§2.6) | OPEN | ❌ **NOT ADDRESSED** — `tasks/__init__.py:156-159` is still a 3-line wrapper with no state writes and no logging on failure (Celery's default exception logging only). |
| 7e | Redundant `concept_doi unique=True` (§2.6) | OPEN | ❌ **NOT ADDRESSED** — migration `0033` still alters `concept_doi` to `unique=True`. The value is `format_doi(identifier)`, deterministic on the PK, so the PK already enforces uniqueness; the index is redundant. |
| 7f | `update_dandiset_doi` `VersionDOIMissingError` TODO (§2.6) | OPEN | ❌ **NOT ADDRESSED** — `services/doi/__init__.py:115` still raises `VersionDOIMissingError` with the same `# TODO: DOI: Remove once DOI is required in all versions` comment. No follow-up issue linked. |

</details>

<details>
<summary>2. The remaining inline import is not actually a circular-import case</summary>

The fix commit added a comment claiming the inline import in
`services/embargo/__init__.py:83-84` is required to avoid a cycle:

```python
if doi_configured():
    # Inline import to avoid circular: tasks → services.embargo → tasks
    from dandiapi.api.tasks import create_dandiset_doi_task
    transaction.on_commit(lambda: create_dandiset_doi_task.delay(ds.id))
```

This justification is **incorrect**. The same module already imports
`from dandiapi.api.tasks import unembargo_dandiset_task` at **line 19**
(module level). If a real cycle existed, line 19 would already break.

Verifying the chain by inspection:
- `services/embargo/__init__.py` line 19 → loads `tasks/__init__.py`
- `tasks/__init__.py` lines 11-30 → imports from `mail`, `manifests`,
  `models`, `services.doi`. **No top-level import from `services.embargo`.**
- `tasks/__init__.py` lines 36, 183 → only inline imports of
  `services.embargo` symbols, deferred until the task body executes.

So adding `create_dandiset_doi_task` to the existing line 19 import works:

```python
from dandiapi.api.tasks import create_dandiset_doi_task, unembargo_dandiset_task
```

The simplest test: replace lines 83-84 with the consolidated import and
run `python -c "import dandiapi.api.services.embargo"`. If the comment
were correct, that would `ImportError`; it doesn't.

The CLAUDE.md inline-import rule exists precisely to avoid the
"if-X-then-import" pattern, where the conditional gives the *appearance*
of a circular-import workaround without actually requiring one. Recommend
either fixing this and removing the comment, or — if there's an empirical
reason the import has to stay deferred (eager-import side effects in
testing, signal registration, etc.) — replacing the misleading comment
with the actual reason.

</details>

<details>
<summary>3. 429 / Retry-After handling (re: yarikoptic's line comment on tasks/__init__.py:143)</summary>

The line comment asks: "how do we know it is 4xx? ... 429 Too Many Requests
... would have Retry-After. Anything here (and similar above) would handle
those as to sleep accordingly and retry again?"

Quick path-trace through the current code on `tasks/__init__.py:143`:

1. `services/doi/utils.py:131-138` configures `urllib3.Retry` with
   `status_forcelist=[429, 500, 502, 503, 504]`, `total=3`,
   `backoff_factor=1.0`. urllib3's `Retry` honors `Retry-After` by default
   (`respect_retry_after_header=True`, capped at `BACKOFF_MAX=120s`). So a
   429 with `Retry-After: 30` causes urllib3 to sleep 30 s. Three urllib3
   retries are attempted.
2. After urllib3 exhausts retries on 429s, `requests` raises
   `RetryError(MaxRetryError(...))` — a subclass of `RequestException`.
3. Celery's `autoretry_for=(requests.exceptions.RequestException,)` then
   catches it and retries up to 5 more times with `retry_backoff=True,
   retry_backoff_max=300`. So Celery's schedule is roughly 1, 2, 4, 8, 16
   seconds (plus jitter), capped at 300 s.

So 429 **is** retried — but with two caveats worth fixing:

**Caveat 1: Celery layer ignores `Retry-After`.** Once urllib3 hands the
exception up, the `Retry-After` header value is gone. Celery retries on
its own exponential schedule (1, 2, 4, 8, 16 s) regardless of what
DataCite asked for. If DataCite returns `Retry-After: 600`, the Celery
retries all fire within ~30 s and likely all 429 again — burning the
retry budget in <1 minute and marking the DOI `failed`, despite DataCite
explicitly telling us when it would be safe to come back.

The fix is to inspect the final response in `_create_version_doi` /
`create_dandiset_doi` / etc. and, on a 429, raise via
`self.retry(countdown=retry_after_seconds)` instead of letting urllib3
swallow it. Sketch:

```python
# in services/doi/utils.py
class DataCiteRateLimited(DandiError):
    def __init__(self, retry_after: int):
        self.retry_after = retry_after
        super().__init__(message=f'DataCite rate-limited, retry after {retry_after}s')

# in raise_datacite_exception (or a new helper)
if response.status_code == 429:
    retry_after = int(response.headers.get('Retry-After', '60'))
    raise DataCiteRateLimited(retry_after=retry_after)

# in the celery task — bind=True is already set
except DataCiteRateLimited as exc:
    raise self.retry(countdown=exc.retry_after, exc=exc)
```

This lets the `except RequestException` / `except Exception` split stay
exactly as it is (which is correct for *non-rate-limited* errors), and
adds a third explicit branch for 429.

**Caveat 2: how does the current code distinguish 4xx from "should
retry"?** It doesn't, at the Celery layer — it distinguishes
`RequestException` (transient/retryable, anything urllib3 surfaces as a
network problem or `RetryError`) from "everything else" (`DataCiteAPIError`
from the explicit non-OK branch in `_create_version_doi` etc.).

The "everything else" path *only* fires when:
- urllib3 retries succeeded enough to get a final response back
  (i.e., status NOT in `status_forcelist`), AND
- `response.ok` was False (i.e., 4xx other than 429).

Concretely: 400, 401, 403, 404, 410, 422 → `DataCiteAPIError` →
non-retryable → `failed`. That's correct — these errors do not get
better with retry. The only 4xx that *should* retry is 429, which urllib3
already lifts out of this path via `status_forcelist`. So the current
two-branch split is correct, with the gap that the Celery-level retry
schedule doesn't honor DataCite's `Retry-After` hint.

**TL;DR for the comment**: the 4xx vs non-4xx split is
correct-by-construction (urllib3 hoists 429 out before Celery sees it).
The remaining issue is that Celery's retry schedule throws away the
`Retry-After` value during sustained rate limiting. Worth a follow-up
commit; not a blocker for the bug fix already landed.

</details>

<details>
<summary>4. Verification of the 4xx fix</summary>

Reviewed `tasks/__init__.py` lines 95-150. The new logic:

```python
except requests.exceptions.RequestException:
    if self.request.retries >= self.max_retries:
        Version.objects.filter(...).update(doi_state='failed')
    raise
except Exception:
    # Non-retryable error (e.g., DataCiteAPIError from 4xx) — mark failed immediately
    Version.objects.filter(...).update(doi_state='failed')
    raise
```

This is exactly the fix recommended in §2.5 of the prior review. Three
nits worth a follow-up (not blocking):

- The `Exception` branch is reached for *anything* that isn't a
  `RequestException` — including `Dandiset.DoesNotExist` /
  `Version.DoesNotExist` if a row is deleted between the queue entry and
  the task pickup. In that case the `Version.objects.filter(...).update`
  becomes a no-op (filter matches nothing). Benign, but logging the
  exception type would help in debugging.
- The `RequestException` branch updates `doi_state='failed'` only at
  `retries >= max_retries`. Celery raises `Retry` internally between
  attempts — confirm with the Celery version pinned (`celery==5.5.3`)
  that `Retry` is *not* a subclass of `Exception` reachable here. (It
  isn't in 5.x; flagging only for completeness.)
- `create_dandiset_doi_task` and `create_published_version_doi_task` now
  have near-identical try/except boilerplate. Consider extracting a
  context manager `track_doi_state(version_or_dandiset, …)`.

These are minor — the fix is correct.

</details>

<details>
<summary>5. Lingering test gap</summary>

The 4xx fix in §1 is exactly the kind of state-machine logic that *needs*
a test, both as regression protection and as documentation of intent. A
proposed parametric test (using `responses` or `pytest-mock` to stub
DataCite):

```python
@pytest.mark.parametrize(
    ('http_status', 'expected_state', 'expected_celery_retry'),
    [
        (201, 'draft',    False),
        (422, 'failed',   False),  # 4xx — non-retryable, immediate
        (503, 'failed',   True),   # 5xx — retried then failed
        # 'connection_error', 'failed', True),
    ],
)
def test_create_dandiset_doi_task_state_machine(...):
    ...
```

This single test would have caught the original 4xx-stuck-pending bug
*and* will catch any future regression where someone consolidates the
two except clauses back into one. Eight lines of parametrize covering
the entire matrix.

The other test additions enumerated in §2.1 of the prior review are
unchanged; none have been added.

</details>

<details>
<summary>6. Recommendations</summary>

In priority order, what's left:

1. **Add at least one test** for the state machine fix (§5 above) and
   one for the destroy/hide flow. The state-machine test is ~20 LOC and
   self-contained — it's the smallest possible step that pays for itself.
2. **Move `hide_published_version_doi` to a Celery task on the `doi` queue**
   (`hide_published_version_doi_task`), then `transaction.on_commit(...)`
   it from the destroy view *after* `version.delete()`. Same pattern as
   the other DOI tasks. Keeps the request handler off the network.
3. **Decide what "dedicated `doi` queue" means** and pick one:
   - keep current Procfile and remove "dedicated" from the description, OR
   - add a `doi-worker:` line analogous to `checksum-worker:` and remove
     `,doi` from the default `worker:` line.
4. **Fix the third inline import** in `services/embargo/__init__.py:84` —
   the "circular import" comment is wrong (§2 above).
5. **Honor `Retry-After` for 429** (§3 above) — promote rate-limit
   responses to a typed exception and use `self.retry(countdown=...)`.
6. **Update the design doc** (or PR description) to reflect the schema
   additions and the deliberate "DOI string always set" deviation
   (`§17` of the original table).
7. **Drop `unique=True`** on `concept_doi` in migration 0033 (or write
   a one-line "why it's load-bearing" comment).
8. **Add a structured log line** in `update_dandiset_doi_task` on success
   and exception so failed concept-DOI updates surface in the operator
   dashboard.

Items 1-3 are blocking. Items 4-8 are follow-ups (5 is small; the others
are minor).

</details>

---

The state-machine fix is solid. The rest of the high-priority feedback
from the prior review is untouched. Posting as **request changes** with
the same severity calibration as before.

dandiapi/api/tasks/__init__.py
satra: Good catch. Tracing what actually happens for 429:

- **Transport layer** (`services/doi/utils.py:datacite_session`): the `urllib3.Retry` adapter has `status_forcelist=[429,500,502,503,504]` and `respect_retry_after_header=True` (the urllib3 default), so a 429 with `Retry-After` is honored — up to `total=3` retries within the single HTTP call, capped at 120s per wait. After that urllib3 raises `MaxRetryError`, wrapped as `requests.exceptions.RetryError` (a `RequestException` subclass).
- **Celery layer**: `autoretry_for=(RequestException,)` then re-runs the task with `retry_backoff=True` (exponential 1s, 2s, 4s, 8s, 16s). At this layer, the original `Retry-After` is gone — Celery does not see it.

So the answer is: 429 *is* retried with `Retry-After` honoring, but only at the transport layer. If the rate limit persists past the urllib3 budget, the Celery layer retries on its own schedule, ignoring the server's hint.

The `except Exception` branch in this hunk is reached only when the response was *delivered* but had a non-2xx status (so `raise_datacite_exception` raised `DataCiteAPIError`, not a `RequestException`). That is by definition a 4xx other than the ones in `status_forcelist` — i.e., 4xx-not-429. The split is correct, but the comment is misleading; I'll tighten it.

Verified by new test `test_create_dandiset_doi_429_then_success` in `dandiapi/api/tests/test_doi_http.py` (commit 65f21895).

Follow-up issue worth filing: typed `DataCiteRateLimited` exception that surfaces `Retry-After` from the response and does `self.retry(countdown=...)` at the Celery layer. Not blocking this PR.

satra:
**Update — addressing review (commit 65f21895):**

Code:
- ✅ **Concept-DOI create is now idempotent** (PUT to deterministic DOI URL instead of POST). Safe under retry even if DataCite accepted but DB write failed. Mirrors the version-DOI path.
- ✅ **`doi_configured()` reads settings at call time** (was importing at module load) — this was a latent test-time bug; `override_settings` now works.
- ✅ **Hoist inline imports** in `hide_published_version_doi_task` to module level (no circular-import risk; the prior comment was wrong).
- ✅ **Drop redundant `unique=True`** on `Dandiset.concept_doi` — value is derived from PK, so PK uniqueness already implies it. Migration 0033 renamed and AlterField for `concept_doi` removed (still on this PR branch, never deployed).
- ✅ **PR description**: replaced "dedicated `doi` queue" with `\`doi\` queue` to match Procfile (`-Q celery,doi -B`). The DOI tasks share a worker with the default queue and beat, by design.

Tests:
- ✅ **New `test_doi_http.py` (10 tests, AI-generated)** using `responses` to exercise the actual `datacite_session()` retry adapter — covers the highest-risk surface that was previously untested:
  - 200 happy path (verifies PUT, not POST)
  - 503-then-success (retry adapter works)
  - 422 → `DataCiteAPIError` (no retry, fail-fast)
  - 429 with `Retry-After` → eventual success (urllib3 honors header)
  - 404 silenced on delete and hide
  - End-to-end task state transitions: `pending → draft` (success) and `pending → failed` (4xx)
  - Auth/JSON-API headers applied per request

Threads:
- 🔗 yarikoptic's [429/Retry-After question](https://github.com/dandi/dandi-archive/pull/2799#discussion_r3161905660) — replied with full transport-vs-Celery analysis and filed follow-up issue #2812 for typed `DataCiteRateLimited` exception at the Celery layer.

Deferred (acceptable post-merge):
- 429 honoring at the Celery layer (#2812)
- `doi_state = 'registered'` is never written; choice exists but `destroy` deletes the row before the hide task runs. Either wire it before deletion or drop the choice — minor.
- Findable concept-DOI hide on dandiset deletion is unreachable (`delete_dandiset` rejects dandisets with published versions, so a Findable concept DOI can't get there). Worth a code comment but not load-bearing.

@yarikoptic @asmacdo — ready for re-review.

asmacdo:
I figured I'd have another look at [Lupo](https://github.com/datacite/lupo), the DataCite REST API, to see if it's feasible to test DANDI and the concept-DOI work end to end against a real DataCite instance instead of mocks.

Good news! Lupo "just worked" this time (the repo has moved a lot in the last year: Rails 8.1, MySQL 8.4, OpenSearch 2, and the dev/test environments now default every credential, so an empty `.env` is enough).

It seems possible, but heavy, to add Lupo to CI:
- docker compose with 5 services: lupo web, MySQL, OpenSearch, memcached, MinIO
- ~4.9 GB of images to pull (lupo web alone is 2.36 GB, OpenSearch 1.47 GB)
- time on my machine: ~5 min for the initial pull and start, then ~6 s to create the search indices and ~16 s for `db:setup` (create, schema load, seed). Restarting from a wiped volume with images cached is ~1 s plus those ~22 s.

The Lupo README is nearly sufficient to get started. One step it omits: `rake elasticsearch:create_all_indexes` has to run before `db:setup`, otherwise the first write auto-creates unmapped indices and every list endpoint returns 400. Also worth setting `HANDLE_URL` to a dead host so a local instance doesn't poke DataCite's real test Handle server. Seeded client credentials are the dev defaults, `DATACITE.TESTUSER` / `test_mds_password`, with prefix `10.14454`.

<details><summary>Findable DOIs can be created via the REST API</summary>

```bash
curl -u DATACITE.TESTUSER:test_mds_password \
  -H 'Content-Type: application/vnd.api+json' \
  -X POST localhost:8065/dois \
  -d '{"data":{"type":"dois","attributes":{"event":"publish","doi":"10.14454/dandi-smoke-001","url":"https://example.org/dandi-smoke","creators":[{"name":"DANDI Test"}],"titles":[{"title":"Smoke test"}],"publisher":"DANDI Archive","publicationYear":2026,"types":{"resourceTypeGeneral":"Dataset"}}}}'
# HTTP 201
```

Abbreviated `GET localhost:8065/dois/10.14454/dandi-smoke-001`:

```json
{"data":{"id":"10.14454/dandi-smoke-001","type":"dois","attributes":{
  "doi":"10.14454/dandi-smoke-001",
  "state":"findable",
  "url":"https://example.org/dandi-smoke",
  "titles":[{"title":"Smoke test"}],
  "publisher":"DANDI Archive",
  "types":{"resourceTypeGeneral":"Dataset","schemaOrg":"Dataset","citeproc":"dataset","bibtex":"misc","ris":"DATA"},
  "registered":null
}}}
```

`registered` is null because Handle registration is attempted after commit, fails against the dead host, and is only logged; the DOI stays findable.
</details>

<details><summary>Draft vs findable behaviour, as observed</summary>

| action | result |
|---|---|
| POST with `event: publish` | 201, state `findable` |
| PUT new title on a findable DOI | 200, still findable |
| DELETE a findable DOI | 405 Method not allowed |
| POST a draft, then DELETE it | 201, then 204 |

The 405 is the controller, not the model: findable DOIs cannot be deleted through the API at all, only hidden (`event: hide`, findable -> registered).
</details>

<details><summary>The DB is resettable</summary>

Full reset, about 25 s with images cached:

```bash
docker compose down -v && docker compose up -d
# wait for elasticsearch: healthy
docker compose exec web bundle exec rake elasticsearch:create_all_indexes RAILS_ENV=development
docker compose exec web bundle exec rake db:setup RAILS_ENV=development
```

In place, keeping seeds and indices (the model has no destroy guard, only the controller does):

```bash
docker compose exec web bundle exec rails runner 'Doi.where(aasm_state: "findable").find_each(&:destroy)'
```
</details>

I'm not planning to wire this into CI right now, but it's a workable dev setup for exercising the DOI lifecycle for real.

<sub>Investigation and write-up assisted by Claude Code.</sub>

PR #2800, opened 2026-04-30
Add REUSE.toml

yarikoptic:
Note that I kept top level `LICENSE` as well to ease discovery etc.  There is also a copy under web/ I didn't touch

Overall - value here is small since it is all nicely falls under 1 license/copyright statement ATM, but still might be nice for standardization etc. WDYT? no strong feelings

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,7 @@ repos:
   rev: v2.3.0
   hooks:
   - id: codespell
+- repo: https://github.com/fsfe/reuse-tool
+  rev: v6.2.0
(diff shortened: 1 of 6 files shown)


yarikoptic:
before I bother resolving conflicts, I would like to hear from @dandi/archive-maintainers  on either they have any objections. I think it would be valuable overall and unless I hear otherwise I will refresh/proceed within a week time or so

jjnesbitt:
My two cents is if we aren't encountering any issues with our current licensing scheme, I am loath to add extra dependencies and tooling to solve a problem we don't really have.

PR #2809, opened 2026-05-01
Get `fsspec` from `dandi[extras]`

brianhelba:
`fsspec[http]` was added by #1545 as a transitive dependency of the CLI. Enabling the `[extras]` extra causes `dandi` to depend on it automatically.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ dependencies = [
   "djangorestframework==3.17.1",
   "drf-extensions==0.8.0",
   "drf-yasg==1.21.15",
-  "fsspec[http]==2026.3.0",
   "jsonschema==4.26.0",
   "more_itertools==11.0.2",
   "psycopg[binary]==3.3.3",
@@ -58,7 +57,9 @@ development = [
   "watchdog==6.0.0",
   "werkzeug==3.1.8",
 ]
-cli = ["dandi==0.74.3"]
+cli = [
+  "dandi[extras]==0.74.3",
+]
 
 [dependency-groups]
 dev = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -58,7 +57,10 @@ development = [
   "watchdog==6.0.0",
   "werkzeug==3.1.8",
 ]
-cli = ["dandi==0.74.3"]
+cli = [
+  "dandi==0.74.3",
```
brianhelba: An alternative is to depend on `dandi[extras]`, which will pull in `fsspec[http]`, but also `duecredit`: https://github.com/dandi/dandi-cli/blob/2fc63ee6bf0f28094415baad4c7745cb652615e7/pyproject.toml#L95

brianhelba: Maybe this can just be fixed in the CLI?

jjnesbitt: What is there to fix in the CLI? Wouldn't we just change our dependency to `dandi[extras]==0.74.3`?

brianhelba: Just splitting out the superfluous (for us) dependency on `duecredit`.

jjnesbitt: I guess I don't see the inclusion of that package as a reason to not use the `extras` install group, although ideally the `fsspec` dependency would be put in a group that is more specific to why it's needed.

brianhelba:
@jjnesbitt Ok, I've updated the approach of this PR, per your feedback.

It's worth looking at `uv.lock`, just to see the new packages added, from the extra dependency on `duecredit`.

jjnesbitt (approved):

PR #2817, opened 2026-05-06
Update stats bar to show the total number of Dandisets that are not empty and unembargoed

kabilar:
Since the search bar is right above the stats bar, it implies that a user can search across that total number of Dandisets.

image

So here I propose only showing the total number of non-empty, unembargoed Dandisets in the stats bar.

### DIFF
--- a/web/src/views/HomeView/StatsBar.vue
+++ b/web/src/views/HomeView/StatsBar.vue
@@ -49,8 +49,11 @@ const stats = computed(() => [
 
 // equivalent of async created method in options API
 dandiRest.stats().then((data) => {
-  dandisets.value = data.dandiset_count;
   users.value = data.user_count;
   size.value = data.size;
+});
+
+dandiRest.dandisets({ page_size: 1, empty: false }).then((response) => {
+  dandisets.value = response.data.count;
 });
 </script>


bendichter:
@kabilar 

How about a compromise:

X public dandisets
(+Y embargoed dandisets)

and we only count non-empty Dandisets

kabilar:
> @kabilar
> 
> How about a compromise:
> 
> X public dandisets (+Y embargoed dandisets)
> 
> and we only count non-empty Dandisets

Thanks Ben.  I was thinking about that too.  The implementation is a bit more complicated.  The `dandisets/` endpoint requires a user to be logged in (and possibly need admin privileges?) to query the number of embargoed dandisets.  So we would need to update the `dandisets/` endpoint to not require a user to be logged in or update the `stats/` endpoint to provide this info.

bendichter:
@kabilar — that's true for `/api/dandisets/?embargoed=true` (which does raise `UnauthorizedEmbargoAccessError` for anonymous), but that endpoint isn't actually how the public count is sourced. There are two separate code paths:

**Public homepage count → `/api/stats/`** (anonymous, no auth). It returns a snapshot row populated by a scheduled celery task:

```python
# dandiapi/api/tasks/scheduled.py
@shared_task(soft_time_limit=60)
def compute_application_stats() -> None:
    ApplicationStats.objects.create(
        dandiset_count=Dandiset.objects.count(),
        published_dandiset_count=Dandiset.published_count(),
        ...
    )
```

That bare `Dandiset.objects.count()` runs server-side without any auth context, so it counts ALL dandisets including embargoed. The frontend's `web/src/views/HomeView/StatsBar.vue` reads it via `dandiRest.stats().then(d => dandisets.value = d.dandiset_count)`. No login required, no admin required.

**Listing endpoint → `/api/dandisets/`** is a separate request handler that *does* gate embargoed visibility on auth. That's what your comment was looking at, but it's not what feeds the homepage count.

So for the "X public (+Y embargoed)" breakdown, the cleanest path is to extend the existing snapshot rather than touch the `/dandisets/` endpoint:

1. Add `embargoed_dandiset_count` (and optionally `unembargoed_nonempty_dandiset_count`) fields to the `ApplicationStats` model — one migration.
2. Update `compute_application_stats` to populate them, e.g.:
   ```python
   embargoed_dandiset_count = Dandiset.objects.exclude(embargo_status='OPEN').count()
   ```
3. Update `StatsBar.vue` to render the breakdown.

No `/dandisets/` permission changes, no admin-only endpoint, and the count stays anonymous-friendly because it lives in the snapshot rather than a per-request query. Same place we already source `dandiset_count` from today.

kabilar:
Thanks @bendichter.  Yea, updating the `stats` endpoint is what I was thinking too.  I'll get back to this pull request (or feel free to take it over).

kabilar:
Closing for now.  Will reopen when I have a chance to make these changes.

PR #2820, opened 2026-05-07
Wider, responsive advanced-search help popover (closes #2818)

### LINKED ISSUE #2818: Make "Advanced search operators" pop-up wider!
There is no reason for it to be this fixed width narrow AFAIK -- should be dynamic to the window, potentially occupying like 80% of the width 

Image

bendichter:
Closes #2818. Stacked on top of #2814 — until that merges, this PR's diff includes the parent's changes; afterward GitHub will rebase and show only the popover diff.

> **Prior conversation:** this PR was originally opened against the parent branch on the fork at [bendichter/dandi-archive#2](https://github.com/bendichter/dandi-archive/pull/2) (now closed) — see that thread for context.

The popover was hard-capped at `max-width: 420px`, which forced the description column to wrap awkwardly even when there was plenty of viewport room (per yarikoptic's screenshot in #2818).

## Change

Replace the fixed cap with responsive sizing:

```css
min-width: 420px;
max-width: min(80vw, 720px);
```

- `min-width: 420px` keeps the popover from collapsing on a single short example.
- `max-width: 80vw` lets it grow with the window — up to 80% of viewport, as requested in #2818.
- `max-width: 720px` caps the absolute width on very wide monitors so the popover doesn't sprawl.

Also pinned the operator-example column with `white-space: nowrap` so strings like `created_after:2024-01-01` and `technique:"patch clamp"` stay readable as code instead of wrapping mid-token.

## Test plan

- [x] `npm run lint -- src/components/DandisetSearchField.vue` passes
- [x] Manually opened the popover at narrow (~500px) and wide (~2000px) viewports — fits content cleanly at both, no horizontal scroll, examples don't wrap.

### DIFF
--- a/web/src/components/DandisetSearchField.vue
+++ b/web/src/components/DandisetSearchField.vue
@@ -35,8 +35,7 @@
             </v-icon>
           </template>
           <v-card
-            max-width="420"
-            class="pa-3"
+            class="pa-3 advanced-search-help"
           >
             <div class="text-subtitle-2 mb-2">
               Advanced search operators
@@ -51,7 +50,9 @@
                   v-for="op in operatorHelp"
                   :key="op.example"
                 >
-                  <td><code>{{ op.example }}</code></td>
+                  <td class="example-cell">
+                    <code>{{ op.example }}</code>
+                  </td>
                   <td class="text-caption">
                     {{ op.description }}
                   </td>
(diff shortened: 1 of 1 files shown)


bendichter:
much nicer looking

old:

image


new:

image

yarikoptic (approved):
tried, works.

diff looks sane

PR #2821, opened 2026-05-08
Add owner: operator to advanced search

bendichter:
Adds an **`owner:`** operator to the advanced-search syntax (which landed in #2814).

`owner:VALUE` returns dandisets owned by users matching `VALUE`. The lookup is case-insensitive against `User.username`, `User.email`, `User.first_name`, `User.last_name`, or `"first_name last_name"` (so the display name shown in the dandiset list also works). Multiple users may match (common with last-name lookups); we union dandisets owned by any of them. Unknown user → 0 results, not 400.

```
owner:alice
owner:alice@example.com
owner:Smith              # any user named Smith
owner:"Jane Doe"         # full display name
```

> The unquoted `owner:me` magic alias for "the current user" was discussed and **deferred to a follow-up PR**. Reasons in the [thread on the contributor PR](https://github.com/dandi/dandi-archive/pull/2822#discussion_r3213508174). For now, `owner:me` is treated as a literal lookup for users named "Me" (matches none in practice).

<details><summary><b>Implementation notes</b></summary>

- `_apply_owner_filter` in `dandiapi/api/services/search/filters.py` does the lookup. Direct query against `DandisetUserObjectPermission` (rather than guardian's `get_objects_for_user`) so we can union across multiple matched users in a single query and bypass the superuser-sees-everything default — `owner:admin` returns only what admin explicitly owns, not the entire archive.
- OpenAPI help text on `DandisetQueryParameterSerializer.search` lists the operator alongside the others.
- Frontend popover entry: `owner:"Jane Doe"`.

</details>

<details><summary><b>Test plan</b></summary>

- [x] `tox -e lint`
- [x] `tox -e test -- dandiapi/api/tests/test_dandiset.py -k "advanced_search_owner"` — three consolidated tests cover username/email/case-insensitivity/full-name/multi-user-union/unknown-user/superuser-non-inflation/cross-key combination.
- [x] Manual: `curl 'http://localhost:8000/api/dandisets/?search=owner:%22Jane%20Doe%22'`.

</details>


dandiapi/api/tests/test_dandiset.py
```diff
@@ -2086,3 +2086,150 @@ def test_advanced_search_species_respects_embargo_visibility(api_client):
...
+
+
+@pytest.mark.ai_generated
+@pytest.mark.django_db
+def test_advanced_search_owner_by_email_matches(api_client):
+    alice = UserFactory.create(username='alice', email='alice@example.com')
+    alice_ds = DandisetFactory.create(owners=[alice])
+    DraftVersionFactory.create(dandiset=alice_ds)
+
+    assert _search_ids(api_client, 'owner:alice@example.com') == {alice_ds.identifier}
```
yarikoptic: could you please ask your agent to remember to make "each test matter more" kinda... I will try to figure out where/how I did it to mine later -- e.g. these two tests could be 1, and likely others could also be folded in where more testing done on the same "setup" thus making overall testing faster and without loss of any coverage.  I do understand that there is a tradeoff (individual aspect tests might be easier to debug etc) but still -- my Allergy to duplication is triggering AIlergy too

dandiapi/api/tests/test_dandiset.py
```diff
@@ -2086,3 +2086,150 @@ def test_advanced_search_species_respects_embargo_visibility(api_client):
...
+
+@pytest.mark.ai_generated
+@pytest.mark.django_db
+def test_advanced_search_owner_by_full_name_matches(api_client):
+    # The dandiset list shows owners by display name (first + ' ' + last).
+    # `owner:"Super User"` should match a user with that full name even
+    # though their username is an email address.
+    user = UserFactory.create(
+        username='ben.dichter@gmail.com',
+        email='ben.dichter@gmail.com',
```
yarikoptic: might better to use some `alice.m.nobody@example.com` from above instead of leaking personal real emails

dandiapi/api/services/search/filters.py
```diff
@@ -104,6 +109,52 @@ def _apply_asset_filter(queryset, operator: str, value: str):
...
+
+    Multiple users may match (common when only a first or last name is given);
+    we union dandisets owned by any of them. Unknown user → empty result (not
+    an error — a search for a nonexistent owner is a valid 0-hit query).
+
+    Direct query against `DandisetUserObjectPermission` rather than guardian's
+    `get_objects_for_user` so we can intersect across multiple matched users
+    in a single query, and to bypass the superuser-sees-everything default.
+    """
+    if value.lower() == 'me':
```
yarikoptic: now I started to "feel" for the unlikely `Me Someoneyou` being unable to look himself without authorization... I wonder if we could make it more explicit, e.g. `owner:%me%` or `owner:me:` or alike -- any other system like github or google having smth like that?  could also be `owner-me:` keyword . This should make it explicit and it is better than implicit

yarikoptic (changes requested):
Initial HI review. Overall -- love it.

dandiapi/api/tests/test_dandiset.py
bendichter: oh, I generally prefer more modular tests, but yes I can adjust for DANDI projects

dandiapi/api/tests/test_dandiset.py
bendichter: haha good point

dandiapi/api/services/search/filters.py
bendichter: `owner:"me"` should search for the name "Me". I can test that. Also, if you wanted to search for an owner named "Me Someoneyou", you could use `owner:"Me Someoneyou"`.

dandiapi/api/tests/test_dandiset.py
bendichter: Done in 8a65f3b3 — consolidated the 10 owner tests down to 3 denser ones (lookup-paths-and-combinations, me-magic-and-literal-escape, superuser-non-inflation). Same coverage; DB setup runs ~3x instead of ~10x. Saved a memory so my agent prefers the denser style for DANDI PRs going forward.

dandiapi/api/tests/test_dandiset.py
bendichter: Fixed in 8a65f3b3 — switched the fixture user to a generic Super/User name, no personal email.

dandiapi/api/services/search/filters.py
bendichter: Done in 8a65f3b3. Addressed via a parser-level quoted/unquoted distinction:
- `owner:me` (unquoted) → magic alias for the requesting user (anonymous → 400 with a hint pointing at the quoted form)
- `owner:"me"` (quoted) → bypasses the magic and matches the literal user named Me
- `owner:"Me Someoneyou"` → matches the full-name display

Implementation: `ParsedSearch.operators` is now a list of `Operator` dataclasses (`key`, `value`, `quoted`) instead of bare 2-tuples; only the `owner` filter inspects `quoted` for now, but the same signal could be used by other operators that ever want a magic-alias escape hatch. New integration test `test_advanced_search_owner_me_magic_and_literal_escape` pins all three cases plus the anonymous-400 path.

GitHub uses `@me` for the same idea, but that's because `@` is GitHub's user sigil — we don't have that convention in DANDI, so the `@` would just be an arbitrary literal here. The quoted-form escape feels more in line with the Gmail-style operator syntax we already have (which uses `"foo:bar"` as the colon escape).

yarikoptic (approved):
conditional approval on keeping it more to the point as to remove that `Operator` which seems to be a "new refactoring " -- would be better as a separate PR for easier digestion.

mvandenburgh (approved):
I resolved the merge conflicts and pushed a correctness fix. 

I'd like to evaluate how this search service is structured at a high level and maybe refactor things a bit, but this PR follows the existing conventions and structure so I can take a pass on that after it and others are merged

PR #2826, opened 2026-05-13
Operator-autocomplete dropdown for advanced search (stacked on #2822)

bendichter:
**Draft.** Stacked on #2822.

GitHub-issues-style autocomplete for the dandiset list search box. Click in → dropdown of every available operator. Type → list narrows to operators whose names match the prefix at the cursor. Click / Enter / Tab → inserts `name:` at the cursor with the caret placed right after the colon, ready for the value.

| Behavior | Trigger |
|---|---|
| Open dropdown | Focus the input |
| Filter list | Typing (prefix-matched against the token under the cursor) |
| Select & insert `name:` | Click, Enter, or Tab |
| Move highlight | ↑ / ↓ |
| Dismiss | Esc, blur, or typing past a `:` |

The token-at-cursor uses the same whitespace-bounded definition the backend parser uses, so what the user sees in the dropdown matches what the parser will recognize. Once the user types `:`, autocomplete is suppressed — this PR doesn't try to suggest *values*, only *operator keys*.

## Files

- **`web/src/components/advancedSearchOperators.ts`** (new) — canonical operator catalog (name, description, value example) plus `tokenAtCursor()` and `suggestionsFor()` helpers. Mirrors the backend allowlist; backend stays the source of truth (it validates and returns "Did you mean?" suggestions for unknown keys).
- **`web/src/components/DandisetSearchField.vue`** — adds a `v-menu` anchored to the wrapping form, controlled manually via `autocompleteOpen`. Keyboard handlers on the input drive selection + completion. The existing `?` help popover is unchanged; the two are complementary (popover = static cheat-sheet, autocomplete = interactive).

## Test plan

- [x] `npm run lint` clean
- [x] `vue-tsc --noEmit` clean
- [ ] Manual: click into the search box on the dandiset list; the dropdown appears with all operators. Typing `aut` narrows to `author:`. Pressing Enter inserts `author:` with the caret after the colon. Typing the value and submitting performs the search as before.

## Known limits / future

- **No value autocomplete yet.** Once the user types `key:`, the dropdown closes. The discovery-API idea @yarikoptic suggested in #2822 (`?search=species:?` returning candidate values) would unlock this — but that's a separate backend addition.
- **Operator catalog is duplicated** between the frontend (`advancedSearchOperators.ts`) and backend (`operators.py`). The discovery API would also let the frontend fetch this dynamically. For now, the duplication is intentional: keeps the autocomplete fast (no network round-trip per keystroke) and the backend remains the validation source of truth.
- **No quote-balancing assist.** Typing `technique:"` doesn't auto-insert the closing quote (could be a small future polish).

Once #2822 merges, this PR's base will collapse to master automatically.


bendichter:
https://github.com/user-attachments/assets/0f8dea7c-eda0-4122-8b7b-33606cc70c50

bendichter:
@yarikoptic

bendichter:
**Update — value-mode autocomplete (5920f05b)**

The dropdown now switches into value mode when the caret moves past `:` for any operator with a known vocabulary:

| Operator | Source |
|---|---|
| `species:` | live fetch from `/search/species?species=PREFIX` (debounced 200 ms, sequence-checked so a slow earlier reply can't overwrite a faster later one) |
| `approach:` | static list (4 values from real metadata) |
| `technique:` | static list (~11 common measurement techniques) |
| `standard:` | static list (NWB, BIDS, NIfTI) |
| `file_type:` | static list (nwb, image, text, video) |

Multi-word suggestions get auto-quoted on insert (`technique:"spike sorting technique"`). All other operators (dates, contributors, owner, affiliation) close the dropdown after the colon — same as before.

UX flow now reads end-to-end without the user ever typing past a colon by hand: pick `species` from the key dropdown → caret lands after `:` → value dropdown appears immediately with live matches → click → search.

bendichter:
https://github.com/user-attachments/assets/431b7625-05a5-4095-9ed3-6e0a6d52ee37

bendichter:
I know you are going to complain about a lot of this being hard-coded. I agree, but I just wanted to get the UX down first

yarikoptic:
Thanks for taking it on. Let's finalize that https://github.com/dandi/dandi-archive/pull/2822 to make this reviewable

PR #2684, opened 2026-01-02
All backup options design doc

CodyCBakerPhD:
A complete summary of all researched options for backup

Quick link: [full document](https://github.com/CodyCBakerPhD/dandi-archive/blob/all_backup/doc/design/all_backup_options.md)

Quick link: [(cost summary table)](https://github.com/CodyCBakerPhD/dandi-archive/blob/all_backup/doc/design/all_backup_options.md#cost-summary)

PR #2830, opened 2026-05-18
Expand S3 doc with all options and archival note

CodyCBakerPhD:
Updated the S3 backup documentation to include detailed sections on proposed solutions, requirements, advantages, limitations, and cost analysis for various backup strategies.

Rebase of https://github.com/dandi/dandi-archive/pull/2684 from my fork, now going to the archive destination with the reference note


kabilar (approved):
This was moved to `archive` since we will not be implementing a backup solution.

PR #2838, opened 2026-05-29
Make `Related resource` titles clickable links on Dandiset landing page

kabilar:
## Context

I have been on calls where users who were navigating the DLP didn't see the link icon that is off to the right of the resource.

image

## Changes

Here I propose making the related resource titles clickable links that open in a new tab and removing the link icon.

cc @satra @ayendiki

### DIFF
--- a/web/src/components/DLP/MetadataCard.vue
+++ b/web/src/components/DLP/MetadataCard.vue
@@ -35,10 +35,15 @@
                 cols="9"
                 class="text-grey-darken-3"
               >
-                <span v-if="item.name || item.identifier || item.id">
-                  {{ item.name || item.identifier || item.id }}
-                  <br>
-                </span>
+                <slot
+                  name="itemTitle"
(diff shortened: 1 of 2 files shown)


kabilar:
Preview that displays a resource with a URL and one without a URL:

https://deploy-preview-2838--sandbox-dandiarchive-org.netlify.app/dandiset/211293?page=1&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=true&pos=1

bendichter:
I like this idea. My only minor request is that I don't really like the default styling for links and I think we should follow the styling in https://docs.dandiarchive.org/

bendichter:
example: https://deploy-preview-2838--sandbox-dandiarchive-org.netlify.app/dandiset/218045?page=2&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=15

image

LGTM!

PR #2841, opened 2026-06-01
Fix search within "My Dandisets" to include owned/embargoed dandisets

### LINKED ISSUE #2834: UI: It's not possible to search within "My dandisets"
Clicking "My Dandisets" goes to: https://dandiarchive.org/dandiset/my and shows embargoed dandisets and a search bar. The "My Dandisets" is highlighted.

Image

When I use the search bar, it goes to e.g. https://dandiarchive.org/dandiset/search?search=Progressive+Mouse+Model

The highlight says on "My Dandisets", so it looks like I am searching across only my dandisets, but the search results are only for public dandisets, omitting all embargoed dandisets that I own.

Image

The search results for the "My Dandisets" page are the exact same as for the public page:

Image

Image

This creates two major problems:
1. It appears to the user that they are searching across dandisets they own, but the results are for public dandisets
2. A user is completely unable to conduct a search that includes embargoed dandisets they own

bendichter:
## Summary

Fixes #2834.

Searching from the **My Dandisets** page silently fell through to the public search route, so results only ever showed public dandisets — the user's embargoed dandisets were omitted, and there was no way to search across them.

The backend already supports `user=me` + `search` + `embargoed=true` together (the dandiset `list` action runs `get_queryset()`, honoring `user=me`/`embargoed`, then applies the search filter). The bug was entirely in the frontend.

## Changes

- **`DandisetSearchField.vue`** — `performSearch` now stays on the current route for the `myDandisets` page (previously only `searchDandisets`), instead of pushing to the public search route. This keeps the search scoped to the user's own dandisets.
- **`DandisetsPage.vue`** — added a `searchActive` computed (`props.search || !!route.query.search`). The `search` param is now sent whenever a search is active, not only on the dedicated public search view. Combined with the existing `user: 'me'` and `embargoed: true` params on this page, search now spans owned + embargoed dandisets. The "N results found" header and inline search-error alert use the same flag.
- **`MyDandisetsView.vue`** — empty state now shows "No results found in your dandisets." during a search, instead of the misleading "You haven't created any dandisets yet."

## Testing

`yarn lint` and `yarn type-check` pass for the changed files. Opening as a **draft** to test signed-in behavior on the deploy preview, since signing in locally is difficult.

To verify:
1. Sign in, go to **My Dandisets**.
2. Search for a term that matches an embargoed dandiset you own.
3. Confirm the URL stays on `/dandiset/my?search=...`, the embargoed dandiset appears, and the result count is shown.

### DIFF
--- a/web/src/components/DandisetSearchField.vue
+++ b/web/src/components/DandisetSearchField.vue
@@ -108,14 +108,9 @@ function performSearch(evt: Event) {
     // nothing has changed, do nothing
     return;
   }
-  if (route.name !== 'searchDandisets') {
-    router.push({
-      name: 'searchDandisets',
-      query: {
-        search: currentSearch.value,
-      },
-    });
-  } else {
+  if (route.name === 'searchDandisets' || route.name === 'myDandisets') {
+    // Stay on the current page so the search keeps its context. On "My Dandisets"
+    // this scopes the search to the user's own dandisets (including embargoed).
     router.replace({
       ...route,
       query: {
@@ -123,6 +118,13 @@ function performSearch(evt: Event) {
         search: currentSearch.value,
       },
     } as RouteLocationRaw);
+  } else {
+    router.push({
+      name: 'searchDandisets',
+      query: {
+        search: currentSearch.value,
+      },
+    });
   }
 }
 </script>
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -142,7 +142,7 @@
       </v-btn>
     </v-toolbar>
     <v-alert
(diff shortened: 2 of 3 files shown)


bendichter:
I tested this and can confirm that it works. It correctly filters to only dandisets I own, and it includes embargoed dandisets. Anyone else's results will vary because it depends on what dandisets you as a user own

yarikoptic:
FWIW, tried on preview - worked for me. Filed 
- https://github.com/dandi/dandi-archive/issues/2845

since ideally we get it all green before merging

yarikoptic (approved):

PR #2843, opened 2026-06-01
Add search operator autocomplete dropdown (#2828)

### LINKED ISSUE #2828: search operators dropdown (UX)
We have added a few GitHub-style operators to the dandi archive search bar, e.g. species:, contributors: created_after:. The problem now is that these operators are not very discoverable. A user might not know they exist or which ones are available.

To solve this problem, we want to create a dropdown that appears under the search bar and recommends operators. It should show all of the available operators, and should filter based on string matching as the user types. Clicking on an operator should select it. up/down/tab should also work to select, which should add that operator to the search bar text.

Similar to github.com/issues:

Image

bendichter:
## Summary

Closes #2828.

The search bar supports GitHub-style operators (`species:`, `approach:`, `created_after:`, etc.), but they weren't discoverable — a user had no way to know they exist or which are available.

This adds an autocomplete dropdown beneath the search field that lists the operators and filters them as you type, similar to github.com/issues.

## Behavior

- **Discoverable:** focusing the field shows the full operator list with descriptions.
- **Filters as you type:** typing an operator prefix (e.g. `spe`) narrows to matching operators (`species:`), matched against the token under the cursor — so operators work mid-query too (`neuropixels spe…`).
- **Keyboard:** ↑/↓ to move the highlight, **Enter** or **Tab** to insert the highlighted operator (keeps focus so you can type the value), **Esc** to dismiss. While browsing the full list (empty token) nothing is pre-highlighted, so **Enter still submits** the search.
- **Mouse:** click an operator to insert it.

## Implementation notes

- Built into the existing `DandisetSearchField.vue` (the search field actually in use). Suggestions are derived from the existing `operatorHelp` array, so the help-icon table and the dropdown share a single source of truth.
- The dropdown is teleported to `<body>` and positioned manually beneath the input, because the toolbar uses `overflow: hidden` and would otherwise clip it; it tracks scroll/resize. Focus stays in the input (list items use `mousedown.prevent`) so keyboard interaction is uninterrupted.

## Notes

- The operator-parsing **backend is already on `master`** and live on the sandbox API; this is a frontend-only change.
- Draft pending review of the UX. Lint and type-check pass.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -73,6 +73,10 @@ jobs:
       # API server env vars
       DJANGO_SETTINGS_MODULE: dandiapi.settings.development
       DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
+      # Close DB connections per-request so the threaded `runserver` doesn't
+      # accumulate connections under parallel Playwright load and exhaust
+      # Postgres's max_connections ("sorry, too many clients already").
+      DJANGO_CONN_MAX_AGE: "0"
       DJANGO_CELERY_BROKER_URL: amqp://localhost:5672/
       DJANGO_MINIO_STORAGE_URL: http://minioAccessKey:minioSecretKey@localhost:9000/django-storage-testing
       DJANGO_DANDI_WEB_APP_URL: http://localhost:8085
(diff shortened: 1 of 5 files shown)


bendichter:
image


image

yarikoptic:
I think it begs for at least some basic tests added to e2e or frontend alone. Note -- update to `master` branch should address CI timeout (#2849 fixed that)

yarikoptic (approved):
As frontend only, seems to work, and  actually quite short, I think we better proceed merging it and giving it real testing!

PR #2844, opened 2026-06-01
Remove `standard:` search operator

bendichter:
## Summary

The `standard:` search operator (e.g. `standard:NWB`) never returned results against real data, and it isn't providing enough value to justify keeping a half-working operator around. **This PR removes `standard:` entirely.** It can be re-added later, wired to the correct metadata source.

### Why it was broken

`standard:` matched `$.dataStandard[*].name` against **asset** metadata, but assets don't carry a `dataStandard` field — the data standard is reported on the **Version's** `assetsSummary.dataStandard`, so the jsonpath never matched.

Verified against the sandbox API:
- `standard:nwb` / `standard:NWB` → **0 results**
- (sibling operators that *do* live on assets work, e.g. `species:mouse` → 134, `technique:two-photon` → 12)

Rather than relocate it to match the version metadata, we're dropping it for now since it's low-value, and re-introducing it cleanly later is straightforward.

## Changes

- `parser.py` — removed `standard` from `OPERATOR_KEYS`; an unknown `standard:` token now falls through to the standard 400 + "did you mean?" handling.
- `filters.py` — removed `standard` from `_ASSET_OPS`/`_NAME_PATH_OPS`, and reverted `_jsonpath_name_match` to its single-column form (the column generalization existed only to serve `standard`).
- `serializers.py` — dropped `standard` from the API search help text.
- `DandisetSearchField.vue` — removed the `standard:nwb` example from the frontend operator hints.
- `test_dandiset.py` — removed `test_advanced_search_standard_matches` and the now-unused `version_metadata` param on the `_seed_dandiset_with_asset` helper.

## Testing

`ruff check` passes on the changed files. The remaining search operator tests (`species`/`approach`/`technique`/`file_type` + the unknown-operator suggestion path) are unaffected.

### DIFF
--- a/dandiapi/api/services/search/filters.py
+++ b/dandiapi/api/services/search/filters.py
@@ -60,34 +60,29 @@ def _annotate_latest_published_created(queryset):
 
 
 # Each entry maps an operator to a Postgres jsonpath that selects the names
-# we want to match against, within an asset's metadata. `[*]` wildcards mean
-# we match if ANY element of the array satisfies the predicate — important for
-# assets that list multiple species, approaches, etc. Paths MUST be trusted
-# constants: they're interpolated into the SQL.
+# we want to match against. `[*]` wildcards mean we match if ANY element of
+# the array satisfies the predicate — important for assets that list
+# multiple species, approaches, etc. Paths MUST be trusted constants:
+# they're interpolated into the SQL.
 _NAME_PATH_OPS = {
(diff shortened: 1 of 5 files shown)


yarikoptic:
rebase/merge -- I fixed up frontend e2e testing

yarikoptic (approved):
per request during call -- let's just indeed remove for now unless you also find a fix quickly ;-)

Issue #2845, opened 2026-06-01
Frontend CI started to fail


yarikoptic:
Last succesfull https://github.com/dandi/dandi-archive/actions/workflows/frontend-ci.yml?query=is%3Asuccess

<img width="2710" height="608" alt="Image" src="https://github.com/user-attachments/assets/767027b4-f35e-451f-9695-4ee9abe11786" />

attn @bendichter since most active contributor atm and your PRs affected...

e.g. in https://github.com/dandi/dandi-archive/actions/runs/26763333978/job/78882721140?pr=2841 seems to be stalled with

```
Downloading Chromium 134.0.6998.35 (playwright build v1161) from https://cdn.playwright.dev/dbazure/download/playwright/builds/chromium/1161/chromium-linux.zip
|                                                                                |   0% of 164.9 MiB
|■■■■■■■■                                                                        |  10% of 164.9 MiB
|■■■■■■■■■■■■■■■■                                                                |  20% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■                                                        |  30% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■                                                |  40% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■                                        |  50% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■                                |  60% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■                        |  70% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■                |  80% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■        |  90% of 164.9 MiB
|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■| 100% of 164.9 MiB
```

PR #2847, opened 2026-06-02
ci: pre-fetch Chromium with curl to fix E2E install stall (#2845)

yarikoptic:
## Summary

Attempt 3 of 3 (see #2845) to unstick frontend E2E CI, which has been stalling at `npx playwright install --with-deps` ever since the GitHub-hosted `ubuntu-24.04` runner image moved from `20260518.149.1` to `20260525.161.1` on 2026-05-28/29.

`curl` from the same runner does **not** exhibit the socket-close-after-EOF hang. Pre-fetch the chromium and chromium-headless-shell zips (build v1161, matching the `@playwright/test` 1.51.1 pin) into `~/.cache/ms-playwright/` and write the `INSTALLATION_COMPLETE` marker so playwright sees them as already installed. The subsequent `playwright install --with-deps chromium` then runs apt only — no broken HTTP fetch.

> [!NOTE]
> The hardcoded `1161` build number is tied to `@playwright/test` 1.51.1. If/when playwright is bumped, this PR's build number will need to be bumped alongside it (or removed if the playwright-bump PR lands and resolves the underlying hang).

See #2845 for the full root-cause analysis (runner image bump from `20260518.149.1` -> `20260525.161.1` triggered the hang).

## Test plan

- [ ] CI on this PR: `E2E tests` job no longer hangs at `Install Playwright Browsers`.
- [ ] Logs show "Pre-fetch Chromium" succeeds in seconds and `playwright install` is apt-only.
- [ ] If the playwright-bump and/or container PR also fix the hang, prefer those — this PR's hardcoded `1161` is brittle.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -115,8 +115,32 @@ jobs:
       run: npm ci
       working-directory: e2e
 
+    - name: Pre-fetch Chromium with curl (workaround for #2845)
+      # `npx playwright install` chromium download from cdn.playwright.dev hangs
+      # after 100 % on ubuntu-24.04 runner image >= 20260525.161.1. curl does
+      # not exhibit the same socket-close-after-EOF bug, so we pre-seed
+      # playwright's cache and let `playwright install --with-deps` only run
+      # apt (it sees the browsers already there and skips the download).
+      run: |
+          set -eu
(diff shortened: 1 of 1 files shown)


yarikoptic:
didn't help! went with smaller #2849

PR #2852, opened 2026-06-10
Add WebDAV as an external service in the file browser

kabilar:
cc @satra @ayendiki @balbasty


yarikoptic:
Insofar decided to not bother promoting webdav we have more and potentially sunset altogether, so I will just close this issue

PR #2853, opened 2026-06-11
Guard against empty `loc` in `_encode_pydantic_error`

### LINKED ISSUE #2851: `_encode_pydantic_error` raises IndexError on validation errors with empty `loc`
## Summary

`_encode_pydantic_error` (in `dandiapi/api/services/metadata/__init__.py`) accesses `error['loc'][0]`, which raises `IndexError` whenever a Pydantic validation error has an **empty `loc` tuple**. Pydantic errors raised by a top-level `@model_validator` have `loc == ()`, so such an error crashes `_collect_validation_errors` — and because that happens inside the `except` handler of the validation flow, the exception is unhandled and the validation task fails instead of recording a clean validation error.

```python
def _encode_pydantic_error(error) -> dict[str, str]:
    return {'field': error['loc'][0], 'message': error['msg']}   # IndexError when loc == ()
```

This is latent today (the current `dandischema` models have no top-level model validators on `Dandiset`/`Asset`), but it will be triggered by an upcoming `dandischema` change that consolidates the publish-only validation into a top-level `@model_validator`. The most common path is the publishability check on an empty/zero-byte draft.

## Suggested fix

Handle an empty `loc` the same way `_encode_jsonschema_error` already handles an empty `path`:

```python
def _encode_pydantic_error(error) -> dict[str, str]:
    loc = error['loc']
    return {'field': str(loc[0]) if loc else '', 'message': error['msg']}
```

This is a correct, low-risk robustness fix independent of the `dandischema` change; it should be made regardless.

<details>
<summary>Why model-validator errors have an empty <code>loc</code> (reproduction)</summary>

A `@model_validator(mode="after")` that raises `ValueError` (or `PydanticCustomError`) attaches the error to the model root, i.e. `loc == ()`. Field validators, by contrast, produce `loc == ('<field>',)`, and validators on nested objects produce a non-empty nested path — which is why this has not been hit so far.

```python
from pydantic import BaseModel, model_validator

class M(BaseModel):
    x: int = 0
    @model_validator(mode="after")
    def check(self):
        raise ValueError("boom")

try:
    M()
except Exception as e:
    print(e.errors())
# [{'type': 'value_error', 'loc': (), 'msg': 'Value error, boom', 'input': ...}]
```

So `error['loc'][0]` raises `IndexError` for that error.
</details>

<details>
<summary>When it fires in the validation flow</summary>

`validate_version_metadata` builds publishable metadata (injecting `datePublished`, `publishedBy`, and a computed `assetsSummary`) and validates it as `PublishedDandiset`. After the upcoming `dandischema` consolidation, the "a Dandiset containing no files or zero bytes is not publishable" check runs in a top-level model validator. For a draft with no (non-empty) assets, `assetsSummary` is `{numberOfBytes: 0, numberOfFiles: 0}`, the check raises with `loc == ()`, and:

```python
try:
    validate(..., schema_key='PublishedDandiset', json_validation=True)
except dandischema.exceptions.ValidationError as e:
    return (Version.Status.INVALID, _collect_validation_errors(e))  # IndexError raised here
```

`_collect_validation_errors` calls `_encode_pydantic_error` → `error['loc'][0]` → `IndexError`, raised inside the `except` block, so it propagates unhandled and the validation task fails (rather than the version being marked `INVALID` with a clear message). Previously this same case produced a tidy error with `field='assetsSummary'`.
</details>

<details>
<summary>Pydantic <code>ErrorDetails</code> shape — what is / isn't safe to index (sourced)</summary>

From the `ErrorDetails` TypedDict in `pydantic_core/__init__.py` (verified against pydantic-core 2.27.2 and a newer 2.x):

```python
class ErrorDetails(TypedDict):
    type:  str                      # required
    loc:   tuple[int | str, ...]    # required; an empty tuple is a valid value
    msg:   str                      # required
    input: Any                      # required
    ctx:   NotRequired[dict]        # optional
    # newer pydantic-core also adds:  url: NotRequired[str]   (absent when a PydanticCustomError is used)
```

So `type`/`loc`/`msg`/`input` are always present; `ctx` (and, where it exists, `url`) are not — i.e. the key set itself is version-dependent. The relevant point here: `error['loc']` is always present but `loc` may be empty, so `error['loc'][0]` is the unsafe access; `error['msg']` is safe. (`dandi-cli`'s equivalent code already handles this — it uses `.get("loc", [])` and `"+".join(...)`.)
</details>

candleindark:
Guards `_encode_pydantic_error` against an empty `loc` tuple (which a top-level `@model_validator` produces), mirroring how `_encode_jsonschema_error` already tolerates an empty path. Without this, `error['loc'][0]` raises `IndexError` inside the validation `except` handler and the validation task fails instead of recording a clean error.

This applies the fix proposed in #2851; see that issue for the full analysis, reproduction, and `ErrorDetails` background.

Closes #2851.

## Test plan

No new test ships here. The bug is unreachable under the pinned `dandischema==0.12.1` (no top-level model validators on `Dandiset`/`Asset`), so a unit test would have to hand-build the empty-`loc` error dict the encoder consumes, exercising no real Pydantic/dandischema behavior. The genuine coverage already exists in waiting: `test_validate_version_metadata_no_assets` and `test_validate_version_metadata_empty_zarr_asset` drive the zero-byte / no-file path and will exercise the real `loc == ()` error through this encoder once the dandischema consolidation lands. Updating their expected `field` at that point is the real regression test (tracked in #2854).

- [x] `pre-commit` hooks pass
- [x] CI: existing validation suite passes unchanged

### DIFF
--- a/dandiapi/api/services/metadata/__init__.py
+++ b/dandiapi/api/services/metadata/__init__.py
@@ -26,7 +26,11 @@ logger = get_task_logger(__name__)
 
 
 def _encode_pydantic_error(error) -> dict[str, str]:
-    return {'field': error['loc'][0], 'message': error['msg']}
+    # `loc` may be an empty tuple, e.g. for an error raised by a top-level
+    # `@model_validator`, which attaches the error to the model root. Mirror how
+    # `_encode_jsonschema_error` tolerates an empty path so we don't IndexError.
+    loc = error['loc']
+    return {'field': str(loc[0]) if loc else '', 'message': error['msg']}
 
 
 def _encode_jsonschema_error(error: jsonschema.exceptions.ValidationError) -> dict[str, str]:


yarikoptic (approved):
unguarded indexing is always a bad practice.  Let's proceed

PR #2855, opened 2026-06-16
Add in-browser zip download for small dandisets

bendichter:
## Motivation

Requiring the DANDI CLI to download a dandiset is **overkill for small datasets**. The current Download menu only hands the user a `dandi download ...` command — there's no way to grab a small dandiset straight from the browser.

This adds a **Download .zip** button to the dandiset Download menu that bundles the dataset client-side and saves it through the normal browser download mechanism.

## How it works

- Fetches each asset's blob **directly from S3** (the bucket already serves `Access-Control-Allow-Origin: *`, so browser `fetch` works) and zips them client-side with [`client-zip`](https://github.com/Touffy/client-zip) (~5 KB).
- **No backend changes and no server bandwidth cost** — bytes go S3 → browser, preserving the existing S3-direct model. The button is purely additive; the CLI flow is untouched.
- Files nest under a single root folder matching the zip filename (e.g. `000123-draft/sub-01/...`).
- **Byte-level progress bar** and a **Cancel** button (backed by an `AbortController` that interrupts both the asset listing and in-flight fetches).
- **Zarr assets are skipped** (they're not single blobs) with a user-facing note pointing to the CLI.

## Gating

Shown only for "small" versions: **≤ 2 GiB** and **≤ 1000 files** (constants at the top of the component). The archive is buffered in memory (`client-zip` → `Blob`) before saving, which is why the cap exists.

Video of sandbox testing below

## Status / open questions
- **Memory cap**: the in-memory buffer is the reason for the 2 GiB limit. A follow-up could stream to disk via the File System Access API (`createWritable()`) to lift the cap.
- **Thresholds** (2 GiB / 1000 files) are guesses — feedback welcome.

### DIFF
--- a/web/src/views/DandisetLandingView/DownloadDialog.vue
+++ b/web/src/views/DandisetLandingView/DownloadDialog.vue
@@ -52,6 +52,59 @@
         </v-tooltip>
       </v-card-title>
       <v-list class="pa-0">
+        <!-- PROTOTYPE: in-browser zip download for small dandisets -->
+        <template v-if="browserZipEligible">
+          <v-list-item density="compact">
+            Download directly in your browser
+            ({{ filesize(currentDandiset?.size ?? 0, { round: 1, base: 10, standard: 'si' }) }})
+          </v-list-item>
+          <template v-if="zipInProgress">
+            <v-list-item density="compact">
+              <v-progress-linear
+                :model-value="zipProgress"
+                :indeterminate="zipProgress === 0"
+                color="primary"
+                height="20"
+                rounded
+              >
+                <span class="text-caption">{{ Math.round(zipProgress) }}%</span>
+              </v-progress-linear>
+            </v-list-item>
+            <v-list-item density="compact">
+              <v-btn
+                color="error"
+                variant="outlined"
+                block
+                prepend-icon="mdi-close"
+                @click="cancelZip"
+              >
+                Cancel
+              </v-btn>
+            </v-list-item>
+          </template>
+          <v-list-item
+            v-else
+            density="compact"
+          >
+            <v-btn
+              color="primary"
+              variant="flat"
+              block
+              prepend-icon="mdi-folder-zip"
+              @click="downloadAsZip"
+            >
+              Download .zip
+            </v-btn>
+          </v-list-item>
+          <v-list-item
+            v-if="zipError"
+            density="compact"
+            :class="zipCanceled ? 'text-medium-emphasis text-caption' : 'text-error text-caption'"
+          >
+            {{ zipError }}
+          </v-list-item>
+          <v-divider class="my-2" />
+        </template>
         <v-list-item density="compact">
           Use this command in your DANDI CLI
         </v-list-item>
@@ -131,10 +184,22 @@
 </template>
 <script setup lang="ts">
 import { computed, onMounted, ref } from 'vue';
+import { downloadZip } from 'client-zip';
+import { filesize } from 'filesize';
 import { useDandisetStore } from '@/stores/dandiset';
 import CopyText from '@/components/CopyText.vue';
 import { dandiDocumentationUrl } from '@/utils/constants';
 import { dandiRest } from '@/rest';
+import type { Asset } from '@/types';
+
+// In-browser zip is gated to "small" dandisets: the whole archive is buffered in
+// memory (client-zip -> Blob) before the browser saves it, so keep both bytes and
+// file-count conservative. Bump these once a streaming-to-disk path is added.
+const BROWSER_ZIP_MAX_SIZE = 2 * (1024 ** 3); // 2 GiB
+const BROWSER_ZIP_MAX_FILES = 1000;
+
+// The assets list endpoint returns `zarr`/`blob` slugs that aren't on the TS Asset type.
+type AssetWithStorage = Asset & { zarr: string | null; blob: string | null };
 
 function downloadCommand(identifier: string, version: string): string {
   // Use the special 'DANDI:' url prefix if appropriate.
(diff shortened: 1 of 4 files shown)


bendichter:
https://github.com/user-attachments/assets/5fcd9a07-71e1-41ff-888b-7ca2e16e5b49

bendichter:
@yarikoptic , what do you think? Is this useful?

yarikoptic:
this reminded me about
- https://github.com/con/jsdownloader

and see there
- https://github.com/con/jsdownloader/issues/1
- https://github.com/con/jsdownloader/issues/2

bendichter:
In-browser zip download eligibility

The branch gates the feature on: has data (size > 0), size ≤ 2 GiB, and asset_count ≤ 1000.

```
┌──────────────────────────────────────┬───────┐
│                Metric                │ Count │
├──────────────────────────────────────┼───────┤
│ Total dandisets                      │ 869   │
├──────────────────────────────────────┼───────┤
│ With any data (size > 0)             │ 605   │
├──────────────────────────────────────┼───────┤
│ Eligible (≤ 2 GiB and ≤ 1000 assets) │ 209   │
└──────────────────────────────────────┴───────┘
```

- ~24% of all dandisets are eligible.
- ~35% of dandisets that have data are eligible (the other 264 dandisets are empty placeholders with 0 bytes).

@CodyCBakerPhD

bendichter:
@yarikoptic @CodyCBakerPhD, summarizing our discussion from Monday's call, plus my response to the jsdownloader idea.

**Where we landed on the call:**

- **Is the feature worth having?** Yarik: yes, this solves a real problem. The main open questions were scope of applicability and modularity.
- **How many dandisets does this apply to?** (Cody's question) Answered above: 209 of 869 dandisets are eligible under the current thresholds (~24% overall, ~35% of dandisets that actually contain data). So this isn't an edge-case feature.
- **Thresholds**: the 2 GiB / 1000 file caps are deliberately conservative because the zip is buffered in memory before saving. They can be raised later by streaming to disk (File System Access API), which is noted as a follow-up in the PR description.
- **Yarik's concerns**: (1) the generic download-list-of-URLs logic should live in a reusable library (per con/jsdownloader) so DataLad catalogs, OpenNeuro, etc. could share it; (2) a hard threshold invites "one byte over" complaints; (3) this should generalize beyond whole dandisets to folders, subjects, or any collection of assets ("a dandiset is just a folder").

**On the library question:**

I'd argue the generic library already exists: it's [`client-zip`](https://github.com/Touffy/client-zip) (~5 KB, maintained, MIT), which is what this PR uses. It takes named byte streams and produces a zip client-side, which is exactly the reusable core. What this PR adds on top is almost entirely DANDI-specific: asset pagination via our REST client, Zarr filtering, eligibility gating on version metadata, and Vuetify UI. The truly generic glue (sequential fetch with byte-level progress tracking and an `AbortController`) is about 60 lines, which I don't think clears the bar for a separately versioned and released package. My general rule is to generalize once something is needed in three places, and I'm happy to help extract a library at that point.

**Middle path I'd propose:** I'll refactor the generic part into a composable inside this repo, roughly `useZipDownload(entries: {name, url, size}[])`, with `DownloadDialog.vue` feeding it dandiset assets. That gives us the modular seam Yarik is asking for: a folder- or subject-level download button in the file browser later just feeds it a filtered asset list, and if a third consumer ever shows up, publishing it as a standalone package becomes mechanical rather than speculative.

**On the threshold complaint-magnet concern:** when a dandiset is over the limit, nothing is lost. The CLI instructions still show, exactly as today. The button is purely additive, and we can tune messaging (e.g., "too large for browser download, use the CLI") rather than architecture.

If that plan sounds reasonable, I'll do the composable refactor in this PR and file a follow-up issue for the folder-level button and the streaming-to-disk cap raise.

bendichter:
One known gap: the zip path uses a raw `fetch()`, which doesn't attach the API auth token that our axios client adds, so asset downloads from an embargoed dandiset would fail with a 401. I'll either hide the button on embargoed versions or route the requests through the authenticated client as part of the refactor.

bendichter:
I've pushed the refactor described above. The generic logic now lives in a composable, `web/src/composables/useZipDownload.ts`, which takes a zip filename and a function returning a list of `{name, url, size}` entries, and handles the sequential fetching, byte-level progress, cancellation, and client-side zipping. `DownloadDialog.vue` keeps only the DANDI-specific parts: asset pagination, Zarr filtering, eligibility gating, and the UI. This gives us the modular seam discussed on the call: a folder- or subject-level download button in the file browser would just feed the same composable a filtered asset list, and extracting a standalone package later becomes mechanical if a third consumer shows up.

Two other changes in the same commit:

1. The embargo gap I noted earlier is closed by hiding the button on embargoed dandisets. The eligibility check now requires `embargo_status === 'OPEN'`. The composable uses a plain unauthenticated `fetch`, so embargoed asset downloads would have failed with a 401. Routing those requests through the authenticated client is possible but depends on CORS behavior for embargoed presigned URLs that I haven't verified, so I went with the conservative option. Embargoed dandisets still get the CLI instructions as before.

2. I removed the `length` option previously passed to `downloadZip`. client-zip interprets that option as a predicted byte length for the archive, not a file count, so the value was wrong. It was harmless in practice because we buffer the response to a Blob and never use the Content-Length header, but there was no reason to keep it.

Still to do: file the follow-up issue for the folder-level download button and for streaming to disk to raise the 2 GiB cap.

web/src/views/DandisetLandingView/DownloadDialog.vue
```diff
@@ -192,4 +257,95 @@ const customDownloadText = computed(() => {
...
+    && !!identifier.value
+    && version.dandiset.embargo_status === 'OPEN'
+    && version.size > 0
+    && version.size <= BROWSER_ZIP_MAX_SIZE
+    && version.asset_count > 0
+    && version.asset_count <= BROWSER_ZIP_MAX_FILES;
+});
+
+// Fetch every asset in the version, following pagination.
+async function fetchAllAssets(
```
naglepuff, line 296: If the initial request for assets (line 291, `page === 1`) `404`s, then this function returns an empty array of assets. This means that in `downloadAsZip`, a `404` throws an error with the message "No download files in this dandiset..," which is misleading, since the error message doesn't match what actually happened.

This function should probably throw an error with a more clear message if it fails to fetch on the first page of assets.

web/src/views/DandisetLandingView/DownloadDialog.vue
naglepuff, line 1: It's unclear whether or not this is intended, but what should the behavior be if a user clicks "Download" but navigates away from the DLP? Currently the download will still happen, consuming resources, and the user will be unable to cancel. (Additionally, if they have configured their browser to prompt them before saving downloaded files, they might get confused if a Save As dialog pops up for a Dandiset they're no longer viewing).

If this is intended, that's fine. Otherwise we could call `cancel` in an `unmount` hook to cancel the download on navigate away.

naglepuff (changes requested):

web/src/views/DandisetLandingView/DownloadDialog.vue
bendichter: Fixed in b0a1ef60. If the first page request fails, the helper now throws "Failed to fetch assets for the current Dandiset" instead of returning an empty list, so the "no downloadable files" message only appears when the dandiset really has none.

web/src/views/DandisetLandingView/DownloadDialog.vue
bendichter: Good catch, that was not intended. In dae2e63c the download is cancelled in an unmount hook and also on a watch of the dandiset identifier and version, so navigating away or switching versions aborts the fetch. The dialog now says so next to the progress bar.

naglepuff (approved):

PR #2856, opened 2026-06-22
Add Anatomy metadata card to Dandiset landing page

bendichter:
## Summary

Adds a dedicated **Anatomy** section to the Dandiset Overview tab that always renders — surfacing anatomical metadata when present and explicitly flagging it when missing. This addresses the request (from the 2026-06-22 metadata-quality discussion) for a prominent, per-dandiset indicator of missing metadata, focused first on anatomy.

The card is placed between **Funding information** and **Protocols** in the Overview tab.

## Behavior

- **Provided:** green "Provided" chip; each anatomical region from the dataset's `about` field (entries with `schemaKey === "Anatomy"`) is listed with its name, ontology prefix + local id (e.g. `UBERON: 0002437`), and a link to the OBO PURL term page.
- **Missing:** amber warning card stating "No anatomical information provided" and explaining that adding indexed anatomy (UBERON terms) makes the dataset discoverable by anatomical location.

## Ontology links

Identifiers resolve prefix-agnostically to OBO PURLs:
- `UBERON:0002437` → `http://purl.obolibrary.org/obo/UBERON_0002437`
- `CL:0000242` → `http://purl.obolibrary.org/obo/CL_0000242`
- Full URLs pass through unchanged.

## Testing

Verified locally against sandbox (`npm run dev` → `api.sandbox.dandiarchive.org`):

| State | Dandiset | Result |
|-------|----------|--------|
| Provided | `100268` | Green card; link → `http://purl.obolibrary.org/obo/UBERON_0002437` |
| Missing | `200017` | Amber warning card |

`eslint` and `vue-tsc` type-check both pass.

### DIFF
--- a/web/src/components/DLP/AnatomyCard.vue
+++ b/web/src/components/DLP/AnatomyCard.vue
@@ -1,0 +1,189 @@
+<template>
+  <v-card
+    prepend-icon="mdi-brain"
+    variant="outlined"
+    height="100%"
+    :color="hasAnatomy ? undefined : 'warning'"
+  >
+    <template #title>
+      <div class="d-flex align-center justify-space-between">
+        <span>Anatomy</span>
+        <v-chip
+          v-if="hasAnatomy"
+          size="small"
+          color="success"
+          variant="flat"
+          prepend-icon="mdi-check-circle"
+        >
+          Provided
+        </v-chip>
+        <v-chip
+          v-else
+          size="small"
+          color="warning"
+          variant="flat"
+          prepend-icon="mdi-alert-circle"
+        >
+          Missing
+        </v-chip>
+      </div>
+    </template>
+
+    <!-- Anatomy information is present -->
+    <v-list
+      v-if="hasAnatomy"
+      :style="`column-count: ${columnCount};`"
+      class="px-5 pb-4"
+    >
+      <div
+        v-for="(item, i) in anatomy"
+        :key="i"
+        class="my-1 d-inline-block"
+        style="width: 100%;"
+      >
+        <div
+          class="pl-2 my-1 py-1"
+          :style="`border-left: medium solid ${borderLeftColor}; line-height: 1.25`"
+        >
+          <v-row
+            no-gutters
+            class="align-center justify-space-between mr-1"
+          >
+            <v-col cols="9">
+              <span class="text-grey-darken-3 font-weight-medium">
+                {{ item.name || item.identifier || 'Unnamed region' }}
+              </span>
+              <br>
+              <span
+                v-if="item.identifier"
+                class="text-caption text-grey-darken-1"
+              >
+                <strong>{{ ontologyName(item.identifier) }}: </strong>{{ ontologyLocalId(item.identifier) }}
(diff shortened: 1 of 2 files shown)


bendichter:
https://deploy-preview-2856--sandbox-dandiarchive-org.netlify.app/dandiset/218234?pos=4

image

https://deploy-preview-2856--sandbox-dandiarchive-org.netlify.app/dandiset/218045?page=2&sortOption=0&sortDir=-1&showDrafts=true&showEmpty=false&pos=15

image

bendichter:
@satra what do you think of this for "stink" around dandisets that don't have anatomy info?

bendichter:
@satra raised that this would cause problem for organoids. We decided on allowing them to get around this by adding a reference to organoids: https://www.ebi.ac.uk/ols4/ontologies/ncit/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCIT_C172259

satra:
it should use this specific thing: http://purl.obolibrary.org/obo/NCIT_C172259

satra:
they could be more specific too if it came from human cells.

web/src/components/DLP/AnatomyCard.vue
```diff
@@ -0,0 +1,189 @@
...
+      >
+        <div class="font-weight-bold">
+          No anatomical information provided.
+        </div>
+        <div class="text-caption mt-1">
+          This Dandiset does not specify which brain regions or anatomical
+          structures it covers. Adding indexed anatomy (e.g. UBERON terms) in the
+          metadata editor makes this dataset discoverable by anatomical location.
+        </div>
+      </v-alert>
```
mvandenburgh, line 83: This PR introduces a second, visually different way to communicate missing data. We already do something similar for when Funding Information is missing:

Image

Should we standardize on one visual design for this type of "encouraged but not technically required by the schema" metadata field? 

One other feedback that comes to mind is that having the bright orange text with a caution symbol and "Missing" chip for a field that isn't strictly required seems visually unappealing. My eyes immediately go to it when I open the DLP because it is so different and stands out from the other elements. Someone who ops to publish their dandiset without anatomy info would have that warning forever on their DLP, despite it technically passing validation.

naglepuff: @mvandenburgh @bendichter I've pushed 16ea74a, which is a suggestion for reconciling the differences pointed out between the existing funding and new anatomy metadata card. Now, both cards use the same visual indicator that metadata is missing/not provided. It moves away from using the warning color (implying the Dandiset metadata may be invalid) for a neutral color, but still uses a `v-alert` with an icon to draw attention to the fact that metadata is missing.

image

(I've also removed the Missing/Provided chip, which may have called too much attention to itself and was somewhat redundant)

Let me know if there are any additional thoughts.

bendichter: Looks good. Attracting attention was sort of the point, though I agree we should have internal consistency in our design.

mvandenburgh (approved):

PR #2865, opened 2026-07-16
Unify multipart upload endpoints

jjnesbitt:
This PR unifies the zarr and asset blob multipart upload endpoints into the singular existing assetblob upload endpoint. The prior change in #2784 was merged to master, but not deployed to production. 

This change is intended to be backwards compatible with the CLI, as it simply extends the fields present in both the upload initialize and upload complete endpoints. The existing CLI PR (https://github.com/dandi/dandi-cli/pull/1839) that was pointed at the previous change in #2784 can either be reworked, or replaced with a new PR.

### DIFF
--- a/dandiapi/api/views/upload.py
+++ b/dandiapi/api/views/upload.py
@@ -42,7 +42,9 @@ class DigestSerializer(serializers.Serializer):
 
 class UploadInitializationRequestSerializer(serializers.Serializer):
     contentSize = serializers.IntegerField(min_value=1)  # noqa: N815
-    digest = DigestSerializer(required=False)
+    digest = DigestSerializer()
+
+    # Unioned fields
     dandiset = DandisetIdentifierField(required=False)
     zarr_id = serializers.UUIDField(required=False)
     chunk_key = serializers.CharField(required=False)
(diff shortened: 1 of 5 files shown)


dandiapi/api/views/upload.py
```diff
@@ -101,6 +110,19 @@ class UploadCompletionResponseSerializer(serializers.Serializer):
...
+    # Zarr fields, null for asset blob uploads. The inherited AssetBlob
+    # fields are likewise null for zarr uploads.
+    zarr_id = serializers.UUIDField(allow_null=True, default=None)
+    chunk_key = serializers.CharField(allow_null=True, default=None)
+
+    class Meta(AssetBlobSerializer.Meta):
+        fields = [*AssetBlobSerializer.Meta.fields, 'zarr_id', 'chunk_key']
+        # Default the inherited AssetBlob fields to null so the zarr case only
+        # needs to supply its own fields.
+        extra_kwargs = {field: {'default': None} for field in AssetBlobSerializer.Meta.fields}
```
jjnesbitt, line 123: This is sort of ugly, but the intent was to avoid simply duplicating the fields from `AssetBlobSerializer`, in case anything were to change in the future.

jjnesbitt:
Closed in favor of #2869

PR #2869, opened 2026-07-27
Re-work zarr multipart upload

jjnesbitt:
Supersedes #2865 

This reworks the zarr multipart upload procedure by firmly distinguishing between multi-part zarrs, and single-part zarrs. This is done via a `multipart` flag on the zarr model itself. This allows all existing single-part zarrs to remain as such, while enabling multi-part upload capability, without any inherent size threshold.

This also unifies the zarr and asset blob multipart upload endpoints into the singular existing asset blob upload endpoint. The original implementation in https://github.com/dandi/dandi-archive/pull/2784 has been since reverted (see https://github.com/dandi/dandi-archive/pull/2878). It was never deployed to production.

This change is intended to be backwards compatible with the CLI, as it simply extends the fields present in both the upload initialize and upload complete endpoints. The existing CLI PR (https://github.com/dandi/dandi-cli/pull/1839) that was pointed at the previous change in https://github.com/dandi/dandi-archive/pull/2784 can either be reworked, or replaced with a new PR.


yarikoptic:
> This is done via a `multipart` flag on the zarr model itself. This allows all existing single-part zarrs to remain as such, while enabling multi-part upload capability, without any inherent size threshold.

Just to make sure -- all prior uploaded zarrs will stay single-part, and only new ones, and new uploaded with the newer version of dandi-cli will be "multipart"?

jjnesbitt:
> Just to make sure -- all prior uploaded zarrs will stay single-part, and only new ones, and new uploaded with the newer version of dandi-cli will be "multipart"?

Correct.

dandiapi/api/models/upload.py
```diff
@@ -40,16 +48,42 @@ class Upload(models.Model):  # noqa: DJ008
...
+        ]
+
+    @property
+    def embargoed(self):
+        dandiset = self.dandiset or self.zarr.dandiset
+        return dandiset.embargoed
+
+    @classmethod
+    def initialize_zarr_multipart_upload(cls, etag, size, zarr: ZarrArchive, chunk_key: str):
+        object_key = zarr.s3_path(chunk_key.lstrip('/'))
```
mvandenburgh, line 67: This changes an invariant we implicitly have that `Upload` rows are 1:1 with S3 blobs. With this change, I believe you could have a situation where you create multiple `Upload` objects that point at the same S3 object.

One thing this breaks is Upload garbage collection - see https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/services/garbage_collection/upload.py#L59; just having one stale Upload pointing at an S3 object is enough for it to get deleted out from under others that point at it

dandiapi/api/views/upload.py
```diff
@@ -15,14 +15,15 @@
 from rest_framework.response import Response
```
mvandenburgh, line 15: In general, it seems you're mixing a lot of zarr-only logic into the upload view and model class. Which clashes with the motivation behind the separate `zarr` and `api` apps. My main concern is that it's becoming difficult to trace the upload path of AssetBlobs and Zarrs independently.

I'm not sure how to best address this. The first thing that comes to mind is to move the zarr upload logic into the `zarr` app, and just accept some slight duplication, but I saw you mentioned you already tried that in the older PR. I'm curious what you think about the negatives of that approach vs. this one.

dandiapi/api/models/upload.py
```diff
@@ -40,16 +48,42 @@ class Upload(models.Model):  # noqa: DJ008
     class Meta:
         ordering = ['created']
         indexes = [models.Index(fields=['etag'])]
+        constraints = [
+            # Ensure that there's exactly one of dandiset or zarr
+            models.CheckConstraint(
+                name='dandiset-zarr-xor',
+                condition=models.Q(dandiset__isnull=True, zarr__isnull=False)
+                | models.Q(dandiset__isnull=False, zarr__isnull=True),
```
jjnesbitt, line 56: TODO: Add constraint for unique blobs per dandiset/zarr

jjnesbitt: And/or split up the models

dandiapi/api/models/upload.py
jjnesbitt: @mvandenburgh With the merge of #2887 is this still an issue?

mvandenburgh (approved):

dandiapi/zarr/views/upload.py
```diff
@@ -106,17 +106,20 @@ def zarr_upload_initialize_view(request: AuthenticatedRequest) -> HttpResponseBa
...
+        if zarr_archive.status in [ZarrArchiveStatus.UPLOADED, ZarrArchiveStatus.INGESTING]:
+            raise ValidationError(ZarrArchive.INGEST_ERROR_MSG)
+
+        upload, initialization = ZarrUpload.initialize_multipart_upload(
+            etag,
+            content_size,
+            zarr=zarr_archive,
+            chunk_key=data['chunk_key'],
+            content_type=data['content_type'],
+        )
```
mvandenburgh, line 122: It's worth pointing out that this holds a row lock through a synchronous network call. I need to refamiliarize myself with the upload flow here, so this might be a dumb question, but can the finalize endpoint handle the locking instead?

dandiapi/zarr/views/upload.py
```diff
@@ -46,6 +46,7 @@ class ZarrUploadInitializationRequestSerializer(serializers.Serializer):
     zarr_id = serializers.UUIDField()
     chunk_key = serializers.CharField(validators=[validate_zarr_path])
     contentSize = serializers.IntegerField(min_value=1)  # noqa: N815
+    content_type = serializers.CharField(required=False, default='application/octet-stream')
```
mvandenburgh, line 49: Need a max_length on this CharField

Issue #2219, opened 2025-03-18
Zarr stuck in "Ingesting"


yarikoptic:
Reported by @gmazzamuto , one of the authors of the 000026, that he fails to update one asset , even fails to delete it
https://api.dandiarchive.org/api/dandisets/000026/versions/draft/assets/fc330324-6693-48cd-87aa-9192ab9caeaf/

looking at the asset, indeed that Zarr seems got stuck in Ingesting state:

```shell
❯ curl --silent -X 'GET' 'https://api.dandiarchive.org/api/zarr/23123eef-62df-46b8-92ae-c2102110977c/' | jq .
{
  "name": "sub-I48/ses-SPIM/micr/sub-I48_ses-SPIM_sample-BrocaAreaS04_stain-NeuN_SPIM.ome.zarr",
  "dandiset": "000026",
  "zarr_id": "23123eef-62df-46b8-92ae-c2102110977c",
  "status": "Ingesting",
  "checksum": null,
  "file_count": 0,
  "size": 0
}
```

when I tried to send DELETE request too in the web interface, got `HTTP 204 No Content` (so it seems succeded) 

![Image](https://github.com/user-attachments/assets/11c46cb5-e0d5-46a8-8026-2f3bf4592d3a)

<details>
<summary>and subsequent repeated DELETE - 404, so seems like I did manage to delete it </summary> 

![Image](https://github.com/user-attachments/assets/5fe4abab-6982-4cc1-9c65-7b0e0edb8f54)
</details>

The https://api.dandiarchive.org/api/dandisets/000026/versions/draft/assets/fc330324-6693-48cd-87aa-9192ab9caeaf/ is still showing the asset but 

<details>
<summary>atpath says not found</summary> 

```shell
❯ curl -i -X 'GET' 'https://api.dandiarchive.org/api/webdav/assets/atpath/?dandiset_id=000026&version_id=draft&path=sub-I48%2Fses-SPIM%2Fmicr%2Fsub-I48_ses-SPIM_sample-BrocaAreaS04_stain-NeuN_SPIM.ome.zarr'
HTTP/1.1 404 Not Found
Report-To: {"group":"heroku-nel","max_age":3600,"endpoints":[{"url":"https://nel.heroku.com/reports?ts=1742307466&sid=c46efe9b-d3d2-4a0c-8c76-bfafa16c5add&s=pL562qIu%2FawpFJsf5aSTNFUunwj6yK1lpZVRliVnprw%3D"}]}
[...]

PR #2872, opened 2026-07-28
Reduce zarr stuckness

danlamanna:
This PR partially addresses #2219.

It removes an accidentally introduced overbroad lock (#2111), and allows the ingestion of zarr archives to retry tasks when they fail due to deadlock, which is what happened with the latest slate of stuck zarrs.

The nature of ingesting zarrs makes for one long transaction (long as in seconds), which provides a large window of opportunity for simultaneous requests against the same resources to depend on each other.

I'd like to do a more thorough analysis of operations that conflict with each other, but for now this should significantly reduce the "zarr stuck in ingesting" problem - and otherwise there should be a deadlock error in sentry.

### DIFF
--- a/dandiapi/zarr/tasks/__init__.py
+++ b/dandiapi/zarr/tasks/__init__.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 from celery import shared_task
 from celery.utils.log import get_task_logger
-from django.db import transaction
+from django.db import OperationalError, transaction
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
Thanks for catching this! That was definitely an oversight on my part. 

I'm going to merge, but hold off on releasing since #2869 will hopefully be merged in the near future, which reworks some un-released changes in master.

PR #2877, opened 2026-08-04
Redirect hub.dandiarchive.org to docs.dandiarchive.org/user-guide-using/dandi-hub

asmacdo:
## CAUTION: ORDER MATTERS 

0.  Merge the graveyard page https://github.com/dandi/dandi-docs/pull/243
1. **This PR merges and deploys.** Open question for @waxlamp: retarget this at `release`?
2. **Add `hub.dandiarchive.org` as a domain alias** on the redirect project in Netlify.
3. **Add the Route53 record** in dandi-infrastructure: CNAME `hub` -> `redirect-dandiarchive-org.netlify.com`
   github.com/dandi/dandi-infrastructure/pull/282

Redirects `hub.dandiarchive.org` to the DANDI Hub retirement notice in the docs, per @yarikoptic's suggestion.

302 rather than 301 in case we need to use the subdomain again.

Two questions:

1. The gui netlify project (the one that does the redirect) deploys  `release`, which is ~120 commits behind `master`, so merging here doesn't activate anything until the next release. Should I open this against `release`  instead?
2. I've deleted the R53 route for now. ~Once this merges, I'll add the DNS record back, and then add `hub.dandiarchive.org` as a domain alias on the redirect project in Netlify.~ (infra PR instead)

### DIFF
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -16,5 +16,5 @@ status = 308
 # The DANDI JupyterHub has been taken down.
 [[redirects]]
 from = "https://hub.dandiarchive.org/*"
-to = "https://notebooks.dandiarchive.org/"
+to = "https://docs.dandiarchive.org/user-guide-using/dandi-hub/"
 status = 302
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -13,7 +13,7 @@ from = "https://api-staging.dandiarchive.org/*"
 to = "https://api.sandbox.dandiarchive.org/:splat"
 status = 308
 
-# The DandiHub JupyterHub has been taken down.
+# The DANDI JupyterHub has been taken down.
 [[redirects]]
 from = "https://hub.dandiarchive.org/*"
 to = "https://notebooks.dandiarchive.org/"
--- a/redirector/netlify.toml
+++ b/redirector/netlify.toml
@@ -12,3 +12,9 @@ status = 301
 from = "https://api-staging.dandiarchive.org/*"
 to = "https://api.sandbox.dandiarchive.org/:splat"
 status = 308
+
+# The DandiHub JupyterHub has been taken down.
+[[redirects]]
+from = "https://hub.dandiarchive.org/*"
+to = "https://notebooks.dandiarchive.org/"
+status = 302


kabilar:
Thanks @asmacdo.

> 1. The gui netlify project (the one that does the redirect) deploys  `release`, which is ~120 commits behind `master`, so merging here doesn't activate anything until the next release. Should I open this against `release`  instead?

It would be ideal to get this released soon.  Please discuss with @waxlamp regarding the best approach.

> 2\. I've deleted the R53 route for now. Once this merges, I'll add the DNS record back, and then add `hub.dandiarchive.org` as a domain alias on the redirect project in Netlify.

It would be best to submit this change to `dandi-infrastructure`.

redirector/netlify.toml
```diff
@@ -12,3 +12,9 @@ status = 301
 from = "https://api-staging.dandiarchive.org/*"
 to = "https://api.sandbox.dandiarchive.org/:splat"
 status = 308
+
+# The DandiHub JupyterHub has been taken down.
```
kabilar:
```suggestion
# The DANDI JupyterHub has been taken down.
```

yarikoptic (changes requested):
redirect is a nice idea but I think it should be redirected to some page describing why/where instead of dumping into a new cold bucket....

I think anyways we need to add to https://docs.dandiarchive.org/developer-guide/creating-dandi-instance/dandi-hub/ the announcement that as of Aug 1 Hub is being deprecated and users are invited to use google collab and visit our notebooks.d.o -- that is where it should be, not here!

redirector/netlify.toml
```diff
@@ -12,3 +12,9 @@ status = 301
 from = "https://api-staging.dandiarchive.org/*"
 to = "https://api.sandbox.dandiarchive.org/:splat"
 status = 308
+
+# The DANDI JupyterHub has been taken down.
+[[redirects]]
+from = "https://hub.dandiarchive.org/*"
+to = "https://notebooks.dandiarchive.org/"
```
yarikoptic, line 19:
```suggestion
to = "https://docs.dandiarchive.org/user-guide-using/dandi-hub/"
```

asmacdo:
Back to draft until https://github.com/dandi/dandi-docs/pull/243 merges

waxlamp:
@asmacdo, @kabilar: there might be a simpler option here: revive the infrastructure for hosting a web application at hub.dandiarchive.org, but replace it with a simple, static page that explains that the hub has been deprecated, and either redirects the user (after a delay) to wherever they should go next (the notebooks site, or the docs), or else just presents them with both links (and they can choose).

That resolves the issue of where to send the user (that Yarik raised), it provides a more graceful experience for the user, and it avoids adding redirect logic to this repository (which won't work right away; we would need to add back the DNS entry and have it point to the Netlify app, etc. etc.).

Let me know what you think.

asmacdo:
I have added the redirect in an infrastructure PR, and can add the netlify when we are ready. 

I lean that way so we don't have another thing we have to host (even a small static page) but I don't have a strong opinion. 

Whatever we do I'd like to get to consensus and be done with it.

yarikoptic:
I am ok either way (redirect or a page), only that with page we would gain one more page to maintain etc. IMHO pointing to docs page should be enough and would remove the need for one more thing to maintain

waxlamp:
Closing in favor of #2888. Long story and it's late on a Friday afternoon 😩

PR #2878, opened 2026-08-05
Revert "Add support for zarr multi-part uploads (#2784)"

jjnesbitt:
This reverts the original multi-part zarr implementation that was merged to master and deployed to sandbox, but never deployed to production. 

That original implementation will be superseded by #2869.


mvandenburgh:
Is reverting the migration going to cause problems on staging? Did it already get deployed there?

jjnesbitt:
> Is reverting the migration going to cause problems on staging? Did it already get deployed there?

It did get deployed there. My intent is to roll the migrations back to `0031` in staging before merging this PR.

mvandenburgh (approved):
LGTM. Looks like there is a linting failure to address before merge.

PR #2881, opened 2026-08-05
Fix zarr path traversal bug

mvandenburgh:
The zarr file endpoints interpolate a user-supplied path into an S3 key of the form `zarr/<zarr_id>/<path>`. django-storages' `clean_name` runs `posixpath.normpath` (which collapses `..`) *before* its `safe_join` guard, and the storage `location` is empty, so a `..` segment escapes the zarr prefix and resolves to an arbitrary key in the shared bucket. ([This is intended behavior](https://github.com/jschneier/django-storages/blob/85928d63673190af4d689c9cad0e2ffaca21480a/storages/backends/s3.py#L520)) 

This lets anyone with a dandi account overwrite/delete arbitrary objects in the bucket, as well as potentially read embargoed data if they have the object key.

This PR adds a validator that rejects any `..` segment to the relevant serializers.

### DIFF
--- a/dandiapi/zarr/tests/test_zarr.py
+++ b/dandiapi/zarr/tests/test_zarr.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from django.core.files.base import ContentFile
 import pytest
 from zarr_checksum.checksum import EMPTY_CHECKSUM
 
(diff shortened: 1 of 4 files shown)


dandiapi/zarr/tests/test_zarr.py
```diff
@@ -495,3 +496,60 @@ def test_zarr_explore_head(api_client):
     resp = api_client.head(f'/api/zarr/{zarr_archive.zarr_id}/files/', {'prefix': filepath})
     assert resp.status_code == 302
     assert f'/test-zarr/{zarr_archive.zarr_id}/{filepath}?' in resp.headers['Location']
+
+
+# Paths that would traverse out of the zarr's S3 prefix and reach arbitrary bucket objects.
+_TRAVERSAL_PATHS = [
```
jjnesbitt: Not really a big deal but should this var just be inlined where it's used below?

jjnesbitt (approved):
LGTM besides my one comment

dandiapi/zarr/tests/test_zarr.py
mvandenburgh: Yeah it should be, I missed that. There was another test that used it too, but it was removed when I rebased this on #2878 because it related to the code removed in that PR.

PR #2883, opened 2026-08-05
Fix for search in-progress indicator on the subsequent searches not being shown

bendichter:
## Problem

The dandisets listing gives no feedback while a search is running. The spinner is gated on `dandisets` being undefined, and the previous response is never cleared when a new request starts, so in practice the spinner appears only on the very first page load. Any search run from an already-populated list leaves the old results sitting on screen, unchanged, until the new response arrives.

That is confusing at any latency and actively misleading at the latency we currently have: the advanced search operators can take upwards of ten seconds, during which the page looks like it has already answered the query. #2882 addresses the slowness itself; this addresses the fact that the remaining wait is invisible.

## Change

Track an explicit `loading` flag rather than inferring it from whether results exist. While a request is in flight:

- an indeterminate progress bar shows under the toolbar
- the result count is replaced with "Searching..."
- stale results stay mounted but are dimmed and non-interactive, so they do not read as the answer to the query being run

Keeping the old results visible rather than blanking the list avoids a full-page content flash on every pagination click, which is the common case and is usually fast.

## Out-of-order responses

The same change fixes a latent bug. Requests were not sequenced, so with a slow backend a fast follow-up search could be overwritten by an earlier, slower response landing after it. Requests now carry a monotonic id and only the newest may write results.

While I was in there I also hoisted the request parameters above the first `await`. They were already read synchronously, so this is not a behavior change, but it makes the `watchEffect` dependency tracking obvious rather than incidental.

## Testing

Frontend type check and lint pass. Verified manually against a local instance: the progress bar and spinner appear for the duration of a slow request and clear when it lands. I first reproduced the original bug by pointing the dev server at the production API, where `species:mouse` takes about 13 seconds, and confirmed the indicator now covers that whole window.

This is independent of #2882 and touches a different file, so the two can merge in either order.

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -164,6 +164,7 @@
     <div
       v-else-if="searchActive && djangoDandisetRequest"
       class="mx-4 mx-md-8 mt-4 text-h6"
+      role="status"
     >
       <template v-if="loading">
         Searching&hellip;
@@ -180,6 +181,7 @@
       v-if="dandisets && dandisets.length"
       :dandisets="dandisets"
       :class="{ 'results-stale': loading }"
+      :inert="loading"
     />
     <v-container v-else>
       <v-row
@@ -285,6 +287,11 @@ watchEffect(async () => {
   latestRequest += 1;
   const thisRequest = latestRequest;
   loading.value = true;
+  // Clear the previous query's error, or its alert keeps rendering and
+  // suppresses the loading state below it for the whole of this request.
+  // `searchError` is only read in the template, so writing it here doesn't
+  // feed back into this effect.
+  searchError.value = null;
   try {
     const response = await dandiRest.dandisets(params);
     if (thisRequest !== latestRequest) {
@@ -354,10 +361,10 @@ watch(queryParams, (params) => {
 </script>
 
 <style scoped>
+/* Purely visual; `inert` on the list itself is what stops interaction. */
 .results-stale {
   opacity: 0.5;
   transition: opacity 0.2s ease;
-  pointer-events: none;
 }
 
 .btn-group--sort-options {
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -141,6 +141,17 @@
         </v-menu>
       </v-btn>
     </v-toolbar>
+    <!--
+      Immediate feedback that a request is in flight. Searches can take
+      several seconds, and without this the previous results sit on screen
+      with nothing to indicate that anything is happening.
+    -->
+    <v-progress-linear
+      :active="loading"
+      indeterminate
+      color="primary"
+      height="3"
+    />
     <v-alert
       v-if="searchActive && searchError"
       type="error"
@@ -154,11 +165,21 @@
       v-else-if="searchActive && djangoDandisetRequest"
       class="mx-4 mx-md-8 mt-4 text-h6"
     >
-      {{ djangoDandisetRequest.count }} {{ djangoDandisetRequest.count === 1 ? 'result' : 'results' }} found
+      <template v-if="loading">
+        Searching&hellip;
+      </template>
+      <template v-else>
+        {{ djangoDandisetRequest.count }} {{ djangoDandisetRequest.count === 1 ? 'result' : 'results' }} found
+      </template>
     </div>
+    <!--
+      Stale results stay mounted (rather than blanking the page) but are dimmed
+      so they don't read as the answer to the query being run.
+    -->
     <DandisetList
       v-if="dandisets && dandisets.length"
       :dandisets="dandisets"
+      :class="{ 'results-stale': loading }"
     />
     <v-container v-else>
       <v-row
@@ -168,7 +189,7 @@
       >
         <v-col>
           <v-progress-circular
-            v-if="!dandisets"
+            v-if="loading || !dandisets"
             indeterminate
           />
           <slot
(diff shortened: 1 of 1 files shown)


bendichter:
old behavior:

https://github.com/user-attachments/assets/f0373dc9-e5ec-4894-a8a9-52e4b8e45d29

bendichter:
New behavior: 


https://github.com/user-attachments/assets/40b6999d-543e-4239-b834-41642153e961

web/src/components/DandisetsPage.vue
```diff
@@ -241,23 +262,40 @@ const sortField = computed(() => sortingOptions[sortOption.value].djangoField);
...
+    user: props.user ? 'me' : null,
+    search: searchActive.value ? route.query.search : null,
+    starred: props.starred ? true : null,
+    draft: props.user ? true : showDrafts.value,
+    empty: props.user ? true : showEmpty.value,
+    embargoed: props.user,
+  };
+  latestRequest += 1;
+  const thisRequest = latestRequest;
+  loading.value = true;
```
yarikoptic-gitmate, line 287: `searchError` is never cleared when a new request starts, only on success. Because the alert at L156 is `v-if` and the count/`Searching…` row is its `v-else-if`, an error left over from the previous query suppresses the new loading state entirely.

Repro: search `species:` (invalid syntax → 400, alert renders) → search `species:mouse`. For the ~10s the new request is in flight the page still shows *"Unsupported operator…"* and never shows `Searching…`. The progress bar is the only hint, and the visible text is actively wrong — the same failure mode this PR is fixing.

```suggestion
  loading.value = true;
  searchError.value = null;
```

(`searchError` is only written here, so this doesn't create a tracking loop in the `watchEffect`.) Gating the alert on `!loading` instead would also work — after a 400, `djangoDandisetRequest` is set to an empty page, so the `v-else-if` branch is reachable either way.

---

web/src/components/DandisetsPage.vue
```diff
@@ -154,11 +165,21 @@
...
+      </template>
     </div>
+    <!--
+      Stale results stay mounted (rather than blanking the page) but are dimmed
+      so they don't read as the answer to the query being run.
+    -->
     <DandisetList
       v-if="dandisets && dandisets.length"
       :dandisets="dandisets"
+      :class="{ 'results-stale': loading }"
```
yarikoptic-gitmate, line 182: `pointer-events: none` only stops the mouse. `DandisetList` renders each row as a `v-list-item` with `:to`, i.e. a real `<a href>`, so while results are dimmed a keyboard user can still tab into them and hit Enter to navigate to a result from the *previous* query — exactly the confusion the dimming is meant to prevent.

`inert` covers pointer, keyboard and AT in one attribute and is supported everywhere Vuetify 3 targets:

```suggestion
      :class="{ 'results-stale': loading }"
      :inert="loading"
```

You could then drop `pointer-events: none` from `.results-stale` and leave it as the purely visual `opacity`/`transition`.

---

web/src/components/DandisetsPage.vue
```diff
@@ -154,11 +165,21 @@
       v-else-if="searchActive && djangoDandisetRequest"
       class="mx-4 mx-md-8 mt-4 text-h6"
     >
-      {{ djangoDandisetRequest.count }} {{ djangoDandisetRequest.count === 1 ? 'result' : 'results' }} found
+      <template v-if="loading">
```
yarikoptic-gitmate, line 168: Optional, but this is the one spot where it's nearly free: neither `Searching…` nor the result count is announced to screen readers, and Vuetify sets `aria-hidden="true"` on the progress bar whenever `active` is false, so an AT user gets no signal at all that a query ran. Adding `role="status"` to the enclosing `<div>` (L166) makes both transitions announce politely.

---

yarikoptic-gitmate (commented):
Reviewed the diff and checked it out locally. The approach is right and the implementation is sound — an explicit `loading` flag instead of inferring it from `dandisets === undefined` is the correct fix, and the monotonic request id is a genuine bug fix, not just scaffolding for the spinner.

Verified locally on `41c7973`: `npm run type-check` and `eslint` both clean.

A few things I checked that are easy to get wrong and that this gets right:

- Hoisting `params` above the `await` is indeed behavior-preserving — the object literal was already evaluated synchronously as the call argument — but it does make the `watchEffect` tracking intentional rather than accidental. Worth keeping.
- The `finally` guards on `thisRequest === latestRequest`, so a stale response landing late can't flip `loading` off while a newer request is still running. That's the part people usually miss.
- `v-progress-linear` with `:active="false"` renders at `height: 0` in Vuetify 3.7, so there's no layout shift when it's idle.
- `.results-stale` reaches `DandisetList` because it has a single root (`v-list`), so the class falls through and the `scoped` attribute lands on it. If that component ever grows a second root node the dimming silently stops working — not a problem today, just noting the coupling.
- The request id also quietly cleans up the double-fire from `watch([showDrafts, showEmpty])` resetting `page`.

Three comments inline. The first is a real user-visible gap in the state this PR is adding (stale error suppresses the new loading state); the other two are accessibility follow-ups on the dimming and the status text. None of them are structural.

<details>
<summary>Notes on things I looked at but am not asking you to change</summary>

- On first load both the linear bar and the pre-existing `v-progress-circular` show at once. Redundant, but the circular one is doing the "empty page" job and the bar is doing the "toolbar" job, so I'd leave it.
- The three separate `thisRequest !== latestRequest` checks are a little repetitive, but collapsing them costs more than it saves — the `try`/`catch`/`finally` branches genuinely need different handling.
- `throw err` in an async `watchEffect` still produces an unhandled rejection for non-400 errors. Pre-existing, and the staleness guard actually improves it by swallowing errors from abandoned requests.
- No test added. The `e2e/` Playwright suite could assert the indicator via a route-level delay, but the timing dependence probably isn't worth the flake here — the two screen recordings cover it well enough.

</details>

---

web/src/components/DandisetsPage.vue
bendichter: Done in 68d142e, `searchError` is cleared when each request starts.

web/src/components/DandisetsPage.vue
bendichter: Done in 68d142e. The list gets `inert` while loading, and `.results-stale` now only sets the opacity.

web/src/components/DandisetsPage.vue
bendichter: Added `role="status"` to the count row in 68d142e.

PR #2887, opened 2026-08-12
Rework "Zarr multipart flag rework"

jjnesbitt:
This PR reworks #2869 to split up the zarr upload endpoints and models from that of asset blobs. This results in a much cleaner API design that is still backwards compatible with the CLI, as the existing single-part upload endpoints aren't removed.


dandiapi/api/models/upload.py
```diff
@@ -110,6 +136,12 @@ def initialize_multipart_upload(
             'parts': multipart_initialization.parts,
         }
 
+    def abort(self) -> None:
+        super().abort()
+
+        # This upload owns its object key outright, so the object can be deleted along with it
+        self.blob.delete(save=False)
+
```
mvandenburgh, line 144: Things like this make me realize we never implemented a service layer for uploads. Obviously, that's completely out of scope for this change, but it might be worth revisiting at some point.

dandiapi/api/services/embargo/__init__.py
```diff
@@ -30,6 +31,11 @@
 logger = logging.getLogger(__name__)
 
 
+def _has_active_uploads(ds: Dandiset) -> bool:
+    """Return whether the dandiset has any in-progress asset blob or zarr chunk uploads."""
+    return ds.uploads.exists() or ZarrUpload.objects.filter(zarr__dandiset=ds).exists()
```
mvandenburgh: We should also include this logic here https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/views/dandiset.py#L612-L648. The logic below prevents a dandiset with active zarr chunk uploads from being unembargoed on the backend, but the Unembargo button in the UI uses the uploads list endpoint to determine whether to allow clicking it.

mvandenburgh: Along similar lines, there's also this https://github.com/dandi/dandi-archive/blob/master/dandiapi/api/models/version.py#L89-L91, which is used to populate the "active uploads" frontend view.

dandiapi/zarr/views/upload.py
```diff
@@ -0,0 +1,189 @@
...
+    from collections import OrderedDict
+
+    from s3_file_field._multipart import TransferredPart
+
+logger = logging.getLogger(__name__)
+
+
+class ZarrUploadInitializationRequestSerializer(serializers.Serializer):
+    zarr_id = serializers.UUIDField()
+    chunk_key = serializers.CharField()
```
mvandenburgh: Similar to the bug #2881, I believe a user could abuse this with a relative path using `..`, as it's used later directly in the construction of the S3 path.

mvandenburgh (changes requested):
I agree this is cleaner. Overall looks great, just a few bugs and places that need to be updated.

dandiapi/zarr/views/upload.py
jjnesbitt: Good catch, I thought I had covered that but clearly not.

dandiapi/api/services/embargo/__init__.py
jjnesbitt: Your second point is easily fixed via https://github.com/dandi/dandi-archive/pull/2887/commits/0ec60be163c44119c4a4c217f2b19e0a32698e93.

The first point however is a bit more intrusive, as now we're mixing two different models within one endpoint. However, we do still need to address it. I'll see what works out best, but we'll have to change that endpoint in some way.

jjnesbitt:
@mvandenburgh I've pushed https://github.com/dandi/dandi-archive/pull/2887/commits/58ee4ebe78fb928fdb44be0f7722f6d19b583697 to address your [comment](https://github.com/dandi/dandi-archive/pull/2887#discussion_r3778052700), let me know what you think. 

I wanted to avoid having separate endpoints for uploads and zarr uploads in this case, since the dandiset level concern of "are there active uploads" should really be agnostic to that.

dandiapi/api/views/dandiset.py
```diff
@@ -623,16 +644,19 @@ def uploads(self, request, dandiset__pk):
         # Special case where a "safe" method is access restricted, due to the nature of uploads
         self.require_owner_perm(dandiset)
 
-        uploads: QuerySet[Upload] = dandiset.uploads.all()
+        # Get combined list of Upload and ZarrUpload, ordered by creation time.
+        blob_uploads = Upload.objects.filter(dandiset=dandiset).values('upload_id', 'created')
+        zarr_uploads = ZarrUpload.objects.filter(zarr__dandiset=dandiset).values(
+            'upload_id', 'created'
+        )
+        unioned_uploads = blob_uploads.union(zarr_uploads).order_by('created')
```
mvandenburgh:
```suggestion
        unioned_uploads = blob_uploads.union(zarr_uploads, all=True).order_by('created')
```

This will make the ORM produce a slightly faster query using `UNION ALL` instead of `UNION`, i.e. skip the duplicate check, which is not needed because these elements are already guarenteed to be unique due to the DB constraint on `upload_id`.

dandiapi/api/views/dandiset.py
```diff
@@ -177,6 +177,26 @@ def filter_queryset(self, request: Request, queryset: QuerySet, view: APIView) -
...
 
 
+def _serialize_uploads(rows: list[dict]) -> list[dict]:
+    """Serialize a combined list of Upload and ZarrUpload."""
+    upload_ids = [row['upload_id'] for row in rows]
+
+    # Upload ID is a UUID, so filtering each model by the entire
+    # list will only return that model's respective entries.
+    serialized_zarr_uploads = DandisetZarrUploadSerializer(
+        ZarrUpload.objects.filter(upload_id__in=upload_ids), many=True
```
mvandenburgh:
```suggestion
        ZarrUpload.objects.select_related('zarr').filter(upload_id__in=upload_ids), many=True
```

Due to https://github.com/dandi/dandi-archive/pull/2887/changes/58ee4ebe78fb928fdb44be0f7722f6d19b583697#diff-aa8cb34bf2d9aebf4ab9746d962f63172c33b0846f5ec6b6cea1416126dc43efR451

mvandenburgh (approved):
LGTM. Some minor DB optimizations are possible, see below.

PR #2893, opened 2026-08-20
e2e: reduce flakiness of playwright tests

naglepuff:
### Problem
While spinning up for development on the `dandi-archive` repo, I noticed inconsistent test runs for the playwright end-to-end tests.

### Changes
This PR implements several changes to the e2e test suite to improve performance (by, e.g. utilizing `beforeAll` hooks to reduce API activity), make selectors behave more correctly/consistently, increase certain timeouts to help race conditions, and reduce the number of workers in order to decrease contention and load on the test server.

Testing was done locally by running the project in "native development" mode. Over the course of 30 pre-change runs of the e2e test suite I saw a pass rate of 88%. With the changes the pass rate is now 99.7% (over 20 runs). This should reduce time spent retrying failed tests in CI, and generally improve this set of tests.

### DIFF
--- a/e2e/tests/dandisetLandingPage.spec.ts
+++ b/e2e/tests/dandisetLandingPage.spec.ts
@@ -43,7 +43,9 @@ test.describe("dandiset landing page", async () => {
     await newPage.waitForTimeout(250);
     await newPage.getByRole("dialog").getByRole("button").nth(1).click();
     await newPage.getByRole("button", { name: "Done" }).click();
-    await expect(newPage.getByText(otherUserName)).toHaveCount(1);
+    // Saving the owner change and re-rendering the owners list is an async
+    // round-trip; give it more room than the default 5s expect timeout.
+    await expect(newPage.getByText(otherUserName)).toHaveCount(1, { timeout: 15000 });
     await context.close();
   });
   test("navigate to an invalid dandiset URL", async ({ page }) => {
--- a/e2e/tests/dandisetLandingPage.spec.ts
+++ b/e2e/tests/dandisetLandingPage.spec.ts
@@ -4,6 +4,9 @@ import { faker } from "@faker-js/faker";
 
 test.describe("dandiset landing page", async () => {
   test("add an owner to the dandiset", async ({ page, browser }) => {
+    // This test does 2 full user signups, a dandiset registration, and a dialog
+    // interaction; under concurrent workers the default 30s budget is too tight.
+    test.slow();
     // Create a user to add as an owner later
     const {
       email: otherUserEmail,
--- a/e2e/tests/dandisetLandingPage.spec.ts
+++ b/e2e/tests/dandisetLandingPage.spec.ts
@@ -43,8 +43,8 @@ test.describe("dandiset landing page", async () => {
     await newPage.waitForTimeout(250);
     await newPage.getByRole("dialog").getByRole("button").nth(1).click();
     await newPage.getByRole("button", { name: "Done" }).click();
-    // Saving the owner change and re-rendering the owners list is an async
-    // round-trip; give it more room than the default 5s expect timeout.
+    // With high concurrency, the default timeout of 5s was not enough, raising the
+    // timeout reduces flakiness of this assertion.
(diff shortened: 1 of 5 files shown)


e2e/tests/dandisetLandingPage.spec.ts
```diff
@@ -37,7 +43,9 @@ test.describe("dandiset landing page", async () => {
     await newPage.waitForTimeout(250);
     await newPage.getByRole("dialog").getByRole("button").nth(1).click();
     await newPage.getByRole("button", { name: "Done" }).click();
-    await expect(newPage.getByText(otherUserName)).toHaveCount(1);
+    // Saving the owner change and re-rendering the owners list is an async
+    // round-trip; give it more room than the default 5s expect timeout.
+    await expect(newPage.getByText(otherUserName)).toHaveCount(1, { timeout: 15000 });
```
mvandenburgh: Did you actually see this request/ui rendering take more than 5 seconds? That seems abnormally high, even when running the server inside github actions

e2e/tests/dandisetLandingPage.spec.ts
```diff
@@ -12,7 +15,10 @@ test.describe("dandiset landing page", async () => {
     } = await registerNewUser(page);
     const otherUserName = `${otherUserFirstname} ${otherUserLastName}`;
     const otherUserInitials = `${otherUserFirstname.charAt(0)}${otherUserLastName.charAt(0)}`;
-    await page.getByRole("button", { name: otherUserInitials }).click();
+    // Right after signup the header can transiently render the avatar button more
+    // than once (exact cause not isolated), producing a strict-mode violation on
+    // this locator; .first() sidesteps it regardless of the mechanism.
+    await page.getByRole("button", { name: otherUserInitials }).first().click();
```
mvandenburgh: Did you try pointing claude at this issue? It seems like something that we'd ideally fix, instead of hacking around buggy behavior in tests. 

If it's too much of a rabbit hole, we should include `TODO:` in this comment so that it is grep-able, and file an issue.

e2e/tests/dandisetLandingPage.spec.ts
```diff
@@ -47,24 +55,37 @@ test.describe("dandiset landing page", async () => {
   });
 
   test.describe("how to cite tab", () => {
+    // These tests only read the "How to Cite" tab of a single dandiset, so a fresh
+    // user + dandiset is registered once for the whole block instead of once per
+    // test. Registering per-test (7x) under 4-way parallelism was overloading the
+    // backend and causing frequent beforeEach timeouts; this also requires the
+    // tests to run serially, since they now share one page.
+    test.describe.configure({ mode: "serial" });
```
mvandenburgh: I think this comment could be simplified/reworded to be less LLM-speak. e.g., "since they now share one page.", over-emphasis on describing previous behavior in great detail with numbers, etc., are things that I think a human wouldn't mention in a code comment but which LLMs seem to love doing 😄

mvandenburgh (changes requested):
Looks good, I just had a question about the response times + suggestions to clean up some LLM-speak in the code comments

e2e/tests/dandisetLandingPage.spec.ts
naglepuff: This assertion failed due to a timeout at least once when I was doing my initial investigation. I reverted this change locally now, ran the test suite 10 times and saw 1 failure in the same spot (again due to timeout). With the timeout set to 15s, this test passed 10/10 times as part of the full suite of tests.

I updated this comment as well as part of 93e62b2

e2e/tests/dandisetLandingPage.spec.ts
naglepuff: see 93e62b2

e2e/tests/dandisetLandingPage.spec.ts
naglepuff: See 49012cd. Clearly I need to practice my "trust, but verify" when it comes to working with LLMs. The "multi-render" was never the issue, and the failure had to do with how `getByRole` works with the `name` option.

mvandenburgh (approved):

PR #2899, opened 2026-08-24
Current State of Garbage Collection design doc

mvandenburgh:
The garbage collection effort has been paused for nearly a year. The last time it was worked on, it was blocked on instrumenting a backup solution for the sponsored bucket; at the time, we decided that backup must come first before garbage collection can be safely rolled out. However, per the doc changes in #2830, we will not be implementing a traditional data backup solution due to the prohibitive cost of the available options. 

The document in this PR does two things:
1) Recaps the current state of garbage collection and what is left in terms of implementation
2) Presents an alternative solution for the backup issue that adds zero additional AWS billing costs, and argues why the approach provides sufficient "backup" capabilities and safety guarantees that will allow garbage collection to be safely deployed.

If we get consensus that the proposed mechanism is good enough to unblock garbage collection (Asset/AssetBlob/Upload, no Zarr yet), I'll make a follow-up to this PR with the final pieces required to enable that.

### DIFF
--- a/doc/design/state-of-garbage-collection.md
+++ b/doc/design/state-of-garbage-collection.md
@@ -40,11 +40,23 @@ Even a worst-case GC bug that "deleted" every object in the bucket could not per
 
 What a backup potentially adds beyond what is mentioned above is protection against threats like malicious behavior of DANDI admins or compromise of the AWS account by outside actors. However, we've already established that these scenarios are outside of the threat model, and aren't something the backup design exists to prevent in the first place.
 
+## Proposal: develop `fsck`-style scripts to monitor GC health
+
+In addition to implementing a longer recovery window, we propose to create an "fsck" style script that can report inconsitencies in the garbage collection state between the application database and the S3 bucket. Such a script can run on a schedule and would report the following items:
+
+- Live `Asset` objects with no corresponding `AssetBlob`
+- Live `AssetBlob` objects with no corresponding live S3 object
+- `GarbageCollectionEventRecord` objects with no corresponding deleted S3 object
+- Live S3 objects that are not already slated for garbage collection
+
+Any of the above categories would indicate an error in the garbage collection logic, or some other issue that needs to be addressed. If such problems are reported within the recovery window, they can be corrected by hand.
+
 ## What remains to get GC fully live (assuming no funding for backup/we're relying on undelete)
 
-1. **Extend the recovery window (if applicable)** - if we decide to accept extended recovery window instead of backup, we'll need a Terraform change to the trailing-delete lifecycle rule (30 → whatever we decide on days [here](https://github.com/dandi/dandi-infrastructure/blob/5ead0bd081b18082ccd8741ece8da19197cb631e/terraform/modules/dandiset_bucket/main.tf#L296-L299)), keeping the GC code's restoration window aligned with it.
-2. **Merge `Asset` GC** - review and land PRs [#2367](https://github.com/dandi/dandi-archive/pull/2367) and [#2368](https://github.com/dandi/dandi-archive/pull/2368).
-3. **Enable the daily task** - uncomment the beat schedule and monitor the first runs.
+1. **Implement the `fsck` script(s)** - this will help us keep the system healthy and, in particular, prevent catastrophic data loss.
+2. **Extend the recovery window (if applicable)** - if we decide to accept extended recovery window instead of backup, we'll need a Terraform change to the trailing-delete lifecycle rule (30 → whatever we decide on days [here](https://github.com/dandi/dandi-infrastructure/blob/5ead0bd081b18082ccd8741ece8da19197cb631e/terraform/modules/dandiset_bucket/main.tf#L296-L299)), keeping the GC code's restoration window aligned with it.
+3. **Merge `Asset` GC** - review and land PRs [#2367](https://github.com/dandi/dandi-archive/pull/2367) and [#2368](https://github.com/dandi/dandi-archive/pull/2368).
+4. **Enable the daily task** - uncomment the beat schedule, schedule the `fsck` scripts to run, and monitor the first runs.
 
 ---
 
--- a/doc/design/state-of-garbage-collection.md
+++ b/doc/design/state-of-garbage-collection.md
@@ -44,7 +44,6 @@ What a backup potentially adds beyond what is mentioned above is protection agai
 
 In addition to implementing a longer recovery window, we propose to create an "fsck" style script that can report inconsistencies in the garbage collection state between the application database and the S3 bucket. Such a script can run on a schedule and would report the following items:
 
-- Live `Asset` objects with no corresponding `AssetBlob`
 - Live `AssetBlob` objects with no corresponding live S3 object
 - `GarbageCollectionEventRecord` objects with no corresponding deleted S3 object
 - Live S3 objects that are not already slated for garbage collection
--- a/doc/design/state-of-garbage-collection.md
+++ b/doc/design/state-of-garbage-collection.md
@@ -1,0 +1,136 @@
+# State of Garbage Collection
+
+> **TLDR:** The garbage collection machinery is designed, implemented, and largely merged. However, the automated daily task has never been switched on because unattended deletion was deemed unsafe without a backup of the archive, and S3 Replication based backup has been deemed too expensive. We propose unblocking GC by extending the already-deployed S3 "undelete" recovery window (e.g., to 180 days), which protects against application bugs comparably to a backup, without incurring any additional AWS costs.
+
+## Current state of garbage collection
+
+Garbage collection (GC) for DANDI has been designed and built incrementally over several years ([overview of the GC types](https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-1.md)). To summarize the current state:
+
+- **S3 trailing delete ("undelete"): deployed.** With [S3 trailing delete](https://github.com/dandi/dandi-archive/blob/master/doc/design/s3-trailing-delete.md) already deployed, the production bucket is versioned: deleting an object only places a delete marker, a lifecycle rule permanently expires deleted objects 30 days later, and the `PreventDeletionOfDeleteMarkers` bucket policy denies `s3:DeleteObjectVersion` to all principals, preventing all IAM users and roles (including the DANDI API server) from being able to permanently delete objects; only the bucket lifecycle rule is permitted to do that. This affords a strong level of protection against accidental data loss via, e.g., mistakes by principals and bugs in application code.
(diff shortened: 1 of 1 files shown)


doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,136 @@
...
+**Storage Costs** (backup bucket in us-east-2):
+
+- Glacier Deep Archive storage: ~$0.00099/GB/month
+  - 1 PB = 1,024 TB × $0.99/TB = **$1,014/month**
+
+**Data Transfer Costs**:
+
+- Same-region data transfer between S3 buckets is free
+
+**Retrieval Costs** (only incurred when disaster recovery is needed):
```
yarikoptic: any plan for how we will test that "retrieval" so we have some assurance that it would work when/if needed?  otherwise we would pay 1k$/mo without knowing if could be used

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Future Costs
+
+The DANDI Archive is expecting a ramp-up in data volume of around 1 PB of new data over each of the next five years (plus an expected one-time contribution of 0.5 PB from LINC), culminating in a total nearing 6.5 PB (see [doc/archive/s3-backup.md](https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md) for the full cost survey based on this projection).
+
+Using the Deep Glacier rate of $11.88/TB/year, the projected additional cost of backup over the ramp-up is (from the [cost survey's projections](https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md#future-costs-over-time)):
+
+| Year 0 (1 PB) | Year 1 (2.5 PB) | Year 2 (3.5 PB) | Year 3 (4.5 PB) | Year 4 (5.5 PB) | Year 5 (6.5 PB) | Cumulative |
+| ------------- | --------------- | --------------- | --------------- | --------------- | --------------- | ---------- |
+| $11,880 / year | $29,700 / year | $41,580 / year | $53,460 / year | $65,340 / year | $77,220 / year | $279,180 |
```
yarikoptic, line 128: formatting suggests use of an agent... I would appreciate disclosures (see e.g. commits of mine) on use of AI

CodyCBakerPhD:
30 days is a pretty short window, but going longer might start to infringe on bucket use practices (e.g., we are also not permitted to 'fully backup' the bucket under the sponsored account)

Should there be a regular special 'check-in' where we go and look for potential errors following GC enabling to ensure we catch and correct any such issues before the temporary backup expires?

waxlamp:
> 30 days is a pretty short window, but going longer might start to infringe on bucket use practices (e.g., we are also not permitted to 'fully backup' the bucket under the sponsored account)
> 

This is a good point, one worth considering for the longer term. If pressed, I would argue that retaining deleted data for a period of *N* days directly supports [Term 1.1(e)](https://aws.amazon.com/opendata/open-data-sponsorship-program/terms/), *make reasonable efforts to optimize the End User experience, including building and maintaining applicable tools to enable use of Program Content*, by preventing accidental data loss. I agree that it becomes a debate on tradeoffs between safety and cost, over whether 7, 30, 90, or 180 days is the right period of time to hang onto stuff.

As we've done previously, I'd say we should go for the retention period we're comfortable with, in good faith, and scale it back to something more mutually acceptable if we get pushback. But ultimately I'll defer to @satra and @yarikoptic on this question.

> Should there be a regular special 'check-in' where we go and look for potential errors following GC enabling to ensure we catch and correct any such issues before the temporary backup expires?

Good idea. I added a bit of text to this doc in #2900 regarding adding safeguarding scripts to detect issues of data integrity surrounding GC. We can run that script on a cron and supplement that with "regular special check-ins" as we see fit.

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+- **`Upload` / `AssetBlob` GC: implemented and merged, but not enabled.** [Design](https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-uploads-asset-blobs-2.md) and [implementation (PR #2087, merged January 2025)](https://github.com/dandi/dandi-archive/pull/2087). The logic for orphaned uploads and asset blobs to be deleted after a 7-day grace period is implemented, with every deletion being recorded in `GarbageCollectionEvent` audit tables so it can be restored during the trailing-delete window. The beat cron it runs on is currently disabled and has never been turned on in production.
```
yarikoptic, line 10:
```suggestion
- **`Upload` / `AssetBlob` GC: implemented and merged, but not enabled.** [Design](https://github.com/dandi/dandi-archive/blob/master/doc/design/garbage-collection-uploads-asset-blobs-2.md) and [implementation (PR #2087, merged January 2025)](https://github.com/dandi/dandi-archive/pull/2087). The logic for orphaned uploads and asset blobs to be deleted after a 7-day grace period is implemented, with every deletion being recorded in `GarbageCollectionEvent` audit tables so it can be restored during the trailing-delete window. The cron it runs on is currently disabled and has never been turned on in production.
```

or what meaning of "beat"?

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+- **Manual GC command: merged.** [`collect_garbage` (PR #2343)](https://github.com/dandi/dandi-archive/pull/2343) lets an operator run GC on demand, with explicit confirmation prompts. This is currently the **only** way GC runs.
```
yarikoptic, line 11: was it ran?

so it triggers full "GC" for whatever there is GC or what specific GC does it do?

doc/design/state-of-garbage-collection.md
yarikoptic: TODO for myself: checkout `GarbageCollectionEvent` events

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+- `s3:DeleteObjectVersion` - the operation that actually destroys data. The existing `PreventDeletionOfDeleteMarkers` bucket policy denies this action to all principals, including the DANDI API server and any credentials it holds.
+
+Even a worst-case GC bug that "deleted" every object in the bucket could not permanently destroy anything: all of it would remain recoverable for the full window. Permanent deletion happens only via the S3 lifecycle rule. Actually destroying data would require an administrator to first deliberately remove the protective bucket policy, the same level of deliberate administrative action it would take to destroy a backup bucket in another AWS account.
+
+**Tradeoffs of accepting no true backup**
+
+What a backup potentially adds beyond what is mentioned above is protection against threats like malicious behavior of DANDI admins or compromise of the AWS account by outside actors. However, we've already established that these scenarios are outside of the threat model, and aren't something the backup design exists to prevent in the first place.
+
+## Proposal: develop `fsck`-style scripts to monitor GC health
```
yarikoptic, line 43: yeay -- finally we agree to agree!  IIRC I labeled `fsck`-style candidates among issues etc

yarikoptic (commented):
minor comments while reading... let's discuss

doc/design/state-of-garbage-collection.md
mvandenburgh: [Celery beat](https://docs.celeryq.dev/en/latest/userguide/periodic-tasks.html#introduction)

doc/design/state-of-garbage-collection.md
mvandenburgh: sure, i can fixup the commits and add that. the tables were indeed generated by an agent.

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+- **Glacier Deep Archive requires 12-48 hours to retrieve data**: the exact amount of time depends on the retrieval mechanism (see [https://aws.amazon.com/s3/storage-classes/glacier/](https://aws.amazon.com/s3/storage-classes/glacier/)), but either way, instantaneous recovery is not possible. However, this latency is acceptable for purposes of catastrophic data recovery; it also indicates why the trailing delete mechanism continues to be valuable (it offers a much faster "data recovery" in the cases where it can be used).
+
+### Cost
+
+**Below are the additional costs introduced by this backup feature** for a 1 PB primary bucket (assuming both the primary bucket and backup bucket are in us-east-2). All of this information was gathered from the "Storage & requests", "Data transfer", and "Replication" tabs on [https://aws.amazon.com/s3/pricing/](https://aws.amazon.com/s3/pricing/).
+
+**Storage Costs** (backup bucket in us-east-2):
+
+- Glacier Deep Archive storage: ~$0.00099/GB/month
+  - 1 PB = 1,024 TB × $0.99/TB = **$1,014/month**
```
CodyCBakerPhD: This shares the same inaccuracy as was fixed in the https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md

I recommend just linking to the rigorous calculations instead of duplicating here: https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md#cost-details

CodyCBakerPhD: (1 PB =/= 1024 TB)

doc/design/state-of-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+
+Using the Deep Glacier rate of $11.88/TB/year, the projected additional cost of backup over the ramp-up is (from the [cost survey's projections](https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md#future-costs-over-time)):
+
+| Year 0 (1 PB) | Year 1 (2.5 PB) | Year 2 (3.5 PB) | Year 3 (4.5 PB) | Year 4 (5.5 PB) | Year 5 (6.5 PB) | Cumulative |
+| ------------- | --------------- | --------------- | --------------- | --------------- | --------------- | ---------- |
+| $11,880 / year | $29,700 / year | $41,580 / year | $53,460 / year | $65,340 / year | $77,220 / year | $279,180 |
+
+At the year-5 volume, backup costs about $77,220/year (~$6,400/month) — far below the ~$31,000/month projected under the earlier 30 PB growth forecast, but still beyond available funding as of the May 2026 review.
+
+Note that these figures are purely the *additional* cost of backup; the cost of the primary bucket itself is excluded, since it is covered by the AWS Open Data Sponsorship Program.
```
CodyCBakerPhD, line 132: Again recommend just citing the table from https://github.com/dandi/dandi-archive/blob/master/doc/archive/s3-backup.md#future-costs-over-time

(which actually was not an agent but just poor old me typing it out from a spreadsheet I made)

doc/design/state-of-garbage-collection.md
CodyCBakerPhD: (though 1 PiB == 1024 TiB)

And oh wow AWS just freely refers to GB === GiB in their docs

Image

And actually means that unit for pricing schemes

amazing lol

doc/design/state-of-garbage-collection.md
CodyCBakerPhD: I did originally propose that we should test it out on a very small number of relatively small assets to ensure it works (and indeed costs) as expected

doc/design/state-of-garbage-collection.md
mvandenburgh: https://github.com/dandi/dandi-archive/pull/2899/commits/0ccf4978a0e950f8709c64181575b70dccbee5d4

Issue #748, opened 2021-09-21
Display some sort of "warning" that you're on the staging instance


waxlamp:
As we move to copy dandisets for testing from production, we may become confused more easily about which instance we're on. If we're on staging, that fact should be displayed as, e.g., a chip in the navbar or something like that.

PR #2903, opened 2026-08-25
Centralize `/api/info` calls + cleaner production/sandbox identification

mvandenburgh:
The frontend previously identified the deployment by string-matching hardcoded hosts (window.location.origin or metadata.url). The server already names itself via /api/info/ instance_config.instance_name (`DANDI` for production, `DANDI-SANDBOX` for sandbox, `DEV-DANDI` in development, and `DANDI-ADHOC` as the default when nothing is provided), so this PR updates the web client to cache that in a Pinia store and derive isProduction/isSandbox from it. 

The motivation for this change is that we currently have several areas of the frontend where we need to do things conditionally depending on whether we're running in production or sandbox, and all of them handle it by hardcoding URL strings and/or dynamically calling the `/api/info` endpoint. #2873 needs this behavior too, and adds more undesirable one-off logic for this. Also, we've discussed having some sort of indicator banner on sandbox to identify it as such (#748); this would also enable doing that in a clean way.

Downstream archives like EMBER can tweak the production/sandbox identifier strings as needed. I opted not to do this in this PR, but we could even make them configurable via environment variables as we do for similar values on the server.

This also converges the four separate dandiRest.info() call sites on one shared request.

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -79,12 +79,12 @@
 </template>
 
 <script setup lang="ts">
-import { ref, onMounted, type PropType } from 'vue';
+import { computed, type PropType } from 'vue';
 import { useRoute } from 'vue-router';
 import moment from 'moment';
 import { filesize } from 'filesize';
 import StarButton from '@/components/StarButton.vue';
-import { dandiRest } from '@/rest';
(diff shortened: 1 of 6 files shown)


naglepuff (approved):
LGTM!

PR #2912, opened 2026-09-02
Fix publish-time race between DOI minting and manifest writing

yarikoptic-gitmate:
Split out of #2862 so it can land on its own; that PR keeps the broader `sync_manifest_files` command and the design note.

Addresses (partially) #2759.

## The race

`_publish_dandiset` registered two independent `transaction.on_commit` callbacks, with `write_manifest_files.delay()` registered *first* and the DOI minting second. `on_commit` callbacks run in registration order, so the manifest task was handed to Celery before `_create_doi` had run its DataCite request and `version.save()` — a worker could write `dandiset.jsonld` from a version whose metadata still lacked the DOI, and nothing re-ran the manifest task afterwards. The published manifest keeps `doi: null` and a non-DOI `citation` forever.

The fix is just the ordering: register `_create_doi` first, then the manifest write. `robust=True` on the DOI callback keeps a DataCite failure from blocking the manifest write, preserving the old behavior where the two were independent.

Two regression tests cover it: the manifest carries the minted DOI on a normal publish, and a DataCite failure still leaves the version published with a (DOI-less) manifest on S3.

## Repairing the versions already affected

`rewrite_manifests_missing_doi` (added by @jjnesbitt) rewrites `dandiset.jsonld` / `dandiset.yaml` for published versions whose on-S3 manifests don't carry the DOI their metadata has — the two manifests that embed it. A version is also treated as stale if its manifest is missing or unreadable, since rewriting is the fix either way.

It deliberately touches nothing but S3: versions whose *metadata* has no `doi` are reported and skipped rather than re-saved, keeping published versions immutable.

<details>
<summary>Failure modes</summary>

- `create_doi` raises → Django logs the callback failure (`robust=True`), `version.doi` stays NULL, and the manifest is still written.
- A version with a `doi` column but no `doi` in its metadata can't be repaired from the manifest side alone; the command warns about those instead of modifying the published version.

Future hardening, not included here: make the DOI step a real Celery task with `autoretry_for=(requests.RequestException,)` so transient DataCite failures recover on their own instead of needing an operator sweep.
</details>

## Test plan

- [x] `ruff check` / `ruff format --check` clean (ruff 0.16.3, as pinned on master).
- [x] CI green on `9bef509` — `test`, both `test (3.13, …)` matrix jobs, E2E, lint and build-image.

https://claude.ai/code/session_01V31tV7K7jzN5HFbizSo9pF

### DIFF
--- a/dandiapi/api/services/publish/__init__.py
+++ b/dandiapi/api/services/publish/__init__.py
@@ -198,38 +198,16 @@ def _publish_dandiset(dandiset_id: int, user_id: int) -> None:
 
         validate(new_version.metadata, schema_key='PublishedDandiset', json_validation=True)
 
-        # After the published version is committed, mint the DOI first and
-        # only then write the manifest files. Doing these as two independent
-        # ``on_commit`` callbacks raced — the manifest task could run before
-        # the DOI was saved, leaving ``dandiset.jsonld`` on S3 without
-        # ``doi`` (and with a non-DOI ``citation``) and never rewritten. See
-        # https://github.com/dandi/dandi-archive/issues/2759.
-        #
-        # Failures at either substep are logged distinctly but do not abort
-        # the manifest write: even a DOI-less published version is better
-        # represented by an on-S3 manifest than by nothing.
-        # ``sync_manifest_files --fix-doi`` can later remint the DOI and
-        # rewrite the manifest if either substep below was skipped or failed.
-        def _create_doi_and_write_manifests(version_id: int):
+        def _create_doi(version_id: int):
             version = Version.objects.get(id=version_id)
-            new_doi: str | None = None
-            try:
-                new_doi = doi.create_doi(version)
-            except Exception:
-                logger.exception('Failed to mint DOI for version %s', version_id)
-            if new_doi is not None:
-                try:
-                    version.doi = new_doi
-                    version.save()
-                except Exception:
-                    logger.exception(
-                        'Minted DOI %s but failed to persist it on version %s',
-                        new_doi,
-                        version_id,
-                    )
-            write_manifest_files.delay(version_id)
+            version.doi = doi.create_doi(version)
+            version.save()
 
-        transaction.on_commit(lambda: _create_doi_and_write_manifests(new_version.id))
+        # Call _create_doi before writing manifest files, so that the new DOI is included in the
+        # manifests. If DOI creation fails, proceed to writing manifests anyway, to maintain
+        # existing behavior.
+        transaction.on_commit(lambda: _create_doi(new_version.id), robust=True)
+        transaction.on_commit(lambda: write_manifest_files.delay(new_version.id))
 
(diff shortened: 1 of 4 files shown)


dandiapi/api/services/publish/__init__.py
```diff
@@ -195,16 +195,16 @@ def _publish_dandiset(dandiset_id: int, user_id: int) -> None:
...
         def _create_doi(version_id: int):
             version = Version.objects.get(id=version_id)
             version.doi = doi.create_doi(version)
             version.save()
 
-        transaction.on_commit(lambda: _create_doi(new_version.id))
+        # Call _create_doi before writing manifest files, so that the new DOI is included in the
+        # manifests. If DOI creation fails, proceed to writing manifests anyway, to maintain
+        # existing behavior.
+        transaction.on_commit(lambda: _create_doi(new_version.id), robust=True)
```
mvandenburgh, line 206: Why is `robust=True` used? Do we want to still write the manifest files if the DOI creation raises an exception?

yarikoptic-gitmate: Yes — deliberately, to preserve the behavior we have today.

Before this PR the two callbacks were independent (manifest registered first, DOI second), so a DataCite failure never affected the manifest write. Django runs `on_commit` callbacks in registration order and *discards the remaining ones* as soon as one raises, so simply swapping the order would introduce a new failure mode: a DataCite outage would leave a version that is already published (the transaction has committed by the time the callbacks run) with no `dandiset.jsonld` / `assets.jsonld` on S3 at all. `robust=True` makes Django log the failed callback and continue, so the worst case stays what it is now — a manifest without the DOI.

That also matters for recovery: `rewrite_manifests_missing_doi` can repair a manifest that's missing the DOI, but it skips versions whose metadata has no `doi`, so a version left with *no* manifest by an aborted callback chain wouldn't be fixed by the sweep.

The failure is still visible — Django logs it at ERROR with `exc_info`, which Sentry picks up — and `test_publish_task_writes_manifest_without_doi_when_minting_fails` pins the behavior.

Happy to drop `robust=True` if you'd rather a DataCite failure surface as a failed Celery task, but since the version is published either way, failing loudly buys an alert at the cost of the manifests.

---

mvandenburgh: okay, just wanted to make sure `robust=True` doesn't suppress exceptions from Sentry. 

I suppose we do want the existing behavior; if we don't set `robust=True` and a DOI call fails, the Celery task would fail and the dandiset would be stuck in `PUBLISHING` status.

mvandenburgh (approved):

yarikoptic:
woohoo -- should I (or that gitmate fella) merge @jjnesbitt ?

yarikoptic:
ok, with power vested in me, since no objections voiced and even gaining an approval ...

PR #2918, opened 2026-09-07
Fix embargo funding and end date parameters on Dandiset creation

bendichter:
The embargo fields on the Dandiset creation form have not been reaching the server. The form passes `funding_source`, `award_number`, and `embargo_end_date` to `createEmbargoedDandiset`, which reads `fundingSource`, `awardNumber`, and `embargoEndDate`, so all three were undefined. The helper then sends camelCase query parameters, while `CreateDandisetQueryParameterSerializer` reads snake_case, so even the values it did have would have been ignored. The only parameter that got through was `embargo=true`. Every embargoed Dandiset created through the form since #2462 has therefore received the default two-year embargo rather than the grant end date, and no funder was recorded.

This aligns the three layers on the names the server serializer reads. The client no longer adds the funder as a contributor itself, since `create_embargoed_dandiset` does that when it receives the funding source. The award number is now required when an award is given, which matches the server-side validation; the form already asks for a project name when there is no number.

One related server-side issue I have not touched here: `create_embargoed_dandiset` builds the contributor list from the raw request metadata rather than the normalized metadata, so the contact person the server adds by default is dropped when a funding source is supplied. The existing test asserts that result. Now that the funding source actually reaches the server, that is worth fixing separately.

Split out from #2917, which builds on this branch.

https://claude.ai/code/session_0181P1YxVoFttcfqrdWr2mqu

### DIFF
--- a/web/src/rest.ts
+++ b/web/src/rest.ts
@@ -14,11 +14,7 @@ import type {
   DandisetSearchResult,
   IncompleteUpload,
 } from '@/types';
-import type {
-  Dandiset as DandisetMetadata,
-  DandisetContributors,
-  Organization,
-} from '@/types/schema';
+import type { Dandiset as DandisetMetadata } from '@/types/schema';
 import { useDandisetStore } from '@/stores/dandiset';
 import qs from 'querystring';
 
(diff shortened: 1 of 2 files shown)


web/src/rest.ts
```diff
@@ -233,34 +229,23 @@ const dandiRest = {
...
-        name: embargoData.fundingSource,
-        schemaKey: 'Organization',
-        awardNumber: embargoData.awardNumber,
-        roleName: ['dcite:Funder'],
-      };
-      const contributor: DandisetContributors = [...(metadata.contributor || []), award];
-      updatedMetadata = { ...metadata, contributor };
+    if (embargoData.embargoEndDate) {
+      params.embargo_end_date = embargoData.embargoEndDate;
     }
```
mvandenburgh, line 261: ~~Is removing this intentional? i.e., we no longer want to add the award as a `Contributor` at dandiset creation time?~~

Sorry, I missed that part in the description.

mvandenburgh (commented):
We should have a test for this to avoid regression. I opened #2921 for that

mvandenburgh (approved):

PR #2919, opened 2026-09-11
Restrict asset path deletion to ancestors of deleted asset

jjnesbitt:
The existing implementation did an un-indexed filter on `aggregate_files=0`, which resulted in a full table scan.

### DIFF
--- a/dandiapi/api/asset_paths.py
+++ b/dandiapi/api/asset_paths.py
@@ -179,33 +179,35 @@ def _delete_asset_paths(asset: Asset, version: Version):
     if leaf is None:
         return
 
-    # Fetch parents
-    parent_ids = (
+    # Fetch ancestors. Materialize the IDs now, as the relation rows backing this queryset are
+    # cascade-deleted below, and they're needed afterwards to scope the delete.
+    ancestor_ids = list(
         AssetPathRelation.objects.filter(child=leaf)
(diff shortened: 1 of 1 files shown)


satra:
@jjnesbitt - was a little surprised when i saw hierarchy mentioned in this PR. for clarification, our asset model is flat with only reference to dandiset versions (published or draft), right? path is a key in that model, and any notion of hierarchy is artificial for display purposes? or are we representing hierarchy now in some manner different from a str key?

jjnesbitt:
> @jjnesbitt - was a little surprised when i saw hierarchy mentioned in this PR. for clarification, our asset model is flat with only reference to dandiset versions (published or draft), right? path is a key in that model, and any notion of hierarchy is artificial for display purposes? or are we representing hierarchy now in some manner different from a str key?

Since [2022](https://github.com/dandi/dandi-archive/pull/1312) we've had an additional component of the archive called Asset Paths, which creates a way for us to track and display the path information of assets. While there's no hierarchical relationship between assets themselves, the typical folder/file structure that existed on disk when the dandiset was uploaded is still present in the paths of assets. So Asset Paths allow us to view and query that information for many uses, including it's original purpose, the file browser.

satra:
thank @jjnesbitt

mvandenburgh (approved):

PR #2920, opened 2026-09-11
Add bulk asset deletion endpoint

jjnesbitt:
This adds the bulk deletion endpoint, to improve efficiency of deleting many assets at once, and reduce/solve the shared ancestor locking problem. I've made this endpoint synchronous, as I think that's a good place to start. If runtime starts to become undesirable, we can always run it in a celery task instead. I was trying to avoid the extra overhead of that to start, along with either a lack of progress tracking for the end client, or us needing to create a system to track progress for this one task.


jjnesbitt:
All integration tests seem to be failing due to being denied access to pull images from dockerhub.

jjnesbitt:
Another thought I had is that we could entirely replace the existing asset deletion endpoint with this implementation, instead hosting it at `DELETE /dandisets/<dandiset-id>/version/<version>/assets/`. It could simply absorb the single asset deletion case, while still accommodating up to 1,000 assets at once . Let me know what you think about that.

dandiapi/api/views/serializers.py
```diff
@@ -15,8 +15,12 @@
 from dandiapi.api.models import Asset, AssetBlob, AssetPath, Dandiset, Upload, Version
 from dandiapi.search.models import AssetSearch
 
+# The maximum number of assets that may be deleted in a single bulk delete request
+MAX_BULK_DELETE_ASSETS = 1000
```
mvandenburgh, line 19: Is this based on any data? I agree it's a reasonable limit in terms of API design, though I'm curious how much memory is used when materializing a queryset of 1000 assets, and whether we need to do it in batches instead of all at once (the materialization within the network thread, that is). 

1000 `Asset` objects doesn't seem like it would consume much... claude says 512 KB to 2 MB for general `Model` instances, but I don't know offhand if there are any surprises in how `Asset` objects are serialized in memory. 

I'm guessing it's fine, we've just been burned by issues like that before so it always catches my attention when I see a variable sized queryset being fully materialized in memory.

jjnesbitt: Are you worried that handling 1000 assets at once is too _little_, or too many?

I don't think 1000 assets materialized in memory is much of a burden at all. But to your point, there's actually no reason to actually load assets into memory. I think just using asset IDs is sufficient, so I could make that change.

Regarding data, I did some rough experimenting locally and found that over 1000 assets per batch saw diminishing returns. But I didn't measure memory consumption.

mvandenburgh: I'm worried it's too many, mainly due to how dynamic the size of the asset metadata json can be. However, that point is moot if we just use asset IDs.

naglepuff (approved):

PR #2922, opened 2026-09-14
fix: pull MinIO from Quay in the dev stack

yarikoptic-gitmate:
MinIO deleted the `minio/minio` Docker Hub repository in September 2026, so `docker compose up` currently fails for every developer with `pull access denied ... repository does not exist`. This points `docker-compose.yml` at Quay, which still serves the archived community images, and pins the tag.

CI was already moved off Docker Hub in #2541 (to `bitnamilegacy/minio:latest`); the dev stack was missed.

<details>
<summary>Background, verification, and the remaining exposure</summary>

**Timeline.** In October 2025 MinIO stopped publishing free pre-built binaries and container images for the community edition — source-only from then on. The images stayed up, stale but pullable, for about eleven months. Then between 2026-09-09 and 2026-09-12 the `minio/minio` and `minio/mc` Docker Hub repositories were removed outright:

```
hub.docker.com/v2/repositories/minio/minio/  -> 404
hub.docker.com/v2/repositories/minio/mc/     -> 404
```

**Why Quay rather than `bitnamilegacy`, as CI uses.** This compose file passes `command: ["server", "/data", "--console-address", ":9001"]`, which is upstream MinIO's entrypoint contract. Bitnami's image wraps it differently — the CI workflows' own comment says as much ("This image does not require any command arguments"). `quay.io/minio/minio` is the same publisher and the same image CI used before, so nothing else in the service definition has to change. I pulled the image config to confirm:

```
Entrypoint: ["/usr/bin/docker-entrypoint.sh"]
Cmd:        ["minio"]
Env:        MC_CONFIG_DIR=/tmp/.mc
release:    RELEASE.2025-09-07T16-13-09Z
```

`mc` is still bundled and configured, so the `mc ready local` healthcheck is unaffected.

**Why pin.** Quay's community tags stopped moving when MinIO went source-only — the last community push was 2025-09-07 and the only 2026 tags are `*.hotfix.*` builds from the paid stream. `quay.io/minio/minio:latest` and `RELEASE.2025-09-07T16-13-09Z` are the same manifest (`sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e`), so `latest` is a frozen 2025 image wearing a name that implies otherwise. An explicit tag makes the next registry surprise a deliberate bump.

**Still exposed, not addressed here:**

- `.github/workflows/cli-integration.yml` runs `docker compose -f "$DANDI_PATH/tests/data/dandiarchive-docker/docker-compose.yml"` — dandi-cli's compose file, which has the same dead image. Both matrix legs are red: the `master` leg clears when dandi/dandi-cli#1921 merges, the `release` leg not until dandi-cli cuts a PyPI release (the compose file ships in the wheel).
- `backend-ci.yml` and `frontend-ci.yml` pull `bitnamilegacy/minio:latest`. That still resolves, but Bitnami's legacy namespace is itself an archive — last push 2025-08-19 — so it is a frozen image on borrowed time. Worth folding into the same decision rather than discovering it the same way we discovered this one.
- The community web console was stripped from MinIO's open-source build before this release, so `--console-address :9001` likely serves little. That predates this change — whatever was last pulled from Docker Hub was also console-less — but it is worth a look while someone is in here.

**Not verified locally:** no Docker daemon in this environment, so the stack was not brought up. The image config above was read from the Quay registry API, and the same digest is running green across dandi-cli's full CI matrix.
</details>

https://claude.ai/code/session_01AxQytU7QLTU3qDrUWCzBwn

---
_Generated by [Claude Code](https://claude.ai/code/session_01AxQytU7QLTU3qDrUWCzBwn)_


yarikoptic:
I reran dandi-cli jobs -- now that it is released etc, all good!  if no objections voiced, since current one simply broken, I will merge within 1 day

PR #2928, opened 2026-09-18
Add table visualizer for .tsv assets

CodyCBakerPhD:
As we enter the upcoming future of transitioning from the ad-hoc DANDI layout to proper BIDS for more modalities, we expect to start seeing more table files uploaded (`.tsv`-only, though the viewer is left to accept `.csv` because 'why not')

This PR improves the default viewing behavior for tables to allow quick introspection of such file contents

This has been requested from multiple external sources, including @rightbower with respect to their lovely BICAN cross-archive mapping tables, and @erikjohnson24 from EMBER

Good examples are found at https://deploy-preview-2928--sandbox-dandiarchive-org.netlify.app/dandiset/218795/draft/files?location=

Example screenshot: 
image


bendichter:
This looks great! I love the search, sorting, and pagination features.

bendichter:
The "table" button in the asset table is working great. The "download" button in the asset table now also opens a table. I expected it to just download the asset without opening the table. The "download" button within the table view works as expected.

CodyCBakerPhD:
> The "download" button in the asset table now also opens a table. 

My bad - good catch - that was one of the last things added but didn't check it on both sides

CodyCBakerPhD:
OK should be better now

bendichter:
The UX looks good to me! Confirmed that that issue is now fixed. I can really assess the quality of the code.

PR #2930, opened 2026-09-18
Only support bulk asset deletion

jjnesbitt:
This PR extends https://github.com/dandi/dandi-archive/pull/2920 by entirely replacing the existing single asset deletion endpoint. This ensures that the new bulk asset deletion code path is mandatory opt-in, and can't simply be ignored by an older client.


naglepuff (approved):
Changes all look good to me

yarikoptic:
I wonder if we should go that far that fast to break API and thus potentially tools without necessarily a clear necessity ATM.  Let's first implement / deploy bulk asset in cli, then throttle individual asset deletion with returning a message back to user to consider bulk deletion end point or smth like so tools which potentially use would still work?!

